Issue #10, opened 2015-04-14
Interface harmonization of in-memory (vg.cpp) and on-disk (index.cpp) graph representations


ryan-williams:
(ported from README.md)

ekg:
@adamnovak @glennhickey @edawson and I discussed this on a call the other day.

We should establish a list of functions that we'd like to have any vg-style object support. This should be pretty easy now that we've made five or six implementations of vg-handling libraries. If we can do this then the vg API can be implemented abstractly, and flexibility provided by giving different storage interfaces to the same system. For example, an xg-backed VG object should look basically the same as one that's backed by hash tables, although it couldn't be writeable. So perhaps there should be static and dynamic versions of the API, with the static one a read-only subset of the dynamic.

faithokamoto:
We have the `HandleGraph` and family now which I think solves what Erik was talking about.

PR #24, opened 2015-04-14
update README about open issues

ryan-williams:


ekg:
I like the idea of listing features here rather than open issues. Maybe this whole section should move to the wiki? It is motivating to see both what has been done and what remains.

ryan-williams:
yea, I'm open to what you prefer; agreed that an enumeration of some things that are supported is useful, and the checked items here are a reasonable proxy for that

ryan-williams:
which wiki are you referring to btw? does one exist, or are you suggesting initializing the github-repo-wiki for this project?

ekg:
What about this: let's keep the completed todo items as features, and for the new ones add links to the issues. Thoughts?

A few people have mentioned that the todo list is actually helpful, so I'm reluctant to give it up.

ryan-williams:
Either way works for me. I've updated this PR to put the "to-do" items in their own list and link to associated issues. There are other issues that are/were not in that list (anything before #10), I can add them if you like. Does this also suggest that people should add new issues filed to the README.md?

I can also leave it alone / close this, whatever you like.

PR #50, opened 2015-07-08
Integrating with GCSA2

adamnovak:
These commits pull in [jltsiren/gcsa2](https://github.com/jltsiren/gcsa2) as a submodule, along with its sdsl-lite dependency. They also add a new "vg index -g" option to build a GCSA2 index instead of a rocksdb index.

I also fixed a bug that I think was present before in "vg kmers -g", where it was creating a source and sink node for each input graph, instead of one global source and one global sink. Multiple source/sink nodes were crashing gcsa2's build_index. Now "vg index -g" works on multiple vg graphs, with a unit test to test it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,3 +34,9 @@
 [submodule "protobuf"]
 	path = protobuf
 	url = https://github.com/ekg/protobuf.git
+[submodule "gcsa2"]
+	path = gcsa2
+	url = https://github.com/jltsiren/gcsa2.git
+[submodule "sdsl-lite"]
+	path = sdsl-lite
+	url = https://github.com/simongog/sdsl-lite.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,8 @@ before_install:
   - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.9 60 --slave /usr/bin/g++ g++ /usr/bin/g++-4.9
   - gcc --version && g++ --version
   - sudo apt-get install -qq bc jq samtools
+  - echo "yes" | sudo add-apt-repository ppa:kalakris/cmake
+  - sudo apt-get update -qq
+  - sudo apt-get install cmake
 install: make get-deps
 script: make && make test
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,10 @@ LIBSNAPPY=snappy/libsnappy.a
 LIBROCKSDB=rocksdb/librocksdb.a
 SPARSEHASH=sparsehash/build/include/sparsehash/sparse_hash_map
 LIBHTS=htslib/libhts.a
-INCLUDES=-I./ -Icpp -I$(VCFLIB)/src -I$(VCFLIB) -Ifastahack -Igssw/src -Iprotobuf/build/include -Irocksdb/include -Iprogress_bar -Isparsehash/build/include -Ilru_cache -Ihtslib -Isha1
-LDFLAGS=-L./ -Lvcflib -Lgssw/src -Lprotobuf -Lsnappy -Lrocksdb -Lprogressbar -Lhtslib -lvcflib -lgssw -lprotobuf -lhts -lpthread -ljansson -lncurses -lrocksdb -lsnappy -lz -lbz2 -lrt
+LIBGCSA2=gcsa2/libgcsa2.a
+SDSLLITE=sdsl-lite/Make.helper
+INCLUDES=-I./ -Icpp -I$(VCFLIB)/src -I$(VCFLIB) -Ifastahack -Igssw/src -Iprotobuf/build/include -Irocksdb/include -Iprogress_bar -Isparsehash/build/include -Ilru_cache -Ihtslib -Isha1 -Isdsl-lite/install/include -Igcsa2
+LDFLAGS=-L./ -Lvcflib -Lgssw/src -Lprotobuf -Lsnappy -Lrocksdb -Lprogressbar -Lhtslib -Lgcsa2 -Lsdsl-lite/install/lib -lvcflib -lgssw -lprotobuf -lhts -lpthread -ljansson -lncurses -lrocksdb -lsnappy -lz -lbz2 -lrt -lgcsa2 -lsdsl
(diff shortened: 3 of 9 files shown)


ekg:
It looks like the build is now failing as you need to verify that the graph complexity reduction is still working as expected. See https://travis-ci.org/ekg/vg/builds/70151732#L2952. If you can verify that this is the correct behavior please update the test.

adamnovak:
Sorry, I'll have a look at it. I was getting a different hash than is in the tests even on master (7a13c8b67c2b1fcd89192e37ae946264 from the file at http://hgwdev.sdsc.edu/~anovak/halIssues/pre.gcsa2), so I didn't notice it had broken.

What's the actual file that produces the expected hash, for comparison?

ekg:
In the test directory (on my version, not this patch):

    vg construct -r small/x.fa -v small/x.vcf.gz | vg mod -pl 8 -e 4 - | vg kmers -g -k 8 - | sort | md5sum
    # 7caff8f7497ad9cf2762e2d95d54a241

This is the same as yours:

    curl -s http://hgwdev.sdsc.edu/~anovak/halIssues/pre.gcsa2 | sort | md5sum
    # 7caff8f7497ad9cf2762e2d95d54a241

The trick is the sorting. The kmers generation steps are all parallel so they don't necessarily make the same order.

Will this help track down what's happening?

adamnovak:
I suspect we're using different sort commands? I have this one:

```
[anovak@kolossus vg]$ sort --version
sort (GNU coreutils) 8.4
Copyright (C) 2010 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and Paul Eggert.
```


The file I put up is already sorted, according to my sort. It hashes to `7a13c8b67c2b1fcd89192e37ae946264`, and sorting it again (which doesn't change the order) still produces `7a13c8b67c2b1fcd89192e37ae946264` for me.

I get the same results on OS X, which also ships GNU Sort (although I need to use md5 instead of md5sum).

I figured out what the actual issue with the graph structure was; I had changed `add_start_and_end_markers` to return `Node`s owned by the caller, which caused trouble when the other code went to delete them from the graph. I have now changed back to the old semantics, and wrote extra code to deal with the deletion of the graphs in the loop over graphs when finding the GCSA kmers. Assuming Travis uses your sort, it should work now.

I have also added a check to make sure the user passes the graphs in in the right order. I can't know the max node ID over all the input graphs, and I have to make the head and tail nodes after seeing only the first, so if you use multiple graphs, the first one has to have the largest IDs.

adamnovak:
I can replicate your sort on another machine.

```
[anovak@rhea ~]$ curl -s http://hgwdev.sdsc.edu/~anovak/halIssues/pre.gcsa2 | sort | md5sum
7caff8f7497ad9cf2762e2d95d54a241  -
[anovak@rhea ~]$ sort --version
sort (GNU coreutils) 8.21
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and Paul Eggert.
```

They do indeed sort the files differently. Here are both sorted files: http://hgwdev.sdsc.edu/~anovak/halIssues/orders.tar.gz

adamnovak:
Here's the problem:

```
[anovak@rhea ~]$ curl -s http://hgwdev.sdsc.edu/~anovak/halIssues/pre.gcsa2 | LC_ALL=C sort | md5sum
7a13c8b67c2b1fcd89192e37ae946264  -
[anovak@rhea ~]$ curl -s http://hgwdev.sdsc.edu/~anovak/halIssues/pre.gcsa2 | LC_ALL=en_US.utf8 sort | md5sum
7caff8f7497ad9cf2762e2d95d54a241  -
```

```
[anovak@kolossus ~]$ curl -s http://hgwdev.sdsc.edu/~anovak/halIssues/pre.gcsa2 | LC_ALL=C sort | md5sum
7a13c8b67c2b1fcd89192e37ae946264  -
[anovak@kolossus ~]$ curl -s http://hgwdev.sdsc.edu/~anovak/halIssues/pre.gcsa2 | LC_ALL=en_US.utf8 sort | md5sum
7caff8f7497ad9cf2762e2d95d54a241  -
```

My sort order is "C" locale, and yours is "en_US.utf8" locale. I made #51 to enforce your ordering, so your hashes would be correct, although I'm not sure your ordering is what one would expect, since "en_US.utf8" tries to be smart about the handling of English text.

ekg:
Great! Thanks for figuring this out. Perplexion resolved.

ekg:
Can't merge. Will investigate. Hopefully this isn't because I pulled in #51.

adamnovak:
It is, there's a conflict because I touched adjacent lines. I'll rebase it.

ekg:
Word, thanks.

ekg:
It looks like `make -j` is failing due to an unspecified build order for gcsa2 and sdsl-lite. Checking now.

ekg:
I don't get it... it looks like it should pick it up correctly here: https://github.com/ekg/vg/blob/master/Makefile#L14

pgrosu:
Maybe the `Make.helper.cmake` file is not pre-processed in time via `cmake` to generate the `Make.helper` file.

~p

ekg:
It seems to work if I make twice.

This is probably an issue when upgrading the repo to use the new submodules
with git pull.

pgrosu:
That makes perfect sense, since it probably generates the `Make.helper` the first time, and then finds it the second time.

adamnovak:
sdsl-lite should definitely build first, since it's marked as a dependency.
Can you see if I did something wrong there?

When I get a chance I will see what happens on a completely clean checkout.

ekg:
If it is building OK on Travis then it should be fine, but actually... That
is building single threaded so maybe it will mask the issue.

PR #51, opened 2015-07-09
Enforce a single sort order for tests by setting the locale

adamnovak:


ekg:
Possibly we could do this in the test directory makefile ;)

In any case it sorts out the issue!

PR #52, opened 2015-07-10
JSON Input for "vg view"

adamnovak:
This PR provides a -J option for `vg view`, allowing it to input JSON in the same format that it outputs with -j.

In order to get it to writer that JSON as a VG-format graph, you need to do `vg view -Jv`, since the default output format is GFA.

I have a unit test for this: it makes a VG graph, turns it into JSON, runs the JSON through jq just to prove it's JSON, and then re-constructs a byte-for-byte identical VG-format graph.

In the process of doing this, I have made an attempt to regularize the format options to `vg view`, and to add a bit more error checking to detect when the user asks for impossible or unimplemented conversions.

I also had to replace the `pb2json` library with the related `json2pb` library, which is the same idea (and grew out of a fork of `pb2json`), except it implements conversion in both directions. It already had the custom fix from vg's `pb2json` to handle empty repeated fields nicely. I took the MIT-licensed files from master in https://github.com/shramov/json2pb, but added a method allowing you to read multiple protobuf objects from a file pointer (since we need to do that in the process of importing).

This builds on top of #50 , so there shouldn't be a merge conflict after that is merged.

I need this functionality so that I can manually write up some test graphs in JSON, so I can start adding support for the directionality flags on edges.


adamnovak:
Looks like Travis features jansson 2.2, and json2pb requires at least jansson 2.4 for `json_boolean`. I'll see if I can convince Travis to use a newer version.

PR #59, opened 2015-07-23
Support for the `from_start` and `to_end` edge flags.

adamnovak:
These commits should bring proper support for the `from_start` and `to_end` flags on VG `Edge` messages.

As a consequence of this, many functions have had to be extended with support for node orientation when doing things like traversals, because allowing edge bidirection lets you start out traversing a node forward and end up traversing the next node backward. This is accomplished through replacing some `Node*` pointers with a new `NodeTraversal` object, and pairing some `int64_t` node IDs with `bool` flags for orientation.

The semantics of the orientation is that it is a "backward" flag: it is false for the natural forward orientation of a node, where the start is on the left and the end is on the right, and true for the "backward" orientation, where the start is on the right and the end is on the left.

In general, indexes of edges "to" and "from" nodes have been replaced by index of edges on the "start" or "end" of nodes. If no edges use the `from_start` or `to_end` flags, these will have the same contents, and the total number of index entries (two per edge) is unchanged. There is possibly an undesirable amount of checking orientation flags and using that to determine what index to go to.

Note that support for cycles is still not implemented. While much of the groundwork has been laid, for example by inserting a topological sort (required to orient nodes) before aligning with `gssw`, there are still some structural issues with edge indexing that prevent multiple edges between a given pair of nodes, and the notions of head and tail nodes still need to be adapted to a graph that may loop back on itself. This PR provides support only for bidirected acyclic graphs with unambiguously identified heads and tails, as a generalization of the previously supported DAGs.


ekg:
I'm merging this with the expectation that we will also extend to the general case of edges that go from either the start or end of one node to either the start or end of another node. This would yield four possible edge types--- whereas now we only have two. Does this make sense?

ekg:
@adamnovak *great* job. Thank you very much for the effort on this front. Things had already gone far enough down the DAG path to make this improvement difficult. Onward to cyclic graphs!

adamnovak:
I'm not quite sure what you mean; this already allows for four types of
edge (you can have any combination of from_start, to_end, both, or neither).

ekg:
Great! In that case disregard my comment.

ekg:
Sorry for my confusion. I see now what you were referring to: https://github.com/ekg/vg/blob/9a1ebd4d79e209474b7c2c94244caf92174ad4ea/vg.hpp#L43-L63

Solid. Thanks again!

PR #61, opened 2015-07-25
vg index: ensure failure when given non-existent input filename(s)

mlin:
otherwise they're silently ignored which is dangerous when taking in a lengthy list of filenames.


mlin:
@ekg The test failure here seems to be unrelated to the diff and something you were fighting previously e.g. https://github.com/ekg/vg/commit/080d140b710f78605490ea6df599d8e6968670bf ... any suggestions, is it nondeterministic somehow?

PR #63, opened 2015-07-29
Add some OS X support logic to Makefile

glennhickey:
These are some hacks that should get VG building on Mac, without affecting the Linux build.  
This is pull request 3 of 3 to get VG building on the Mac. 

IMPORTANT:  You need to update VG's vcflib submodule to point to the latest vcflib (that merges my 2nd pull request).   And you need to update *that* vcflib to point to the latest tabixpp submodule (that merges my 1st pull request).   I am unable to change the submodule linking myself via pull request, without swapping in my own forks for these, which would probably be best to avoid. 

Thanks!


glennhickey:
Thanks for this!

Just a reminder that the Mac build won't work unless

vcflib https://github.com/ekg/vcflib/pull/97
vcflib/tabixpp https://github.com/ekg/tabixpp/pull/7

get merged too (and submodules updated in vg/)....

ekg:
Thanks! Now I need to update the submodules.

ekg:
@glennhickey, Would you please verify that the updated master now works for you?

glennhickey:
OK, I just re-cloned and built from scratch successfully on my Mac.
Thanks!

PR #65, opened 2015-08-04
Add support for cycles

adamnovak:
These commits add support for cyclic graphs.

This necessitated a lot of changes, described fully in the commit messages below. Briefly:
* Cycles are removed for GSSW alignment.
* The VG class now indexes edges by the node sides they touch, instead of the node IDs.
* The topological sort has been extended to do something smart with cyclic graphs; the general problem of "topologically sorting" a cyclic graph is NP-hard, so this could be improved with better heuristics.
* GCSA2 indexing now operates on a graph where everything has been doubled out to describe the forward and reverse strands. This is necessary for proper handling of reversing edges, since GCSA2 doesn't have native support for them. In this paradigm, a reversing edge connects the forward strand of node A to the reverse strand of node B, and visa versa, while a non-reversing edge connects forward strand to forward strand and reverse strand to reverse strand.
* Kmers that cross start-to-start edges are now produced correctly from kmer iteration.
* Kmer iteration has been revised to flip kmers with negative offsets onto the negative strand when necessary, and to be able to produce all kmers starting on both strands of a node. The deduplication logic has been revised to support this.
* Nodes along paths are now pointed to in a way that is unambiguous if the path loops through the same node twice.


ekg:
Wonderful work. Thanks @adamnovak!

PR #68, opened 2015-08-10
Add warnings for 0-ID nodes and duplicate nodes/edges

adamnovak:
Stop writing the same edge in multiple chunks, and only warn when we
want to check an input graph for duplicates.

Fix test data to not duplicate edges.

Make sure to serialize dangling edges when only the larger-id node is
in the graph.


adamnovak:
Both these errors (0-ID nodes and duplicate edges) were made by the sg2vg tool and not caught, so I figured I should add code to catch them.

This may make importing graphs that do contain duplicate nodes or edges for good reasons unnecessarily spammy.

PR #74, opened 2015-08-14
Make chop cut off consistently-sized pieces like was probably intended

adamnovak:
Right now it's trying to chop nodes at progressively larger offsets, and can eventually try to chop nodes at offsets greater than their lengths.


ekg:
I ran into something similar but stashed it away. Thanks for the catch.

PR #76, opened 2015-08-17
Several bugfixes

adamnovak:
Here are several bugfix commits that I didn't want to split out into their own branches.

* In the mapper, `balanced_kmers` wasn't generating the very last kmer if it ended at the end of the input sequence. Now it is.
* `VG::is_valid` no longer requires heads or tails, since some cyclic graphs don't have that.
* `VG::create_edge` can now create self loops.
* Non-reversing self loops are now properly removed when feeding the graph into GSSW.
* Self-loops are now no longer spuriously requested (and, previously, thrown out by `create_edge`) when attaching a node to all heads or tails, or when creating the GCSA2 start/end node.
* `vg view` now validates the graph when doing conversion.
* The code for getting the edges for a graph range from the index has been fixed up to account for the fact that the key for a node no longer sorts before all the keys for its edges.

### DIFF
--- a/index.cpp
+++ b/index.cpp
@@ -1477,7 +1477,7 @@ void Index::get_range(int64_t from_id, int64_t to_id, VG& graph) {
             graph.add_edge(edge);
         } break;
         case 'e': {
-            // Key describes an edge on the start of a node
+            // Key describes an edge on the end of a node
             Edge edge;
             int64_t id1, id2;
             char type;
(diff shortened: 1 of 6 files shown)


ekg:
Travis CI is throwing off these errors again https://travis-ci.org/ekg/vg/builds/76189651#L2906-L2911. I'll check the build locally.

ekg:
It passes locally.

ekg:
Your comments in 0a48b8d69e are great. This was all written in a hurry, and needed a pass through after it stabilized to describe the approach.

adamnovak:
Was the problem the banded alignment test? That one fails locally every
once in a while for me. Maybe it's a threading thing?

ekg:
I should have copied down the error rather than just linking to it. Does
the link still work?

adamnovak:
The link shows it passing all the tests. On the Travis history page, the
last failed build is supposed to be 13 days ago. No idea where the failure
info went, if anywhere.

Issue #80, opened 2015-08-21
Extracting a valid subgraph with vg find


mlin:
Probably a novice roadbump: I have `11.vg` for the whole chromosome and I'd like to extract a subgraph and then align a sequence to it. Based on [05_vg_find.t#L32](https://github.com/ekg/vg/blob/faa5e9fa2994a695f589ff790d3488aee62230b6/test/t/05_vg_find.t#L32) I'm trying this:

```
$ vg find -p 11:1910000-1970000 -c 256 vg/11.vg -d vg/index > 11_1910000_1970000.vg
$ vg view 11_1910000_1970000.vg -g | wc -l
graph invalid: edge index=12754 (174413169->174413171) cannot find node (from) 174413169
[vg view] warning: graph is invalid!
28866
[...]

PR #81, opened 2015-08-22
vg find -p: remove orphaned edges in the output subgraph

mlin:
Per discussion in #80, turned out a rather trivial change. I noticed the other code paths in `main_find` similarly call `remove_orphan_edges` so guessing this was just an oversight.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1844,6 +1844,7 @@ int main_find(int argc, char** argv) {
         if (context_size > 0) {
             index.expand_context(graph, context_size);
         }
+        graph.remove_orphan_edges();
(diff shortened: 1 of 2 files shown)


mlin:
Hm, I seem to have bad luck with sporadic test failures in PRs here. The same merge (as far as I can tell) passed in https://travis-ci.org/dnanexus-rnd/vg/builds/76921460

ekg:
Looks like the same intermittent errors we've been seeing on travis. @adamnovak thinks it might be some kind of race--- seems plausible but it's hard to trigger.

PR #82, opened 2015-08-26
Add a -n option to annotate edges with paths that use them in dot output

ekg:
Looks good. This is to test it out.


ekg:
@adamnovak this is great!

Some ideas for improvement:

1. Also annotate nodes in the same way.
2. Use newlines to condense the annotations.
3. Establish color/symbol combinations using a hash of the sequence name, and/or make this configurable.

Here's what I'm getting.

screenshot from 2015-08-26 10 20 05

I don't know what font it's using by default. I'm on Ubuntu 15.04 if that helps.

ekg:
Oh yeah, and we need a legend...

PR #86, opened 2015-08-28
Include .h instead of .cpp

adamnovak:
It works either way, but I figure this way is better.


ekg:
The test failures are very frustrating. I remain continually perplexed

adamnovak:
The right way to track it down is probably to run the test in a loop under gdb or something.

Which test is the problematic one?

ekg:
I've run it thousands of times and can't get it to fail.... :frowning:

PR #87, opened 2015-08-28
Build on OSX

glennhickey:
This is a minor change to the Makefile to keep allow building on OSX by disabling static link flag on that platform (some details about this here [here](http://stackoverflow.com/questions/3801011/ld-library-not-found-for-lcrt0-o-on-osx-10-6-with-gcc-clang-static-flag)).

PLEASE NOTE: For vg to build on OSX, the xg submodule will also need to be updated to a version that includes [xg pull request 4](https://github.com/ekg/xg/pull/4).


ekg:
Thanks @glennhickey. I don't use OSX and so I can't test on it. This would have been a problem for others.

glennhickey:
No problem.  So far all this OS X stuff has been super minor which is
pretty lucky.  Personally, it's a huge help to be able to work directly on
my Mac when possible (as opposed to remotely on a linux machine)

PR #89, opened 2015-08-29
Use gcsa + xg in mapper

ekg:
To comment on changes.

### DIFF
--- a/mapper.cpp
+++ b/mapper.cpp
@@ -2,9 +2,10 @@
 
 namespace vg {
 
-Mapper::Mapper(Index* idex, gcsa::GCSA* g)
+Mapper::Mapper(Index* idex, gcsa::GCSA* g, xg::XG* xidex)
     : index(idex)
     , gcsa(g)
+    , xindex(xidex)
     , best_clusters(0)
     , cluster_min(2)
     , hit_max(100)
@@ -25,6 +26,17 @@ Mapper::Mapper(Index* idex, gcsa::GCSA* g)
     , min_kmer_entropy(0)
     , debug(false)
 {
+    // Nothing to do. We just hold the default parameter values.
+}
+
+Mapper::Mapper(Index* idex, gcsa::GCSA* g) : Mapper(idex, g, nullptr)
+{
+    if(idex == nullptr) {
+        // With this constructor we need an index.
+        cerr << "error:[vg::Mapper] cannot create a RocksDB-based Mapper with null index" << endl;
+        exit(1);
+    }
+    
     kmer_sizes = index->stored_kmer_sizes();
     if (kmer_sizes.empty() && gcsa == NULL) {
         cerr << "error:[vg::Mapper] the index (" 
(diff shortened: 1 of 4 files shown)


mapper.cpp
```diff
@@ -56,11 +86,22 @@ void Mapper::align_mate_in_window(Alignment& read1, Alignment& read2, int pair_w
...
     VG* graph = new VG;
-    index->get_range(first, last, *graph);
     
-    // We also want to get nodes connected to these nodes along edges, down to the specified depth.
-    // This is useful for graphs that have unhelpful node ID assignments.
-    index->expand_context(*graph, context_depth);
+    // Now we need to get the neighborhood by ID and expand outward by actual
+    // edges. How we do this depends on what indexing structures we have.
+    if(xindex) {
+        Graph new_stuff;
```
ekg: Note that there is the function [XG::neighborhood](https://github.com/ekg/xg/blob/3c7002cf707af09f999b7e1dea7a6cacfec8c8dc/xg.cpp#L923). This looks fine though.

mapper.cpp
```diff
@@ -462,8 +503,25 @@ vector<Alignment> Mapper::align_threaded(Alignment& alignment, int& kmer_count,
...
+        // We fill this in only once if we're using GCSA indexing
+        gcsa::range_type gcsa_range;
+        
+        // Work out the number of *bytes* of matches for this kmer with the appropriate index.
+        uint64_t approx_matches;
+        if(gcsa) {
+            // A little more complicated. We run the search and count the range size
+            gcsa_range = gcsa->find(k);
+            // Measure count and convert to bytes
+            approx_matches = gcsa::Range::length(gcsa_range) * sizeof(gcsa::node_type);
```
ekg: I'm not sure we should be using the approximate size of matches in kilobytes, as this is a nice precise estimate. Maybe either use approx_matches or exact_matches. I can change this after merge.

mapper.cpp
```diff
@@ -474,7 +532,31 @@ vector<Alignment> Mapper::align_threaded(Alignment& alignment, int& kmer_count,
...
-        index->get_kmer_positions(k, kmer_positions);
+        
+        if(gcsa) {
+            // We need to fill in this vector with the GCSA nodes and then convert.
+            std::vector<gcsa::node_type> gcsa_nodes;
+            gcsa->locate(gcsa_range, gcsa_nodes);
+            
+            for(gcsa::node_type gcsa_node : gcsa_nodes) {
+                if(gcsa::Node::rc(gcsa_node)) {
+                    // We found a kmer on the reverse strand. The old index
```
ekg: Yeah, this is a little confusing. There may be situations where we use a forward-only index. (I'd like there not to be but it's not clear yet.)

mapper.cpp
```diff
@@ -474,7 +532,31 @@ vector<Alignment> Mapper::align_threaded(Alignment& alignment, int& kmer_count,
...
+           index->get_kmer_positions(k, kmer_positions);
+        } else {
+            cerr << "error:[vg::Mapper] no search index present" << endl;
+            exit(1);
+        }
+        
+        
         // ignore this kmer if it has too many hits
         // typically this will be filtered out by the approximate matches filter
         if (kmer_positions.size() > hit_max) kmer_positions.clear();
```
ekg: This could be moved earlier in the GCSA case, as we get the precise estimate up-front.

mapper.cpp
```diff
@@ -739,11 +821,22 @@ vector<Alignment> Mapper::align_threaded(Alignment& alignment, int& kmer_count,
...
             if (debug) cerr << "getting node range " << first << "-" << last << endl;
             VG* graph = new VG;
-            index->get_range(first, last, *graph);
             
-            // We also want to get nodes connected to these nodes along edges, down to the specified depth.
-            // This is useful for graphs that have unhelpful node ID assignments.
-            index->expand_context(*graph, context_depth);
+            // Now we need to get the neighborhood by ID and expand outward by actual
+            // edges. How we do this depends on what indexing structures we have.
+            // TODO: We're repeating this code. Break it out into a function or something.
```
ekg: Yup, it's used approximately the same in three places.

ekg:
The failure: `cannot align because node predecessors cannot provide seed` seems to result from xg passing duplicate nodes into vg. This should be fixed in xg. I'll dig further tomorrow.

ekg:
I'm working on this and I've some issues. I've got the mapper working.

However, xg is not threadsafe, for reasons I'm rather confused by, and the mapper only works single-threaded. Investigating.

PR #90, opened 2015-08-29
trying to understand this failure

ekg:
This test is frequently failing, so let's slow things down and see if travis can give us any hints.

### DIFF
--- a/test/t/07_vg_map.t
+++ b/test/t/07_vg_map.t
@@ -55,7 +55,11 @@ a=$(vg map -f graphs/2086553952_1469228759.mag -d graphs/199754000:199755000.vg.
 b=$(vg map -f graphs/2086553952_1469228759.mag -d graphs/199754000:199755000.vg.index -B 500 -J | jq '.path.mapping[0].position.offset' -c)
 is $a $b "banded alignment works correctly even with varied band size"
 
-is $(vg map -f graphs/2086553952_1469228759.mag -d graphs/199754000:199755000.vg.index -B 1000 -J | jq -c '.path.mapping[0].position.offset') 29 "unitig mapping position is as expected"
+sleep 1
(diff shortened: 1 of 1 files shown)


ekg:
So this is [what travis says](https://travis-ci.org/ekg/vg/builds/77810926#L8301):

```text
not ok 13 - unnamed test
#          got: 'banded alignment works correctly even with varied band size'
#     expected: ''
29
```

Basically, it produces the correct output but fails the test due to some kind of issue with the command syntax.

I wonder if this is some kind of problem with bash-tap. But if so, why have I never seen it anywhere but travis?

ekg:
AH great, I got it to fail:

```zsh
➜  test git:(happy-travis) ✗ while true; do n=$(vg map -f graphs/2086553952_1469228759.mag -d graphs/199754000:199755000.vg.index -B 1000 -J | jq -c '.path.mapping[0].position.offset'); if [ $n -ne 29 ]; then echo failed; else echo ok >>x; fi; done 
[: unknown condition: -ne
[: unknown condition: -ne
[: unknown condition: -ne
[: unknown condition: -ne
[: unknown condition: -ne
[: unknown condition: -ne
```

3048 tests.

ekg:
It might have something to do with (from running under valgrind)

```text
==31236== Conditional jump or move depends on uninitialised value(s)
==31236==    at 0x61BDB2: __pthread_disable_asynccancel (cancellation.S:98)
==31236==    by 0x619547: pthread_join (pthread_join.c:94)
==31236==    by 0x6CC425: JoinAllThreads (env_posix.cc:1594)
==31236==    by 0x6CC425: rocksdb::(anonymous namespace)::PosixEnv::~PosixEnv() (env_posix.cc:1093)
==31236==    by 0x877DF1: __run_exit_handlers (in /home/erik/vg/vg)
==31236==    by 0x877E42: exit (in /home/erik/vg/vg)
==31236==    by 0x86E3FF: (below main) (in /home/erik/vg/vg)
```

PR #94, opened 2015-09-02
Add a test that reveals GCSA2 indexing not working

adamnovak:
First off, this PR is meant to fail tests.

I've found this bug in the GCSA2 indexing while trying to maybe use it as a faster aligner for the graph bake-off. I'm going to move forward with the rocksdb aligner rather than trying to fix it now, since it's important that I stop messing about with test runs and get to running a nontrivial number of samples, but I figured that I ought to make a note of the problem here.


ekg:
Do you have any sense of what's broken?

adamnovak:
Unfortunately not really. It still manages to work on the other reversing
edge test case, so maybe it has something to do with node order?

ekg:
I'm also running into problems with graphs that have cycles in them.

adamnovak:
I've found the time to look into this, and it seems to be related to 2df58205279335c46e2d49edbd8f7c11ec472126 and the distinct start/end nodes. There's code in that commit to specially handle attaching the end node when the start node is used to break into a cycle. But if you have a graph that has, for example, a graph with two heads and no tails, like this graph I've added here, the start node will be connected but the end node won't. The end node will then produce duplicate and conflicting GCSA2 kmer entries, saying that the overall sing and source aren't connected to anything. This confuses GCSA2 and leads to a crash or an invalid index.

I think the right thing to do is to make the addition of the start and end nodes to the graph only actually add those nodes if they end up connected to anything. A graph with two heads and no tails will get only the start node added, for example. If you're not only looking at the forward kmers, this should work out fine. If you are looking only at the forward kmers, I'm not sure what ought to happen.

PR #96, opened 2015-09-07
Add vg pileup command to transpose read alignments to alignment columns

glennhickey:
I'm cobbling together a rudimentary vg variant caller for the purposes of evaluating graph construction and mapping methods in our bakeoff.  This PR is the first step for this, and adds basic functionality similar to samtools pileup:

* NodePileup record added to protobuf defs
* vg pileup command added to convert GAM to pileup format
* vg view changed to support pileups
* vg view changed to support json input of GAMs (only change that affects existing functionality)
* simple pileup test added.


ekg:
Thanks for this, it's a huge step!

Where next?

glennhickey:
Actually, let's wait until this is more mature (ie been used in next iteration of mapping bakeoff), before merging.  Worried I'll have to play around with the pileup logic / structure more than I'd expected in the meantime..  Will send updated PR later.

ekg:
Ah, ok. I was ready to merge this in. I'm happy to work on it with you in
the main repo. The danger here is that if we don't merge what you've done,
I'll do nearly the same thing and we'll end up forked.

glennhickey:
OK, great.  Maybe better to merge in sooner after all.  Re-opening!

For where next?  That would be "vg call".  My idea is to get something really simple (find SNPs with basic thresholding function) that we can use as a metric in our graph alignment bakeoff.  (comparing call output is of course a giant question, am hoping to use surject as a starting point).  Then iterate from there.  

For the existing pileup PR, I have two main concerns that you can likely help with:

1. I'm easily confused by strandedness, especially when there are two flags (is_reverse in both Mapping and Alignment).  When encountering Alignment.is_reverse() == true, does [this logic to flip it around make sense](https://github.com/glennhickey/vg/blob/master/pileup.cpp#L236-L256)? 

2. I've tried to mimic the [samtools pileup format](http://samtools.sourceforge.net/pileup.shtml) as closely as possible in the [new protobuf records](https://github.com/glennhickey/vg/blob/master/vg.proto#L142-L155).  But I'm far from a samtools expert and have no idea how to represent an insertion before the 0th offset of a node.  The current code spits out a warning and ignores these cases (which is fine for the naive snp caller, but obviously not ideal in genera).

ekg:
I'm gonna merge so as to avoid any forkage.

Some notes for future development:

When the alignment and graph construction end of things stabilizes, I plan to write the classifier needed to drive the "caller" using a [caffe](https://github.com/BVLC/caffe). Possibly a [recurrent neural network](https://en.wikipedia.org/wiki/Recurrent_neural_network) is a better model for an alignment-driven genotyper. A simpler bayesian model could be adapted from freebayes with minimal little effort.

glennhickey:
Cool, thanks. I'll PR my simple stone-age SNP caller as a placeholder.
Would be nice to get at least the freebayes logic working for this project
-- maybe a hackathon topic for new york.

ekg:
Your stone-age caller is still better than anything we've got in vg right now :)

I like the idea of a hackathon project focused on graph-based on calling. Let's try to sort that out in NY.

The genotyping component of freebayes is independent of the input parsing. I'm chartered to work on freebayes as well (moonlight hours), and planning a revamp of the software for better maintainability. I can keep my eyes open as I do so to how I could ensure the two bits are decoupled enough to use as an API in the context of another project.

PR #103, opened 2015-09-17
Syntactically correct RDF/JSON for vg (for discussion not for merging at this time)

JervenBolleman:
This pull request is not ready for accepting, but I open it as a starting point for further discussion.

The aim is to have the Variation Graph in an RDF serialisation so that SPARQL/Triplestore databases can load the variation graph and SPARQL queries can be executed on it.

The shell script uses a multitude of jq and sed scripts to translate the JSON output into RDF/JSON as used by the apache jena project. This [RDF/JSON](http://jena.apache.org/documentation/io/rdf-json.html) can then be translated or loaded directly into any SPARQL/RDF store.

Including the rather slow, but useful for testing, jena commandline sparql tool.
e.g. translate the vg json to n-triples and then use SPARQL to rebuild the linear dna sequence for a path.
```
vgtordf.sh x.json | riot --syntax="RDF/JSON" --output="N-Triples" > x.nt
sparql --data x.nt 
"PREFIX :<http://base/> PREFIX rdf:<ttp://www.w3.org/1999/02/22-rdf-syntax-ns#> 
 SELECT ?path (group_concat(?sequence; separator='') as ?pathSeq)
 WHERE {?step :path ?path; 
              :node ?node ;
              :step ?order.
        ?node rdf:value ?sequence} 
 GROUP BY ?path 
 ORDER BY ?order"
```
On the same data one can do for example to give the average "node" sequence length.
```
PREFIX rdf:<http://www.w3.org/1999/02/22-rdf-syntax-ns#>SELECT (avg(?len) as ?avglen) WHERE  { ?x  rdf:value ?s . BIND (strlen(?s) as ?len)}
```

The RDF is syntactically correct, but is only useable as a demo that the technology "works". i.e. wrong URI's are generated etc...

RDF/JSON is not a standardised format, JSON-LD would normally be the preferred solution for having a Semantic Web view on existing JSON. As far as I understand JSON-LD can not be used here as "IDs" are implicit indexes in JSON arrays.

Otherwise this small script touches upon some of the questions raised in https://github.com/ga4gh/schemas/issues/287, i.e. should there be a focus on data instead of on API's.
Especially taking into account that while we can have a SPARQL interface data does not need to be stored in a RDF database. For now it means that this data can be stored into content addressable RDF stores such as dydra.

### DIFF
--- a/test/t/17_vgtordf.t
+++ b/test/t/17_vgtordf.t
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+#
+BASH_TAP_ROOT=../bash-tap
+. ../bash-tap/bash-tap-bootstrap
+
+PATH=..:$PATH # for vg
+
+vgtordf.sh cyclic/all.json | jq .
+  ' 
--- a/test/t/17_vgtordf.t
+++ b/test/t/17_vgtordf.t
@@ -5,5 +5,6 @@ BASH_TAP_ROOT=../bash-tap
 
 PATH=..:$PATH # for vg
 
-vgtordf.sh cyclic/all.json | jq .
-  ' 
+plan tests 1
+vgtordf.sh <(vg view -V  graphs/199754000\:199755000.vg  -j) | jq . 
+is $? 0 "Basic syntax check passes"
--- a/vgtordf.sh
+++ b/vgtordf.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+((
+jq " .path[]|{(\"\(.name)\"):{\"rdf:type\":[{\"value\":\"Path\",\"type\":\"uri\"}]}} " $1|sed '$d';
+jq ".edge[]|{(\"\(.from | tostring)\"):{\"before\":[ {\"value\":(\"\(.to |tostring)\"), \"type\" : \"uri\"}]}}" $1 | sed -e "s_^\}_,_"| sed -e 's_^[\{]__'|sed '1d;$d';
+jq ".node[]|{(\"\(.id | tostring)\"): {\"rdf:value\":[ {\"value\":.sequence, \"type\" : \"literal\"} ] }}" $1|sed -e "s_^\}_,_"| sed -e 's_^[\{]__'|sed '1d;$d';
+echo ',';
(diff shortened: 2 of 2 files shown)


JervenBolleman:
Some background context:

The idea is that the variation graph is very interesting by itself, but will often be used in combination with other datasets including health data.
Having the VG as RDF allows users to combine in a single technology stack both the VG and for example FHIR data or annotated reference genomes, as well as allowing federalised queries etc.
As an off the shelf technology stack RDF/SPARQL will not have the same performance as a custom code base.
Yet it being standard with multiple implementations (including columnar stores) it could be used by some people.

The graph having RDF output does depend on vg existing and doing the hardwork of building the graph. i.e. it does not replace just extends vg for some usecases.

JervenBolleman:
There is also an option for a [binary RDF](using http://afs.github.io/rdf-thrift/rdf-binary-thrift.html) output apache thrift

ekg:
This is super interesting. How can I help?

JervenBolleman:
The next steps, for me, will be to formalize the URl structures/model in collaboration with the ga4gh and DBCLS. At that point it would be great to have your C++ code directly output this RDF/JSON (or a nicer serialisation).
My C++ code skills are non existent so I won't be much help there :(

We are also thinking about having stable(or near stable) ID generation so that graphs can be merged without collisions but without going for gUUID/random strings. We have some ideas for that but it requires much more work to get something real.

ekg:
I'd like to merge this, lest it rot (for some reason). @JervenBolleman do you object?

JervenBolleman:
I don't object but maybe we should have a minimal test case. i.e. just making sure that the transformation is executed and does not return non zero exit status.

ekg:
Would you be able to add one to this PR?

JervenBolleman:
At least one making sure the json is "correct"

JervenBolleman:
I added one test, I will see if it passes on Travis (might happen). My workstation does not have the dependencies installed for compiling testing VG. So I will see this evening on my laptop if the following it actually works.

JervenBolleman:
Ok found a json file for testing that generates a correct json file. The test pases on my mac os x laptop and validates the script can read the json vg view -j produces and produces valid json as well.
Requires a recent bash to run as it uses the nice job substitution syntax.
Noticed that if a path edge or nodes list is missing this script won't work.
Probably easiest long term is to for us to sit together and make the C++ code do the work

PR #105, opened 2015-09-22
Add (very basic) vg call

glennhickey:
vg call takes as input a vg pileup and produces an alignment for each snp in gam, which can be merged back into the graph with vg mod -i.  The logic is presently based off the samtools mpileup paper (but only supports single sample for time being).  This is a quick prototype for the purposes of our graph/mapping bake-off, but can hopefully help with baseline testing when more sophisticated approaches are implemented...


caller.hpp
```diff
@@ -0,0 +1,199 @@
...
+#include "hash_map.hpp"
+#include "utility.hpp"
+#include "pileup.hpp"
+
+namespace vg {
+
+using namespace std;
+
+// Super simple variant caller, for now written to get bakeoff evaluation bootstrapped.
+// Idea: Idependently process Pileup records, using simple thresholding to make calls.
```
ekg: You're no longer just thresholding though, no?

main.cpp
```diff
@@ -17,12 +17,128 @@
...
+#include "caller.hpp"
 #include "google/protobuf/stubs/common.h"
 
 using namespace std;
 using namespace google::protobuf;
 using namespace vg;
 
+void help_call(char** argv) {
+    cerr << "usage: " << argv[0] << " call [options] <pileup.vgpu> > out.gam" << endl
+         << "Compute SNPs from pilup data (prototype! for evaluation only). " << endl
```
ekg: All of `vg` is for evaluation purposes only :smile:

caller.hpp
glennhickey: Oops, yeah, I added support for quality scores but didn't update that
comment (maybe others?).  Not sure if worth another pull request but happy
to do it..

ekg: Nah we can fix it later.

PR #106, opened 2015-09-22
Add vg compare to compare kmer sets of two graphs

glennhickey:
This PR is based off of PR #105.   It adds a command, vg compare to compare two graphs.  Presently, it only supports simple kmer set comparison between two rocksdb indexes (which could probably be sped up).  It would also be nice to add support for the newer indexes (xg, gcsa) in the future.  We are also looking at a graph comparison implementation that could get added in the future, too.


jltsiren:
Kmer set comparison should be fast with GCSA, for any k up to the maximum query length. At least if we ignore the fact that the number of kmers can be really large for large values of k. The algorithm is very similar to merging the BWTs of read collections. Some optimizations require additions to the low-level interface, and they only work with the indexed paths. (Finding reverse complements is much slower, if the index does not include reverse complements.)

glennhickey:
That's good news.  I may ping you for help with the API if/when it comes time to writing a GCSA version of the compare.  Thanks!

ekg:
:star2:

ekg:
I should probably merge #105 first.

PR #137, opened 2015-10-30
const-ify references to lambdas

adamnovak:
This lets you pass an inline lambda, because the function no
longer requires the ability to assign to the lambda you are
passing in.

I didn't touch the lambdas in stream.hpp; they should be changed
too, both here and in the stream repo.


adamnovak:
Turns out that consts on stream stuff in vg were actually needed, because some function references get passed down into stream.hpp.

PR #141, opened 2015-11-02
Check Edits and Offsets in vg validate

glennhickey:
vg validate already checks that edges exist between successive nodes in paths.  This PR adds logic to check if the offset and edit fields in the mappings are consistent with these edges.  For example, if I have two consecutive forward mappings, m1 and m2, I'd expect that the region covered by m1 ends at the rightmost base of its node, and the region covered by m2 begins at the leftmost base of its node.  

This is all done under the assumption that offset is always relative to 0 on the forward strand (leftmost base).  So if m2 in above example is reversed, the expected offset would be node_length - 1.

### DIFF
--- a/vg.cpp
+++ b/vg.cpp
@@ -4172,6 +4172,7 @@ bool VG::is_valid(bool check_nodes,
                 paths_ok = false;
                 return;
             }
+            
             for (size_t i = 1; i < path.mapping_size(); ++i) {
                 auto& m1 = path.mapping(i-1);
                 auto& m2 = path.mapping(i);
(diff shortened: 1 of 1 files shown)


glennhickey:
I'd like to add a check for rank values to is_valid() (currently if two consecutive mappings dont have consecutive ranks they just get ignored).  Does something like 

1. sort mappings by rank
2. report error if ith mapping does not have ith rank

make sense?  They only way I see that being a problem is if we want to support cut up paths.  If that's the case currently, is there a way to modify the above to make sense?

ekg:
This could be an optional check. The cut up paths are important in some applications, so I don't know what to do about them. Maybe path cutting is just something that should be allowed. However, if the ranks are contiguous but there is no link, that's a problem. `is_valid` checks for that now.

PR #149, opened 2015-11-09
Reverse paths

adamnovak:
Right now divide_node is messing up any Mappings that happen to be the reverse strand, which is a problem in the BRCA1 graphs in which one of the input sequences was backward relative to the others.

The second commit here fixes that.

The first one goes through and clears and rebuilds the Mapping ranks after a bunch of operations that modify Mappings (and thus invalidate the existing rank values). I unfortunately wasn't able to generate a test case where messed up Mapping ranks actually broke anything; am I doing redundant work here?

### DIFF
--- a/test/reversing/reversing_path.json
+++ b/test/reversing/reversing_path.json
@@ -1,0 +1,35 @@
+{
+    "node": [
+        {"id": 24, "sequence": "GATTACA"},
+        {"id": 25, "sequence": "CATTAG"}
+    ],
+    "edge": [
+        {
+            "from": 24,
+            "to": 25
+        }
+    ],
+    "path": [
+        {
+            "name": "GI262359905",
+            "mapping": [
+                {
+                    "rank": 1,
+                    "position": {
+                        "node_id": 25,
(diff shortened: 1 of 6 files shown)


ekg:
Uh oh we may be overlapping. I'll check the merge manually. I have un
pushed changes on my end.

ekg:
Another thing, for vg msga the mappings are removed from the graph after
divide_node, so we wouldn't have noticed this problem.

The bug I was trying to fix (alignments over reversing edges) is mostly
resolved by fixing a silly mistake in the mapper.

Then, I found other issues with reversed mappings due to inconsistency in
the way that reverse mappings are handled. And reversed alignments just
seem like an anti-pattern, but maybe they are easier to work with. We need
to stick to a sane coordinate system for reversed mappings. The current one
is crazy. I'm stuck trying to figure out how to proceed and I realize it is
necessary to resolve this consistently to do so.

That all said I think you are not wasting your time here because I hadn't
been looking into mishandling of the ranks. I'll merge tomorrow.

PR #155, opened 2015-11-13
Repositionized

ekg:
In this PR I've updated the Position object to include orientation, which generalizes it properly. The is_reverse flag has been removed from Alignment and Mapping. The idea is to leave it attached to Path in that this would allow the determination, but I haven't done that here. There is only one way to represent a given path and still get back the original sequence.

In addition to these changes there are a heap of other fixes:

- resolved a bug with graph context gathering that yielded broken alignments
- MSGA now checks that paths are correctly included by realigning them
- validate understands the new position format
- pileup tests have been hacked into shape to match the new schema (although @glennhickey may need to verify)


ekg:
There has been a pretty significant performance regression for mapping in this series. I trust it can be resolved, but it is noticeable with the tests.

PR #159, opened 2015-11-16
Submodule gone crazy

ekg:
Various fixes.


ekg:
Why is travis not running?

PR #161, opened 2015-11-17
Add .pre-build make dependencies and deconstructor skeleton classes.

edawson:
This fixed up building with multiple threads on my Ubuntu 14.04 system. There's still a lot more work to be done for deconstruct, but I'm going to toss skeletons in there for the moment. We can manually fix up just the makefile if we want to wait until deconstruct has real functionality to add it in.


ekg:
c'mon Travis... reboot your Raspberry PI!

edawson:
Update: Currently, the deconstruction can retrieve a list of mappings. These mappings contain the correct Position. However, their edits seem to be incorrect, so for the moment I'm waiting to resolve this before finishing off mapping-> vcf::Variant conversion.

Oh, and submodules aren't playing nicely. I'm working on that to fix Travis.

edawson:
Okay, So here's a tarball of my test case with a .txt output for Github:
[tiny.txt](https://github.com/ekg/vg/files/49892/tiny.txt)

Warning: it's a tar bomb (though it's only 1 directory and three files).

my run line looks like:
```vg deconstruct -r tiny.fa -g tiny.vg -i tiny.vg.index```

PR #164, opened 2015-11-24
Including cyclic paths in graphs

ekg:
This will fail tests (vg mod) but it's mostly done. There is enough going on that I wanted to push it up.


ekg:
This is a big problem. When merging progresses through multiple nodes, we'll end up with situations like this (ad-hoc debugging output):

```text
------------------
merging mapping groups
r1
{"position": {"node_id": 8}, "edit": [{"from_length": 22, "to_length": 22}], "rank": 1}
r2
{"position": {"node_id": 9}, "edit": [{"from_length": 22, "to_length": 22}], "rank": 2}
merging mapping groups
r1
{"position": {"node_id": 8}, "edit": [{"from_length": 22, "to_length": 22}], "rank": 1}
r2
{"position": {"node_id": 10}, "edit": [{"from_length": 22, "to_length": 22}], "rank": 3}
------------------
vg: src/vg.cpp:1007: std::map<std::basic_string<char>, std::vector<vg::Mapping> > vg::VG::merge_mapping_groups(std::map<std::basic_string<char>, std::map<int, vg::Mapping*> >&, std::map<std::basic_string<char>, std::map<int, vg::Mapping*> >&): Assertion `f != ranked2.end()' failed.
```

The first merge makes the path look non-contiguous from this restrictive perspective. There are a few ways to handle this.

1. Somehow hack knowledge of the previous merges into the mapping merger.

I don't like this because it will be hard to generalize to the different cases. It will look different for reversed mappings that we merge in the forward direction.

2. Re-compact the mapping ranks.

This could be really expensive to do for the entire graph every time we are merging nodes. An alternative is to write something to just do it for a particular node's mappings, because everything is node local.

I think I'm going with 2.

ekg:
Another way to ensure this works out is to take the rank of the "right" mapping. As we are progressing left->right this should ensure that we've got the right semantics to merge at every step.

ekg:
This last bit gets unchop working.

```bash
vg mod -u msgas/q_redundant.vg | vg mod -C - | vg validate - # is happy

# and rendered
```

image

But normalization is flat broke:

```bash
vg mod -n msgas/q_redundant.vg | vg view -dpn -
```

image

I stared at this for a while and realized that the problem is just that I forgot to handle paths in the merging step where we eliminate redundant pieces of nodes.

PR #176, opened 2015-12-10
record VCF records in graph

ekg:


ekg:
@dillonl I'm going to make this optional via a flag in `vg construct` and then we can sort out the reference-matching bits by adding a new test for that.

ekg:
Gah this is rotting. Bump. Still it needs the flag, then we can add it in.

ekg:
Calling @JervenBolleman 

What needs to happen for this to be merged?

1. The behavior should not be default, and it should be enabled by a command line flag.
2. rsids should be pulled in if they are available.

src/vg.cpp
```diff
@@ -1842,6 +1846,15 @@ void VG::from_alleles(const map<long, set<vcflib::VariantAllele> >& altp,
 
             }
 
+            if (alt_node != NULL) {
+                auto nameIter = alleleNames.find(allele);
+                if (nameIter != alleleNames.end()) {
+                    for (auto& name : nameIter->second) {
+                        paths.append_mapping(name, alt_node->id());
```
adamnovak: It looks like this is just putting mappings on the non-reference nodes involved in the alt. If the alt has any matches to the reference separating the alleles representing bubbles, this will produce an invalid path, right?

ekg: That sounds right... I thought we had tested things this far but maybe I was confused.

I didn't merge this because it's not complete so I'd expect there to be problems like this.

PR #199, opened 2016-02-17
Incorporate new GFA parser + updated GFA spec

edawson:
I've updated a few of the tests in the test suite to match the latest GFA spec. The previous md5sum values used a malformed header that looked like `H   HVN:Z:1.0`; it should (and now does) look like `H   VN:Z:1.0`.

The new parser also:  
  - forces lexicographic sort
  - sorts output by sequence elements
  - allows comments, alignments and unaligned regions
  - can parse both paths with ranks and paths without ranks
  - makes adding a line type very easy.
  - parses header lines properly
  - parses extra fields in sequence elements.


ekg:
I need to merge this ASAP, before it gets stale. However, I can't do it today.

At the hackathon in Sendai we need to get a binary distributed, so I'm going to make a release (v1.4) and then a point release including this.

PR #201, opened 2016-02-19
Fix Docker build

adamnovak:


adamnovak:
Docker build now builds and runs.

ekg:
:joy:

Issue #202, opened 2016-02-19
travis could build with clang and gcc


JervenBolleman:

adamnovak:
Looks like there hasn't been any work on this in a while. While it would be nice if we had vg building on Clang, I'm going to close this for now.

PR #203, opened 2016-02-19
#202 try clang and gcc in travis file

JervenBolleman:


JervenBolleman:
This just does not work, sparsehash is not compilable with clang, and this needs to be done by someone with a ubuntu install so they can debug it

PR #206, opened 2016-02-19
Build on OS X for me

adamnovak:
This solves the rocksdb build issue with vzeroupper, and the problem with building snappy and libtoolize. You still can't build statically, though.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/ekg/gssw.git
 [submodule "snappy"]
 	path = deps/snappy
-	url = https://github.com/google/snappy.git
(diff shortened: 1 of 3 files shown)


edawson:
Nicely done! I fiddled with this all last week to no avail. Are you on Yosemite or El Cap?

adamnovak:
I'm on Yosemite. I haven't tried El Capitan yet, but I don't think it should be different (since this requires real GCC and not clang GCC), unless real GCC doesn't work on El Capitan.

JervenBolleman:
Real gcc works on El Capitan. I am on that mac os x version.

PR #210, opened 2016-02-20
Set up Travis for OS X

adamnovak:
Let's see how Travis likes this.

It's not set to fail if OS X fails yet, but that can change later.


adamnovak:
OK, VG builds on OS X with these commits, but tests can't pass because md5sum and shuf are missing, and also because vg occasionally segfaults.

adamnovak:
OK I've gotten the tests to run on my machine now. We'll see if Travis agrees.

ekg:
https://media.giphy.com/media/13EOgFTzFfiUM/giphy.gif

PR #212, opened 2016-02-20
Cleaned up protobuf datamodel comments and (meta)data

macieksmuga:


macieksmuga:
related to https://github.com/vgteam/xg/pull/16

ekg:
Oh this is a nice one. We should pick it back up. The failure was dependency related?

ekg:
@macieksmuga I did some of this but what's missing are the comments that you added (I removed metadata and data fields). Would you please check to see if you can add your comments back in?

ekg:
@macieksmuga did you want to update the comments? If not I'll close this.

ekg:
This is roughly matched by master.

Issue #109, opened 2015-10-01
VG RDF uses local non standard URIs for its own concepts


JervenBolleman:
We should mint URIs for the concept of a node in the graph, path, step in path, and before.
i.e. a  mini ontology for the following concepts.

Something like this

The following rdfs classes are present in the data (can be left implicit via an RDFS construct as they are not that interesting or needed for likely queries).

A `:Node` in a Variation Graph is a resource that represents a stretch of continuous DNA. They have an identity in the graph, as well as an associated IUPAC representation of the DNA molecule.

A `:Path` in the Variation Graph links nodes in the graph into an order (via steps) that represents a linear DNA sequence as found in a single person/assembly.

A `:Step` belongs to a single `:Path` and is ordered by the `:step` predicates value. 

The RDF presentation depends on the following predicates

`:before` is used to link `:Node`s together in the directed graph flowing in the general 5' to 3' direction of the consensus assembly. rdfs domain and range are `:Node`

`:step` has as  domain the `:Step` type and as a range `xsd:positiveInteger` for each `:Path` a step has an unique value.

`:node` links a `:Step` in a `:Path` to the specific `:Node`. has as domain `:Step` and range `:Node`

ekg:
I think this is a good start. We will also need to represent entities and
paths on the reverse strand. This is necessary to support concise
representations of inversions, translocations, and assembly graphs. So
instead of linking nodes our edges link :Sides, which correspond to a node
end, and each step in a path can be on the forward or reverse complement
(:Traversal maybe). Does this make sense?

JervenBolleman:
@ekg Something like this 

So instead of having a ```:before``` predicate we should maybe have 

```linksOn35to53``` i.e. what was ```:before```

```linksOn35to53ofComplementOf``` and use this predicate to say that the end of nodes sequence is added 

We can then also split the ```:step``` predicate into two subproperties, ```:stepOnForward``` and ```:stepOnReverse```.

This makes the query example a bit more complicated.

```
PREFIX :<http://base/> PREFIX rdf:<ttp://www.w3.org/1999/02/22-rdf-syntax-ns#> 
SELECT ?path (group_concat(?sequence; separator='') as ?pathSeq)
WHERE
{?step :path ?path; 
              :node ?node ;
              {
                   :stepOnForward ?order.
                   ?node rdf:value ?sequence
               }  UNION {
                   :stepOnReverse ?order.
                   ?node rdf:value ?toComplement
[...]

PR #217, opened 2016-02-21
#109 Start with making public stable URIs for the VG graph RDF concepts

JervenBolleman:
For now I put them in the biohackathon.org namespace, but this needs to be discussed with all stakeholders. Until that is resolved please don't merge this into master.


JervenBolleman:
Discussed in at [GA4GH schemas issue 572](https://github.com/ga4gh/schemas/issues/572)

Issue #214, opened 2016-02-20
Don't dump binary data to the terminal if we can help it


adamnovak:
vg should detect if standard out is a terminal, and automatically convert to GFA or something, instead of dumping binary data to a terminal.

JervenBolleman:
I am no longer sure this is a good idea. I just spend an hour listening to a horror story of lftp doing this kind of thing and sending my boss on a 3 day goose chase. It made cronjobs and jobs started with nohup act differently than those without that. 

Maybe exit(1) with an error message instead? Being explicit about what is going on?

adamnovak:
I've seen messages on standard error about not dumping binary output to
terminals, notably from gzip:

$ echo "Hello there" | gzip
gzip: compressed data not written to a terminal. Use -f to force
compression.
For help, type: gzip -h

$ echo "Hello there" | gzip -f
no�V �H���W(�H-J� v��A

So it looks like gzip follows this pattern of not even outputting anything.
Though it doesn't have a text format to use...

I think we want to avoid the case where you type your first vg command and
it fills your terminal with garbage. I'm not sure if refusing to output
anything is a better solution. We definitely would want to point users
towards the output-text switch in preference to the dump-binary-to-terminal
switch.

Maybe output should be GFA or JSON by default, with a particular option you
need to use when piping around binary data?

adamnovak:
If we don't want to change the default, the message should be something
like "pass XXXX for human-readable output or YYYY to send binary output to
the terminal".

adamnovak:
We should revive #228 to address this.

PR #228, opened 2016-02-21
Issue #214: don't dump binary data to terminal

buske:
VG serialization now outputs in GFA if stdout is a tty.
This is an improvement, but does introduce some idiosyncrasies:
- `vg construct -r x.fa` output GFA, while `vg construct -r x.fa | less` still shows binary

### DIFF
--- a/test/t/02_vg_construct.t
+++ b/test/t/02_vg_construct.t
@@ -7,7 +7,7 @@ PATH=../bin:$PATH # for vg
 
 export LC_ALL="C" # force a consistent sort order 
 
-plan tests 20
+plan tests 21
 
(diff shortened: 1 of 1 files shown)


ekg:
We should try this out and see if we like it. It's easy enough to revert.

Maybe I am just masochistic that I enjoy dumping garbage to my terminal.

macieksmuga:
I don't think that there's a way to distinguish between `vg construct -r x.fa | less` and `vg construct -r x.fa | <some_process_taking_binary_input>` from within the codebase. With that in mind, would it make sense instead to just add a `-g` flag for GFA output, or something like that, for when such output is immediately desired (say, when playing around with settings, to see how things are working out)?

ekg:
@macieksmuga you were thinking of adding the `-g` flag to all programs that output graphs?

Issue #230, opened 2016-02-23
strange `vg view -v` behavior


ekg:
```
➜  test git:(master) ✗ vg view -v B-3106.vg | vg view -
Unknown line identifier  encountered: �%GWI&/VM�ld<3�����ww�qwwwwwwwww���9g��d$|�oX�޳OU���3��
```

And so on...

Why does it not output vg format?

I can do

```
vg view -Vv B-3106.vg | vg view -
```

And it's happy.

pgrosu:
I notice an extra capital `V` in your command.  The first one has `-v` while yours has `-Vv`

pgrosu:
Maybe the reason is that in the code for the [main_view()](https://github.com/ekg/vg/blob/ccf57ee381bc6a7ebae151b831ccf59723ed61aa/src/main.cpp#L4709-L4719) method when it sees lower-case `v` it picks the input to be `gfa` as `input_type` - which would be empty at that point - as shown in the following portion of code:

```C
        case 'v':
            output_type = "vg";
            if(input_type.empty()) {
                // Default to GFA -> VG
                input_type = "gfa";
            }
            break;

        case 'V':
            input_type = "vg";
            break;
```

It's only afterwards that the default `input_type` empty check is performed, but by that time it's already populated from the `switch` statement:

```C
    // If the user specified nothing else, we default to VG in and GFA out.
    if (input_type.empty()) {
        input_type = "vg";
    }
```

buske:
Indeed, this is the cause. I'm about to make a PR with the fix.

ekg:
Thanks @pgrosu and @buske.

@pgrosu we're at a hackathon in Sendai, thus the lack of communication in public channels. Sorry to take your time reading the source.

pgrosu:
Always happy to help, and many thanks @ekg and @buske for the fix!  No worries, I was bored and I love reading code - I find it cathartic :smile:  Hope you get to do some awesome sightseeing in Japan as well.

Cheers,
~p

adamnovak:
This was fixed in #231

PR #231, opened 2016-02-23
Issue #230: input format should always default to VG

buske:


buske:
Rebased on top of master to resolve CI failure

PR #235, opened 2016-02-24
Mem mapper

ekg:
This introduces a MEM style mapper. (Maximal exact match based.) There are still some problems but the tests should pass. We'll see...


ekg:
Implementing MEM mapping has improved the sensitivity of the mapper quite a bit, and this has then uncovered a handful of bugs as I've pushed vg back through the MHC gene alignment tests that I'm doing. I'm resolving these.

A few todo items remain:

1. The MEM mapper is needlessly aligning SMEMs (singular exact matches) using GSSW. These could be aligned using the graph as represented in GCSA/xg, leaving the dynamic programming to be spent where it can help on the unaligned portions. This will save a lot of time.
2. Sections of the read that are all N still get MEMs. These should be filtered out or we will just confuse ourselves when aligning assembled genomes.
3. The MEM mapper is natively banded and should be adapted to behave as such. At present I am pushing in "bands" to it as if it is the old kmer-based mapper.

ekg:
This is failing on OSX and I'm concerned it'll be pretty difficult to resolve.

I need to find an OSX image on a VM I can use for testing. Any ideas?

ekg:
@adamnovak or @JervenBolleman, any thoughts would be welcome. I think that I'm doing something which produces undefined behavior on OSX.

I once had a Macbook but its battery exploded.

If I could get login on the travis instance for an hour I might be able to figure it out. @adamnovak mentioned something being available but I'm not finding anything.

ekg:
Oh! I've just reproduced the bug on Linux by running a modification of the test!!

ekg:
```
vg msga -f GRCh38_alts/FASTA/HLA/B-3106.fa -B 64 -P 0.9 -K 11 -X 2
```

:boom:

It happened once, and now doesn't happen any more. :frown:

ekg:
The solution was to run the test inside of valgrind.

Thanks to @buske's protobuf mem arena fix this is actually possible now --- before it featured millions of messages. (Maybe... I should have paid attention to those. Hmm.)

ekg:
Looks like I didn't solve all the issues with path corruption. I'll need to make sure the re-normalization of the path storage is happening after each normalization step which involves modification of the paths. I suspect I have still not caught all cases and can use valgrind on the failing OSX tests to hopefully reproduce the issue.

ekg:
My MEM extraction implementation is not complete because it only searches in the forward direction. It could be called forward-MEM. This is still useful, but it is incomplete and easy to fix. It means that we get different MEMs depending on where we start in the read. Also, it means we cannot break the read into SMEMs, because the MEMs aren't well defined and truly maximal. I would like to finish the implementation by deriving SMEMs, so this goes into the TODO items.

adamnovak:
Are you starting searches at multiple places in the read in parallel?

You can tell a match might need to be longer on one side if it bumps up against the start of your search string. Then I guess you can just run over from the previous non-bumped-up-against-the-start search attempt.

I guess the GCSA2 index isn't bidirectional, so you can't just turn around and extend left, though.

ekg:
I'm not starting the searches in parallel at multiple places. I start at
the beginning and search until that forward-MEM is determined. Then I step
to the next position and continue moving down the query. What I want to do
is I think what you are describing, to turn around and search backwards
starting from the beginning of the FMEMs in order to make them into true
MEMs. Then SMEMs can be determined through a scan across the read.

We search both the forward and reverse complement strands at the same time
in that the sampled positions and BWT are built from both strands. I do not
think this makes the index bidirectional in the same sense as the FMD.

The way to progressively search backwards with GCSA2 is to search the
reverse complement. This presents some awkwardness though, because we will
need to start searching for the reverse complement of our MEM from its end
on the forward strand. This seems expensive in that we double the number of
backward searching steps, but it is the only way I can see to derive the
true MEM.

Finding the start and end positions of the MEMs requires a series of steps
through the graph, but this is straightforward to implement so long as the
graph is normalized to not contain the same character on either side of a
bifurcation. Probably this isn't a safe assumption to make, so it may be
necessary to ensure we can handle poorly normalized graphs by implementing
this traversal as a backtracking search.

Are you starting searches at multiple places in the read in parallel?

You can tell a match might need to be longer on one side if it bumps up
against the start of your search string. Then I guess you can just run over
from the previous non-bumped-up-against-the-start search attempt.

I guess the GCSA2 index isn't bidirectional, so you can't just turn around
and extend left, though.

—
Reply to this email directly or view it on GitHub
<https://github.com/vgteam/vg/pull/235#issuecomment-190993903>.

adamnovak:
We don't have a retract operation in GCSA2, do we? Can we add one? Because
I think the Right Way (at least in algorithmics) to find all the
maximal-at-both-ends matches is to extend on the right when you can, and to
retract on the left and try again when you can't extend on the right. Every
time you have to retract, you have a substring match that is maximal, and
couldn't have been longer on either side. This works out to an O(N) scan
through the query string, because you do no more retractions than you have
input characters, and you do no more than twice as many extension attempts
as you do retractions.

Does that match up with the definition of MEM that you're using? Because
these things (I think we called them MUMs in our last paper) can
potentially overlap (due to having a core region with two incompatible
extensions).

jltsiren:
I'm not sure of the terminology used here.

In GCSA2, as in other FM-indexes, searching extends the pattern to the left. If lexicographic range `[sp,ep]` matches pattern `X`, we can find the range matching pattern `cX` for any character `c` as `LF([sp,ep], c)`. This takes a bit less than a microsecond in a whole-genome index.

If the match for pattern `X` cannot be extended further left, we need to remove characters from the right end of the pattern. The operation for this is the `parent()` operation in the suffix tree, which removes one or more characters. We need a representation of suffix tree topology for this operation. We also need the LCP array, if we want to know immediately how many characters the `parent()` operation removed.

The LCP array takes n bytes, increasing the size of the index by about 50%. If we add a range minimum tree over the array, we can simulate suffix tree topology using negligible additional space. Alternatively we can store the balanced parentheses representation of suffix tree topology explicitly in about n/2 bytes. Either way, the `parent()` operation should take a few microseconds. There might also be smaller and/or faster representations, if we only need the `parent()` operation without the entire suffix tree functionality.

With `LF()` and `parent()`, we can find maximal exact matches and solve related problems quite efficiently. It might also be possible to find them directly using the forward/reverse index we already have, but I don't have any idea how the forward/reverse subgraphs of the pruned de Bruijn graph are related. With a full de Bruijn graph, this would be much easier.

adamnovak:
Sorry, I got turned around; the PBWT stuff I have been doing implements
forward search, but the normal thing is backwards search (right to left).

I was trying to say what you said: if GCSA2 stored the LCP array range
minimum tree, you could get all the maximal exact matches efficiently.

ekg:
@adamnovak @jltsiren it sounds like you are talking about the same thing, and it's much more sensible than what the current arrangement would allow.

Based on the current implementation of GCSA2, I was thinking that the way to find the MEM would be to search until our BWT range goes to 0 (I called this our "forward-MEM", and it is defined as the longest match beginning at a certain position), then flip to the reverse complement of our query and repeat the process in the other direction until we go to 0 to generate a true MEM.

I could be misunderstanding something, but this would appear to work. It's just expensive because we nearly double the search time for each MEM.

I have implemented only the forward-MEM. This seems un-ideal for a variety of reasons, but it also appears that it will work. Non-overlapping forward-MEMs are a useful construct for seeding, but won't be as specific as SMEMs.

ekg:
By the way, I am still perplexed about what is making travis OSX builds fail here. I can't get the test to fail locally with valgrind. Still trying things, and looking through diffs to get an idea.

ekg:
I think that I didn't appreciate this fully. We can't get all MEMs efficiently without `parent()` because we will need to try to find the MEM starting at every single position. The forward-reverse search pattern I described will miss some MEMs that start in the middle of the "forward" MEMs*. With `parent()`, finding the MEMs requires < O(2N) operations where N = |query|.

*sorry that my terminology isn't clear. By forward I mean "in the direction of backward searching". Hopefully what I mean comes through.

ekg:
I've got OSX passing again. I had introduced some stupid path-corrupting logic in an earlier commit. I am amazed it didn't cause failures all over the place. Again, I was able to find the problem by running under electric fence. Problematically, doing so seems to change the behavior of vg, thus making some of the tests based on md5sums fail. I would like to run some of the tests using it though, but additionally doing that slows them down by a factor of 100. I'll leave this for future improvement.

adamnovak:
I've also tweaked some path stuff and possibly fixed some bugs in the work
I have been doing on haplotype path loading.

It's still too memory-inefficient to load all the paths from 1kg chr22 on
my machine, but I think I should probably send it up to mainline today
anyway, because it works fine on smaller problems.

ekg:
We can work with smaller regions for the time being, but this will require
some tweaks. So maybe it is best to improve the vg path storage rather than
focus on hacks. We can afford to be relatively slow but need to get things
more compact. Using dynamic wavelet trees could be the key.

adamnovak:
I think it might work well to load the VCF and make variant paths like in that other PR, and then have something that takes the variant-annotated graph and the original VCF and spits out the XG-encoded phasing paths. Maybe handling each sample one at a time, or in small batches, or working just with Path objects instead of the whole paths index.

PR #246, opened 2016-03-03
Test travis fix

pgrosu:


pgrosu:
Interesting...this requires a deeper look :)

ekg:
Hey Paul, what were you working on here? Btw, The problems I've encountered
in the mem mapper PR are mostly reproducible by using different memory
profiling tools.

pgrosu:
Hey Erik,

Well, I was trying a naive probabilistic test, by coming from the other direction :)  So we know that Travis are basically VM images, which can be selected from this list:

https://docs.travis-ci.com/user/languages/objective-c

I wanted to try different ones until one does not exhibit a probable memory corruption - whose memory for some VM implementations can sometimes can be file-mapped memory (i.e. paging file, BIOS, etc).

Basically I was bored, and wanted to try a lazy test after a long night of problem-solving :)

So what did you discover via valgrind that differentiates the VM as compared to the the physical machine?

Thanks,
~p

ekg:
@pgrosu I've found that there are a lot of invalid accesses that seem to be masked by optimization (?) on linux. There are probably some settings of the memory management that affect this. Basically, problems will come up on the OSX build but I can only reproduce them by either running vg in valgrind, which picks up some of the nastier errors, or using electric fence by compiling dynamically with `-lefence`. Sometimes setting `MALLOC_CHECK_=3` in the env helps. It's a bit of a grab bag. At present I'm working along with electric fence, which seems to be the most stringent, to resolve memory access bugs 

If you want to try to find bugs of this type, just add `-lefence` in the linker flags in the Makefile, and recompile with `make` (but not make static, it'll break). Then try running the tests. I suspect there are a bunch of small errors that get masked by a smart compiler. I just found one related to paths, which was boneheaded and I was amazed it even ran. (Something was being destroyed and immediately subsequent accesses to it weren't failing because the memory wasn't overwritten yet.)

pgrosu:
@ekg Definitely would be interested in helping out with debugging this, but I just noticed https://github.com/vgteam/vg/pull/248 passed the [OSX build](https://travis-ci.org/vgteam/vg/builds/113614892), which makes me suspect it was the machine the VM was running on, and that it's just a temporary periodic issue.

PR #249, opened 2016-03-09
Vcf paths

adamnovak:
This is an attempt to load paths from VCF phasing. It works, but memory usage balloons in ways I am still trying to track down when you try to use this with a large number of samples.

I'm not sure it's ready to merge yet, but I figured I should open a pull request.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5173,6 +5173,7 @@ void help_construct(char** argv) {
          << "    -v, --vcf FILE        input VCF" << endl
          << "    -r, --reference FILE  input FASTA reference" << endl
          << "    -P, --ref-paths FILE  write reference paths in protobuf/gzip format to FILE" << endl
+         << "    -B, --phase-blocks    save paths for phased blocks with the ref paths" << endl
          << "    -R, --region REGION   specify a particular chromosome" << endl
          << "    -C, --region-is-chrom don't attempt to parse the region (use when the reference" << endl
          << "                          sequence name could be inadvertently parsed as a region)" << endl
@@ -5200,6 +5201,8 @@ int main_construct(int argc, char** argv) {
     int max_node_size = 0;
     string ref_paths_file;
     bool flat_alts = false;
+    // Should we make paths out of phasing blocks in the called samples?
+    bool load_phasing_paths = false;
 
     int c;
     while (true) {
@@ -5209,7 +5212,9 @@ int main_construct(int argc, char** argv) {
                 //{"verbose", no_argument,       &verbose_flag, 1},
                 {"vcf", required_argument, 0, 'v'},
                 {"reference", required_argument, 0, 'r'},
+                // TODO: change the long option here?
                 {"ref-paths", required_argument, 0, 'P'},
+                {"phase-blocks", no_argument, 0, 'B'},
                 {"progress",  no_argument, 0, 'p'},
                 {"region-size", required_argument, 0, 'z'},
                 {"threads", required_argument, 0, 't'},
@@ -5221,7 +5226,7 @@ int main_construct(int argc, char** argv) {
             };
 
         int option_index = 0;
-        c = getopt_long (argc, argv, "v:r:phz:t:R:m:P:s:Cf",
+        c = getopt_long (argc, argv, "v:r:phz:t:R:m:P:Bs:Cf",
                          long_options, &option_index);
 
         /* Detect the end of the options. */
@@ -5240,6 +5245,10 @@ int main_construct(int argc, char** argv) {
 
         case 'P':
             ref_paths_file = optarg;
+            break;
+            
+        case 'B':
+            load_phasing_paths = true;
             break;
 
         case 'p':
@@ -5282,6 +5291,11 @@ int main_construct(int argc, char** argv) {
         }
     }
 
+    if(load_phasing_paths && ref_paths_file.empty()) {
+        cerr << "error:[vg construct] cannot save phasing paths without a paths file name" << endl;
+        return 1;
+    }
+
     vcflib::VariantCallFile variant_file;
     if (!vcf_file_name.empty()) {
         variant_file.open(vcf_file_name);
@@ -5299,13 +5313,21 @@ int main_construct(int argc, char** argv) {
     reference.open(fasta_file_name);
 
     // store our reference sequence paths
+    // TODO: use this. Maybe dump paths here instead of in the graph?
     Paths ref_paths;
 
-    VG graph(variant_file, reference, region, region_is_chrom, vars_per_region, max_node_size, flat_alts, progress);
+    VG graph(variant_file, reference, region, region_is_chrom, vars_per_region,
+        max_node_size, flat_alts, load_phasing_paths, progress);
 
     if (!ref_paths_file.empty()) {
         ofstream paths_out(ref_paths_file);
         graph.paths.write(paths_out);
+        if(load_phasing_paths) {
+            // TODO: keep the reference paths and throw out the phasing paths.
+            // We may need a way to get the reference path names, or some way to
+            // distinguish inside the Paths class.
+            graph.paths.clear();
+        }
     }
 
     graph.serialize_to_ostream(std::cout);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5323,10 +5323,22 @@ int main_construct(int argc, char** argv) {
         ofstream paths_out(ref_paths_file);
         graph.paths.write(paths_out);
         if(load_phasing_paths) {
-            // TODO: keep the reference paths and throw out the phasing paths.
-            // We may need a way to get the reference path names, or some way to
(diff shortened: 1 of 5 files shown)


adamnovak:
One thing I've thrown in here is some code to auto-magically output stack traces for some signals. Unfortunately it 's not necessarily robust to OpenMP, and really isn't related to the actual feature being written, so I'm planning to pull it out into its own PR.

ekg:
Thanks for the PR. I can use it to see how I can help.

As far as the memory usage of the paths goes... I think maybe it's time to move them to using dynamic wavelet trees within VG. Otherwise, we're going to need to break the construction of whole chromosomes into little pieces, which is not nice. I'll document a design here.

adamnovak:
We could get a lot of benefit out of a more traditional compression solution. Maybe store some of the paths by reference to other paths in some places? For example, where the phase paths match the reference paths, we could just say "follow the reference path for X bases" instead of storing all the mappings for each phase path. And where the phase paths match each other, we could have even longer blocks of reference to other paths.

Since paths are rarely/never modified, except to compensate for node modifications, this could work reasonably well. And it's clear to me how to do this, while it's not really clear to me how to represent the paths as dynamic wavelet trees without just porting over the whole gPBWT from XG.

I would like to see your design for it.

ekg:
> One thing I've thrown in here is some code to auto-magically output stack traces for some signals. Unfortunately it 's not necessarily robust to OpenMP, and really isn't related to the actual feature being written, so I'm planning to pull it out into its own PR.

I missed this. Did it end up merged here? (Can't tell from a quick look at the log.)

ekg:
I haven't yet had a chance to read through what you've done, so I might what you've written in the code.

To get to where we can do testing of gPBWT, we only really need to be able to serialize the paths from the VCF. They don't have to be kept in memory at once. So we could imagine collecting them somehow by making multiple passes over the VCF after construction is complete.

We don't need to read the VCF every time to do this. We can convert it to a more succinct representation, like a set of compressed integer vectors. Maybe a single compressed dynamic wavelet tree that we append to would work well. I don't know if this would be cheaper than keeping one wavelet tree per sample. Or, we could provide VG the PBWT here rather than the VCF with genotypes.

Provided a mapping from the alleles in the VCF to segments of the graph (stored as `vg::Paths` for instance), we can convert a haplotype in the VCF to a path in the graph by getting the Nth column from our matrix and generating a path by looking up the Path corresponding to each allele and the intervening reference-matching segments. The mapping could be implemented as a function that takes us from site-allele in the VCF to Path through the graph.

> Since paths are rarely/never modified, except to compensate for node modifications, this could work reasonably well. And it's clear to me how to do this, while it's not really clear to me how to represent the paths as dynamic wavelet trees without just porting over the whole gPBWT from XG.

The idea with the dynamic wavelet tree is that we can insert in the middle of the structure without paying an enormous penalty as we would by using vectors. Now, to avoid this cost we are storing `std::list`s of Mappings to provide this functionality now. We could store the paths each as a single DWT. That's as far as I'd gotten before.

The Paths stuff is all about working with the paths as collections of mappings: https://github.com/vgteam/vg/blob/master/src/path.hpp#L56-L73

The paths are stored as lists of mappings:

```
    // This maps from path name to the list of Mappings for that path.
    map<string, list<Mapping> > _paths;
```

A set of maps store the indexes to work with them. One gives us a pointer into the path for a given Mapping:

```
    // This maps from Mapping* pointer to its iterator in its list of Mappings
    // for its path. The list in question is stored above in _paths. Recall that
    // std::list iterators are bidirectional.
    map<Mapping*, list<Mapping>::iterator > mapping_itr;
```

One tells us to which path each mapping belongs:

```
    // This maps from Mapping* pointer to the name of the path it belongs to
    // (which can then be used to get the list its iterator belongs to).
    map<Mapping*, string> mapping_path;
```

We also store information about the mapping ranks:

```
    // We need this in order to make sure we aren't adding duplicate mappings
    // with the same rank in the same path. Maps from path name and rank to
    // Mapping pointer.
    map<string, map<size_t, Mapping*>> mappings_by_rank;
```

And we map from node ids to the set of mappings sorted by source path name:

```
    // This maps from node ID, then path name, then rank and orientation, to
    // Mapping pointers for the mappings on that path to that node.
    map<id_t, map<string, set<Mapping*>>> node_mapping;
```

We don't necessarily need to integrate gPBWT to model these, but it could help answer some of the questions we have.

It might be simpler to just directly replicated the functionality these provide using a series of DWTs. As the paths are rarely modified, we'd be happy to pay to increase the editing cost provided the memory size is minimized.

The first step is to decide if we want to be able to store paths in the graph that have edits in them. If not, then we don't need to keep the Mapping objects around. Or, we could keep them only when they are not perfect matches, but this is more complex.

Assuming we just work with perfect matches, we can reduce them to a series of node traversals, something equivalent to `pair<id_t, bool>`, so they can have an integer space representation. We could store them as DWTs. The question of what paths cross a given ID can be answered by storing a mapping from each node into these DWTs. We also back this with DWTs: one for the node+mapping description, and one bitvector for rank/select queries on it. It seems to me that the issue of "mapping ranks" disappears--- it's given by the first DWT. We can always serialize this back to Paths, for transmission, storage, or use in another application.

This isn't a succinct representation as in gPBWT, but it at least seems to be compact. Also it doesn't help us answer the pattern matching queries of gPBWT. But, it's something we can keep mutable, which will support the graph editing operations while preserving paths.

As an aside, at some point it might be easier to just merge XG and VG. It's not clear to me that they should be separate, especially if we want to scale to work comfortably with large graphs.

adamnovak:
No, the stack trace stuff was pulled out of this; I have it sitting in my own branch.

I think I see where you are going now with the wavelet trees. Rather than having them for dynamic bit vectors, you want dynamic integer vectors, so you can basically do what you do now with the linked lists of mappings to account for the extra mappings you need to make when you split a node.

It seems to me like a single dynamic int vector per thread would do nicely here. You don't get any benefit from run-length encoding (though delta encoding might help if node IDs are close together), but if you can do select on a node ID to find all its usages to update them when you break/merge the node, then you can dispense with most of the indexes and just store the node IDs you need. At 8 bytes per node visit, you could store all 5000 phasing paths for chr22 in ~20 gigabytes, which is manageable. (The whole genome's worth of paths would take 1.2 TB, but like you said there's no real reason to keep them all in memory at once).

The only problem is that you might be paying large constant factors every time you e.g. renumber a node.

adamnovak:
I think some of the problems might be solved by a new Paths object API.

First, I think a distinction should be made between Paths, which contain arbitrary Mappings, and Threads, which are just visits to nodes. I *kind* of want to let threads begin and end at offsets on their first and last nodes, so we don't go around breaking nodes in the graph to include things as threads, but I don't know if the complexity there would be worth it.

Next, I think that the Paths object shouldn't be throwing around Mapping pointers nearly so much. What it really should do is have Path and/or Thread iterators, which let you traverse forward and back along a path and get the Mapping (or, for a thread, maybe synthesize a Mapping) for the node you are on. The iterator can internally keep track of the path name and the rank that you are at, so you only need to go forward from path name and rank to mapping, and never backward from Mapping* to anything. This could cut out at least one of the indexes, even if we stick with the current internal representation.

I don't know if we would want to change the serialization, or if we would still want to dump everything as Paths in protobuf. (What we would want to do is not dump the paths in the same way we do now, with all the Mappings for a chunk all together, because the Mappings can be very numerous; I managed to serialize some oversized chunks when testing.)

We could then, internally, swap over to a succinct-ish dynamic int vector representation for visits to nodes for threads, using select to find the visits that are on a node.

If the rank-select stuff turns out to be too slow, we could also try something more special-case, with something like a binary tree where all the nodes are keyed by rank, except that you can insert new elements before/after existing ones in the leaf order and recalculate all the ranks according to the new structure. But alogn with the extra indexes we'd need to find things node number, I don't think that would be much better than what we have now.

ekg:
I'm tempted to add a Threads class, aimed at handling large numbers of haplotypes and annotation paths. We could use it alongside Paths, which are aimed more at describing the sequences that have been embedded in the graph to construct it.

That said, it looks like I can just modify Paths to do this. We don't need to work with the Mapping objects, or I can stash them somewhere when they are not perfect matches for the graph. This and the changes I suggested should make it easy to port things over. But, performance may be abysmal. I think it will simply have to be tested.

ekg:
Here's another concern. We're depending on the DYNAMIC library, which has this limitation according to its README:

> Note that at the moment the library does not feature delete operations (only inserts).

So we'll need to have special semantics to know if something's been deleted, and also a compaction routine to generate a new thread with the deleted nodes removed.

adamnovak:
True. It would be nice if there were a better library; I think in the meantime we might be stuck setting deleted entries to -1 or something.

ekg:
Does libmaus have delete implemented?

adamnovak:
Doesn't look like it, but I haven't seen any real docs for its API.

PR #254, opened 2016-03-12
update xg to version which allocates gPBWT dynamic int vector on heap

ekg:
This resolves segfaults with xg, and hopefully does here as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ CXXFLAGS:=-O3 -msse4.1 -fopenmp -std=c++11 -ggdb
 CWD:=$(shell pwd)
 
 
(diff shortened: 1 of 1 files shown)


ekg:
Looks like I need to set up OSX builds for XG. @adamnovak did you ever see errors like these when testing on your laptop?

```
template argument deduction/substitution failed:
deps/xg/src/xg.cpp:1084:54: note:   mismatched types 'std::initializer_list<_Tp>' and 'long unsigned int'
             start = max(end - len, s_cbv_select(rank));
```

I guess I need to go through and make them explicit? Perhaps -Wall or another flag will help suss them out.

ekg:
Now we need this to work at scale, seamlessly on many haplotypes.

This means upgrading the path storage model as discussed. Threads interface.

adamnovak:
No idea what's up with the type inference (?) on OS X. But thanks for making work.

ekg:
The other weirdness was that I need to allocate the dynamic wavelet tree on
the heap using new.

PR #257, opened 2016-03-14
Raptor

ekg:
Includes raptor as a submodule and runs a build script to set it up.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -52,3 +52,6 @@
 [submodule "deps/DYNAMIC"]
 	path = deps/DYNAMIC
 	url = https://github.com/vgteam/DYNAMIC
+[submodule "deps/raptor"]
+	path = deps/raptor
+	url = https://github.com/dajobe/raptor.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -y install cmake; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew tap homebrew/science; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install jq jansson coreutils md5sha1sum samtools raptor rasqal; brew link raptor;fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install jq jansson coreutils md5sha1sum samtools raptor rasqal gtk-doc; brew link raptor;fi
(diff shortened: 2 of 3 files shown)


ekg:
I've hacked on this and removed the dependency on gtkdocize from the build with some comments and hacks. But, now I'm stuck needing Bison.

```
configure: error: GNU Bison too old
```

Ouch.

I'm going to beg our sysadmins to install the raptor package. Hopefully soon I can test.

JervenBolleman:
There is a little script inside the deps/raptor/script package
that tries to install bison
but it does make install central
do the
./configure --prefix="$HOME/randombin/
make
make install
dance

On Mon, Mar 14, 2016 at 4:35 PM, Erik Garrison <notifications@github.com>
wrote:

> I've hacked on this and removed the dependency on gtkdocize from the build
> with some comments and hacks. But, now I'm stuck needing Bison.
>
> configure: error: GNU Bison too old
>
> Ouch.
>
> I'm going to beg our sysadmins to install the raptor package. Hopefully
> soon I can test.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vgteam/vg/pull/257#issuecomment-196371137>.
>



-- 
Jerven Bolleman
me@jerven.eu

ekg:
I decided to try to get the static build going without building raptor2
locally. I just pulled it in using apt-get install libraptor2.

Thus far I've added:

-lraptor2 -lcurl -lyajl_s -lxslt -lxml2 -lgnutls -lnettle -lhogweed -lgmp
-ltasn1 -lp11 -lgss -lshishi -lgcrypt -lgpgme -lgpg-error

Still no end in sight for its runtime dependencies.

JervenBolleman:
Of course you need to add the just installed bison to your $PATH env.

Also if it is an ancient Fedora, it will likely not work. Cent-OS 6.6. did not build a current raptor. 
You need Cent-OS 7+

ekg:
I'm on a current ubuntu with this problem now. I'm not sure building raptor
is going to resolve all of these dynamic dependencies? We'll need to pull
them all into the static build in any case.

ekg:
I'm stuck because "p11-kit can't be used as a static library".

http://upstream.rosalinux.ru/changelogs/p11-kit/0.20.6/changelog.html

Argh.

We may not get static builds back without force-building it as a static lib.

JervenBolleman:
Should be OK to force build raptor statically. We don't need the runtime
sepsis. That is all in unused functionality

ekg:
@JervenBolleman I don't know if I can. Attempting to build static, and I end up with a wild proliferation of libs.

ekg:
I think we would need to build all of these dependencies as static. That's
a no-go. I'm going to take a look at the RDF features we need and see if I
can work around raptor with something more self contained.

JervenBolleman:
We don't need much just a turtle parser. That is not to hard because all we
need is strings out of them. Just need to do it in a rail friendly style.

ekg:
How do we keep it rail friendly? (What is rail?)

JervenBolleman:
Autocorrect error RAII

Issue #252, opened 2016-03-10
static build broken on Linux in current HEAD


ekg:
I guess there are incompatibilities with raptor's curl and the other curl. @JervenBolleman says he's got a patch coming.

```gcc
/usr/lib/gcc/x86_64-linux-gnu/4.9/../../../x86_64-linux-gnu/libraptor2.a(raptor_www_curl.o):(.text+0x2be): more undefined references to `curl_easy_setopt' follow
/usr/lib/gcc/x86_64-linux-gnu/4.9/../../../x86_64-linux-gnu/libraptor2.a(raptor_rss_common.o): In function `raptor_rss_field_conversion_date_uplift':
(.text+0x870): undefined reference to `curl_getdate'
```

JervenBolleman:
no no, I am saying that I am putting apt get install raptor in the makefile getdeps target.

JervenBolleman:
I suspect the solution is to make raptor/librdf a deps in the git submodule sense.
I can't do this on my work machine (autotoos to old...) but will try to do so on my mac.

JervenBolleman:
I have made progress on this see [jervenbolleman/vg/raptor](https://github.com/JervenBolleman/vg/tree/raptor) but can't quite test it on my Cent-OS 7 box due to missing libbzp2 and libjansson static compiled libs in the default rpm repo.

JervenBolleman:
Looks like raptor head has the same problem with the libcurl in ubuntu trusty. Meaning that adding raptor as a dependecy/submodule does not help.

ekg:
I'm stuck now building on an older fedora system due to raptor. I'm not
sure what to do. It sounds like it'd be easiest to pull in your build
script. Then I can at least build the dynamic executable on the linux box.
Static binaries are a nice feature but building is an absolute essential.

ekg:
I'm pulling in your changes here: #257

ekg:
The raptor interface gets described by removing -lraptor2 from my libs and attempting to build static:

```
➜  vg git:(master) ✗ make static
[...]

PR #258, opened 2016-03-14
Rdf build ♥

ekg:
c'mon travis!

I changed the deps and make sure the raptor build was pushing its files to the right place in the includes directory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -y install cmake; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew tap homebrew/science; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install jq jansson coreutils md5sha1sum samtools raptor rasqal gtk-doc; brew link raptor;fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install jq jansson coreutils md5sha1sum samtools raptor rasqal gtk-doc bison; brew link raptor;fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PATH="/usr/local/opt/coreutils/libexec/gnubin:/usr/local/bin:$PATH"; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export LD_LIBRARY_PATH=/usr/local/lib/; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export CFLAGS="-I/usr/local/include/"; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,9 @@ before_install:
   - source ./source_me.sh
 install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make get-deps;fi
-script: make && make test && make static
+script:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make && make test && make static; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then make && make test; fi
 
 os:
   - linux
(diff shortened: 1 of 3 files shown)


ekg:
Will resolve #252

ekg:
This build now works for Linux but fails on OSX due to a bison dependency issue.

```
CMake Error at /usr/local/Cellar/cmake/3.0.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:136 (message):
  Could NOT find BISON: Found unsuitable version "2.3", but required is at
  least "3" (found /usr/bin/bison)
```

ekg:
I don't get the failure. brew suggests that Bison 3.04 is installed: http://braumeister.org/formula/bison

Could someone with access to brew please check what's up?

pgrosu:
It's installed in two locations:

/usr/local/Cellar/bison/3.0.4

/usr/bin/bison (which is 2.3)

You can either run it with `brew link bison --force` or might just need to update the $PATH to read `/usr/local/bin` before `/usr/bin`, as noted at the following link:

http://stackoverflow.com/questions/24835116/bison-latest-version-installed-but-not-in-use

Hope it helps,
~p

ekg:
I've got this far:

```
g++ -O3 -msse4.1 -fopenmp -std=c++11 -ggdb -o bin/vg obj/main.o obj/gssw_aligner.o obj/vg.o cpp/vg.pb.o obj/index.o obj/mapper.o obj/region.o obj/progress_bar.o obj/vg_set.o obj/utility.o obj/path.o obj/alignment.o obj/edit.o obj/sha1.o obj/json2pb.o obj/entropy.o obj/pileup.o obj/caller.o obj/position.o obj/deconstructor.o -static -static-libstdc++ -static-libgcc -I/Users/travis/build/vgteam/vg/include -I. -I/Users/travis/build/vgteam/vg/src -I/Users/travis/build/vgteam/vg/cpp -I/Users/travis/build/vgteam/vg/include/dynamic -ggdb -L/Users/travis/build/vgteam/vg/lib -lvcflib -lgssw -lprotobuf -lhts -lpthread -ljansson -lncurses -lrocksdb -lsnappy -lz -lbz2 -lgcsa2 -lxg -lsdsl -ldivsufsort -ldivsufsort64 -lvcfh -lgfakluge -lraptor2 -L/usr/local/lib
ld: library not found for -lcrt0.o
collect2: error: ld returned 1 exit status
make: *** [static] Error 1
```

My inclination is to avoid the static OSX build. Questions abound already: Where is libcrt0 coming into this?

ekg:
Thanks @pgrosu that bit fixed the bison problem!

The last part is just to avoid testing the static build on OSX. If you or anyone would like to make static OSX builds work, please see what you can do! It claims to be missing libcrt0. Typically this means we'll need to add it as -lcrt0 or such. Fixing this in the `-static` build may lead to another dependency, and so on. I don't now how deep down the rabbit hole we've gone.

ekg:
(Reset the PR to try to convince travis to run the OSX build.)

pgrosu:
@ekg Sorry to reply so late to this as it's been a bit busy lately, and I wanted to think a bit more of the possibilities.  So the easiest one without chaning internally on the gcc code and recompiling it, would be to perform the following based on http://stackoverflow.com/questions/5259249/creating-static-mac-os-x-c-build.  Below is the full example that shows how to build in the library on a Mac statically:

```Bash
$ cat example.h
void aFunction( int a );
$
$ cat example.c
#include <stdio.h>

void aFunction( int a )
{
    printf( "%d\n", a );
}
$
$ cat main.c
#include "example.h"

int main( ) 
{
    aFunction( 3 );

    return 0;
}
$
$ gcc -c example.c
$
$ ar -r libmylibrary.a  example.o
ar: creating archive libmylibrary.a
$
$ gcc main.c -lmylibrary -L. -I. -o main
$
$ ./main 
3
$
```

Then by inspecting the library, we see the function:

```Bash
$ nm libmylibrary.a 

libmylibrary.a(example.o):
0000000000000000 T _aFunction
                 U _printf

$ 
```

Then by inspecting the `main` program we see that the library is embedded:

```Bash
$ nm main
0000000100000000 T __mh_execute_header
0000000100000f60 T _aFunction
0000000100000f30 T _main
                 U _printf
                 U dyld_stub_binder
$
```
When we inspect it's dependencies, there are no dependencies other than the system library that is core for the Mac:

```Bash
$ otool -L main
main:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```

Hope it helps,
~p

PR #259, opened 2016-03-15
SuperBubbly

edawson:
This PR is a bit out of hand since I've fallen behind the main branch, but I should have it ready for merge soon.

I have simply wired in the code from [Superbubble](https://github.com/vgteam/Superbubbles) after some minor makefile/build changes and after hardwiring 64-bit support. Then I added the requisite lines in the main makefile. I completely destroyed deconstruct and have started adding bits back in, starting from suppressible detection.


ekg:
If you'd like to merge this in, don't go for the most-recent xg HEAD. @adamnovak has done a ton of stuff there and it might make sense to let him finish the merge. matser HEAD is on dd41c32  now.

Makefile
```diff
@@ -54,7 +56,7 @@ STATIC_FLAGS=-static -static-libstdc++ -static-libgcc
 .PHONY: clean get-deps test set-path static
 
 $(BIN_DIR)/vg: $(LIB_DIR)/libvg.a $(OBJ_DIR)/main.o
-	. ./source_me.sh && $(CXX) $(CXXFLAGS) -o $(BIN_DIR)/vg $(OBJ_DIR)/main.o $(LD_INCLUDE_FLAGS) -lvg $(LD_LIB_FLAGS)
+	. ./source_me.sh && $(CXX) $(CXXFLAGS) -o $(BIN_DIR)/vg $(OBJ_DIR)/main.o -lvg $(LD_INCLUDE_FLAGS) $(LD_LIB_FLAGS)
```
ekg: Why did you change the linking order? Was there a problem in supbub?

Makefile
```diff
@@ -15,7 +15,7 @@ CWD:=$(shell pwd)
 
 
 LD_INCLUDE_FLAGS:=-I$(CWD)/$(INC_DIR) -I. -I$(CWD)/$(SRC_DIR) -I$(CWD)/$(CPP_DIR) -I$(CWD)/$(INC_DIR)/dynamic
-LD_LIB_FLAGS:= -ggdb -L$(CWD)/$(LIB_DIR) -lvcflib -lgssw -lprotobuf -lhts -lpthread -ljansson -lncurses -lrocksdb -lsnappy -lz -lbz2 -lgcsa2 -lxg -lsdsl -ldivsufsort -ldivsufsort64 -lvcfh -lgfakluge -lraptor2
+LD_LIB_FLAGS:= -ggdb -L$(CWD)/$(LIB_DIR) -lvcflib -lgssw -lprotobuf -lhts -lpthread -ljansson -lncurses -lrocksdb -lsnappy -lz -lbz2 -lgcsa2 -lxg -ldivsufsort -ldivsufsort64 -lvcfh -lgfakluge -lraptor2 -lsupbub  -lsdsl
```
ekg: Same here for -lsdsl. Does it need to be at the end. I remember some arcane rules but not what they are.

edawson: I don't think -lsdsl needs to be at the end, but it needs to follow -lvg and -supbub which rely on it.

Arcane rule: LD makes a single search pass through the libs, so the lib functions must be provided on the link line after the calls to them.

PR #263, opened 2016-03-15
Insert paths for variant alleles

### LINKED ISSUE #262: Include VCF allele information in constructed graphs
This is basically what we wanted to do in #176. Now that we have code in that can include whole phase paths, it shouldn't be too hard to include per-allele path threads as well.

This combined with a prototype VCF-to-xg-thread importer could be a quicker way to get my gPBWT code to a testable state than taking on #261 right away.

adamnovak:
This basically does the same thing as #176, with different path names, and closes #262.

I'm naming things based on variant IDs, so this won't import non-ID'd variants. I do this because it's (syntactically) possible for a VCF to have multiple variants at the same site, but they really ought to have unique IDs. The problem is that paths for variants with no IDs are not created.

The other issue is dealing with deletions. Since deletions are spelled like GAA -> G, it's not really clear where the path for the deletion allele should go after visiting the leading G. What I do right now is just put a Mapping on the anchoring G and nothing else, and I rely on whoever is interpreting these paths to make the connection that if there's an allele 0 path that's this long, and an allele 1 path that's shorter and stops in a different place, you should treat it as a deletion and go to either the next reference base or something that replaces it.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5121,6 +5121,7 @@ void help_construct(char** argv) {
          << "    -r, --reference FILE  input FASTA reference" << endl
          << "    -P, --ref-paths FILE  write reference paths in protobuf/gzip format to FILE" << endl
          << "    -B, --phase-blocks    save paths for phased blocks with the ref paths" << endl
+         << "    -a, --alt-paths       save paths for alts of variants by variant ID" << endl
          << "    -R, --region REGION   specify a particular chromosome" << endl
          << "    -C, --region-is-chrom don't attempt to parse the region (use when the reference" << endl
          << "                          sequence name could be inadvertently parsed as a region)" << endl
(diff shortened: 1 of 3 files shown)


ekg:
We could encode the reference alleles that are implied in VCF and spelled
out in gVCF. These would begin where the indel representing paths end,
making the situation clearer.

On the other hand we get this for free using the reference.

ekg:
Looks good to me. Have you tried visualizing them?

adamnovak:
I have not tried visualizing them; My admins' GraphViz is still old and likes to emit bogus SVG when there are emojis involved.

Maybe there should be an option to disable the pictographs for compatibility...

adamnovak:
Also I do encode the reference alleles; they are all allele 0.

ekg:
Do you encode the implied reference matches between the variant sites?

adamnovak:
No, those would come in later when I actually start writing out haplotype
threads.

PR #266, opened 2016-03-17
scaling: DAGify fix, MSGA normalization

ekg:
DAGify was not correctly calculating `taget_min_walk_length`, which is the minimum length that we wanted to be able to walk back through each strongly connected component's DAG from tail to head. It always vastly underestimated this, causing effects like the following for even small min walk lengths:

screenshot from 2016-03-17 12-41-52

Due to a quirk of the bug, this caused severe problems when node sizes were short. This hopefully sorts out some of the bugs I ran into on large scale testing.

Another problem I ran into was that normalization is breaking on complex cases. In order not to tear my hair out, I'm removing it from the process for each alignment in MSGA. It can be applied afterwards, and in any case may need to be applied several times to get the right effect: `vg mod -n - | vg mod -n - ...`.

I've fixed up tests that depended on the normalization in MSGA to use an external normalizer.


ekg:
Currently blocked on some silliness in my tests.

This is a stack overflow that occurs in the recursive function to find strongly connected components (Tarjan's):

```
#7994 0x0000000000489b15 in std::_Function_handler<void (long), vg::VG::strongly_connected_components()::{lambda(long)#1}>::_M_invoke(std::_Any_data const&, long) ()
    at /usr/include/c++/4.9/functional:2439
#7995 0x0000000000489b15 in std::_Function_handler<void (long), vg::VG::strongly_connected_components()::{lambda(long)#1}>::_M_invoke(std::_Any_data const&, long) ()
    at /usr/include/c++/4.9/functional:2439
#7996 0x0000000000489b15 in std::_Function_handler<void (long), vg::VG::strongly_connected_components()::{lambda(long)#1}>::_M_invoke(std::_Any_data const&, long) ()
    at /usr/include/c++/4.9/functional:2439
#7997 0x0000000000489b15 in std::_Function_handler<void (long), vg::VG::strongly_connected_components()::{lambda(long)#1}>::_M_invoke(std::_Any_data const&, long) ()
    at /usr/include/c++/4.9/functional:2439
....
```

There are several problems suggested here.

This should not be recursion based. I should rewrite the recursion to use a std::stack rather than the call stack.

Why is VG::align even trying to align against a graph that big? It should have a maximum alignment matrix size. Before alignment, we check if the graph * read length has grown larger than this and if so we bail out. 1G here is way too big. It should probably be in the order of 10-100M.

Also, we should take care not to push large subgraphs into VG::align from the mappper. This looks like a symptom of a read which has many poor SMEMs that match in clusters. We try over and over to align it, and when the alignment doesn't work we attempt crazy amounts of soft clip resolution, which in turn cause more problems by expanding out the graph. If the graph is dense then this will end badly.

ekg:
Using e2e92c2f I am now able to build a (relatively) compact version of the HGVM MHC region:

```
time vg msga -f ~/graphs/hgvm/MHC/MHC.fa -B 128 -K 16 -X 2 -E 4 -D -G -S 0.95 -H 5 -D >MHC.vg
```

These aren't the default parameters for MSGA, but perhaps they should be. I've been using them in testing and find they generate the most compact graphs I can make. In my mind this means they drive good alignments and good indexes.

On a 32-core machine, it takes a bit over 1.5 hours. About a day of CPU is required.

```
real    103m43.080s
user    1539m24.712s
sys     17m56.923s
```

I then normalized the graph using a series of calls.

```
vg mod -n MHC.vg | vg mod -n - | vg mod -z - | vg mod -X 64 - | vg ids -c - >MHC.norm.vg
```

The resulting graph is normalized and has nodes of <64bp long.

I notice that much of this pipeline's time is spent loading and writing the graph, which is unfortunate. A sdsl-lite backed serialization format would not have these problems...

The graph is compact-ish. Although we've put 38M of sequence in (9 haplotypes over the MHC), the resulting graph is 17M and contains only 10.5M of sequence. (The unnormalized one is 10.6M.)

I think my goal with this branch is complete. It's time to merge it.

Issue #267, opened 2016-03-17
vg index: xg construction memory usage increase 


mlin:
I'm noticing memory usage in `vg index` to build the xg index has increased quite a lot in the last few weeks. Previously and for a long time, this operation maxed out at roughly 200 GiB of RAM on the whole-genome 1000G graph. Now I see it blowing through the machine's 244 GiB. Is this expected, anything to be done?

(Edited to remove misleading reference to gcsa2)
(Edited to remove misleading reference to 435c97c)

ekg:
I'm surprised that you are able to get the whole genome indexing to work
directly from vg index. It should be supported, but I am not yet sure
exactly how to implement it. The typical approach I have followed for the
1000G graph is to use vg kmers to make GCSA2 binary deBruijn graphs for
each chromosome, then feed these into the build_gcsa utility. The
documentation does need be updated to reflect the generation of the LCP
array that is used for counting and parent queries. Memory usage should be
well below 100G maximum. I think @jltsiren's tests show under 40G during
build_gcsa.

mlin:
Sorry I got a bit mixed up -- I mean the use of `vg index` to build the xg index, the first step in: https://github.com/vgteam/vg/wiki/working-with-a-whole-genome-variation-graph#xggcsa-index-high-memory-build-low-memory-runtime-fast

So i'm not getting to the kmers or GCSA2...

adamnovak:
You need https://github.com/vgteam/xg/commit/9fda90e46546956195cc8694da0d0282fe0283f3 which came in with #25. I'll make a PR to update vg to use that xg.

mlin:
Thanks @adamnovak - I'm launching a build from your PR now

ekg:
I've merged your PR in @adamnovak.

adamnovak:
Does it actually solve your problem, Mike?

mlin:
I hit the same OOM error with 93d26266 unfortunately. Are there any other quick things to try? If not, I can try some git-bisect although the iteration time is slow unfortunately.

adamnovak:
OK, I am looking at this a bit more closely. I thought you were talking about some older commits where I brought in the succinct thread storage, but it seems that that is not your problem if you've made it work successfully with anything after, say, c621d413b707278194f32540fdedbd7f7b64dd47.

[...]

PR #272, opened 2016-03-21
Vcf threads

adamnovak:
This can't merge yet (as I need to rebase), and I still want to revise the way the little allele paths are handled (to not put them in the XG index itself, because then it wastes loads of time making the tiny things into wavelet trees), but I figured I should open a PR for my better approach to loading phase threads from VCFs.


ekg:
Any way I can help on this? I see you've integrated the callback based xg build, which is something I'm interested in using. I've run into some problems making the xg index for an MHC graph that I've generated (which is curious because the XG index is built internally in the process of long read assembly). I thought it would make sense to sync up with the current xg master, but it looks like you've already done all the integration here.

Is the blocking thing the allele paths?

ekg:
I'm going to rebase, test, and likely merge. I need to keep up to pace with xg development because I'm trying to resolve #267 and some other issues. `XG::from_callback` might be the ticket.

It is OK in my mind if the feature does not scale yet. There is utility in keeping development together.

edawson:
How should I zig with SuperBubbly then? Is it fine if it's on the latest XG?

ekg:
If you can keep level with the master on this. We will update to use the
new xg in #272.

ekg:
I've got the merge ready locally. I'm going to push it into master.

@adamnovak I don't see any tests handling the uses here. That's fine. We can add them when you're satisfied with the implementation.

Some things I'm interested in doing are adding an interface to read the paths file in for vg view (for debugging, coolness), and also a vg view option to dump the paths in json format.

PR #280, opened 2016-03-29
Use a more efficient dice_nodes algorithm

### LINKED ISSUE #279: `dice_nodes` is slow on large nodes
As uncovered in #278, `dice_nodes` is O(n^2) and can sit more or less forever when you start with chromosome-sized nodes.

adamnovak:
This fixes #279.

Unfortunately, the test is going to fail here. There are a lot of hashes in the tests, and this changes how sequences are broken up into nodes and thus changes the hashes of a lot of things.

I was going to just clobber the old hashes, but some of them were in the MEM finding tests, and the MEM-finding results should really be the same. So I am parking this here until I can replace the hashes in those tests with something that I can actually look at and compare with the results I am getting, to see if I really am getting something like what I am supposed to get.

It's really hard to debug hashes not matching if I don't have an example of whatever hashes to the correct hash handy.

### DIFF
--- a/src/path.cpp
+++ b/src/path.cpp
@@ -1561,6 +1561,17 @@ bool adjacent_mappings(const Mapping& m1, const Mapping& m2) {
     return abs(m1.rank() - m2.rank()) == 1;
 }
 
+bool mapping_is_match(const Mapping& m) {
+    for(size_t i = 0; i < m.edit_size(); i++) {
+        if(!edit_is_match(m.edit(i))) {
+            // We have a non-match edit
+            return false;
+        }
+    }
+    // No non-match edits found
+    return true;
+}
+
 // assumes complete description of mapped node by the edits
 double divergence(const Mapping& m) {
     double from_length = 0;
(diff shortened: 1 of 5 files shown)


ekg:
The reason the men finding results might be different is that the graph is structured differently.

For the graph comparison stuff we should be able to make a comparator in vg that determines if two graphs are equivalent. Then we can retain the actual graph.

The hashes are nice due to being lightweight. I understand your frustration. If in doubt I look at the output of the previous version of vg. This has worked fine for me but I think we should do what is most comfortable for everyone.

ekg:
MEM finding results. Heh. Autocorrect.

adamnovak:
What I think might be best to do here would be to replace divide_node with something that can make multiple divisions at once, only copying the relevant part of the node's string each time. Then we could break up big nodes into nodes of the right size all at once, and keep the same division breakpoints that we would have otherwise had (thus not changing the graph structure, and ensuring that if you ask for 100-base nodes, you do end up with predominantly 100-base nodes and not 70-base nodes or something).

I will try to set that up when I get a chance.

adamnovak:
OK, I have changed over to an algorithm that makes exactly the same cuts as would have been made before, but doesn't copy the same node sequence over and over to do it.

If there are multiple mappings, they may have to get copied, but mapping handling when cutting nodes needs an overhaul anyway.

PR #282, opened 2016-03-30
Change over to the scoring parameters used in BWA-MEM

adamnovak:
The current scoring parameters of match = +2, mismatch = -2, gap open = -3, gap extend = -1 have a bit of a problem: if you can replace 2 mismatches with 2 indels and create an extra match, it's free (and the aligner seems to tend to do it).

These scoring parameters don't seem to have that problem.

Unfortunately, changing the default scoring parameters seems to have confused a large number of tests. Some of them just need to be adjusted to assume a perfect score is 1 point per base instead of 2 (or we could double the score values), but some of them (like the problem with vg mod) seem to indicate that these mapping parameters produce different mappings that aren't what the tests are expecting.

Do we actually want to change over to these parameters as the defaults, or should I just make them configurable and use them in my mapping runs?

### DIFF
--- a/src/gssw_aligner.cpp
+++ b/src/gssw_aligner.cpp
@@ -91,6 +91,7 @@ void GSSWAligner::align(Alignment& alignment, bool print_score_matrices) {
                                                     gap_open,
                                                     gap_extension);
 
+    //gssw_graph_print_score_matrices(graph, sequence.c_str(), sequence.size(), stderr);
 
     gssw_mapping_to_alignment(gm, alignment, print_score_matrices);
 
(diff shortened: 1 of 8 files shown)


ekg:
Makes sense to me. But what are the scores used in bwa mem? (Sorry I
haven't read the diff.)

ekg:
I guess we need to investigate the differences in the tests. When in doubt,
visualize. I'm not really free tomorrow but will try to pitch in if this
stays open.

adamnovak:
BWA-MEM is +1 for match, -4 for mismatch, -5 for gap open, and -1 for
extend.

ekg:
It looks like most of the errors are to be expected. We just need to check that the soft clipping is doing the right thing.

I would really like to move score to a float in `[0,1]`, where we rescale things so that the maximum allowed score is 1.

adamnovak:
Looks like the Ubuntu Travis setup decided to break itself. I will kick it, and if it doesn't fix, I might merge this anyway and try and fix that separately. It still works fine on OS X.

PR #288, opened 2016-04-04
Restore surject

adamnovak:
Surject is currently completely broken. This fixes it to actually include mappings in the surjected alignments, and to work on both strands.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -620,7 +620,7 @@ vector<Alignment> reverse_complement_alignments(const vector<Alignment>& alns, c
 }
 
 Alignment reverse_complement_alignment(const Alignment& aln,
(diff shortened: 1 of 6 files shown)


ekg:
I had started to hack at surject but hadn't really gotten far. The code
needs to be moved into the Mapper. Then it needs to be extended to the new
indexes.

ekg:
Thanks for catching the non functionality though. What are you using it for
now?

adamnovak:
We're surjecting reads so we can look at the BAMs on the genome browser and
see if the alignments we are getting out of VG make sense. It's easier than
messing about with GraphViz and jq to subset a GAM.

ekg:
(As you know) the problem is that the surjection isn't the same as the
original alignment. But we don't have the mechanism to efficiently extract
the alignments to a region yet...

PR #292, opened 2016-04-06
Update vg call

glennhickey:
This PR contains a bunch of fixes and hacks to vg call that were made while working to improve results on our pilot regions for the bakeoff paper.  

- Variety of bug fixes
- Filtering heuristics changed, and default values made saner
- Support for edge calls and indel calls added (in addition to SNPs)
- VCF export workflow added

This is still a work in progress...  Am looking at scaling now, as nothings been tested beyond the small pilot regions.


ekg:
Those small pilot regions are pretty big. 5mb is quite passable. Whole
chromosome is hopefully something we can avoid.

ekg:
Thank you _very_ much for this @glennhickey!

I'm looking forward to trying it out.

Issue #294, opened 2016-04-07
mapping only works on graphs with alphabet { A, T, G, C, N }


ekg:
In `vg/test/`:

```bash
cat tiny/tiny.fa | tr '[:upper:]' '[:lower:]' >tinylc.fa
vg construct -m 10 -r tinylc.fa >t.vg
vg index -x t.xg -g t.gcsa -k 16 t.vg
vg map -B 10 -s gaatataatagaactccagaaaatttccaagccttatttg -g t.gcsa -x t.xg -J
```

This yields something like the following:

```
vg: src/path.cpp:1525: std::pair<vg::Path, vg::Path> vg::cut_path(const vg::Path&, size_t): Assertion `seen >= offset' failed.
rev [1]    4796 abort (core dumped)  vg map -B 10 -s gaatataatagaactccagaaaatttccaagccttatttg -g t.gcsa -x t.xg -J
vg: src/path.cpp:1525: std::pair<vg::Path, vg::Path> vg::cut_path(const vg::Path&, size_t): Assertion `seen >= offset' failed.
rev [1]    4796 abort (core dumped)  vg map -B 10 -s gaatataatagaactccagaaaatttccaagccttatttg -g t.gcsa -x t.xg -J
```

The read is simulated from the graph, so we expect to find it perfectly. Note that debugging code I added makes the line number different than the current git master HEAD.

This is not a problem for the graph when it's constructed with upper case bases.

Two questions:

1. Where is the failure happening, and can we resolve it sensibly? (Is it in GSSW?)
2. For now, can we implement a check that the graph bases are at least upper case, and ideally all `{ A, T, G, C, N }`?

ekg:
So it's also odd that instead of not aligning, we literally got a completely empty alignment path. This isn't happening ever with upper-case sequences.

edawson:
It should be easy to implement a check, which should toss an error in construction. The issue comes up again if one aligns to a graph they did not construct though, so maybe we need to check before mapping.

I'll put a warning in `construct` and look into forcing uppercase letters perhaps?

ekg:
No need to force uppercase. Let's just warn.

adamnovak:
I made a change in the GSSWAligner that upper-cases reference and query sequences. If we don't want that, we might want to back out that particular change.

yhoogstrate:
[...]

PR #295, opened 2016-04-07
Added warning to vg.cpp if input fasta is lower case

edawson:
Ran into this on #294. This doesn't fix it per se, but it will toss a simple two line warning. It's one short lambda applied during construction.

### DIFF
--- a/src/vg.cpp
+++ b/src/vg.cpp
@@ -3006,7 +3006,8 @@ VG::VG(vcflib::VariantCallFile& variantCallFile,
     //
     
     std::function<bool(string)> all_upper = [](string s){
-        for (int i = 0; i < s.size(); i++){
+        //GO until [size() - 1 ] to avoid the newline char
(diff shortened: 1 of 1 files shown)


edawson:
This seems to catch newlines as well - Hack incoming.

edawson:
Travis says: `The job exceeded the maxmimum time limit for jobs, and has been terminated.`

Woah.

ekg:
I was just about to write that we should set up binary builds of the libs
and executables. And just fix them until we change commits. Maybe we can
post them to the web or github or something. This would save a lot of the
build time. All but about 8 minutes of it I'd guess.

edawson:
That's a good idea, given the lack of a decent C++ package manager. Maybe a "vgteam/vg_deps" repo with OS X and linux directories and a corresponding `make get-binary-deps` target or something.

ekg:
Could we use ipfs to host binaries of lib/ directory and protoc for each platform?

PR #299, opened 2016-04-11
Make surject work on the reverse strand

adamnovak:
Turns out surject wasn't working for the reverse strand. This should fix (and test) that.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -407,13 +407,14 @@ bam1_t* alignment_to_bam(const string& sam_header,
                          const Alignment& alignment,
                          const string& refseq,
                          const int32_t refpos,
+                         const bool refrev,
                          const string& cigar,
                          const string& mateseq,
                          const int32_t matepos,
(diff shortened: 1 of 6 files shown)


ekg:
I want to port surject to be backed by xg. Probably it should live in the mapper.

adamnovak:
I guess it should, if it's going to depend on doing mapping.

This sounds like an issue that should be addressed in the massive mapper
refactoring. It would be nice to sit down and draw out a UML or something
of how it all should really plug together.

Do we want to keep 3 mapping engines and 2 completely different sets of
indexes? Or should we just throw out RocksDB-based kmer indexing and
non-MEM mapping?

Should all the operations be possible on all the indexes?

Should xg and vg graphs support a common interface? If so, what should that
look like?

-Adam

ekg:
I'm all for paring away the non xg/gcsa based mapping cruft. Even kmer
based can go. The rocksdb index is useful potentially for sorting things
but could be refactored to do so.

PR #310, opened 2016-04-14
Add splitting option to vg filter

glennhickey:
Takes a bed file as input, and uses the xg index to find all node ranges in each region.  The gam is chunked out accordingly, where an alignment gets put into the chunk corresponding to a node range if it contains at least one node in that range.


ekg:
This looks great! Do you have any test cases we could use to pin this functionality down?

glennhickey:
I'll brush up on my jq and commit a test.  Results look sane on the examples I've tried here with bakeoff graphs.  

It's not hard to imagine cases where nodes get lost between chunks if they are not on the reference path.  The options here are to pad the input regions, manually or as an option.  Or perhaps its best to be consistent with vg find and use the context_extend option.  Any thoughts?

ekg:
Maybe extending the context by some amount will ensure we don't loose
anything. It is easier to remove any duplication afterwards.

glennhickey:
Makes sense, added as an option.

PR #314, opened 2016-04-22
Use glenn2vcf -l option to simplify chunked_call

adamnovak:
This should eliminate the need to munge the headers.

I unfortunately don't have a small test case handy; @glennhickey should probably look at this before merging.


glennhickey:
Thanks! Definitely cleaner to put that in glenn2vcf.  Will run it on chr21
just to be sure...

glennhickey:
I've folded this change into #320.

PR #316, opened 2016-04-23
Fix chunked_call

glennhickey:
fix bug where path range instead of node range was used to extract vg chunk from index.

will merge in once chromosome 1 finishes and looks sane.


glennhickey:
Will fix pileup test (which fails on travis) soon.

PR #317, opened 2016-04-24
Another chunked_call fix

glennhickey:
Fix bug in determining path offset of chunk graph.


adamnovak:
@glennhickey Do you know what broke the build here? The `vg pileup` test didn't seem to like this commit.

glennhickey:
The jq command I was using to test if two json files are equal stopped
working after I started splitting the protobuf into smaller pieces (tests
still passed locally though), but I didn't have time to look much being in
a rush to get the precision fda stuff done.

PR #324, opened 2016-05-04
Attempt to fix ssw build on Mac

adamnovak:
Why do we even have ssw and gssw in the same project? Can't gssw do whatever it is we want ssw for?

Anyway, this uses ssw with -rdynamic pulled out, which the Mac build on Travis didn't like.

### DIFF
--- a/src/vg.cpp
+++ b/src/vg.cpp
@@ -2815,7 +2815,7 @@ void VG::from_gfa(istream& in, bool showp) {
     map<string, sequence_elem>::iterator it;
     id_t curr_id = 1;
     map<string, id_t> id_names;
-    std::function<id_t(const string&)> get_add_id = [&](const string& name) {
+    std::function<id_t(const string&)> get_add_id = [&](const string& name) -> id_t { 
         if (is_number(name)) {
             return std::stol(name);
         } else {


ekg:
ggsw fills out score matrices and thus is not memory efficient when
aligning larger sequences. SSW does one fast pass only to find the best
scoring position and then returns to evaluate a banded alignment algorithm
in that area of the reference. It will be much better performing when we
have two sequences. I am using it during overlap resolution for assembly
graph bluntification.

adamnovak:
Would it make sense to make filling out those matrices optional? We would either need to put another branch in the inner loop, or double out all the functions (again).

adamnovak:
I'm going to merge this in, because as is it upgrades a completely failed build to just a failed test.

```
t/21_vg_filter.t ..... 
1..5
ok 1 - vg filter with no options preserves input.
Unable to find region in index: y:1-1
ok 2 - vg filter makes right number of chunks.
not ok 3 - vg filter left chunk has all left nodes
#          got: '2'
#     expected: '3'
ok 4 - vg filter right chunk has no left nodes
ok 5 - vg filter big chunk has everything
Failed 1/5 subtests 
```

PR #335, opened 2016-05-12
Add better guards to prevent rescuing off of unmapped reads

adamnovak:


adamnovak:
This will also fix Travis's Chrome setup and ssw on mac (again).

PR #351, opened 2016-05-18
Add some debugging information to caller's tsv output

glennhickey:
Throw in the original graph positions into the tsv.  This step 1 for speeding up the cross referencing of VCF records and input alignments..


adamnovak:
OK looks good. I will update glenn2vcf to read it.

PR #368, opened 2016-06-02
Base quality adjusted alignments and mapping quality

jeizenga:
Big picture changes:
1. Switched to my fork of GSSW that can compute base-quality adjusted alignments
2. Aligner can how use alignment scores to compute mapping qualities
3. Refactored Mapper to remove / avoid duplicated code in alignment subroutines
4. Aligner is now a persistent object inside Mapper instead of being created and destroyed for each alignment
5. Aligner refactored to be threadsafe
6. Changed paired mapping semantics to avoid confusion between independently mapped ends of fragment and pair mapped ends of a fragment
7. Added command line options for base quality adjusted alignment and mapping quality to vg map


ekg:
Looks amazing! I will read through but I also have a few high level
questions:

Do we incur any runtime performance penalty for the quality adjusted GSSW?
Do you have a summary describing how it works?

There seem like a lot of tricky things. For instance the quality of a large
number of bases may matter for indels.

+++++!!!

jeizenga:
There's shouldn't be a substantial time penalty for the quality adjusted GSSW. The only difference in the algorithm is that the score matrix used for a given base is a function of its base quality. It works by memoizing a score matrix for each base quality score at the beginning of a mapping run (that's why the aligner object is now persistent) and then just looking up the scores the same as you would before. All of the changes to GSSW take place in constructing the gssw_profile and doing the traceback, so the optimized part that fills out DP matrices is unchanged.

I also have a write-up of the math behind my methodology here: https://www.dropbox.com/sh/ra1uh6ghuwmdy82/AAClBVPTUByTgzAYjz5-FMQwa?dl=0

ekg:
@jeizenga thanks, I'll have a read through and then follow up in the source.

What needs to happen technically to merge this? I see there are conflicts. Have you got passing tests on your end?

jeizenga:
@ekg I do have passing tests. I'm working on the merge in my fork right now. Most of conflicts are pretty minor, but it does seem like we diverged a bit on the paired end mapping. 

First, just to be certain: I think what you're doing is pre-filtering your MEMs to only those that could be in a consistent pair, right? And then the algorithm is the same as it was before.

I changed the semantics a bit in my version. I really didn't like our strategy of promoting one pair for consistency because it made the multimaps non-comparable: one pair is a really a pair and the other are just independent mappings. That also makes it impossible to calculate mapping qualities because the alignment scores don't mean the same thing down the whole list of multimaps. What I did was replace the "promote consistent pair" option with a "report consistent pairs" option that applies the consistency filter down the whole list of multimaps so that all of the alignments that are reported are consistent pairs (even allowing one end mapping to be reported in multiple pairs if it's consistent with more than one of the mappings of the other end). I'm still not quite sure if that's going to interact poorly with your changes. I think maybe it would make the most sense to only do the pre-filtering for MEMs if the user calls it with the report consistent pairs option. Otherwise I would assume that they want all of the independent mappings.

I also refactored things a bit to straighten out the spaghetti bowl of alignment functions calling each other. In short, there's one function (align_multi_internal) that gets called by all of the public-facing functions and then calls the private-facing alignment algorithms. It handles all the gnarly logic that the different use cases require so that the alignment functions can be simpler because they don't all need to sort, deduplicate, etc. It looks like you redirected the paired end mapping around the align_multi function, so we'll have to figure out how to remerge the logic there. 

I think those are the main things.

ekg:
Makes sense to me.

The MEM-based paired end resolution I implemented was not tested
completely, but had been in discussion for several weeks. It is more of an
experiment to determine if we could improve performance by estimating
candidate mapping loci within the read pair constraint using Glenn's
extension to xg that enables path based inter-node distance estimation.

The alignment functions had become a mess due to a series of upgrades and
the maintenance of kmer based and MEM based alignment. I think we should
deprecate the kmer based method, and had thought to move to using
align_mem_multi for everything. That noted I will follow your lead on the
cleanup as it was necessary and it sounds like you're doing the most
sensible thing.

It seems like the way forward is to finish the merge and focus on improving
the performance of the mapping. To be competitive we should aim for around
1000 reads/CPU/s on a 32 machine. When I last worked on this I got to
around 500 reads/CPU/s. To properly test this I need to complete two
outstanding tasks: add pair simulation to sampler.cpp, and complete the
"check" functionality in the mapper interface. This would let us check if
we get the same alignment as before by re-mapping a GAM stream and
comparing the mapping results for the before and after. Where the pair was
simulated we would know the truth. Otherwise we could use this to compare
different parameterizations. I think this kind of testing harness is
essential if we want to optimize and debug the method completely.

jeizenga:
Okay, it looks like the problem here is that it's trying to build with your GSSW repo instead of mine. I had to change some of the function signatures in my fork so it's throwing compiler errors. Do you know how to make it point to my fork of GSSW (I'm still pretty bad at GitHub)? I thought I changed it in the .gitmodules file, but that seems to have not worked.

ekg:
We should update to gssw in vgteam. I can try to fix.

jeizenga:
Okay, I made a pull request over at gssw

ekg:
I just finished and debugged an implementation of paired end resolution that will require us to always provide an estimated fragment length and have embedded paths in the graph. I'm moving to merge this, and I've run into a nasty merge conflict. Working on that.

ekg:
In the current HEAD (post merge), we need to supply a fragment size whenever doing paired end resolution e.g. `-W 750`. This is required to check that the pairs are consistent. We should adjust away from the node space distance checking.

jeizenga:
I'm all for moving to a more sequence-based notion of distance rather than node IDs. It looks like Travis is still throwing the same compiler error though, as if it's using the old GSSW master. They're merged on both though, right? Any idea why that would be happening?

ekg:
I'm not sure. Doing a clean build now to check.

adamnovak:
Travis seems to be upset about "rs" not being a homebrew package (because it's a built-in command on OS X). I'm trying to fix it, but Travis is being weirdly slow, and some of the tests are intermittently failing.

adamnovak:
It ought to build happily on Travis now.

ekg:
It seems my build has succeeded. But it is also failing some tests I had not seen fail during local dev. Test 21 in the vg map tests.

ekg:
We have moved back to some blocking code in xg because I've run into another non threadsafe function in sdsl-lite. The locking is implemented in xg. It may diminish performance at high parallelism. The workaround is patching sdsl-lite.

jeizenga:
Hmm. The tests are still passing on my local build.

ekg:
Mine too. The failure for me is on the Sanger compute farm.

ekg:
@jeizenga I'm testing this out:

```
vg construct -r minigiab/q.fa -v minigiab/NA12878.chr22.tiny.giab.vcf.gz -m 64 >giab.vg           
vg index -x giab.xg -g giab.gcsa -k 11 giab.vg
vg map -x giab.xg -g giab.gcsa -f minigiab/NA12878.chr22.tiny.fq.gz -t 4 -J | jq .mapping_quality
```

These are the scores I get. They all seem pretty high. Is this expected for a small genome?

```
793
2147483647
2147483647
2147483647
165
337
2147483647
556
805
799
2147483647
2147483647
```

adamnovak:
2147483647 is just -1 as an unsigned int. I suspect that something is broken.

jeizenga:
No, that can happen. It's the signed int max, and it will return that value if the mapping algorithm can't find another multimap (+inf is the mathematically correct value in that case). However, another reasonable design choice would be to treat it like the next best multimap has score 0.

The other scores are 1.4 * 10 * (max score - next best score), so 800 only corresponds to a ~57 point difference in the alignment scores. That seems reasonable to me. If we want the scores lower, we would have to change the scale factor for the Phred units. Right now the formula is -10 * log(1 - P_err).

ekg:
@jeizenga have you compared the mapping qualities output by a few other programs and those now made by vg?

Ideally, we could compare the empirical mapping qualities from a simulation experiment with those produced by vg map.

> The other scores are 1.4 * 10 * (max score - next best score) / log(10), so 800 corresponds to a ~131 point difference in the alignment scores. That doesn't seem unreasonable to me. If we want the scores lower, we would have to change the scale factor for the Phred units. Right now the formula is -10 * log(1 - P_err).

What does the 1.4 constant mean here?

jeizenga:
@ekg I have not, but that seems reasonable. The 1.4 constant is a normalizing factor that that's based on the scores (1.4 is just the value for the default score parameters). It ensures that the scores can be interpreted probabilistically and also adjusts for the scale of the scores (i.e., if we multiplied all the score parameters by 10, it would shrink to .14).

ekg:
How do you derive this? If it is already in code I missed it so please
share a link to the source.

jeizenga:
@ekg There's a discussion of the equation in the math write up I linked near the top of the thread. Check out the section around equations (15)-(17) and then (23). The implementation is inside gssw:
https://github.com/ekg/gssw/blob/master/src/gssw.c#L2786

ekg:
Thanks! That plus the writeup has explained how this works. I'm doing what
I can to catch up my intuition.

PR #370, opened 2016-06-09
Remove "rs" packages on Travis

adamnovak:
Maybe now the mac build will work


adamnovak:
I'm leaving it on Ubuntu where needed, but removing it on OS X where it's built-in.

PR #371, opened 2016-06-10
Deconstruct (FINALLY)

edawson:
Deconstruct finally outputs VCF (with the exception of position). It works fine for single SNPs and short INDELs. For adjacent SNPs, it requires an accurate position relative to the start of the reference so that it can distinguish between adjacent positions (rather than calling them as multi allelic with duplicate calls).

I need to change about three lines of logic once there are accurate positions to ensure the output is perfectly sane. It's just a matter of using a vector or map to keep adjacent variants (which form a single superbubble) separate. Other than that, this should finally be close to done.


edawson:
Oh, this also includes a skeleton for an sv detector that is far from finished and a modification to the schema to allow string annotations to nodes and edges. I'd like to use these to store variant information.

ekg:
The position is available in the xg index. You can get all the path
relative positions for a given node. As the start and end of the VCF record
are on the reference path, this should work for all things we can project
into VCF.

I'm not sure I understand the problem with multi nucleotide polymorphisms
or adjacent SNPs. Wouldn't you just get all paths across the superbubble
which would contain both SNPs and the MNPs? This is perfectly valid in VCF
without any special casing.

If the bubble starts and ends on the reference then there is a direct
projection into VCF where the REF is the reference path through it,
position is the offset of the reference subpath within the reference path,
and each walk through the bubble is an ALT. Combinatorial expansion is
allowed. Don't stress about normalizing in the VCF projection as this mixes
concerns and makes it harder to get right. We can always use external
normalization routines. Also, if we have read evidence then we really do
want to think about whole traversals of each bubble rather than isolated
variants.

edawson:
Yep, found the code for getting positions and that bit is well.

I am not sure I understand the non-normalized VCF bit either. Here's the tiny test example, which has two adjacent SNPs:
[graphname.pdf](https://github.com/vgteam/vg/files/310207/graphname.pdf)

The set of possible paths through these nodes are:
1-> 2-> 4-> 6
1-> 2-> 5-> 6
1-> 3-> 4-> 6
1-> 3-> 5-> 6

2 & 4 are on the reference path (as are 1 and 6).

Does this mean the VCF record would simply show an MNP at this location, e.g.:
`x   9   .   GC   AT   `

or is there a way to represent multiple references in VCF, like so:
`x   9   .   G,C   A,T`

I'm assuming enumerating all paths would be wrong?:
`x   9   .   GC   AT`
`x   9   .   GA   GT`
`x   9   .   AC   AT`
`x   9   .   AT   AC`

I'd like to print this, as I think it is the most correct:
`x   9   .   G   A`
`x   10   .   C   T`

ekg:
You just represent one reference sequence and then all of the alternative
alleles of haplotypes across it. So in this case:

Pos ref alts
x 9 GC AT,GT,AC

This is the standard way of expressing this kind of thing. We can then
decompose it using normalization stuff that works on the VCF.

Changing the REF field would make the records invalid.

If we have a few phased haplotypes in the graph then it is _more_ correct
to list the haplotypes across successive SNPs than not. The decomposed view
is lossy when we don't have phased genotypes, and phased genotypes have
their own representational problems.

edawson:
Alright, I've fixed it to output adjacent SNPs correctly. INDELs are output as:
`x   9    .    A   .`
`x   20  .    .    A`
Positions are printed, and it looks pretty sane. I could perhaps rework the header.

I will work on making sure dagify and unroll work next (the graph will have to be reindexed).

ekg:
Do you have any tests of deconstruct or superbubble detection? We're not getting anything to work. I thought I'd leveled up to working superbubble detection in vg, but it seems I also forgot to add a test and so it's regressed.

edawson:
How broken is it? I'll push up a bash-tap file and post one here. `deconstruct -s` and `deconstruct -x tiny.xg tiny.vg` work on my end, as does the 1mb1kg test case.

edawson:
Just pushed up three simple tests that pass on my machine. Here's the vcf output of deconstruct from the 1mb1kg graph.
[1mb1kg.vcf.txt](https://github.com/vgteam/vg/files/311685/1mb1kg.vcf.txt)

edawson:
Why are the tests suddenly failing?

ekg:
@edawson isn't that test supposed to be over 1MB? I see only 1KB of results.

edawson:
I assumed the 1mb1kg test case was a megabase?

On Thursday, June 23, 2016, Erik Garrison <notifications@github.com> wrote:

> @edawson <https://github.com/edawson> isn't that test supposed to be over
> 1MB? I see only 1KB of results.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/vgteam/vg/pull/371#issuecomment-228004971>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ACeZUbbiuCeiV9GoblRiW2GPWBw5e1_lks5qOlljgaJpZM4IzXfh>
> .
>


-- 
Sent from my iPhone

ekg:
Right, so it's confusing that there is onl 1KB of results in the file you posted.

edawson:
Weird. Here it is again, looks more like a megabase.
 
[z.vcf.txt](https://github.com/vgteam/vg/files/329723/z.vcf.txt)

edawson:
I'm going to merge this today to get my fork back on track, unless there are objections.

PR #382, opened 2016-06-17
Support for clang-omp build

buske:
Explicit template required for casting bit to direct to ostream.

This is a re-PR of https://github.com/vgteam/vg/pull/380 from a branch within VG repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,10 +29,13 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export LD_LIBRARY_PATH=/usr/local/lib/; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export CFLAGS="-I/usr/local/include/"; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export LIBRARY_PATH=$LD_LIBRARY_PATH; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then which g++-4.9 || (brew unlink gcc && brew install gcc49); fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then mkdir -p ./bin; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ln -sf `which g++-4.9` ./bin/g++; fi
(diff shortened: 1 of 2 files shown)


adamnovak:
Can you get Travis to build with clang-omp?

buske:
I have the travis commit ready, but it will break the build until https://github.com/jltsiren/gcsa2/pull/23 is merged. Shall I add that to this PR?

buske:
The previous gcsa2 PR was merged, but changes on master since then break the build, so this PR is blocking on https://github.com/jltsiren/gcsa2/pull/24.

ekg:
Weirdly, it's failing with something like

```
/Users/travis/build/vgteam/vg/src/vg.hpp:18:10: fatal error: 'gcsa.h' file not found
#include "gcsa.h"
```

buske:
Yeah, not sure what's causing that. Also seeing
```
In file included from src/mapper.cpp:2:
In file included from /Users/travis/build/vgteam/vg/src/mapper.hpp:10:
/Users/travis/build/vgteam/vg/src/index.hpp:9:10: fatal error: 'rocksdb/db.h' file not found
#include "rocksdb/db.h"
         ^
1 error generated.
make: *** [obj/mapper.o] Error 1
```

The build succeeds on my system but I haven't tried a completely fresh install. Could be caused by parallelizing `make` and some underspecified targets. Will look into it. :)

adamnovak:
Could be a cache problem; all the include files are copied over to the
cache, and the Makefile probably isn't smart enough to know when they need
to be updated. I will clear out the Travis cache.

adamnovak:
Actually, that makes no sense as caches are per-compiler. I already cleared
the caches; maybe it will help anyway.

adamnovak:
OK, it now builds on Travis, but many of the tests fail.

buske:
Based on the travis logs, it still looks like it's failing due to missing header files, no? I'm going through the makefile and adding unstated dependencies now. I believe this should make the caching more robust.

adamnovak:
It looks like "clang-omp" is no longer a Homebrew package, and normal LLVM handles it.

ekg:
I wonder if it's time to explore this again. clang builds would be great!

ekg:
This is still interesting to me. Should we reinitiate? It may be easier than starting from this branch. Sorry @buske

buske:
No worries! :)

PR #385, opened 2016-06-18
Make Cactus/supbub a command line option.

adamnovak:
We probably want to turn it around and only use Cactus eventually, because supbub gets the wrong answers on dangling softclip-induced nodes. But this at least lets us try both, for now.


ekg:
I have seen some strange things out of both. We should continue to investigate and debug.

PR #386, opened 2016-06-18
Non redundant edits and translation overlay

ekg:


ekg:
Adding in stuff for translation overlay. We can now merge as many layers of translations as we want into one translation.

This will help for instance when applying multi-step editing algorithms to graphs where we need to retain the translation. One that comes to mind is in genotyping, when we might want to homogenize the reads with respect to the induced graph that they imply. Soft clipped segments often occur due to SNPs that aren't in the graph, and we can theoretically smooth these into the graph by realigning them after including the independently generated alignments.

PR #397, opened 2016-06-22
updates xg

ekg:
Just does the xg update. @adamnovak is there anything else we need to do on this end?


adamnovak:
There are some other associated changes needed; I have added them to this branch.

Note that this is going to break compatibility with old xg indexes, because the old ones have a DYNAMIC-based gpbwt bit vector in them, and these new ones are going to be reading and writing SDSL-based bitvectors at the same location.

adamnovak:
This is the failed test, for `vg mod` on OS X, only in the push tests:

```
not ok 20 - normalization works on a graph with an inversion
#          got: 'd9f8a778898a0e547fa0dc82a0827879'
#     expected: '84138fe8bd1015fb6b80278c2ed8f7c6'
```

I'm going to go ahead and merge this; I think we have an unrelated nondeterminism problem in `vg mod` normalization.

ekg:
It shouldn't be related. Normalization is not working too well these days.

ekg:
Thanks for the note about compatibility breakage.

Any thoughts about an efficient method to build gPBWT on non-DAGs?

adamnovak:
You could probably put together something that computes a partial result
for each node, and then loops around and updates the depending nodes only
when the nodes they depend on change. However, that sounds really tricky to
implement, and I'm not sure what the runtime bound would be (since you
might be stuck rebuilding the whole rank-select B vector for each node when
you update it).

PR #399, opened 2016-06-22
Fix Cactus memory leak, document option

adamnovak:
Cactus superbubble finding was leaking allocated node IDs when things were merged.

Also, the -C option for vg genotype wasn't in the help text.


glennhickey:
Good catch!

ekg:
I'd already added the -C option to the help text locally. However it
doesn't resolve some of the problems I have found on a micro test. I'll
move over to the "make genotyping good" issue.

Issue #396, opened 2016-06-21
Cactus Superbubble Stack Overflow


adamnovak:
Here's a graph that makes the recursive cactus-to-superbubbles code in VG run out of stack and segfault (after several thousand recursive calls to `cactus_recurse()`.

[break_cactus_nopaths.vg.zip](https://github.com/vgteam/vg/files/326777/break_cactus_nopaths.vg.zip)

Assuming all that recursion is correct, the recursive function might need to be replaced with a loop that keeps a stack on the heap, so as not to exhaust space on the stack.

adamnovak:
@glennhickey do you know what to do about this?

glennhickey:
I suspected this would happen, but am surprised it did on such a small
graph.  Will look into it.  It shouldn't be a big deal to refactor the code
to not be recursive, but I'm pretty sure the underlying issue here is
something else...

adamnovak:
This has been fixed.

PR #400, opened 2016-06-22
Make cactus bubbles more robust to dangling nodes

glennhickey:
Dangling nodes can cause cactus bubbles to be wrong, or crash (#396).  Try to fix this by choosing better source and sink (Still testing)

### DIFF
--- a/source_me.sh
+++ b/source_me.sh
@@ -8,8 +8,12 @@ export PATH=`pwd`/bin:`pwd`/scripts:$PATH
 export CC=$(which gcc)
 export CXX=$(which g++)
 
+#
+#  disable until file arguments work as in normal bash :(
+#
 # add bash autocompletion
-if test -n "$BASH_VERSION"
-then
-	 . ./autocomp.bash
-fi
+#if test -n "$BASH_VERSION"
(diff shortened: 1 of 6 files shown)


adamnovak:
Is there a real difference between `vg mod -a` and `vg stats -C`?

glennhickey:
mod -a returns the topology of the cactus graph (viewable in graphviz).
This is handy to help debug cactus-related stuff, and generally not super
obvious by looking at the ID sets of bubbles returned by stats -C.

glennhickey:
Current version should fix crash but is bandaid solution.  As discussed, source and sink should get picked up from the original (non-augmented graph).  Furthermore, the dangling nodes don't really help call anything, so they should get ignored by bubble construction (ideally removed entirely).

ekg:
Don't we want to work from the augmented graph? From there we can always
find our way back to the original graph.

glennhickey:
Yeah, the idea was just to use the translator to find the augmented source
and sink with the help of the original graph (where there's less ambiguity
in this department).  Still not sure if this is better than just using the
reference path (as done currently)

PR #403, opened 2016-06-23
Make genotyper good

ekg:


ekg:
@glennhickey @adamnovak note the refactoring here. I'd like to move the genotyper logic into the genotyper. This is a first step.

adamnovak:
This is probably a good idea. It does seem to be segfaulting in the OS X build, though.

adamnovak:
Closes #404; I can't reproduce that segfault anymore with these fixes.

Issue #423, opened 2016-07-12
Make superbubbles bigger when there's low entropy/evidence


adamnovak:
If your site is low-entropy (i.e. all repeats), or if your reads don't give you any traction towards one allele or the other, you ought to expand your site (by collecting sibling superbubbles) until you have something big enough to really genotype.

You know your site is big enough when you have enough entropy/uniqueness at the ends, or when it's clear that reads really do argue in favor of one allele or the other, even when aligning in the context of the surrounding graph.

@ekg could you elaborate on how the entropy measurement stuff in Freebayes works?

ekg:
It's pretty simplistic. When we see repeated sequence, we progressively extend the window right (keeping the same start position) until we find a window with entropy greater than some fixed number of shannons per base.

adamnovak:
I think this is outdated now, since call and genotype have been reworked a few times over and superbubbles are gone.

PR #424, opened 2016-07-12
New xg and min context support

adamnovak:
This pulls in the new xg that has the fixed context expansion.

It also requires recurrent support for nodes being used as context around a site undergoing indel realignment. This was meant to try and clean up at the genotyping level some of the extra indel copies created by not having the xg fix mentioned above. I'm not sure it really helps.

What we really want to do is #423 and have some kind of expansion of the site to include all the stuff in the surrounding context that makes it ambiguous.


adamnovak:
Isn't Travis going to test this?

PR #432, opened 2016-07-29
Merge Glenn's vg call improvements and other fixes being used for assembly realignment evaluation

adamnovak:
In particular, this PR bumps the max iterations of softclip extensions from 3 to 10 (because it was too low and was producing sofclips in the middle of banded alignments). It also add some more stats to vg stats to count up the number of edits of different types in a GAM file.


adamnovak:
@glennhickey did you change the way vg pileup works? It looks like the test that compares vg pileup's output to the expected result is failing because we're actually getting a semantically different pileup.

glennhickey:
Looks like ce163b9e2ca20cb8274aa5ba270e9c94f0f8d2d0 broke the test.  Will
fix soon.  I really need to look through the call branch and glenn2vcf to
make sure no crap I added while trying vainly to improve it gets put
upstream.

glennhickey:
I fixed this in
https://github.com/glennhickey/vg/commit/be71d6067dd497bc52880ddde9d75302bd04132a

adamnovak:
OK, I'm going to merge this in, if there aren't any objections.

adamnovak:
Actually, I'm going to hold off because the new `-S` option can induce crashes...

ekg:
On vg call?

adamnovak:
No on `vg genotype`.

PR #433, opened 2016-08-02
MEM threader

ekg:
Align using the MEMs as direct exact matches, and fix up the intervening unaligned sequences using dynamic programming only on the small unaligned bits.

We first use a greedy clustering process to organize the SMEMs  into a set of (potentially nested) SMEM clusters. We ensure that these clusters approximately match the position of the SMEMs in the read and their positions in any of the embedded reference paths in the graph.

We then order them by length, and for each one generate an alignment. This allows us to produce multi-mappings. The alignment is taken to be the perfect matches and the intervening sequences are considered to be insertions or soft clips. A separate function (patch_alignment) patches the alignment so that the intervening sequences are realigned locally to the graph which is implied to be near the ends of the flanking SMEM(s).


jeizenga:
Just starting to look through this, but an initial observation: I see that you switched `map<int64_t, gssw_node*> nodes` from a local variable in `Aligner::create_gssw_graph` to a member variable. I don't think that's going to work. When I added in the base quality adjusted alignments, I refactored the `Aligner` to be a persistent object in the `Mapper`, whereas prior to that one was created every time we did an alignment. Since one `Aligner` is shared by every alignment the `Mapper` is doing, the member variables need to be threadsafe (that's why I turned some of the member variables into local variables in the first place). Anyway, I'm willing to bet that adding in gssw_nodes from multiple subgraphs will mess up what you're trying to do here.

ekg:
@jeizenga I'll look into switching the nodes back. I need to check my commit log to remember why I made that change. I think I'll need another solution.

ekg:
I found a problem with my match-walking.

```
on position 106678+5
C vs C i+1= 29
 on 30
C == C
on position 106678+6
A vs A i+1= 30
 on 31
A == A
on position 106678+7
A vs A i+1= 31
 on 32
A == A
on position 106678+8
we are going into a new node
2
next : 106679+0 G (looking for G)
 on 33
G == G
on position 106679+0
we are going into a new node
1
next : 106680+0 G (looking for T)
vg: src/mapper.cpp:1852: vg::Alignment vg::Mapper::walk_match(const string&, vg::pos_t): Assertion `found_next' failed.
```

I had a hunch this would be a problem, but then forgot abuot it.

The match walking search is greedy, and so can fail when we choose to take the wrong path.

Here from node 106678 we can go to `G` two ways. However, if we travel through 106679 we add an extra `G`. We should go over the deletion.

image

This needs to be turned into a backtracking DFS on the graph. Where we encounter a non-match when we expect a match, we should go back to the last branch point and continue from alternative branches. We'll need to mark visited branches.

ekg:
@richarddurbin I think this demonstrates why unidirectional graphs for the forward and reverse complement of the graph can be easier to work with. We could implement a normalization in which we only put branches in the graph when the first base on each branch is different.

ekg:
@jeizenga I believe I made this change to support the traceback initialization, which has been mostly reverted. I don't think it's necessary to maintain it. I'll think about another way to implement it.

richarddurbin:
Yes. There are still issues I think about when paths rejoin after a bubble, and (related) the forward to reverse-complement mapping is not one-to-one. 

Sent from my iPhone

> On 4 Aug 2016, at 4:49 pm, Erik Garrison <notifications@github.com> wrote:
> 
> @richarddurbin I think this demonstrates why unidirectional graphs for the forward and reverse complement of the graph can be easier to work with. We could implement a normalization in which we only put branches in the graph when the first base on each branch is different.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 



-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

ekg:
For the moment, I will change the exact path walking to fork whenever we have two allowed candidates, and to stop walking paths when a mismatch with the SMEM is encountered. This will guarantee that we emit at least the path that the SMEM referred to. However, I don't know what should be done when we get two or more walked paths for the same SMEM. I guess one could be picked without loss.

jeizenga:
@ekg 
Some questions / comments that might make our lives easier down the line:

- The way the segment joining is factored seems convoluted to me. We turn all the MEMs into alignments, then we join them into one alignment, then we detect the intervening segments as insertions, then we split it back into smaller alignments and align the intervening segments, and then we join the whole thing back up into a single alignment again. Am I following this correctly? Basically, I'm wondering what the point of the first merging step is if we're just going to split the alignments all back up again. Wouldn't it be simpler to leave the MEMs as a vector of alignments or something and then do one merging step at the end? (Full disclosure: I will probably have to refactor it this way later for the algorithm I'm working on anyway, so my motivation for recommending this is partly selfish).
- I see that you are aligning the intervening segments with the overload of Graph::align that doesn't use the Mapper's Aligner member variable. This actually risks making the scores non-comparable if the Mapper is not using the default scoring parameters. Using the aligner will also make it easier to work in base quality adjusted alignments later. I also think that it makes sense to push the MEM-to-alignment conversion into the Aligner so that it can handle scoring consistently.
- I've noticed that the alignment concatenation functions in alignment.cpp don't merge all aspects of the alignments correctly. For instance, they don't bother copying base quality information. I also think it makes more sense to add the scores together inside the merging functions rather than in an ad hoc manner outside of them.
- On lines 2121 and 2138, I think it should be (edit.from_length() - 1) so you don't pay both a gap extend penalty and a gap open penalty on the first base. Not positive though--check me on that.
- When you're trimming the nodes to remove the part of the sequence that's covered by the MEM, you remove the edges into that node. Will it ever happen that the subgraph includes a cycle back onto the node we're trimming? If so, I think this step might cause problems.
- If I'm understanding the MEM clustering step, you check whether a cluster starting with each possible position would form a new cluster. I think this is probably wasteful. Since you always greedily choose the longest MEM covering a position, you will deterministically choose the same MEM at each of the start positions it covers, right? Won't this just make copies of the same cluster with a slightly truncated initial MEM? I think it would be better to start the next attempt to make a cluster at the end of the first MEM in the current cluster. 
- Is there any filter to ensure that a cluster of MEMs is in the correct order relative to the read? It might be in there, but I couldn't find it. If they are not, the banded alignment will do really wonky things in the intervening segments.

I also have some questions / comments looking forward to merging our development. 
- It seems like the exchange format between our modules should probably be Alignment objects--that way you can just pop in the banded alignment in place of the Graph::align that you're currently using. Sound good to you?
- The way that I've currently implemented the banded global aligner, it only works on a DAG and it will produce the optimal alignment from any source node to any sink node. This handles cases where the mismatch that breaks the MEM happens at a SNP site or something so that the alignment can take disjoint paths between the MEMs. Will that work with the subgraph extraction algorithm you've implemented?
- I'm thinking about implementing an automatic band-width detection algorithm where the aligner will automatically detect the tightest band would allow paths through every node to reach a sink. Do you think that would work with the subgraphs you're producing?

ekg:
@jeizenga I'll respond to what I can in your comments. Where possible I'm already taking action to implement the suggestion.

----

> The way the segment joining is factored seems convoluted to me. We turn all the MEMs into alignments, then we join them into one alignment, then we detect the intervening segments as insertions, then we split it back into smaller alignments and align the intervening segments, and then we join the whole thing back up into a single alignment again. Am I following this correctly? 

This is correct; although rather than splitting it's more like we're copying (in `Mapper::patch_alignment`) the alignment from one to another. If we find an unaligned segment, we try to align it in the region of the graph that apparently lies between the end points of the surrounding SMEMs (of which there may be 1 or 2 depending on where we are in the read).

> Basically, I'm wondering what the point of the first merging step is if we're just going to split the alignments all back up again. Wouldn't it be simpler to leave the MEMs as a vector of alignments or something and then do one merging step at the end? (Full disclosure: I will probably have to refactor it this way later for the algorithm I'm working on anyway, so my motivation for recommending this is partly selfish).

I won't argue it's the cleanest way, but it helped me during development to generate a valid Alignment object, as this is something that I can inspect with using the tools in vg. Also, I had the idea that we might want to just work with the exact-match only Alignment objects in some other context--- like maybe we don't even need to do the refinement in `patch_alignment`. In theory, we could refine the alignments in a bundle, when we have all the reads that localize somewhere. However, this might break mapping quality and quality adjusted alignment, which are important downstream. In exchange, alignment could be twice as fast. (Right now, alignment patching is taking up 56% of `vg map` runtime during MEM-threading.)

I'll see what I can do to match your side as I begin integrating #444.

----

> I see that you are aligning the intervening segments with the overload of Graph::align that doesn't use the Mapper's Aligner member variable. This actually risks making the scores non-comparable if the Mapper is not using the default scoring parameters. 

Thanks, I should fix that.

> Using the aligner will also make it easier to work in base quality adjusted alignments later. I also think that it makes sense to push the MEM-to-alignment conversion into the Aligner so that it can handle scoring consistently.

I will hopefully understand how to do this when I've dug into the global banded aligner.

----

> I've noticed that the alignment concatenation functions in alignment.cpp don't merge all aspects of the alignments correctly. For instance, they don't bother copying base quality information. I also think it makes more sense to add the scores together inside the merging functions rather than in an ad hoc manner outside of them.

So that's two things to fix: 1) correctly handle the base quality information if it exists, 2) add the scores inside of merging ops.

-----

> On lines 2121 and 2138, I think it should be (edit.from_length() - 1) so you don't pay both a gap extend penalty and a gap open penalty on the first base. Not positive though--check me on that.

I actually don't know what to do here. I thought we payed both an open and extend penalty on the first base, but I may be confused.

----

> When you're trimming the nodes to remove the part of the sequence that's covered by the MEM, you remove the edges into that node. Will it ever happen that the subgraph includes a cycle back onto the node we're trimming? If so, I think this step might cause problems.

That is a problem. We could resolve it in various ways, but I can't think of anything nice immediately.

----

> If I'm understanding the MEM clustering step, you check whether a cluster starting with each possible position would form a new cluster. I think this is probably wasteful. Since you always greedily choose the longest MEM covering a position, you will deterministically choose the same MEM at each of the start positions it covers, right? Won't this just make copies of the same cluster with a slightly truncated initial MEM? I think it would be better to start the next attempt to make a cluster at the end of the first MEM in the current cluster.

The MEM-clustering runs across unique start positions. For each unique starting reference position of any MEM (on any sequence), we generate the cluster that would be obtained by stepping to the next exact match that starts at least after the end position of this exact match. The greedy heuristic is that we take the longest exact match at the position as the basis of the cluster, which might miss shorter ones that start at the same position.

I see how this could be a problem. I can generate a cluster starting from each one of the MEMs at the position instead. A test case would make this easier, so I'll try to come up with one.

--- 

> Is there any filter to ensure that a cluster of MEMs is in the correct order relative to the read? It might be in there, but I couldn't find it. If they are not, the banded alignment will do really wonky things in the intervening segments.

There is a filter on this in the cluster generation:

```
if (!cluster.empty() && cluster.back().end > mem->begin
    || in_cluster.count(mem)) {                        
    ++x; // step to the next position                  
...
```

This ensures that we step along and look for the next MEM if the last member of the cluster ends after the next candidate start _in the read_. At the other side of the loop, we are taking the next MEM whose reference starting position is after ours.

I am not being rigorous about this, but it appears to handle most cases. I can think of a few which it doesn't handle, that are related to the overlapping MEM problem with the greedy selection heuristic above.

As I can't prove that this produces MEMs in the correct order, I guess a check at the end is required before sending them to the banded global aligner.

---

> It seems like the exchange format between our modules should probably be Alignment objects--that way you can just pop in the banded alignment in place of the Graph::align that you're currently using. Sound good to you?

Yes, this is great.

---

> The way that I've currently implemented the banded global aligner, it only works on a DAG and it will produce the optimal alignment from any source node to any sink node. This handles cases where the mismatch that breaks the MEM happens at a SNP site or something so that the alignment can take disjoint paths between the MEMs. Will that work with the subgraph extraction algorithm you've implemented?

I'm not completely sure. I need to see how the interface works before this is clear. That said, we can get to a DAG (or check if we have a DAG) pretty easily, so there should be a way to graft things together.

---

> I'm thinking about implementing an automatic band-width detection algorithm where the aligner will automatically detect the tightest band would allow paths through every node to reach a sink. Do you think that would work with the subgraphs you're producing?

Very interesting. I think we can make that work, but maybe it would be most useful in the context of aligning the subgraph touched by the MEM cluster, rather than the subgraphs in the "gaps" between the exact matches.

---

OK, I'm continuing on! Continue responding inline. I'll work on what I can from here.

ekg:
This is mostly where I want it to be in terms of first implementation and not breaking all the existing tests and code. There are some outstanding issues though that comprise a few hours more work:

- I haven't tested the quality adjusted alignments. I will need to add the quality scores from the alignments into the alignment patches.
- I am not yet using the alignment scorer.
- I haven't tried to integrate with the banded aligner.
- I still haven't resolved the issue of overlapping MEMs. This will be an issue where there are multiple candidate alignments that overlap. (MEM threads that are "woven" together.)

ekg:
So as to not let development fray, I'm going to merge this in.

The issues above still pertain to the mem-threading though.

PR #434, opened 2016-08-02
Linearization algorithm based on max flow/min cut

mzueva:
Hello!
This is implementation of vg sorting using max flow/min cut algorithm from the paper.
Sorting now is implemented as a separate tool "vg sort".
For first implementation I've used a simple Ford–Fulkerson algorithm for finding min cut in the graph and a matrix representation of a graph for it. Surely we'll change it, when we figure out which max flow algorithm implementation is the best for this task.
I have some questions on the sorting algorithm, it would be great to discuss them and make some changes to the code:
1. As we have discussed earlier, we need to determine the approach for assigning weights to edges (joins). Now it is implemented as a number of paths, that go through the edge (path goes through edge, if both adjacent nodes are mapped to that path). This approach is relatively fast, but it needs several paths in vg file. This was considered as “OK for now”. How do you see the further development of this approach?
2. We have implemented the backbone determining in the following way:
   a. for the whole graph we use a path from the “-ref” parameter as a backbone;
   b. we use a greedy algorithm for recursive calls for in- and out-growths; we start from the backbone and go in depth always selecting the edge with the maximum weight.
   This is not always optimal, but it is fast. What would you say? Can you suggest any other approach?
3. We have used cactus and brca1 graphs for testing. Can we use some extra data for this to be more sure that testing is sufficient? 
4. For comparing various approaches we probably need to add some graph metrics – like weighted feedback, cutwidth, etc. Do you have any ideas for such metrics?

For now there is only one unit test, it is based on example, described in the paper. It checks, that order of the nodes after sorting the same as expected. I think we need more small graphs with some "corner cases" to make testing sufficient.

Looking forward to your answers and review!

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -8538,6 +8538,83 @@ int main_test(int argc, char** argv){
     return vg::unittest::run_unit_tests(argc, argv);
 }
 
+void help_sort(char** argv){
+    cerr << "usage: " << argv[0] << " sort [options] -i input_file -r reference_name > sorted.vg " << endl
+         << "options: " << endl
+         << "           -g, --gfa           input in GFA format" << endl
+         << "           -i, --in            input file" << endl
+         << "           -r, --ref            reference name" << endl
+         << endl;
+}
+
+int main_sort(int argc, char *argv[]) {
+
+    //default input format is vg
+    bool gfa_input = false;
+    string file_name = "";
+    string reference_name = "";
+    
+    int c;
+    while (true) {
+        static struct option long_options[] =
+            {
+                {"gfa", no_argument, 0, 'g'},
+                {"in", required_argument, 0, 'i'},
+                {"ref", required_argument, 0, 'r'},
+                {0, 0, 0, 0}
+            };
+
+        int option_index = 0;
+        c = getopt_long (argc, argv, "i:r:g",
+                         long_options, &option_index);
+
+        /* Detect the end of the options. */
+        if (c == -1)
+            break;
+
+        switch (c)
+        {
+        case 'g':
+            gfa_input = true;
+            break;
+        case 'r':
+            reference_name = optarg;
+            break;
+        case 'i':
+            file_name = optarg;
+            break;
+        case 'h':
+        case '?':
+            /* getopt_long already printed an error message. */
+            help_sort(argv);
+            exit(1);
+            break;
+        default:
+            abort ();
+        }
+    }
+  
+    if (reference_name.empty() || file_name.empty()) {
+        help_sort(argv);
+        exit(1);
+    }
+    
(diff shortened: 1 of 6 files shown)


src/main.cpp
```diff
@@ -8538,6 +8538,83 @@ int main_test(int argc, char** argv){
     return vg::unittest::run_unit_tests(argc, argv);
 }
 
+void help_sort(char** argv){
+    cerr << "usage: " << argv[0] << " sort [options] -i input_file -r reference_name > sorted.vg " << endl
```
yhoogstrate: Writing help functions to stderr is bad practice. One of the few sane use case for this is when an invalid argument is passed through, e.g. 'vg sort --awwfdfsdvs'.

src/main.cpp
```diff
@@ -8538,6 +8538,83 @@ int main_test(int argc, char** argv){
...
+            reference_name = optarg;
+            break;
+        case 'i':
+            file_name = optarg;
+            break;
+        case 'h':
+        case '?':
+            /* getopt_long already printed an error message. */
+            help_sort(argv);
+            exit(1);
```
yhoogstrate: this should be `exit(0)`, the exit(1) at line 8599 is good though because there an error occurred

src/main.cpp
```diff
@@ -8538,6 +8538,83 @@ int main_test(int argc, char** argv){
...
+            break;
+        default:
+            abort ();
+        }
+    }
+  
+    if (reference_name.empty() || file_name.empty()) {
+        help_sort(argv);
+        exit(1);
```
yhoogstrate: How about `throw std::invalid_argument( "err msg missing file name" );` ?

src/main.cpp
```diff
@@ -8538,6 +8538,83 @@ int main_test(int argc, char** argv){
...
+    }
+    
+    ifstream in;
+    in.open(file_name.c_str());        
+    VG* graph = nullptr;
```
yhoogstrate: There is no obvious need to use pointers here. Try to avoid this whenever possible.

src/main.cpp
adamnovak: All the vg commands return 1 when asked for help. It looks like most other tools do return 0, though.

ekg:
Is everything from the paper now in the PR? Anything we can help with?

mzueva:
@ekg Hello, Erik! I'm sorry. but this pull request is outdated, since we moved the development into another branch (https://github.com/mzueva/vg/tree/f_sort_hash/). I'll clean up the code and publish a new pull request shortly.

mzueva:
PR with new implementation of graph sorting: https://github.com/vgteam/vg/pull/537

PR #444, opened 2016-08-12
Banded global alignment

jeizenga:
I added an object that can perform banded global alignment of a sequence against a graph. It also has an interface through the Aligner object for consistent scoring. There's a pretty exhaustive set of unit tests too.


src/banded_global_aligner.cpp
```diff
@@ -0,0 +1,1698 @@
+//
+//  banded_global_aligner.cpp
+//  
+//
+//  Created by Jordan Eizenga on 7/25/16.
```
adamnovak: But why? What's it for?

src/banded_global_aligner.hpp
```diff
@@ -0,0 +1,148 @@
...
+#include <list>
+#include "vg.pb.h"
+
+
+using namespace std;
+
+namespace vg {
+    enum matrix_t {Match, InsertCol, InsertRow};
+    
+    class BandedAlignmentBuilder {
```
adamnovak: What is this for?

src/banded_global_aligner.hpp
```diff
@@ -0,0 +1,148 @@
...
+        bool matching;
+        Node* current_node;
+        int64_t edit_length;
+        int64_t edit_read_end_idx;
+        
+        void finish_current_edit();
+        void finish_current_node();
+        
+    public:
+        BandedAlignmentBuilder(Alignment& alignment);
```
adamnovak: This modifies the alignment in place, I think. Right?

src/unittest/banded_global_aligner.cpp
```diff
@@ -0,0 +1,1406 @@
...
+#include "gssw_aligner.hpp"
+#include "vg.hpp"
+#include "banded_global_aligner.hpp"
+
+using namespace google::protobuf;
+
+namespace vg {
+    namespace unittest {
+        
+        TEST_CASE( "Banded global alignment works when read matches exactly",
```
adamnovak: I think it's better to avoid "works" as the test name, because it's not that useful an indication of what broke when it fails.

src/unittest/banded_global_aligner.cpp
```diff
@@ -0,0 +1,1406 @@
...
+                                                                                   graph.graph,
+                                                                                   band_width);
+            
+            
+            banded_aligner.align(aligner.score_matrix, aligner.nt_table, aligner.gap_open,
+                                 aligner.gap_extension);
+            
+            const Path& path = aln.path();
+            
+            // is a global alignment
```
adamnovak: Maybe all these different categories of checks should be their own SECTIONs so if they fail you get better output?

src/unittest/banded_global_aligner.cpp
```diff
@@ -0,0 +1,1406 @@
...
+            REQUIRE(path.mapping(1).edit(0).from_length() == 2);
+            REQUIRE(path.mapping(1).edit(0).to_length() == 2);
+            REQUIRE(path.mapping(1).edit(0).sequence().empty());
+            
+            REQUIRE(path.mapping(2).edit(0).from_length() == 6);
+            REQUIRE(path.mapping(2).edit(0).to_length() == 6);
+            REQUIRE(path.mapping(2).edit(0).sequence().empty());
+        }
+        
+        TEST_CASE( "Banded global alignment works when a there is a multi-base lead deletion",
```
adamnovak: You mean like the alignment starts with a deletion or something?

adamnovak:
OK @jeizenga I have put comments all over this that maybe you should read and maybe fix/improve? But I am going to merge it in anyway because they can all be improved later.

src/banded_global_aligner.cpp
ekg: @adamnovak I think the objective is to enable a faster alignment mode when we have a good idea of the overall structure of the alignments.

I guess you mean we should add commentary into the source. That would be helpful.

adamnovak: Yes exactly. Git can keep track of who made things when, but the why isn't
captured.

Issue #127, opened 2015-10-20
Better explain coordinate system in vg.proto


glennhickey:
From existing docs it's not clear what a Position's offset means in the case of a reverse Mapping.  Is it always forward relative or strand relative? In other words, which of the following would be used to represent a mapping over the reverse of a node of length 10 (from right-to-left of node)?  

"mapping": [ {
"is_reverse": true,
          "position": {
            "node_id": 1000,
            "offset": 0
          }]

OR 

"mapping": [ {
"is_reverse": true,
          "position": {
            "node_id": 1000,
            "offset": 9
          }]
          
I think it's the second, but would be nice to confirm, and add details to vg.proto...

ekg:
I think it's the second. I can't remember how this is implemented though.

glennhickey:
Yeah looking like the second.  vg validate will show me the way.

ekg:
Uh oh vg validate isn't that smart yet...

glennhickey:
Doh!

But I've been distracted by another problem that I didn't notice before vg
validate:  Paths with cycles aren't supported.  When importing them,
duplicated mappings are silently deleted, then validate complains about
missing edges in the resulting path.  Though spelled out in a comment in
path.cpp, this took me way too long to find.

// problem, won't allow us to keep multiple identical mapping to the same
node,
// as will happen with looping paths

For some reason, I was under the impression cycle support had been
added.... Do you think this is difficult to add?  Many of our graphs have
paths with cycles...

ekg:
I thought I had fixed this, but I guess not. Sadness.

It should be easy enough to fix. xg supports it.

ekg:
I wonder if this is causing other issues.

glennhickey:
OK, I added an issue for it.

Luckily, we haven't been using vg paths for anything so far (though that
will change), as far as I'm aware.  The only exception was the surjected
bams I was using for the variant comparison, and this issue my be related
to my nonsense results for that.

glennhickey:
No idea, but it's encouraging at least that vg validate picked this up
(however indirectly) right away.

ekg:
There are two ways to represent a reverse alignment. You can represent the path in "forward" orientation and then set the alignment `is_reverse` flag. Or, you can have a path in which the mappings marked with the `is_reverse` flag.

I prefer the second way. It seems more conceptually clear. However, the first is simpler for DAG-like graphs. This needs to be resolved and implemented or there will be a lot of confusion. Suggestions?

[...]

PR #446, opened 2016-08-12
Add alignment ambiguity trimming to vg filter

adamnovak:
This adds a `-D` or `--defray-ends` option to `vg filter` that should let you find and clip off instances where the end of a read has been arbitrarily assigned to one path out of many sequence-identical paths available in the graph.

This commit also addresses #127 by updating the comments in `vg.proto`, and fixing `reverse_complement_alignment()` to use offsets just like `vg map`'s output has been using them. This might confuse other uses of `reverse_complement_alignment()`, so we ought to keep an eye on the test results.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ endif
 STATIC_FLAGS=-static -static-libstdc++ -static-libgcc
 
 # These are put into libvg.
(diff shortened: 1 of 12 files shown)


ekg:
Should we bring this in?

adamnovak:
I think @glennhickey is rebasing it for me.

We can bring it in; Benedict still suspects that it's broken somehow, but
I'm going to run another negative control and hopefully put that to bed.

-Adam

glennhickey:
Yeah, I have a rebased branch I'm working on now with this in it.  Will do
PR this week hopefully final vg call changes.

PR #452, opened 2016-08-23
Adds SV construction and basis for SV detection

edawson:


ekg:
What are the xg bugs you had to hack around?

ekg:
You're referring to the problem where the id space needs to be contiguous. We should fix this. I guess it's a bug in the documentation at this point, so I can't disagree.

edawson:
Brief description of what's in this PR:
   1. SV handling during construction for variants containing an SVTYPE of {INV, INS, DEL} in the corresponding info field in the vcf. It is preferable to use flat alternates when doing this.
   2. sift (in progress) - There was a namespace collision with filter, and scrub didn't describe the functionality well, so the command has finally settled on the name `sift`. This command is used to select or filter Alignments based on selected properties (SV signatures, depth, etc). It is still a work in progress, and test coverage is currently low.
   3. srpe (in progress) - split read / paired-end SV detection from GAM. This takes a graph, a GAM, and a GAM index. Alignments which are flagged as triggering any SV filters (e.g. discordant fragment length, inverted pairs, etc) are collected. The subgraph around a given alignment is pulled out, a candidate SV allele is generated, and the alignments are remapped. The allele with the best total mapping quality is kept and the allele Path is added to the original graph using `vg::edit`. It will eventually also call variants at this stage and variant supports will be based on read remappings.
   4. homogenize (in progress) - SV detection and graph correction for augmented graphs. SVs in graph augmented with short reads show up as tips. The homogenize process takes all tips that are mappable and remaps them within the graph. The edges created are then introduced and the genotyping process is repeated. This sounds similar to the above because it is - it just takes place closer to the genotyping step. It's a bit slower, and probably less correct.

Makefile
```diff
@@ -350,6 +359,16 @@ $(SUBCOMMAND_OBJ_DIR)/construct.o: $(SUBCOMMAND_SRC_DIR)/construct.cpp $(SUBCOMM
 
 $(SUBCOMMAND_OBJ_DIR)/simplify.o: $(SUBCOMMAND_SRC_DIR)/simplify.cpp $(SUBCOMMAND_SRC_DIR)/subcommand.hpp $(SRC_DIR)/vg.hpp $(SRC_DIR)/progressive.hpp $(SRC_DIR)/utility.hpp $(DEPS)
 	 +$(CXX) $(CXXFLAGS) -c -o $@ $< $(LD_INCLUDE_FLAGS) $(LD_LIB_FLAGS) $(ROCKSDB_LDFLAGS)
+	
+$(SUBCOMMAND_OBJ_DIR)/homogenize_main.o: $(SUBCOMMAND_SRC_DIR)/homogenize_main.cpp $(OBJ_DIR)/homogenizer.o $(OBJ_DIR)/filter.o $(OBJ_DIR)/mapper.o $(OBJ_DIR)/bubbles.o $(OBJ_DIR)/vg.o
```
adamnovak: Could you add the _main suffix to the other subcommands? Or take it off of these? I don't think there's a risk of name collision with other .o files that aren't subcommands, because they go in a different output directory under obj.

src/constructor.hpp
```diff
@@ -246,7 +260,7 @@ class Constructor : public Progressive {
      * one FASTA file. Reference and VCF vectors may not contain nulls.
      */
     void construct_graph(const vector<FastaReference*>& references, const vector<vcflib::VariantCallFile*>& variant_files,
-        function<void(Graph&)> callback);
+        function<void(Graph&)> callback, bool do_svs, const vector<FastaReference*>& insertions);
```
adamnovak: I think do_svs maybe should be a configuration flag on Constructor.

Why are the insertion FASTAs separate from the reference FASTAs?

Also, the callback probably should stay as the last parameter, so you don't have a long inline lambda and then more arguments to the function it gets passed to after it.

src/homogenizer.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#ifndef HOMOGENIZE_HPP
+#define HOMOGENIZE_HPP
+using namespace vg;
+using namespace std;
+
+namespace vg{
+    class Homogenizer{
+        public:
+            /* Locates tips in the graph
```
adamnovak: This ought to be a /** doc comment.

src/srpe.hpp
```diff
@@ -0,0 +1,40 @@
...
+#define SRPE_x
+#include "filter.hpp"
+#include "index.hpp"
+#include "vg.pb.h"
+#include "vg.hpp"
+#include "gcsa.h"
+#include "alignment.hpp"
+#include "mapper.hpp"
+
+/**
```
adamnovak: This needs to move inside the namespace, or the doc generator will think it's documentation for the vg namespace itself.

src/subcommand/homogenize_main.cpp
```diff
@@ -0,0 +1,209 @@
+#ifndef HOMOG_MAIN_x
```
adamnovak: There shouldn't be include guards in a cpp file!

src/srpe.hpp
```diff
@@ -0,0 +1,40 @@
+#ifndef SRPE_x
```
adamnovak: Probably should be "VG_SRPE_H" to match what we do in vg.hpp.

src/subcommand/sift_main.cpp
```diff
@@ -0,0 +1,318 @@
+#ifndef SIFT_MAIN_x
```
adamnovak: Again, include guards in CPP files don't make sense.

src/homogenizer.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <vector>
+#include "vg.hpp"
+#include "translator.hpp"
+#include "filter.hpp"
+#include "mapper.hpp"
+#include "bubbles.hpp"
+#include "vg.pb.h"
+#include "types.hpp"
+
+#ifndef HOMOGENIZE_HPP
```
adamnovak: I'm trying to prefix all the new include guards with "VG_" to namespace them.

src/unittest/srpe_filter.cpp
```diff
@@ -0,0 +1,349 @@
+/**
```
adamnovak: Add a \file to make this a file doc comment.

```
/** \file
 * This documents the file
 */
```

Otherwise you end up documenting the first thing in the file (the vg namespace).

Makefile
edawson: I can remove it, though I'll have to rename something to avoid a collision. Originally this happened because my *_main were in the src directory; this made organization a bit more sane, but the new subcommand system is awesome.

edawson: I think on second thought that I will rename the other subcommand files to include the _main suffix.

PR #455, opened 2016-08-25
Make ultrabubble algorithm to handle bridge

benedictpaten:
Generalizing algorithm so that reported structure contains all ultrabubbles between the source and sink node, and reporting the result in a tree of nested snarls.


benedictpaten:
[note this is not yet ready to merge]

glennhickey:
moved to #488

PR #461, opened 2016-09-01
Pinned alignment, pinned multi-alignment, banded global multi-alignment

jeizenga:
Added functions in the Aligner object to perform pinned alignment (the end of a sequence guaranteed to align to the end of a node, the other end allowed to end anywhere or soft clip) for aligning read tails in the MEM threader. Also added multi-alignment feature for pinned alignment and banded global alignment, which finds the top k suboptimal alignments in addition to the optimal alignment. All new modules have unit tests for vg test.


ekg:
Going for squash and merge. I'll probably continue doing this unless there are objections.

ekg:
What do we need to do to integrate this?

PR #464, opened 2016-09-02
update dependencies to threadsafe sdsl-lite and xg

ekg:
This is needed for non-blocking operation of the mapper.


ekg:
Could someone on OSX check what might be going on here? Travis is not happy...

kislyuk:
@ekg, I think the error is in the snappy dep build, I've seen this before - you need to add the following to your `.travis.yml` `before_install:`

```
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export LIBTOOL=`which glibtool` LIBTOOLIZE=`which glibtoolize`; fi
```

ekg:
Thanks @kislyuk but that didn't work. I'm independently trying: https://github.com/mkrufky/libdvbtee/issues/22

PR #465, opened 2016-09-03
Fragment length

ekg:
Working on getting fragment length calculations in the mapper.


ekg:
After talking with @jeizenga about this, I've decided to follow this design:

When `vg map` starts, we use pairs where we have two unambiguous mappings to estimate read fragment distances. When a read pair doesn't map unambiguously, we save it into a queue. When we've aligned some number of reads (or run through all our input reads) we calculate statistics to describe the read pair distance distribution, then realign the queued reads using a read pair threshold of several standard deviations above the mean of our pair distribution. This allows each mapper thread to run in parallel without blocking on a process to estimate this distribution.

Issue #485, opened 2016-09-26
banded alignment unstable on two node graph


ekg:
I'm experiencing instability in the banded aligner.

I add this to src/unittest/banded_global_aligner.cpp

```C++
 SECTION( "Banded global aligner produces correct alignment for a graph with two nodes") {

    VG graph;

    QualAdjAligner aligner = QualAdjAligner();

    Node* n0 = graph.create_node("A");
    Node* n1 = graph.create_node("G");

    graph.create_edge(n0, n1);

    string read = string("C");
    Alignment aln;
    aln.set_sequence(read);

    int band_width = 1;
    BandedGlobalAligner banded_aligner = BandedGlobalAligner(aln,
                                                             graph.graph,
                                                             band_width);


    banded_aligner.align(aligner.score_matrix, aligner.nt_table, aligner.gap_open,
                         aligner.gap_extension);

    //cerr << pb2json(aln) << endl;                                                      
    const Path& path = aln.path();
    cerr << pb2json(path) << endl;

    // follows correct path                                                              
    REQUIRE(path.mapping(0).position().node_id() == 1);

}
```

Sometimes it runs.

```
-> % vg test
{"mapping": [{"position": {"node_id": 1}, "edit": [{"from_length": 1}], "rank": 1}, {"position": {"node_id": 2}, "edit": [{"to_length": 1, "from_length": 1, "sequence": "C"}], "rank": 2}]}
===============================================================================
All tests passed (4959 assertions in 13 test cases)
```

Often it doesn't:

```
-> % vg test
error:[BandedGlobalAligner] traceback stuck at node boundary
vg: src/banded_global_aligner.cpp:1274: void vg::BandedGlobalAligner::BAMatrix::traceback_internal(vg::BandedGlobalAligner::BABuilder&, vg::BandedGlobalAligner::AltTracebackStack&, int64_t, int64_t, vg::BandedGlobalAligner::matrix_t, bool, int8_t*, int8_t*, int8_t, int8_t, bool, int8_t): Assertion `0' failed.

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
vg test is a Catch v1.5.6 host application.
Run with -? for options

-------------------------------------------------------------------------------
Banded global aligner produces correct alignments with all types of edits
  Banded global aligner produces correct alignment for a graph with two nodes
[...]

PR #487, opened 2016-09-26
Banded alignment bug fix and multi-alignment

jeizenga:
I believe this resolves https://github.com/vgteam/vg/issues/485, which was caused by an edge condition across node boundaries that I hadn't thought of. I have created a new unit test for this case. The PR also includes my multi-alignment code for the banded aligner, which I guess I hadn't PR'ed yet.

### DIFF
--- a/src/banded_global_aligner.cpp
+++ b/src/banded_global_aligner.cpp
@@ -221,6 +221,10 @@ void BandedGlobalAligner::BAMatrix::fill_matrix(int8_t* score_mat, int8_t* nt_ta
     int64_t ncols = node->sequence().length();
     int64_t band_size = band_height * ncols;
     
+#ifdef debug_banded_aligner_fill_matrix
+    cerr << "[BAMatrix::fill_matrix]: allocating matrices of height " << band_height << " and width " << ncols << " for a total cell count of " << band_size << endl;
+#endif
+    
     const string& node_seq = node->sequence();
     const string& read = alignment.sequence();
     const string& base_quality = alignment.quality();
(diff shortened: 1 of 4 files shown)


jeizenga:
Actually, I did PR this https://github.com/vgteam/vg/pull/461. I don't know why those commits are showing up again. I probably Githubbed something wrong.

ekg:
It may be because of the squash merge commit. I am not going to do that
more. To resolve it make sure to do a git rebase on master before branching
off again.

ekg:
There are still problems.

For instance, try "CAATTT" against this graph:

```json
{"node": [{"sequence": "TA", "id": 72571}, {"sequence": "A", "id": 72572}, {"sequence": "G", "id": 72573}, {"sequence": "TTTG", "id": 72574}], "edge": [{"from": 72571, "to": 72572}, {"from": 72571, "to": 72573}, {"from": 72572, "to": 72574}, {"from": 72573, "to": 72574}]}
```

ekg:
When aligning with bandwidth 1. I'll write a test case.

ekg:
I get this error:

```
error:[BandedGlobalAligner] traceback stuck in insert row matrix interior
vg: src/banded_global_aligner.cpp:906: void vg::BandedGlobalAligner::BAMatrix::traceback_internal(vg::BandedGlobalAligner::BABuilder&, vg::BandedGlobalAligner::AltTracebackStack&, int64_t, int64_t, vg::BandedGlobalAligner::matrix_t, bool, int8_t*, int8_t*, int8_t, int8_t, bool, int8_t): Assertion `0' failed.
```

But, I can't seem to trigger it during tests. It looks like a problem with my use, or some invalid memory access? I'm confused.

jeizenga:
Definitely not a problem with your use. That code should never execute if the aligner is working correctly. If it's non-deterministic, then I'd say the usual culprits: uninitialized variables, invalid memory access, bad thread syncing. Probably not the last option in this case. Can you give me the invocation that you see it on?

PR #488, opened 2016-09-27
Update cactus bubbles to fixed ultrabubble logic

glennhickey:
Generalized algorithm will work properly on cycles and obviate need for picking source and sink (constant source of problems).  More cleaning and testing needed before merge.

(Includes some very minor vg call fixes that seem to have been forgotten about)


adamnovak:
@glennhickey now that the ultrabubbles paper work is done, are you going to rebase this and merge it in?

glennhickey:
That's the plan.  It breaks vg genotype in its current state though, so I
have to go in and make sure edit doesn't leave a bunch of stubs for the
softclips.  Unlike the old version, stubs now break bubbles.

glennhickey:
Git fail.  Retrying in #521

PR #493, opened 2016-09-29
Map good

ekg:
The "MEM threader" (which walks exact matches in the reads and does alignment only on the unmatched bits between them) isn't yet working perfectly. This branch/PR is all about improving it.


src/mapper.cpp
```diff
@@ -996,15 +1003,20 @@ Mapper::mems_pos_clusters_to_alignments(const Alignment& aln, vector<MaximalExac
...
                         }
                     }
                 }
                 cerr << mem.sequence() << " ";
             }
             cerr << endl;
         }
+    };
+
+    if (debug) {
```
yhoogstrate: I see the code base contains 2 ways of dealing with this: `#ifdef debug` and `if(debug)`. Would #ifdef not be desired?

ekg: There is some tension between enabling quickly dropping into debug mode and saving the useless branching by using the `#ifdef`. However, I agree. It's best to enable the debugging at compile time.

This PR isn't in the best state. I'll clean this up. There are also a lot of extraneous debugging lines in the current state of things.

jeizenga:
@ekg I have the banded global aligner (hopefully) fully debugged on z.pairs.gam.1000.2, but I was working on the map-good branch so I need these commits to be merged into master before I PR so I don't duplicate them.

ekg:
Ok, I'll get this done tomorrow. Should require minor cleanup on my end to
get the build working again.

ekg:
@jeizenga I mentioned a bit of development needed to properly generate the positional clusters, but didn't describe it.

I'm about to implement it. For the sake of writing out my plan, here's the idea.

Currently in the version of the mapper being improved here, we use the positional path indexes to generate clusters of MEMs from the reads that are approximately similar to their arrangement in the read. The way this is done is pretty simple. There is a hard distance threshold (3x read length) and we can take any MEM from that window beyond the current partially formed cluster as we're building it out. Then the series of MEMs, ordered along the read, are turned into a partial alignment. This partial alignment is then "patched" using @jeizenga's banded global aligner on the pieces of the read and graph between the ends of the MEMs.

This is mostly fine for short reads in a long reference, where there are large distances between repeats. However, it falls on its face when we have long reads and a similar-length graph (as is typical during the assembly of the bakeoff regions / MHC genes). It greedily generates twisted alignments that quickly jump across the entire reference sequence.

To resolve this I've sketched out a different positional clustering approach. The idea is to build a graph that embeds all possible alignments of the read through its MEMs against the graph. We can think of this as a markov model (the MEM-MM), where the states are particular MEM hits (one read and one graph position each) with transition probabilities proportional to both the distance in the read and the graph between the end of one MEM and the beginning of the next. To keep things sane in terms of the state space, I'm going to impose a band width such that we don't encode transitions between MEMs in the model where the MEMs are beyond some distance from each other in the read. The best alignment (according to the MEMs and the distance scoring) would be the Viterbi path through the markov model. 

It seems natural that secondary alignments can be found as well in the MEM-MM, but the best way to do so isn't clear to me. Removing and re-estimating the Viterbi path for each alignment could be expensive, and although I'll try it for the purpose of experimentation I suspect it won't help performance. Probably multiple clusters in the global multi-mapping sense can be extracted by avoiding secondary paths that overlap with our primary/better clusters.

Local multi-mapping is best handled by DAG alignments that @jeizenga is working on. The issues that are driving me to this are basically the same, but I'm reacting to an immediate problem for the aligner. I think I don't need the DAG alignment model to try what I'm doing, but I think there is a natural way to map the results of the MEM-MM into it.

jeizenga:
This sounds good to me. 

There is a way to get the second best (or in general k-th best) path out of the Viterbi algorithm without too much modification. You do it all in the traceback step. The key insight is that the second-best path must either a) be present in the DP matrix, or b) have been overwritten by the best path (since no other path has high enough score to overwrite it). Moreover, if it was overwritten, then the score of the second-best path is equal to the optimal score minus the difference between the score of the cell where it was overwritten and the score that would have been entered there if it hadn't been overwritten. This is because the second best path would have taken the same route to the end state as the optimal path from that point on.

The way it works then is like this. You check the difference between the optimal path and the other paths that could have extended into the same cell at each step of the traceback. The smallest difference is the where the second best path "deflects" from the optimal path. Once you've finished the traceback, you start it over again. This time, you take suboptimal traceback at the deflection point, and then that gives the second-best path. This procedure also generalizes to the top-k paths.

I have this implemented in the BandedGlobalAligner if you want to check it out.

ekg:
So this gets at something I'm curious about: could we extend the
BandedGlobalAligner to work in MEM space? Not a completely formed idea but
it would seem we would be able to exploit mechanisms like this that you've
already implemented. I could also copypasta.

jeizenga:
It would definitely take some work to retool it to work outside the banded aligner. You might be able to rework the fundamental data structure I made for this algorithm though. Look at the the API of the AltAlignmentStack. It keeps track of suboptimal tracebacks as lists of "Deflections" from the optimal traceback. As you go back through a traceback you "propose" new deflections, and it decides if they're better than any of the tracebacks it's already storing. Once you're done with a traceback, you ask it for the next one and it should have it there waiting for you.

It would probably also be instructive to take a look at a couple of the places where I call propose_deflection.

jeizenga:
I had a thought. How does the MEM threader handle MEMs that land in repeat units? It seems to me that correctly handling the translation of the MEM into the unrolled graph is really key. I think the Viterbi algorithm you're working might be a good approach to this problem, so I wanted to suggest that you keep this case in mind while you're working on it.

ekg:
If the right transition cost function is given to the MEMMarkovModel then
the resulting threads can weave backwards and create loops. However I am
not able to get consistently good results on simple cases with this
approach, and am tending to not enable this by default.

PR #499, opened 2016-10-04
Debugged traceback in banded global aligner

jeizenga:
Debugged traceback at node boundary in banded global aligner. Edited test to catch this case.


jeizenga:
@ekg To merge into map-good

ekg:
Seems this got merged in.

PR #505, opened 2016-10-11
Set rpath when building so we can find libraries without LD_LIBRARY_PATH

### LINKED ISSUE #180: Linking to dependencies
I'm running into compilation issues, which HPC- IT staff tells me is because "a lot of external dependencies are installed in in `lib/`, but these libraries are not used during linking. In particular, this is one of the linking lines:

````
g++ -O3 -std=c++11 -fopenmp -g -o bin/example cpp/example.pb.o 
obj/main.o -I./ -Isrc -Iobj -Icpp -L./ -L/usr/local/lib/ -lprotobuf -lz 
&& cp src/*.h* include/
```

The external protobuf library compiled in `$(TOPGIT)/lib` is not used, and 
instead the one installed in `/usr/local/lib` is."

I hope this text makes sense, and I hope there is a possible fix?

adamnovak:
Closes #180 by setting rpath (actually RUN_PATH, I believe) in built vg executables, so they prefer to dynamically link against stuff in vg's lib directory over system libraries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ CWD:=$(shell pwd)
 
 
 LD_INCLUDE_FLAGS:=-I$(CWD)/$(INC_DIR) -I. -I$(CWD)/$(SRC_DIR) -I$(CWD)/$(UNITTEST_SRC_DIR) -I$(CWD)/$(CPP_DIR) -I$(CWD)/$(INC_DIR)/dynamic -I$(CWD)/$(INC_DIR)/sonLib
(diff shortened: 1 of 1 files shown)


adamnovak:
It's weird that this actually breaks the build on OS X. Maybe there's a default rpath setting we're overriding that's needed to find gcc's libstdc++?

If we really want it to *work* on OS X we'll need to tinker with the libraries themselves: http://stackoverflow.com/a/4514831/402891. But it shouldn't *break*.

adamnovak:
Looks like the OS X failures were actually unrelated. But I don't think it makes sense to do it on OS X and Linux the same way. So I'm going to merge it in like this.

Issue #62, opened 2015-07-25
RocksDB conditional dependency on lz4 shared library


mlin:
Documenting a minor build annoyance:

The revision of RocksDB used as of this writing has a [slighty too-clever configure script](https://github.com/facebook/rocksdb/pull/74/files) that enables LZ4 compression if liblz4 is present on the build system. The vg Makefile does not include `-llz4` in the final link step. Consequently:

* If liblz4 isn't installed, everything is fine.
* If liblz4 is installed, then librocksdb.a depends on it, and the final link step of vg fails.

It looks like the RocksDB configuration has been [simplified recently](https://github.com/facebook/rocksdb/commits/master/thirdparty.inc) so the solution may be as simple as upgrading to a recent revision.

ekg:
Great, so hopefully I can upgrade past this. Soon we will only need rocksdb
for sorting reads, but this is still worth sorting out.

adamnovak:
It looks like this isn't building on Travis where liblz4 isn't installed. It fails because the linker can't find `-llz4`.

PR #512, opened 2016-10-16
Upgrade RocksDB

mlin:
- Upgrade RocksDB from an ancient version to v4.11.2.
- Switch RocksDB submodule from [ekg/rocksdb](https://github.com/ekg/rocksdb) back to upstream [facebook/rocksdb](https://github.com/facebook/rocksdb). I think Erik's fork was in use for a small patch which was long since [merged upstream](https://github.com/facebook/rocksdb/pull/488).
- Fix an issue where vg doesn't link if the build system has liblz4 installed (#62)
- Tighten some error handling during metadata interactions in RocksDB. Some undefined behavior of RocksDB had changed between the versions, causing a test to fail without these adjustments. Unfortunately there are still quite a few code paths that ignore potential errors (e.g. most invocations of `db->Put()` in index.cpp).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,7 +18,7 @@
 	url = https://github.com/ekg/lru_cache.git
 [submodule "rocksdb"]
 	path = deps/rocksdb
-	url = https://github.com/ekg/rocksdb.git
+	url = https://github.com/facebook/rocksdb.git
 [submodule "htslib"]
 	path = deps/htslib
 	url = https://github.com/samtools/htslib.git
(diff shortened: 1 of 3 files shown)


ekg:
Thanks @mlin!

Maybe we can wrap put and get with an error handler? This needs love
although it doesn't yield much of a research reward.

mlin:
@ekg Yea; well, this diff makes _some_ progress :) Most cases would only error in exceptional circumstances (disk full etc.), which can just throw. Can we write off some of the tech debt by retiring vestigial features? (like the k-mer index as I understand it)

Left you another q on [gitter](https://gitter.im/vgteam/vg)

ekg:
I think we would do well to let go of the rocksdb kmer index. Notes on
gitter.

adamnovak:
Do we want to merge this in as is, or do we want to drop the kmer index before updating rocksdb?

mlin:
If there are no detailed comments on the diff, I think we can merge this as a strict improvement to master, and I'll separately work on dropping the rocksdb kmer index and exception handling.

PR #525, opened 2016-10-28
Break out construct

adamnovak:
This breaks the construction methods of the `vg::VG` type into their own compilation unit. Next I'm going to rework them so they aren't part of the VG object at all, somehow.

I also break the `vg construct` subcommand out into its own subcommand compilation unit, and set up a system whereby subcommands can auto-register just by being linked into the binary (like with the unit tests). I'm planning to refactor all of the subcommands to use this system, so `main.cpp` will stop being huge, and so we won't have to touch quite so many places to add subcommands.

@ekg does this design make sense to you? Eventually the `vg help` output will be generated from the subcommand names and descriptions, but for now I have left it manually defined.

I want to get this merged in before it's really done, because I don't want to try and deal with conflicts in the construct code between my moves to new files and other people's edits.


ekg:
This design sounds amazing! I fully support.

PR #532, opened 2016-11-02
Add -f's to rm calls

adamnovak:
Ought to fix clean builds as @ifiddes and @JervenBolleman recommend.

Maybe we should test clean builds somehow?


adamnovak:
One of these checks is broken and testing a branch in the repo that I deleted before Travis could build it.

Issue #518, opened 2016-10-26
Refactor construct code out of main (and out of vg::VG)


adamnovak:
Lots of code used in the vg construction process lives in `main.cpp`. It should be moved to its own file.

If code can be moved from `vg.cpp` to that file as well, that would be ideal. `vg.cpp` really shouldn't depend on vcflib at all.

adamnovak:
This has been done.

PR #534, opened 2016-11-04
`vg construct` streaming re-implementation

adamnovak:
I wanted some new features in construct, like the ability to re-name contigs. I decided to go and refactor construct out of the VG class.

I've really mostly ended up reimplementing it, but I worked out a way to do it streaming: you only need to load overlapping clumps of variants, instead of all the variants for a chromosome at a time. This should use a lot less memory, and is much faster in my testing on the `1mb1kgp` example (2.5 seconds vs 9.5 seconds).

I'm not sure how this reimplementation will scale to really big graphs, because it's currently single-threaded. You need to have the previous chunk constructed before you can re-number the next chunk. I didn't copy over the cool stuff from the old construct logic where graphs were merged as they were completed. It would be possible to do that while streaming, but you'd only be able to queue things up in the order they come in/go out, so the benefit might be less.

It would be easy to parallelize across chromosomes, because each can stream through independently; I'm going to do that.

I've laid the groundwork for taking in multiple VCFs and FASTAs (although it's not hooked up to the UI), and vg construct will now work with VCFs that aren't indexed, as long as they are sorted.

Because of the way I generate the ref nodes for a clump of variants after the alt nodes, the IDs of the ref nodes end up being larger than those for the alt nodes, where before they were smaller. This broke a lot of the hardcoded node IDs in tests, so I had to change them out.

I also have slightly different behavior for some sets of variants. In particular, I daisy-chain deletions, adding an edge combining the deletions when multiple deletions are right next to each other.


adamnovak:
@edawson What do you think?

adamnovak:
This closes #518 by the way.

Since I've been forging ahead on top of this, and since I don't want to get too far out of sync, I'm going to go ahead and merge this in the next few days if there's no feedback to the contrary.

PR #537, opened 2016-11-07
Linearization algorithm based on max flow/min cut (Final)

mzueva:
Hello!
This is final implementation of vg sorting using max flow/min cut algorithm from the paper. 
Sorting now is implemented as a separate tool "vg sort". 

In this implementation we use Ford-Fulkerson algorithm to find the min cut edges, remove them from the graph and walk along the backbone (some primary path in the graph) finding in- and -out growths and inserting them before and after backbone nodes. Weights are assigned to edges as a number of paths, that go through the edge (path goes through edge, if both adjacent nodes are mapped to that path). 

We have implemented the backbone determining in the following way:
a. for the whole graph we use a path from the “-r” input argument as a backbone;
b. we use a greedy algorithm for recursive calls for in- and out-growths; we start from the backbone and go in depth always selecting the edge with the maximum weight.

In this pull request we also provide tool "vg fsort" (fast sort), this is implementation of graph sorting using Eades algorithm, described in the paper for Recomb. 

Sorting tools take three input arguments:
1.  -r ref - name of the initial backbone, required
2.  -i in.vg - input graph with *.vg or *.gfa extension
3.  -g - flag that indicates *.gfa input

The output is binary vg file, written to stdout.
All out code is located now in a separate file flow_sort.cpp. 
Looking forward to your comments and review!
I've closed the previous PR: https://github.com/vgteam/vg/pull/434 since it's outdated.


ekg:
Cool!!! I'd move to merge but then combine the sort and fsort command line
arguments. Is this tenable? I just want to keep the top level API as
minimal as reasonable.

mzueva:
Hi, Erik. This is completely reasonable, I've merge the tools into one "sort", sorting algorithm is managed by the -f flag (--fast). Default algorithm is David's max-flow sorting and Eades algorithm is launched with -f flag. Please, take a look at the latest commit.

glennhickey:
@mzueva Was just about to merge this but see the commit merging fsort and sort into one tool has been reverted.  Is there a reason why?  Can you please merge them back?  Thanks

mzueva:
@glennhickey I'm sorry, it's a mistake. I've returned that commint it back

PR #546, opened 2016-11-12
Rewrite stream::for_each_parallel

### LINKED ISSUE #530: performance problem with stream::for_each_parallel<Alignment>
@adamnovak @glennhickey @ekg 
As discussed on the call earlier today; I've seen some drastic performance problems using `stream::for_each_parallel` on GAM alignments. I've hit this while fiddling with RocksDB, and it also seems to manifest using Glenn's new parallel version of vg filter (which ran much slower than the old serial version!)

I've pushed a branch [mlin-stream-for-each-parallel-perf-repro](https://github.com/vgteam/vg/tree/mlin-stream-for-each-parallel-perf-repro) to isolate the problem using a trivial subcommand `vg count` that counts the alignments. `vg count -t 0` uses `stream::for_each`, while `-t N` for N>0 uses `stream::for_each_parallel` with that number of threads.

Here's a small GAM (155MB) I'm using with it: [TSNano_1lane_L008_13801_NA12878_R1_001.18.gam](https://dl.dnanex.us/F/D/YBvZXJ5k2B6jJZzb31y2KyggK440Qb3zY3Gg4Zk3/TSNano_1lane_L008_13801_NA12878_R1_001.18.gam)

On a 4-core desktop I get:

```
$ time vg count -t 0 TSNano_1lane_L008_13801_NA12878_R1_001.18.gam
945336

real	0m4.600s
user	0m4.524s
sys	0m0.068s
$ time vg count -t 1 TSNano_1lane_L008_13801_NA12878_R1_001.18.gam
945336

real	0m10.414s
user	0m10.356s
sys	0m0.052s
$ time vg count -t 2 TSNano_1lane_L008_13801_NA12878_R1_001.18.gam
945336

real	0m11.625s
user	0m21.948s
sys	0m1.092s
$ time vg count -t 4 TSNano_1lane_L008_13801_NA12878_R1_001.18.gam
945336

real	0m14.399s
user	0m48.956s
sys	0m1.048s
$ time vg count -t 8 TSNano_1lane_L008_13801_NA12878_R1_001.18.gam
945336

real	0m58.169s
user	2m37.132s
sys	0m5.712s
$ time vg count -t 16 TSNano_1lane_L008_13801_NA12878_R1_001.18.gam
945336

real	3m53.226s
user	11m15.684s
sys	0m9.280s
```

Given this trivial counting function, I wouldn't necessarily expect a wall speedup from parallelization; but, this goes the other way rather dramatically!

mlin:
Fixes #530. Use OpenMP tasking to avoid performance issues due to mutex contention and busy-waiting. Offload protobuf parsing onto worker threads. Raise protobuf errors.

### DIFF
--- a/src/readfilter.cpp
+++ b/src/readfilter.cpp
@@ -598,7 +598,7 @@ int ReadFilter::filter(istream* alignment_stream, xg::XG* xindex) {
                 // flush buffer (could get fancier and allow parallel writes to different
                 // files, but unlikely to be worth effort as we're mostly trying to
                 // speed up defray and not write IO)
-#pragma omp critical
+#pragma omp critical (ReadFilter_flush_buffer)
                 {
                     flush_buffer(tid, chunk);
                 }
--- a/src/stream.hpp
+++ b/src/stream.hpp
@@ -133,76 +133,101 @@ void for_each_parallel(std::istream& in,
                        const std::function<void(T&)>& lambda,
                        const std::function<void(uint64_t)>& handle_count) {
 
-    ::google::protobuf::io::ZeroCopyInputStream *raw_in =
-          new ::google::protobuf::io::IstreamInputStream(&in);
-    ::google::protobuf::io::GzipInputStream *gzip_in =
-          new ::google::protobuf::io::GzipInputStream(raw_in);
-    ::google::protobuf::io::CodedInputStream *coded_in =
-          new ::google::protobuf::io::CodedInputStream(gzip_in);
+    // objects will be handed off to worker threads in batches of this many
+    const uint64_t batch_size = 1024;
+    // max # of such batches to be holding in memory
+    const uint64_t max_batches_outstanding = 256;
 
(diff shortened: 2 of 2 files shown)


mlin:
@glennhickey FYI this should fix the worst problems we hit with parallelized `vg filter` actually being a lot slower+costlier than the previous serial version (due to #530). However, it's still not that much *faster* than the serial version (well-utilizing only ~2 threads). I speculate the bottleneck is now serialized gzip compression of the output. We can get to that in due course.

There are some constants in the new version implying that up to 256K `Alignment` records may be held in memory at once. I think I recall discussion of some outlier records that are megabytes in size, but, the *mean* size is quite small, right? We can make these adjustable in the future of course.

adamnovak:
This is ready to merge right?

adamnovak:
I'm going to merge this in, since it's been sitting for a while.

PR #556, opened 2016-11-17
Make pinned alignment start from any source/sink node

jeizenga:
Change in the pinned alignment API so that now pinning left/right toggles whether it is pinned to source/sink nodes, but it will automatically find the optimal one.

### DIFF
--- a/src/gssw_aligner.cpp
+++ b/src/gssw_aligner.cpp
@@ -119,7 +119,8 @@ gssw_graph* Aligner::create_gssw_graph(Graph& g, bool add_pinning_node, gssw_nod
 }
 
 void Aligner::align_internal(Alignment& alignment, vector<Alignment>* multi_alignments, Graph& g,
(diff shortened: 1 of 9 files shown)


jeizenga:
@ekg I'm not entirely sure what's failing on the build here, but I think it has more to do with the Travis back end than my code. It's complaining because it doesn't have functions from the newest sonLib update, but that update is included in this PR, it just seems to be not rebuilding it. One of the tests also seems to be not generating the protobuf files. Not sure what's going on there.

I notice that the main map-good PR is also having build issues. Have you changed anything around in the Makefile that might have affected this?

jeizenga:
@ekg This now depends on https://github.com/ekg/gssw/pull/10 or else it will not compile. I'll need you to merge that first so I can update the dependency in this PR. I'll look into this sonLib conflict in the meantime.

jeizenga:
I did this a different, better way later.

Issue #563, opened 2016-11-20
assertion failure in `vg mod -N` on newly constructed graphs


mlin:
Hey @adamnovak, I'm hitting a problem with my usual 21+Y integration test, I think related to the recent `vg construct` overhaul. Specifically `vg mod -N` hits this assertion failure:

```
$ curl -s https://dl.dnanex.us/F/D/YJJb0ByKv92vfPQGYxq9Z2gbp8yq4JQ8Kp3j5b6b/21.vg | vg mod -N - > /dev/null
vg: src/vg.cpp:1155: vg::VG::remove_non_path()::<lambda(const vg::Path&)>: Assertion `has_edge(s1, s2)' failed.
```

 I also note that when I `vg view` that graph i get:

```
graph path '21' invalid: edge from 504158 end to 504159 start does not exist
graph path '21' invalid: edge from 1394159 end to 1394160 start does not exist
graph path '21' invalid: edge from 1399320 end to 1399321 start does not exist
graph path '21' invalid: edge from 1866566 end to 1866567 start does not exist
graph path '21' invalid: edge from 3474228 end to 3474229 start does not exist
graph path '21' invalid: edge from 3580632 end to 3580633 start does not exist
[vg view] warning: graph is invalid!
```

I've bisected this to between d49b351 and 3590b1c. The version linked above is constructed with the latter version while I don't hit these issue with the former.

I'm poking around but hoping you've got the instinct where to look...

adamnovak:
That does sound like it's probably my code's fault. I must not be
constructing the reference path correctly based on some of the variants
that exist in your VCF.

Looking at the VCF and the generated graph for the first issue, it looks
like there's a variant (21:15847470, esv3646432) that has a long ref, and
maybe its reference node is getting broken up by my code to stop nodes from
being too long when they are generated. But that code is supposed to link
the two parts with an edge, and they aren't being linked.

I had a look at the code, and it turns out I forgot to actually write the
code to link up those broken parts. I've added ti in
https://github.com/vgteam/vg/pull/564 which I should be able to merge soon.

-Adam

mlin:
Thanks @adamnovak! This fixed my 21+Y test, and I also just kicked off a whole-genome graph & index build with  2e9de45. It actually has been a long time since we've done such a build on master (vs. the bakeoff freeze branch) so this will be interesting to see.

PR #564, opened 2016-11-21
Fix missing construct edges

adamnovak:
@mlin was having some problems with missing edges with the new vg construct; it turns out I forgot some code to actually generate edges when an allele gets broken for being too long. This fixes that.


adamnovak:
Closes #563 (hopefully).

PR #568, opened 2016-11-28
RocksDB whispering

mlin:
* Make `vg index --store-alignments` efficient enough for WGS GAMs by wandering through the [RocksDB options jungle](https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide) to tune buffer sizes, concurrency, and write amplification.
* We no longer perform a total compaction of the RocksDB database upon initial completion of the load. The tradeoff is more work to do during subsequent reads from the index, as RocksDB needs to search within a number of sorted tables instead of just one. We do however keep this number small using background compactions while the load is running, so the practical impact should be minimal compared to the frustration of waiting for the single-threaded total compaction (which can still be done using `vg index --compact`).
* `vg index --store-alignments` would crash on any 'unmapped' read. For now I've avoided the crash by making it ignore/discard such reads and printing a warning (as a temporary FIXME measure). As a next step, we should decide how these should be keyed in the database.

This is a start- lots more to do!

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -9,12 +9,10 @@ Index::Index(void) {
     start_sep = '\x00';
     end_sep = '\xff';
     write_options = rocksdb::WriteOptions();
-    mem_env = false;
-    use_snappy = false;
-    // We haven't opened the index yet. We don't get false by default on all platforms.
-    is_open = false;
+    // disable write-ahead logging when writing to RocksDB. This is for write durability
+    // in the event of power failure etc. which is not really relevant to our use case.
+    write_options.disableWAL = true;
     db = nullptr;
-    //block_cache_size = 1024 * 1024 * 10; // 10MB
     rng.seed(time(NULL));
 
     threads = 1;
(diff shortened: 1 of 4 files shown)


mlin:
@ekg the load performance remains quite dependent on storage bandwidth. make sure to build the rocksdb index on a locally attached scratch SSD if possible rather than NAS. once built of course it can be copied and accessed from NAS.

ekg:
@mlin, this really worked well.

Before, on a yeast data set:

```
eg10@vr-4-1-16 [17:37:26] [/lustre/scratch115/realdata/mdt3/projects/graphs/yeast/erik] 
-> % time vg index -N -d Y12+SK1.gam.index aln.gam
vg index -N -d Y12+SK1.gam.index aln.gam  6673.46s user 3173.51s system 254% cpu 1:04:22.31 total
```

After:

```
eg10@vr-4-1-16 [18:41:50] [/lustre/scratch115/realdata/mdt3/projects/graphs/yeast/erik] 
-> % time vg index -N -d Y12+SK1.gam.index.1 aln.gam
vg index -N -d Y12+SK1.gam.index.1 aln.gam  3330.13s user 1416.78s system 838% cpu 9:26.16 total
```

6 times faster!

PR #575, opened 2016-11-30
Add vg map option to be stricter about alignment pairs

glennhickey:
Specifically, this new option (-O), will only report a paired alignment if both read-alignments of the primary pair are also individually top-scoring for their respective reads.  

vg filter (pre-parallelization) used to do something similar by filtering any primary mapping with a higher-scoring secondary mapping.  This logic noticeably helped whole-genome calling results.  The new option here makes a bit more sense in that it treats both fragments together when filtering, so hopefully it helps at least as much.     

@mlin I added this (but using -a instead of -O) to my freeze-with-parallel-filter branch here https://github.com/glennhickey/vg/commit/70200fab8d69c960eff05367df86389588fecd82

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6057,11 +6057,11 @@ int main_map(int argc, char** argv) {
                  &max_mem_length,
                  &band_width,
                  &pair_window,
-                 &only_top_pairs,
+                 &top_pairs_only,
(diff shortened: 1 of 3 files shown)


mlin:
Thanks @glennhickey - I'm rerunning NA12878 on the 1000G and primary graphs over the weekend.

PR #577, opened 2016-11-30
Fix gpbwt

### LINKED ISSUE #572: Empty paths are discarded
If a `Path` has no `Mapping`s in it, it gets discarded by `vg view` and other commands that work with `vg::VG` objects.

```
$ echo '{"node": [{"id": 1, "sequence": "A"}], "path": [{"name": "apath"}]}' | vg view -Jv - | vg view -j -
{"node": [{"sequence": "A", "id": 1}]}
```

Empty paths are potentially useful for storing zero-length alts of deletion variants imported from VCF. You have one path defining the reference bit that's deleted, and another path, stored as an alternate allele for the same variant, that doesn't visit any nodes, meaning you would skip from the end of the node before the ref alt to the start of the node after the ref alt.

adamnovak:
This fixes and adds a basic test for the thread generation for the gPBWT. Previously it was very broken even on simple graphs, and probably has been since my moving of vg index to its own file, and/or my rewrite of vg construct.

This also fixes #572 in a fairly inelegant way, by dumping all 0-length paths into the first chunk of a graph when serializing it. @ekg do we actually want to have 0-length paths? My thread generation code is designed to treat all paths it can't find as 0-length paths, so we could continue dropping/disallowing them.


adamnovak:
I'm going to merge this in, and we can talk about empty paths on the Tuesday call and back it out if we don't want it.

PR #583, opened 2016-12-03
Improve gPBWT construction

adamnovak:
The xg changes that enable compression have already been merged, but these changes fix some issues with graph construction that were coming up when encoding the gPBWT, and fixes some oversights in the phase thread construction algorithm for the gPBWT.

I've also refactored out a PathIndex that can be used to efficiently find nodes and positions in a reference path, which we now use in the gPBWT construction as well as the genotyper.


ekg:
Is this just merge-broke? Or is there another problem?

adamnovak:
I just have xg pointing to a commit that's not present upstream.

adamnovak:
I'm tweaking the expected node and edge counts on the 1MB graph, since my changes to construct seem to have made them different. All the construct unit tests still pass, but it's possible these changes are undesirable. If pipeline performance drops, this might be a place to look.

PR #590, opened 2016-12-08
copy the stream::for_each_parallel into batched and not-batched versions

ekg:
The current for_each_parallel on master appears to hurt performance when the runtime of the lambda is long relative to the parsing time. This is in contrast to the issue that @mlin and @glennhickey resolved where performance was very bad in the old version when processing time in the lambda was short.

This PR proposes to split the solutions into two functions and select the one appropriate for the current task depending on the expected runtime of the lambda. @mlin's improved version is now called `stream::for_each_parallel_batched`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1911,7 +1911,7 @@ int main_pileup(int argc, char** argv) {
             int tid = omp_get_thread_num();
             pileups[tid].compute_from_alignment(aln);
         };
-        stream::for_each_parallel(alignment_stream, lambda);
+        stream::for_each_parallel_batched(alignment_stream, lambda);
     });
 
     // single-threaded (!) merge
@@ -4256,7 +4256,7 @@ int main_stats(int argc, char** argv) {
         };
 
         // Actually go through all the reads and count stuff up.
-        stream::for_each_parallel(alignment_stream, lambda);
+        stream::for_each_parallel_batched(alignment_stream, lambda);
 
         // Calculate stats about the reads per allele data
         for(auto& site_and_alleles : reads_on_allele) {
--- a/src/stream.hpp
+++ b/src/stream.hpp
@@ -133,6 +133,90 @@ void for_each_parallel(std::istream& in,
                        const std::function<void(T&)>& lambda,
                        const std::function<void(uint64_t)>& handle_count) {
 
+    ::google::protobuf::io::ZeroCopyInputStream *raw_in =
+          new ::google::protobuf::io::IstreamInputStream(&in);
+    ::google::protobuf::io::GzipInputStream *gzip_in =
+          new ::google::protobuf::io::GzipInputStream(raw_in);
+    ::google::protobuf::io::CodedInputStream *coded_in =
+          new ::google::protobuf::io::CodedInputStream(gzip_in);
+
+    uint64_t count;
+    bool more_input = coded_in->ReadVarint64((::google::protobuf::uint64*) &count);
+    bool more_objects = false;
+    // this loop handles a chunked file with many pieces
+    // such as we might write in a multithreaded process
+    std::list<T> objects;
+    int64_t object_count = 0;
+    int64_t read_threshold = 5000;
+#pragma omp parallel shared(more_input, more_objects, objects, count, in, lambda, handle_count, raw_in, gzip_in, coded_in)
+    while (more_input || more_objects) {
+
+        bool has_object = false;
+        T object;
+#pragma omp critical (objects)
+        {
+            if (!objects.empty()) {
+                object = objects.back();
+                objects.pop_back();
+                --object_count;
+                has_object = true;
+            }
+        }
+        if (has_object) {
+            lambda(object);
+        }
+
+#pragma omp master
+        {
+            while (more_input && object_count < read_threshold) {
+                handle_count(count);
+                std::string s;
+                for (uint64_t i = 0; i < count; ++i) {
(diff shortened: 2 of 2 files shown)


mlin:
@ekg I'm going through your test procedure and one thing I'm noticing, with *both* master and b6aa404, is that the time to run the mapping seems to vary a lot across repeated invocations on the same input (z.pairs.gam.10k). With either branch, if I run it repeatedly a dozen or so times, the wall time (with -t 4 on my 4-core desktop) is usually around 30sec, but sometimes much faster (<20sec) and occasionally much slower (>2min).

Do you see this as well? Is there something nondeterministic in the actual amount of useful work the mapper does?

(I have several ideas for how to adjust the batch version to make it handle all cases better, without needing two different implementations although of course that gets the job done -- but, the above makes it difficult to isolate any performance issue)

Recording for posterity Erik's benchmark case, the last command being the one discussed above.

```
vg construct -r 1mb1kgp/z.fa -v 1mb1kgp/z.vcf.gz -m 32 -p | vg mod -c - >z.vg                                                   
cat <(vg mod -pl 16 -e 4 z.vg ) <(vg mod -N z.vg) | pv | vg kmers -gB -k 16 -H 1000000000 -T 1000000001 - >z.smooth.graph 2>/dev/null
vg index -i z.smooth.graph -g z.smooth.gcsa -X 2 -Z 400 -p                                                  
vg index -x z.xg z.vg                                                                                       
vg sim -n 10000 -e 0.01 -i 0.005 -l 100 -p 500 -v 50 -s 9999 -x z.xg -a | vg view -a - | pv -l | vg view -JaG - >z.pairs.gam.10k
vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -t 4 -a --compare | pv -l >aln.gam
```

ekg:
@mlin I am getting extremely large variance in runtime with master, but not with this PR.

```
-> % time vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4 --compare >/dev/null
vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4  >   26.29s user 0.06s system 372% cpu 7.075 total
erik@tappy [11:28:52 AM] [~/vg.8/test] [master *]
-> % time vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4 --compare >/dev/null
vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4  >   26.17s user 0.06s system 377% cpu 6.943 total
erik@tappy [11:29:00 AM] [~/vg.8/test] [master *]
-> % time vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4 --compare >/dev/null
vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4  >   26.16s user 0.06s system 379% cpu 6.919 total
erik@tappy [11:29:08 AM] [~/vg.8/test] [master *]
-> % time vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4 --compare >/dev/null
vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4  >   26.18s user 0.06s system 378% cpu 6.941 total
erik@tappy [11:29:15 AM] [~/vg.8/test] [master *]
-> % time vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4 --compare >/dev/null
vg map -x z.xg -g z.smooth.gcsa -iG z.pairs.gam.10k -u 1 -L 10 -a -t 4  >   26.54s user 0.06s system 379% cpu 7.017 total
```

So almost always ~7s on this PR with this specific command line parameters. Point being, it's not variable.

The thing that drove me to this issue was observing the high variability. Sometimes a similar test would be 20s, but other times it would be many minutes.

ekg:
@mlin you may need to rebuild all of the vg objects from scratch between changes to stream.hpp. I noticed that the dependencies don't seem to be set up for stream.hpp, so it's possible you've recompiled but not actually changed the stream functions in some contexts.

I did this to rebuild:

```
source ./source_me.sh
make clean-vg
make -j 3 static
```

ekg:
@mlin I should be clear that I saw the same behavior you're describing after swapping out stream.hpp before I did the `make clean-vg` rebuild.

ekg:
I'm going to merge this so as to not get blocked. Let's sort it out as we can.

mlin:
@ekg OK. I'll keep studying this.

mlin:
@ekg I'm just now noticing the `gam_paired_interleaved_for_each_parallel` function in `alignment.cpp` to process the alignments in pairs. It extracts read pairs expected to come one after the other in the GAM stream, and in particular assumes that the critical section will see these records "in order."

I think this concurrency assumption is problematic under **either** implementation of `stream::for_each_parallel`. That stated, it might be violated more frequently in the batched version than in the original. I think this is the probable cause of the speed variability: using the batched version, the "pairs" are more likely to be wrong, with higher variance across runs, and the mapper takes much longer to deal with wrong pairs.

I'll think about how to refactor some of this.

PR #594, opened 2016-12-09
Make good

adamnovak:
Rework the Makefile to be less verbose and reduce the chances of future conflicts.


ekg:
Since merging this I've had a make fail to produce a binary. I used a high
number of threads.

adamnovak:
That's not encouraging.

Do you know what exactly it got hung up on?

ekg:
I don't and I'm moving a bit too fast to focus on this... I'll see if I can
reproduce next time I do a completely fresh clone.

PR #596, opened 2016-12-11
generalize stream::for_each_parallel to subsume gam_paired_interleaved_for_each_parallel

mlin:
`gam_paired_interleaved_for_each_parallel` had some problematic concurrency assumptions about the order in which `for_each_parallel` would present records to it. This problem was exposed by the recent rewriting of `for_each_parallel` to use batches, but, also existed under the old implementation (albeit less frequently-arising in practice). This diff handles interleaved record pairs in the `for_each_parallel` implementation, with the usual for-each-individual-record as a special case.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -327,31 +327,6 @@ size_t fastq_paired_two_files_for_each(string& file1, string& file2, function<vo
 
 }
 
-void gam_paired_interleaved_for_each_parallel(ifstream& in, function<void(Alignment&, Alignment&)> lambda) {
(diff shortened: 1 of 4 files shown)


mlin:
@ekg i think this will be a cleaner solution for #590

adamnovak:
Looks like we got burned by stale Travis results on this; between when Travis ran the tests and when I hit merge, I merged some other changes from @edawson that used gam_paired_interleaved_for_each_parallel.

I've walked back master, but Github isn't very clever about situations like this and can't reopen PRs.

PR #597, opened 2016-12-13
vg construct: reduce stderr warnings about symbolic alleles

mlin:
@adamnovak @edawson wdyt of this?


adamnovak:
This probably makes sense to have.

PR #612, opened 2017-01-05
Add BED rewriting to vg simplify

adamnovak:
This attempts to rewrite BEDs during `vg simplify` so they point to corresponding places in the new coordinate space for the paths they are on, after pieces of the paths that traverse bubbles change in length.

It doesn't actually seem to work correctly all the time yet.


adamnovak:
The BED rewriting ought to work now; I have some unit tests for the logic.

PR #613, opened 2017-01-06
Make MEM threader optional in MSGA

adamnovak:
I was having some trouble getting a set of sequences to MSGA correctly, because the paths that got added to the graph didn't actually trace out the sequences. I fixed this by disabling the MEM threader.

I can provide the offending file upon request, and I have a bunch of debugging code that tries to dig into the problem in my debug-msga branch.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1731,6 +1731,7 @@ void help_msga(char** argv) {
          << "    -o, --gap-open N      use this gap open penalty (default: 6)" << endl
          << "    -e, --gap-extend N    use this gap extension penalty (default: 1)" << endl
          << "mem mapping:" << endl
+         << "    -l, --no-mem-threader   do not use the mem-threader-based aligner" << endl
          << "    -L, --min-mem-length N  ignore SMEMs shorter than this length (default: 0/unset)" << endl
          << "    -Y, --max-mem-length N  ignore SMEMs longer than this length by stopping backward search (default: 0/unset)" << endl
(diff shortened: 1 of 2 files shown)


ekg:
I have fixed this on the mapqz branch. The MEM Markov model was emitting
candidate alignments with micro inversions, which I have since disabled. I
found it by code review.

I just now checked your test case (all_seqs_upper.fa). It works.

I would suggest changing this PR to add an option to go back to the old
style mapper.

ekg:
For proof here's the all_seqs_upper result.
[all_seqs_upper.vg.zip](https://github.com/vgteam/vg/files/689359/all_seqs_upper.vg.zip)

ekg:
I generated this with this command:

```
vg msga -f all_seqs_upper.fa -X 1 -E 4 -K 16 -t3 -D >all_seqs_upper.vg
```

Probably you could get away with running a straight `vg msga -f all_seqs_upper.fa -D >all_seqs_upper.vg` on a bigger server. I ran out of memory on my laptop so I pruned the graph back.

ekg:
@adamnovak I've now fixed this in master. We should change this PR to make MEM threading optional.

adamnovak:
Alright, I'll make it an option that defaults to the new MEM threader.

When will mapqz be merged in?

ekg:
@adamnovak I merged mapqz in earlier today, so you should be able to update your master and test.

Issue #620, opened 2017-01-12
Construct clobbers its own ID space


edawson:
It looks like construct clobbers graph ID spaces when multiple contigs are present in the fasta/VCF pair. I get a bunch of `warning: [node/edge ID N] appears multiple times. Skipping.` when trying to view the resulting graphs and some node IDs appear on multiple reference paths in the GFA. Building the graphs individually and then combining them after a `vg ids -j` yields a perfectly fine graph, so the chunking logic looks to be fine.

I think it's as simple as tracking the maximum id of the individual contig graphs in the callback function that serializes the graphs to disk and incrementing the next graph's ID's as needed (similar to what is done for chunks right now). Does this sound reasonable as a fix, @adamnovak ?

adamnovak:
Oops.

That sounds like it ought to fix it; all the construction is done in
serial, so it should all get serialized once in order.

PR #623, opened 2017-01-13
Construct multiple contigs

adamnovak:
This ought to fix the bug @edawson found in #620 where construct just doesn't think about multiple calls to `construct_graph` needing to be in the same ID space. Now all the graphs produced by a `Constructor` should be in a shared ID space.


edawson:
@adamnovak any idea why this fails tests?

adamnovak:
Looks like it's failing because I rolled back xg changes when I didn't mean to. I've removed that change; let's see if it works now.

PR #628, opened 2017-01-18
MEM good

jeizenga:
Tune up for the SMEM and sub-MEM reseeding algorithms. This update:
- Switches reseeding semantics so that it returns all minimally-more-frequent sub-MEMs (i.e. MEMs that occur at least once outside of the SMEM) rather than MEMs that are at most half the length of the SMEM
- Introduces a new sub-MEM node filling algorithm that avoids returning any hits that occur inside the containing MEMs
- Moves the node filling routing into the main MEM finding routine so that it does not need to be called separately (this is also necessary for the sub-MEM node filling algorithm)
- Resolves a bug in which false MEMs would be returned that had the same start position as true MEMs but were shorter 
- Moves min and max length filters into the MEM finding routine to avoid redundancies
- Exposes reseeding algorithm through command line options in vg find
- Adds tests for reseeding in 05_vg_find.t

We should check to make sure this doesn't hurt the performance of the mapper, but I think it's unlikely that it will. The MEMs that we find now should only exclude non-supermaximal exact matches if they were redundant with SMEMs, and it should also find some new sub-MEMs that we would miss before. Point being, the actual information content of the MEMs should never decrease. It's possible that the actual MEM finding routine will slow down, but it should get faster to process the MEMs downstream since we will eliminate redundant ones.


jeizenga:
@ekg @adamnovak 
I think these should pass tests now. I went ahead and removed the MSGA test that was based on getting a specific md5 hash from the graph, which is one of the ones this PR was failing. I don't think it's a good test. Both the long sequence aligner and the read aligner are approximations (because of banding and MEM threading respectively), and the algorithms are being actively developed/hopefully improved. It doesn't seem reasonable to me to enforce that they always produce the exact same graph.

jeizenga:
Incidentally, I'm really interested to know how this algorithm compares in speed to the non-reseeding version. Especially if there's a way to break that down by how much is actually spent in the reseeding routine as opposed to downstream processing of the larger number of MEMs.

I have reason to believe that both this algorithm and the current reseeding algorithm will be quite slow whenever there aren't any true sub-MEMs being masked by SMEMs (i.e. most of the time). Probably something like O(log(|graph|) * |read|).

I have another version of the reseeding algorithm that I suspect will be a lot faster, but it provides somewhat different guarantees. My current one finds all MEMs that are contained in the SMEM read sequence, but occur outside of the SMEM somewhere in the graph. The new one would find all such MEMs subject to a minimum length (something like |SMEM| / 2 or |SMEM| - 16). I think this is reasonable since the we should suspect that the SMEM is actually the correct match in the graph. This constraint basically requires a sub-MEM to not only be locally maximal in the graph, but also pretty good compared to the SMEM. Anyway, the new algorithm's complexity more like O(|SMEM| + log(|SMEM|) * \sum |sub-MEM|), where the sum is over only the sub-MEMs with sufficiently high length to be returned. That's a lot better because, again, in most cases there probably won't be any sub-MEMs of length |SMEM| / 2.

benedictpaten:
This is a great example of why we need to setup decent continuous
integration and benchmarks for the project. You should have ways of easily
experimenting with the speed and catching regressions.

PR #633, opened 2017-01-20
Graph the SVs

edawson:
This brings in a fresh vcflib which includes the SV reformatting code, as well as a small loop in the src/subcommand/srpe_main.cpp file which will soon type all variants in the graph or a subset from a vcf. This can also be done with `vg genotype`, and probably with the right incantation to `vg call`.


adamnovak:
So the SRPE stuff is supposed to be unfinished so far, right?

Is SRPE maint to be another genotyper? So we have call, genotype, and SRPE?

If so, maybe we should think about collecting them all into one interface, maybe with selectable algorithms/strategies. Nobody will know which one to use if there are three top-level genotyping commands.

edawson:
Yep, it's supposed to be unfinished - the variant typing works but isn't tuned, and the initial plan for SV detection by local realignment has kind of fallen into irrelevance.

I definitely support bringing everything under one interface and I think we'll have to allow users to select various algorithms. I've had a lot of trouble with specific datasets choking call/genotype, especially when a lot of SVs are present in the variation graph, the reads don't span novel SVs or the number of reads gets really large. I'm not sure how we could determine programmatically which algo would best handle a specific set of reads/variants.

I guess for the moment I should move the variant typing into genotype and maybe look at fleshing out the SV caller in either one of call's/genotype's namespace.

adamnovak:
It would be nice to get the vcflib changes in master like we talked about this morning on the call.

You have a line `else if (!spec_vcf.empty() do_all){` in `srpe_main.cpp` which is breaking the mac build. I have no idea how the Linux build is tolerating it; it doesn't make any sense.

edawson:
Fixed and merged.

PR #634, opened 2017-01-22
Fast sub-MEM algorithm

jeizenga:
I went ahead and implemented the other sub-MEM algorithm I mentioned in https://github.com/vgteam/vg/pull/628 and included it has a command line option. Right now it needs to be turned on, but I think it's worth considering making it the default. We would probably want to do speed tests on real data first.

The main difference between the two algorithms is that the new one doesn't make any calls to the LCP index, which are O(log(|graph|)). That's not as huge a deal when you only need to make a few calls to it for a read because you're looking at mostly true, long SMEMs. In contrast, when you're looking for sub-MEMs most of the MEMs are short false MEMs, which means you have to make a bunch of LCP queries (say, something like |SMEM| / 4 of them).

Instead, the new algorithm looks for "probe matches" with length equal to the minimum MEM length and then tries to extend them into MEMs. If the minimum MEM length is substantially longer than the expected length of a match to random sequence (10-16 bps in most genomes), this leads to an effective pruning strategy. To help ensure this criterion, I boost the minimum MEM length up to |SMEM| / 2 when reseeding. 

I think this PR will still fail the one test in mod that https://github.com/vgteam/vg/pull/628 failed. If I understand it correctly, that's another test where changing the mapping algorithm will pretty much always cause it to fail. Really, it's enforcing that none of the underlying processes change from when the test was introduced, which I don't think is a good way to design tests when we're working with approximations. If nobody objects, I'd like to replace it with something less stringent and add that to this PR.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4364,7 +4364,8 @@ void help_find(char** argv) {
          << "    -j, --kmer-stride N    step distance between succesive kmers in sequence (default 1)" << endl
          << "    -S, --sequence STR     search for sequence STR using --kmer-size kmers" << endl
          << "    -M, --mems STR         describe the super-maximal exact matches of the STR (gcsa2) in JSON" << endl
-         << "    -R, --reseed-length N  reseed super-maximal exact matches of length at least --reseed-length" << endl
+         << "    -R, --reseed-length N  find non-super-maximal MEMs inside SMEMs of length at least N" << endl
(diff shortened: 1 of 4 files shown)


adamnovak:
It looks like it passes all the tests, probably because the mod tests are using the default algorithm and not this one.

I'm going to merge this in so we can all play with this algorithm. Though the command-line switches proliferate unabatedly; maybe we can come up with a better way to select algorithms (like --algorighm XXXXX?).

PR #644, opened 2017-01-31
Add snarls subcommand

glennhickey:
(beginnings) of command interface for snarls.hpp

@jeizenga I can't get ExhaustiveTraversalFinder to work at all.  On tiny.vg, the traversals are returned with empty visits:

 {"visits": [{}, {}]}
 
On larger tests it gets hung up for ever on very simple looking snarls that don't seem at all cyclic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,6 +128,7 @@ SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/index_main.o
 SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/mod_main.o
 SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/annotate_main.o
 SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/chunk_main.o
+SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/snarls_main.o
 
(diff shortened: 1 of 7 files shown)


jeizenga:
Should be fixed in https://github.com/vgteam/vg/pull/645

glennhickey:
@jeizenga I can't paste files for some reason so am emailing you cactus-brca2.vg and out.png

Command runs forever:
 vg snarls cactus-brca2.vg -r t > /dev/null

It's looking for traversals in this snarl:

finding traversals for {"type": 1, "start": {"node_id": 1059}, "end": {"node_id": 1062}}

vg index cactus-brca2.vg -x cactus-brca2.vg.xg
vg find -n 1059 -c 3 -x cactus-brca2.vg.xg | vg view -pd - | dot -Tpng > out.png

shows that this snarl is an innocent-looking snp bubble.

glennhickey:
@shipagarg Let's continue disucssion about implementing simple bubbles on github.  

The simplest way for you to detect your simple bubbles is probably by testing each snarl currently computed for the properites you want.  That could be a function that gets called here after traversals are computed:

https://github.com/glennhickey/vg/blob/path_bubbles/src/subcommand/snarls_main.cpp#L192

From there, we can think about making your simple bubbles part of the general API in genotypekit, either by custom iterators for SnarlManager or implementing a SnarlFinder subclass.

PR #646, opened 2017-02-01
CXX is set in the sourceme script lets not override it again

JervenBolleman:


Makefile
```diff
@@ -12,7 +12,6 @@ CPP_DIR:=cpp
 
 EXE:=vg
 
-CXX:=g++
```
mdkeehan: This one caught me with a fresh install and build from git.

ekg:
We should merge this in. Any reason not to?

adamnovak:
Travis has been in progress for months. I kicked it, so hopefully we should get a test result.

Issue #653, opened 2017-02-09
`vg snarls -u` produces incorrect bubbles


adamnovak:
Shilpa has a graph:

```
[...]

PR #659, opened 2017-02-10
fix typos in vg snarls

glennhickey:
* default bubble size for traversals misrepresented in help
* -u activated superbubbles instead of ultrabubbles (#653)
* -u -b traces still left in vg stats.

### DIFF
--- a/src/subcommand/snarls_main.cpp
+++ b/src/subcommand/snarls_main.cpp
@@ -44,7 +44,7 @@ int main_snarl(int argc, char** argv) {
     bool leaf_only = false;
     bool top_level_only = false;
     bool include_endpoints = false;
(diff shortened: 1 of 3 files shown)


src/subcommand/snarls_main.cpp
```diff
@@ -147,7 +147,7 @@ int main_snarl(int argc, char** argv) {
 
     // old code from vg stats
     if (legacy_superbubbles || legacy_ultrabubbles) {
-        auto bubbles = superbubbles ? vg::superbubbles(*graph) : vg::ultrabubbles(*graph);
```
glennhickey: Shame on the compiler for having allowed this, btw.

Issue #656, opened 2017-02-09
Bluntification duplicates overlapping sequences over reversing edges


adamnovak:
Try bluntifying this GFA with `vg mod -B`:

```
S	2	ACA
S	1	TC
L	1	-	2	+	1M
```

It produces this:

![image](https://cloud.githubusercontent.com/assets/5062495/22806144/80097004-eed5-11e6-8d2a-b563313a3aae.png)

It seems to have duplicated the overlapping sequence and included it in both orientations, when really it should just include it in one orientation.

adamnovak:
The problem with this is it creates a fake bubble, which throws off @shilpagarg's bubble-based analysis.

adamnovak:
Also, we're getting an extra A on the full traversal through. In the original it's `GACA`, and bluntified it's `GAACA`.

shilpagarg:
Here is one more example where bluntification fails, it is in GFA format.
[bluntify_fail_test1.txt](https://github.com/vgteam/vg/files/765347/bluntify_fail_test1.txt)

adamnovak:
I've sketched out an implementation of this that uses the pinch graph library. Now I just need to test it to make sure I have all the orientation flags correct.

shilpagarg:
Thats great! Thanks. I can provide u the test cases tomorrow.

adamnovak:
This has been fixed.

PR #669, opened 2017-02-14
Bluntify good

adamnovak:
This replaces the old bluntify implementation with one based on Benedict's pinch graph library in sonLib.

In setting up the tests for this, I noticed we were linking against the wrong SSW. I had to turn off a warning it gives whenever you do a small alignment, and I found that we were using the one in vcflib instead of the one we pulled in directly as a submodule. I fixed that in vcflib in https://github.com/vcflib/vcflib/pull/182 but that PR needs to merge before Travis will be happy with this one.


adamnovak:
This closes #656.

PR #675, opened 2017-02-17
Make vg add handle large structural variants specially

adamnovak:
This makes vg add smarter about SVs. It tries to align only reasonably-sized things, and falls back to aligning the ends and splicing in the middle for really big variants if an allele is too big or if a good alignment can't be found.

Also right now by default it skips adding duplications; they really ought to be added as cycle-inducing edges and not large pieces of identical sequence that the mapper can't tell the difference between.


ekg:
Could you use the banded aligner to map the SVs?

adamnovak:
I'm using the BandedGlobalAligner class when I can, but it doesn't handle
indels inside cycles well (because they make the bands get huge).

I use the banded functionality of the Mapper sometimes, but it doesn't work
well for big duplications, because it places each band independently and
thus can't tell a duplicate from the original.

Though maybe I want that, if it can find where a cycle-creating edge ought
to go...

Issue #666, opened 2017-02-13
vg map produces invalid GAM


glennhickey:
(indexes are the latest whole genome indexes @mlin sent out.  Here for ucsc folks /hive/users/hickey/toil_vg/genome/vg)

vg map -x vg/index.xg -g vg/index.gcsa -s TGCACCACTGCACTCTAGCCGCAGCAACAGGGCAAGACCTTGTCAAAAAAAAAAAAAAAAAGAAGAAAGGAAGGAAGAATGAAAGAAGAAAAGGAGAGACA

produces an alignment containing this mapping

{"position": {"node_id": 290483617}, "edit": [{"from_length": 31, "to_length": 31}, {"sequence": "G", "from_length": 1, "to_length": 1}, {"from_length": 1, "to_length": 1}], "rank": 5}

The from_length totals 33.  But the node in question has length 32

{"node": [{"id": 290483617, "sequence": "GGCAAGACCTTGTCAAAAAAAAAAAAAAAAAA"}], "path": [{"mapping": [{"position": {"node_id": 290483617}, "rank": 3079143}], "name": "22"}]}

When trying to extract the subgraph, re-index, and reproduce, I get a valid alignment, where the "G" is an insert (from_length = 0), which makes debugging hard.  

But even if the fault lies entirely within the index, vg map needs to throw an error before writing invalid output.

ekg:
Is this an index version mismatch? Where should we be detecting the problem?

glennhickey:
Hoping it's just an index version.  I'm trying to regenerate a new one.

As for detection, if it is an index version issue, this should be detected
when the gcsa is loaded by vg map (which I thought was what happened).

My other point was that vg map should can and should detect invalid edits
before they are written.  At least for this case where from_length >
node_length.

ekg:
The node length check seems reasonable and very cheap to implement. I
haven't been able to work on this yet, but I will also explore this when I
do. Should be early next week.

glennhickey:
Closed by #689 which fixed the faulty gam error (at least in the bakeoff data, will assume in the larger case too), and #682 which added some more checking.

PR #682, opened 2017-02-23
vg add improvements and alignment error checking

adamnovak:
This improves vg add to not use uninitialized memory, adopts Jordan's subgraph extraction to speed up getting large regions without extra heads/tails, and adds some asserts to make sure that alignment and pileup steps explode if invalid alignments past the ends of nodes get generated.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -2719,6 +2719,22 @@ vector<Alignment> Mapper::align_multi_internal(bool compute_unpaired_quality,
         filter_and_process_multimaps(alignments, additional_multimaps);
     }
     
+    for (auto& aln : alignments) {
+        // Make sure no alignments are wandering out of the graph
+        for (size_t i = 0; i < aln.path().mapping_size(); i++) {
+            // Look at each mapping
+            auto& mapping = aln.path().mapping(i);
+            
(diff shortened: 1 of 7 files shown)


glennhickey:
Thanks, those checks alone will be huge time savers.  I've run into similar issues in #666

src/mapper.cpp
```diff
@@ -2719,6 +2719,22 @@ vector<Alignment> Mapper::align_multi_internal(bool compute_unpaired_quality,
...
     
+    for (auto& aln : alignments) {
+        // Make sure no alignments are wandering out of the graph
+        for (size_t i = 0; i < aln.path().mapping_size(); i++) {
+            // Look at each mapping
+            auto& mapping = aln.path().mapping(i);
+            
+            if (mapping.position().node_id()) {
+                // Get the size of its node
+                size_t node_size = xindex->node_length(mapping.position().node_id());
```
ekg: This call is  pretty slow. Please use the cached version instead from `position.cpp`. `xg_cached_node_length` I think.

adamnovak:
OK, I've fixed the issue @ekg pointed out by using the Mapper's built-in XG cache in the node length finding function that it already had. I'm going to merge this in.

PR #694, opened 2017-03-02
VG Call improvements

adamnovak:
Mostly this is just a partial refactoring. I have the TSV generation pulled out into its own class (which I will soon be just sending from Caller to call2vcf, and I have call2vcf in its own Call2Vcf class.

I also deleted the non-multiallelic codepath from call2vcf.

The final design is probably going to have the existing Caller as an Augmentor, call2vcf as the new Caller, and the TSV replaced with this AugmentedGraph class that represents a support-annotated graph.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -374,11 +374,11 @@ $(OBJ_DIR)/pileup.o: $(SRC_DIR)/pileup.cpp $(SRC_DIR)/pileup.hpp $(INC_DIR)/stre
 
 $(OBJ_DIR)/caller.o: $(SRC_DIR)/caller.cpp $(SRC_DIR)/caller.hpp $(SRC_DIR)/vg.hpp $(SRC_DIR)/progressive.hpp $(INC_DIR)/stream.hpp $(SRC_DIR)/json2pb.h $(SRC_DIR)/pileup.hpp $(DEPS)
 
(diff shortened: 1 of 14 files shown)


glennhickey:
Thanks for starting this.  I owe you a cleanup of the earlier steps.

For the pileup, I'm thinking it'd be worth it to change the deletion
representation to just be something like the existing edge pileup with a
pair of offsets.  Just that fix would go a long way to cleaning the code.
Then modularizing the augmentation code which just needs a bunch of cruft
cleaned out.

adamnovak:
So the idea is you would emit a proposed deletion that's between internal
bases of nodes in the augmented graph, rather than breaking the nodes to
accommodate the proposed deletion?

glennhickey:
For this I'm just talking about the pileup, which is always in terms of the
base graph.  Currently deletion edges, which can be between internal bases,
are represented inline in the base pileups with this kind of crazy format:
-1;448277;4;0;485408;5;1.  This was done in a rush to add support for novel
deletions, and made a certain kind of sense in the crappy old caller which
went base by base.  Was proposing moving these to edge pileups, which are
now also supported by the caller.  It would also allow us to support novel
deletions inferred by hops in the mappings (as opposed to edit from/to
lengths).

PR #697, opened 2017-03-03
Added an algorithm to extract the subgraph surrounding a set of positions

jeizenga:
Added an algorithm that can extract the subgraph surrounding a set of positions. The implementation is VG/XG flexible. It also has some options that I want, like the ability to assign different search distances to different positions or to different directions from each position. Some tests are included.


adamnovak:
I'm going to merge this in, but I think it might make sense to break out the different algorithms into their own files (which is what the namespace/folder is for). The one algorithms file is getting pretty long even with just two algorithms in it.

PR #699, opened 2017-03-04
Combi alignments for pairs

ekg:
Add another alignment function for pairs which aligns each read independently with the single ended aligner, then crosses all pairs, scoring them based on their individual scores plus a bonus for alignments that fit into the fragment length distribution that we've observed or set.

This alignment function align_paired_multi_combi implements this model. A flag is added to the Mapper to allow configuring this dynamically but it is not hooked up to the command line interface.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -158,6 +158,7 @@ void Mapper::set_alignment_threads(int new_thread_count) {
     clear_aligners(); // number of aligners per mapper depends on thread count
     init_aligner(default_match, default_mismatch, default_gap_open, default_gap_extension);
     init_node_cache();
+    init_node_start_cache();
     init_node_pos_cache();
 }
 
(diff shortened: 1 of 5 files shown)


adamnovak:
It looks like this makes us fail a couple of the `vg map` tests:

```
not ok 24 - paired read alignments forced to be consistent are closer together in node id space than unrestricted alignments
#          got: '0'
#     expected: '1'
not ok 25 - only primary alignments have mapping quality scores
#          got: '0'
#     expected: '2'
```

ekg:
Yes, it isn't working right yet.

ekg:
I am moving to test this. It passes the tests with the pair fragment length restrictions because I set it to the old style mapper. It appears we do better at mapping reads despite the fact that we don't forcefully limit reads to map according to the fragment length distribution.

PR #703, opened 2017-03-08
correctly compute mapqual, sort on mix of pair correctness and score

ekg:


ekg:
Locally, this is failing the unit tests with:

```
-------------------------------------------------------------------------------
Banded global aligner can align to graphs with empty sources and sinks         
  Banded global aligner can align to a graph with an empty nodes that is both a
  source and sink                                                              
-------------------------------------------------------------------------------
src/unittest/banded_global_aligner.cpp:2292                                    
...............................................................................
                                                                               
src/unittest/banded_global_aligner.cpp:2551: FAILED:                           
due to a fatal error condition:                                                
  SIGSEGV - Segmentation violation signal                                      
                                                                               
test cases:   15 |   14 passed | 1 failed                                      
assertions: 2798 | 2797 passed | 1 failed                                      
```

ekg:
The problem is that I don't really understand how my changes would be affecting that.

ekg:
Well, it looks like the problem is in my local build.

I'll have to set that back up.

Merging based on the linux build success. More :potato: please.

PR #711, opened 2017-03-09
Oriented path distance function in XG

jeizenga:
Updates xg dependency to most recent commit, which includes a new path-based distance estimation function that is orientation-aware and adds unit tests for this function.


jeizenga:
@ekg @adamnovak 
This build failed because it doesn't have vg.pb.o when it tries to build libxg.a. I don't think I touched any part of the Makefile there. Any idea why that would be happening?

edawson:
I hit this too - adding a dependency on libxg.a to position.o, and changing the compile line for lib/libxg.a to:

           . ./source_me.sh && cd $(XG_DIR) && $(MAKE) && cp lib/* $(CWD)/$(LIB_DIR) && cp src/*.hpp $(CWD)/$(INC_DIR)

seemed to fix it.

adamnovak:
Since xg has updated its proto, I kicked off the tests for this PR again.

I don't particularly like this in the test output:

```
t/06_vg_index.t ...... 
1..45
ok 1 - indexing nodes and edges of graph
ok 2 - building an xg index of the graph
...
ok 12 - index can dump alignments
t/06_vg_index.t: line 62: 18995 Segmentation fault      (core dumped) vg map -s "CTGATGAGGCCGCCCCTAGTTTTGTTTCTGGTGCTATGGCTAAAGCTGGTAAAGGACTTC" -d x.idx
ok 13 - FIXME: alignment index does NOT store unmapped reads!
```

I don't think this has anything to do with this PR, though.

adamnovak:
Here's a real failure a bit later in the index tests:

```
ok 17 - vg connects paired-end reads in gam output
[xg] error: path GI568335954 contains duplicate node ranks
t/07_vg_map.t: line 80: 20746 Segmentation fault      (core dumped) vg map -s "ACACCTCCCTCCCGGACGGGGCGGCTGGCC" -d e > /dev/null
not ok 18 - mapping to graphs that can't be oriented without swapping edges works correctly
#          got: '139'
#     expected: '0'
```

jeizenga:
@glennhickey 
Any idea what this is:
```
The command "if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew tap homebrew/science; fi" failed and exited with 1 during .
```

glennhickey:
No idea.  The Travis Mac build flakes out on me too every so often, and I
just restart it until it works.

jeizenga:
I don't think I have permissions to make it re-run. You mind making that happen for me?

jeizenga:
Good for merge?

adamnovak:
Sure!

PR #714, opened 2017-03-10
Sv decent

edawson:
Adds some preliminary things for variant calling (without breaking the existing infrastructure, I hope) and fixes a build issues with XG.


Makefile
```diff
@@ -445,6 +448,8 @@ $(OBJ_DIR)/simplifier.o: $(SRC_DIR)/simplifier.cpp $(SRC_DIR)/simplifier.hpp $(S
 
 $(OBJ_DIR)/nested_traversal_finder.o: $(SRC_DIR)/nested_traversal_finder.cpp $(SRC_DIR)/nested_traversal_finder.hpp $(SRC_DIR)/genotypekit.hpp $(SRC_DIR)/snarls.hpp $(DEPS)
 
+$(OBJ_DIR)/gamsorter.o: $(SRC_DIR)/gamsorter.cpp $(SRC_DIR)/gamsorter.hpp $(DEPS)
+	. ./source_me.sh && $(CXX) $(CXXFLAGS) -c -o $@ $< $(LD_INCLUDE_FLAGS) $(LD_LIB_FLAGS) $(ROCKSDB_LDFLAGS)
```
adamnovak: Is there a reason that gamsorter.o can't build with the default rule? If so there should probably be a comment here so someone doesn't "clean up" this line by deleting it.

src/filter.cpp
```diff
@@ -140,73 +140,92 @@ namespace vg{
...
+
+    //     if (aln_first.name() == "" || aln_first.name() == ""){
+    //         return make_pair(Alignment(), Alignment());
+    //     }
+    //     else{
+    //         return make_pair(aln_first, aln_second);
+    //     }
+
+    // }
+    // pair<Alignment, Alignment> Filter::soft_clip_filter(Alignment& aln_first, Alignment& aln_second){
```
adamnovak: Are you sure you want to push up all this commented-out code?

edawson: Yikes, no. will fix.

adamnovak: You should fix this so this PR can merge and not develop conflicts.

edawson: I think I'm going to merge it into whole-lotta-genotype and just bring everything in at once, rather than multiple PRs having to get brought in.

PR #715, opened 2017-03-10
Whole lotta genotype

edawson:
This isn't ready to be merged quite yet, but will bring:
     1. Updated genotype code reliant on snarls (soon)
     2. Working variant recall, outputting allele depth per each path at a variant allele stored during construction (done)
     3. Genotype likelihoods for variant recall (really soon)


edawson:
This is ready to merge, though I had to revert genotype's new Snarl-based implementation as I couldn't get it to pass tests the way I'd done it. I'll do it the right way and open a fresh PR once this is in and I have time after AACR.

ekg:
Does the snarl based implementation still produce loci calls that can be
used to edit the base graph?

edawson:
Currently it does nothing, but I'm sure we could translate Snarls to Loci. There shouldn't be any information lost in going to Snarls.

edawson:
@ekg @adamnovak If there are no objections I'd like to merge this in a few hours. I just don't want to break everybody's existing PR (all 11 of 'em).

src/subcommand/snarls_main.cpp
```diff
@@ -182,6 +190,19 @@ int main_snarl(int argc, char** argv) {
...
     vector<const Snarl*> snarl_roots = snarl_manager.top_level_snarls();
+    if (fill_path_names){
+        //delete trav_finder;
+       TraversalFinder* trav_finder = new PathBasedTraversalFinder(*graph);
+        for (const Snarl* snarl : snarl_roots ){
+           vector<SnarlTraversal> travs =  trav_finder->find_traversals(*snarl);
+           stream::write_buffered(cout, travs, 0);
+        }
+
+        exit(0);
```
adamnovak: Is this meant to be in here?

src/subcommand/snarls_main.cpp
```diff
@@ -182,6 +190,19 @@ int main_snarl(int argc, char** argv) {
     // Load up all the snarls
     SnarlManager snarl_manager = snarl_finder->find_snarls();
     vector<const Snarl*> snarl_roots = snarl_manager.top_level_snarls();
+    if (fill_path_names){
+        //delete trav_finder;
```
adamnovak: Do you want to actually do this delete?

src/vg.proto
```diff
@@ -245,6 +257,7 @@ message SnarlTraversal {
 
     // A Snarl with the start and end Visits filled in to label which Snarl these traverse.
     Snarl snarl = 2;
+    string name = 3;
```
adamnovak: Are we naming the traversals individually? Or does this refer to the parent snarl?

adamnovak:
Sounds good to me. I think whoever merges things first wins.

You might want to talk with @jeizenga eventually about the snarl naming stuff and where/when we should use that.

src/vg.proto
edawson: The name is for the SnarlTraversal - each alt_path has a direct mapping to a SnarlTraversal. The Snarl names correspond to variant names.

adamnovak:
Can you add that to the comments? Someone reading the protobuf might not
already know about this alt path naming system we have.

PR #721, opened 2017-03-13
Improve the MEM chain model

ekg:
This improves the Markov model that we sample our candidate chains out of by "banding" based on approximate positions in the graph (as given by the node offset in the xg index).

### DIFF
--- a/scripts/map-sim
+++ b/scripts/map-sim
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-if [ $# -ne 6 ];
+if [ $# -ne 7 ];
 then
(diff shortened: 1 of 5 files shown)


ekg:
Here are results with this branch and the following simulation:

```
vg/scripts/map-sim human.8 hs37d5/hs37d5.fa ref/vg/index pan/vg/index 32 "-s 399 -n 500000 -e 0.01 -i 0.002 -l 150 -p 500 -v 50" " "
```

human_sim vg-93f4f26 roc logx 1

So we're doing significantly better than previously.

@jeizenga the big jump at the end is all due to your MEM determination algorithm. Solid work! Until this point its benefit was obscured by several bugs in the MEM clustering. Not only is it more accurate, but it is faster than the previous version!

PR #725, opened 2017-03-14
Call and Simplify bugfixes

adamnovak:
This fixes vg call to not try and call non-ultrabubbles, and fixes vg simplify to not try and use a SnarlManager after modifying its graph.


adamnovak:
I've added some more code to this, including a revision to how stream.hpp works when saving a vg graph. We should now have intelligent chunk auto-sizing, so we should no longer ever see vg outputting a too-large Protobuf message. If there's a single truly massive node, it will throw an error rather than emit something that can't be loaded.

PR #727, opened 2017-03-15
Fix seg fault in banded global aligner

jeizenga:
Resolves the bug we discussed last meeting that was causing the BandedGlobalAligner tests to fail on Ubuntu systems. 

@mlin @edawson (or anyone else running Ubuntu)
Would appreciate if someone pulls this branch and tests it on an Ubuntu system before merging. I had it working in the Docker image Mike linked, but I'm developing on a Mac so I haven't tested it outside Docker.


mlin:
@jeizenga looking good thanks!!

```
$ sudo docker run quay.io/vgteam/vg:v1.4.0-2428-g93e6cfe9-build vg test
index file /tmp/vgXXTTLpRZ.fai not found, generating...
All tests passed (4812 assertions in 99 test cases)
```

PR #728, opened 2017-03-16
Mapper spring cleaning

ekg:
This PR will accumulate a number of command line API breaking changes and also parameter settings to `vg map` that will get it ready for checkpointing tomorrow.

The first commit removes reference to rocksdb index in vg map. The code is still there but now inaccessible. Using the -d argument to vg map we can specify a single base name for both xg and gcsa indexes.


ekg:
I did not mention this in the previous commit, but this also changes some of the default parameters in the mapper quite a bit. It additionally eliminates the concept of cluster mapping quality from the CLI and it is disabled internally.

ekg:
OSX builds are still consistently failing, but I'm going to merge.

PR #732, opened 2017-03-17
Adopt vcflib that can handle an indexed vcf with no variants

adamnovak:
Needs to wait on https://github.com/vcflib/vcflib/pull/191 before it will build.


adamnovak:
@ekg, can you merge this whole chain?

edawson:
@ekg could we do a vgteam vcflib fork for prototyping useful changes?

adamnovak:
I think these changes have gotten pulled in.

adamnovak:
Actually no, looks like vcflib still needs this from tabixpp: https://github.com/ekg/tabixpp/pull/14

@ekg?

PR #737, opened 2017-03-21
Revamped VG graph manipulation algorithms

jeizenga:
- Rewrote `VG::unfold` to have more consistent semantics. No longer has artifacts resulting from arbitrarily preferring the forward strand around inversions. Providing extremely long unfold distances near an inverting edge does not always produce the equivalent unidirected graph anymore.
- Added a new `VG::split_strands` algorithm that converts to the equivalent unidirected sequence graph.
- Added tests for both algorithms.
- Switched all node translators to unordered maps, which are better.

### DIFF
--- a/src/genotyper.cpp
+++ b/src/genotyper.cpp
@@ -651,7 +651,7 @@ vector<Genotyper::Site> Genotyper::find_sites_with_supbub(VG& graph) {
     // Unfold the graph
     // Copy the graph and unfold the copy. We need to hold this translation
     // table from new node ID to old node and relative orientation.
-    map<vg::id_t, pair<vg::id_t, bool>> unfold_translation;
+    unordered_map<vg::id_t, pair<vg::id_t, bool>> unfold_translation;
     auto transformed = graph.unfold(unfold_max_length, unfold_translation);
     
     // Fix up any doubly reversed edges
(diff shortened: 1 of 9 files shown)


src/unittest/vg.cpp
```diff
@@ -143,55 +144,888 @@ TEST_CASE("is_acyclic() should return whether the graph is acyclic", "[vg][cycle
...
+            }
+            else if ((e.from() == node_9 && e.to() == node_6 && !e.from_start() && !e.to_end()) ||
+                     (e.from() == node_6 && e.to() == node_9 && e.from_start() && e.to_end())) {
+                found_edge_3 = true;
+            }
+            else if ((e.from() == node_6 && e.to() == node_5 && !e.from_start() && !e.to_end()) ||
+                     (e.from() == node_5 && e.to() == node_6 && e.from_start() && e.to_end())) {
+                found_edge_4 = true;
+            }
+            else if ((e.from() == node_5 && e.to() == node_3 && !e.from_start() && !e.to_end()) ||
```
adamnovak: As much as I love your thorough test code, maybe this could be compressed through the use of a loop?

src/genotyper.cpp
```diff
@@ -651,19 +651,19 @@ vector<Genotyper::Site> Genotyper::find_sites_with_supbub(VG& graph) {
...
 
     // Now dagify the graph. We need to hold this translation table from new
     // node ID to old node and relative orientation.
-    map<vg::id_t, pair<vg::id_t, bool>> dag_translation;
+    unordered_map<vg::id_t, pair<vg::id_t, bool>> dag_translation;
     transformed = transformed.dagify(dagify_steps, dag_translation);
     
     // Compose the complete translation
-    map<vg::id_t, pair<vg::id_t, bool>> overall_translation = transformed.overlay_node_translations(dag_translation, unfold_translation);
+    unordered_map<vg::id_t, pair<vg::id_t, bool>> overall_translation = transformed.overlay_node_translations(dag_translation, unfold_translation);
```
adamnovak: As part of these changes, do you want to adjust the signature of `map<id_t, vector<Mapping>> PathIndex::parse_translation(const Translation& translation)`? Or is that `Translation`-handling code logically distinct enough from this code that works with "translations" that get passed around in a different format?

And should we think about settling on a single internal format for translations? The `PathIndex` code that works with them uses `Mappings` so it can potentially express translations back to parts of original nodes, whereas this code here only thinks about entire nodes (because that's all it needs to). `Translation` objects can in general map from parts of nodes to parts of other nodes, but in practice they're always emitted following particular, not-documented-in-the-protobuf constraints that prevent that.

PR #744, opened 2017-03-23
Improving MEM chaining and mapping quality

ekg:
This implements a number of changes that relate to the performance (both runtime and accuracy) of the MEM chain model.

Earlier commits focus on working around a problem that led to an explosion in the size of the model when encountering repeats.

Later commits resolved some issues that came up in this adaptation and update the mapping quality calculations to mirror those in bwa mem.


adamnovak:
Looks like it's failing one of the paired end tests:

```
not ok 22 - paired read alignments forced to be consistent have lower score than unrestricted alignments
#          got: '0'
#     expected: '1'
```

src/alignment.hpp
```diff
@@ -98,6 +98,7 @@ Alignment reverse_complement_alignment(const Alignment& aln, const function<int6
 vector<Alignment> reverse_complement_alignments(const vector<Alignment>& alns, const function<int64_t(int64_t)>& node_length);
 int softclip_start(const Alignment& alignment);
 int softclip_end(const Alignment& alignment);
+int query_overlap(const Alignment& aln1, const Alignment& aln2);
```
adamnovak: What is the purpose of this function? I guess getting the overlap between the two Alignments? Should it count just the outermost piece of overlap or all the overlap? How does it count indels and substitutions and things?

src/gssw_aligner.cpp
```diff
@@ -729,16 +742,36 @@ void BaseAligner::compute_paired_mapping_quality(pair<vector<Alignment>, vector<
     if (use_cluster_mq) {
         mapping_quality = prob_to_phred(sqrt(phred_to_prob(cluster_mq + mapping_quality)));
     }
-    
-    // scale mapping quality by half to match bwa mem
+
     mapping_quality /= 2;
```
adamnovak: Now there's no explanation at all for this division.

src/mapper.cpp
```diff
@@ -220,26 +225,30 @@ void Mapper::clear_aligners(void) {
...
     if (mismatch > max_score) max_score = mismatch;
     if (gap_open > max_score) max_score = gap_open;
     if (gap_extend > max_score) max_score = gap_extend;
     
     double gc_content = estimate_gc_content();
 
-    qual_adj_aligners.resize(alignment_threads);
-    regular_aligners.resize(alignment_threads);
+    //qual_adj_aligners.resize(alignment_threads);
+    //regular_aligners.resize(alignment_threads);
```
adamnovak: Do you want to leave commented-out code lying around?

src/vg.proto
```diff
@@ -120,6 +120,7 @@ message Alignment {
     repeated Path fragment = 17; // An estimate of the length of the fragment, if this Alignment is paired.
     repeated Locus locus = 18; // The loci that this alignment supports. TODO: get rid of this, we have annotations in our data model again.
     repeated Position refpos = 19; // Position of the alignment in reference paths embedded in graph
+    double uniqueness = 20; // Meant to encode the uniqueness of the mapping based on its seeds
```
adamnovak: We should explain what uniqueness is here. Is 0 "not unique" and 1 "completely unique"? What does a uniqueness of 0.5 mean?

src/alignment.hpp
ekg: This is the length of the overlap between the two alignments assuming they are both alignments of the same query.

src/vg.proto
ekg: It's the fraction of bases in the read that have non-repetitive seeds. I'll indicate this in the protobuf.

src/mapper.cpp
ekg: I've removed it.

src/gssw_aligner.cpp
ekg: I've removed this as well.

ekg:
Although the push is failing, I think this is due to problems with caching and submodule changes.

I've been able to test this and see that it resolves the performance issues that @mlin ran into while being overall slightly better than v1.5.0 in terms of its accuracy. I am running large scale tests now. I'll merge :crossed_fingers:

mlin:
@ekg Okay I'm kicking off a docker image build and will try with my version of map_sim too. No parameter changes, right?

PR #748, opened 2017-03-27
Improve vg call VCF output and options system

adamnovak:
I've added merging of identical bases on either end of a VCF record (back?) to vg call.

I also have a command line options system I've written, so that we can define options for these thing-doer objects in one place (the class hpp) and have them automatically come out to the command line with their names, descriptions, and default values. This should save a lot of repetitive option parsing code if we apply it in more places.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5435,10 +5435,13 @@ int main_map(int argc, char** argv) {
     };
 
     for (int i = 0; i < thread_count; ++i) {
-        Mapper* m;
+        Mapper* m = nullptr;
         if(xindex && gcsa && lcp) {
             // We have the xg and GCSA indexes, so use them
             m = new Mapper(xindex, gcsa, lcp);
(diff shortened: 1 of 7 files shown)


src/main.cpp
```diff
@@ -5435,10 +5435,13 @@ int main_map(int argc, char** argv) {
     };
 
     for (int i = 0; i < thread_count; ++i) {
-        Mapper* m;
+        Mapper* m = nullptr;
         if(xindex && gcsa && lcp) {
             // We have the xg and GCSA indexes, so use them
             m = new Mapper(xindex, gcsa, lcp);
+        } else {
```
yhoogstrate: I think this does not need to happen in every iteration.

adamnovak: Yeah, I'm not sure why we're creating and tearing down the mapper all the time (or why we need a separate mapper for each thread).

PR #759, opened 2017-04-05
Rewrite vg call to be Visit-based and recursive

adamnovak:
Now vg call will recursively handle all nested sites in a tree, emitting multiple Loci, or multiple VCF records when parents and children are both on a primary path.

This appears to slightly improve calling performance on @glennhickey 's `bakeoff.sh` script, in my testing (which I did before rebasing onto master).

I've also added the XSEE tag back to the generated VCFs, to point to the un-augmented nodes that are relevant for a variant.

I also improved the way we build `libvg.a` in the Makefile, so we don't dump source files into the static library's archive.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,32 +161,35 @@ DYNAMIC_DIR:=deps/DYNAMIC
 SSW_DIR:=deps/ssw/src
 STATIC_FLAGS=-static -static-libstdc++ -static-libgcc
 
+# Dependencies that go into libvg's archive
+LIB_DEPS =
+LIB_DEPS += $(LIB_DIR)/libprotobuf.a
+LIB_DEPS += $(LIB_DIR)/libsdsl.a
+LIB_DEPS += $(LIB_DIR)/libssw.a
+LIB_DEPS += $(LIB_DIR)/libsnappy.a
+LIB_DEPS += $(LIB_DIR)/librocksdb.a
+LIB_DEPS += $(LIB_DIR)/libgcsa2.a
+LIB_DEPS += $(OBJ_DIR)/Fasta.o
+LIB_DEPS += $(LIB_DIR)/libhts.a
+LIB_DEPS += $(LIB_DIR)/libxg.a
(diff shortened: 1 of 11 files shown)


adamnovak:
@jeizenga This has a horrible linear-time `SnarlManager::manage()` function that just scans the vector and finds the correct snarl, so overall this implementation is going to be O(n^2) in the number of snarls.

edawson:
@adamnovak could you fill out the names to be unique (e.g. StartNode_EndNode) and place them in a map name->Snarl ?

adamnovak:
The Snarls inside Visits won't have the names set. I could index it all by
a start node, end node pair, or Jordan's Key Form thing.

adamnovak:
Here's the performance under the `bakeoff.sh` script from toil-vg:

```
real    295m47.050s
user    747m35.055s
sys     7m36.805s
BRCA1   0.9483
BRCA2   0.8812
SMA     0.7903
LRC-KIR 0.7338
MHC     0.8611
```

@glennhickey How does this compare to what you're getting with master?

glennhickey:
This is what I got from  v1.5.0-174-g12a70a7f-t45 (most recent I have on
hand)

BRCA1    snp1kg    0.9452
BRCA2    snp1kg    0.8889
SMA    snp1kg    0.7955
LRC-KIR    snp1kg    0.7401
MHC    snp1kg    0.8634

I'm just finishing a pytest version of this which will make sensible
jenkins output, and keep an archive of benchmarks for each version run.

adamnovak:
So it looks like my changes make everything go down slightly, except BCRA1 which improves a tiny bit.

How much does performance have to drop to block a merge? Do we require strictly improving performance for everything? Or can we merge this now and fix it later?

glennhickey:
I think we have to allow for some wiggle room, especially for the bakeoff
f1s.  I think you're fine to merge in this case.

adamnovak:
Alright. In it goes.

PR #764, opened 2017-04-08
Fix some bugs I introduced in the caller

adamnovak:
My new caller code had a couple of bugs in the way average supports were being calculated, because I forgot to initialize some things.

I also have added some code to switch between average and min support for snarls of different sizes, because we don't want to use average support on adjacent errors (because it makes the errors look good), but we don't want to use minimum support on long possibly-deleted reference segements (because it makes them look really bad if there are any coverage low spots).

The real solution is something clever that accounts for read-explained-ness, but we can't do that in the pileup framework.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -475,13 +475,13 @@ $(UNITTEST_OBJ_DIR)/banded_global_aligner.o: $(UNITTEST_SRC_DIR)/banded_global_a
 
 $(UNITTEST_OBJ_DIR)/pinned_alignment.o: $(UNITTEST_SRC_DIR)/pinned_alignment.cpp $(UNITTEST_SRC_DIR)/catch.hpp $(SRC_DIR)/gssw_aligner.hpp $(DEPS)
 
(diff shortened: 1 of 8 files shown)


adamnovak:
Here are the F1s I get out of this branch:

```
BRCA1	0.9547
BRCA2	0.8812
SMA	0.7967
LRC-KIR	0.7364
MHC	0.8677
```

That's more than I got with what I merged to master last, so I think this is an improvement. The numbers are quite sensitive to the default bias limit parameters; we should try and guess good values across all the regions somehow.

ekg:
What are the default bias limit parameters?

adamnovak:
They used to be 3 for bias towards an alt, and 4 for bias towards a
reference. I upped them both to 4.5.

If the coverage of one alt is 4.5 times the coverage of another, the minor
allele will be disregarded and the caller will emit a homozygous site.

Really this should be replaced with some kind of distribution that feeds
into the quality score.

PR #767, opened 2017-04-12
improve the runtime of the mapper

ekg:
Estimate the maximum mapping quality from the distribution of max LCPs for each MEM and the lengths of the MEMs.

Use this to not try to make too many clusters when we have obviously good or bad reads.

This apparently comes at no cost to accuracy (in yeast).

To test on human...

Alternative improvement to test: only reseed when we see the estimated max mapping quality below a certain value.


ekg:
This branch implements a mapping quality estimator based on the GCSA2 suffix tree. This estimate is used in two ways. 1) it is mixed with the mapping quality via a geometric mean as cluster-mq was before, which helps to down weight over-confident alignments for which the tree suggests many close alternates, 2) it is used to scale the number of alignments that we attempt. The scaling is a trivial division of the maximum number of attempts `--try-up-to` by the estimated quality, with a lower bound at `--try-at-least` (4 by default). This allows us to avoid work when the alignment is very likely to be good. This and other small improvements have allowed us to gain a fair amount of performance.

[human_sim.1M-d857fc5.roc.pdf](https://github.com/vgteam/vg/files/945082/human_sim.1M-d857fc5.roc.pdf)
[human_sim.1M-d857fc5.qq.pdf](https://github.com/vgteam/vg/files/945080/human_sim.1M-d857fc5.qq.pdf)

[yeast_sim.1M-d857fc5.roc.pdf](https://github.com/vgteam/vg/files/945081/yeast_sim.1M-d857fc5.roc.pdf)
[yeast_sim.1M-d857fc5.qq.pdf](https://github.com/vgteam/vg/files/945079/yeast_sim.1M-d857fc5.qq.pdf)

ekg:
Also, this branch removes all the extra mapping functionality that we had floating around in mapper.cpp. Some more cleaning could be in order, but this should trim things down to be less confusing to reviewers or other interested people.

PR #768, opened 2017-04-12
Change second-best traversal selection logic

adamnovak:
Now we choose the second-best traversal based on the coverage of graph elements not used by the best traversal.

I've also added the SVLEN tag back in.


adamnovak:
Here are the performance numbers I got:

```
BRCA1	0.9544
BRCA2	0.8779
SMA	0.8017
LRC-KIR	0.7383
MHC	0.8673
```

PR #774, opened 2017-04-20
Enable large insert calling

adamnovak:
Previously, we would only explore to a max depth of 10 visits when looking for snarl traversals with the `RepresentativeTraversalFinder`, in order to prevent combinatorial explosions when braids were present off the primary path/backbone traversal of a snarl.

This prevented us from calling large inserts, involving more than 10-20 nodes or nested snarls, because the required insert traversals were too long to ever be enumerated.

I tried lifting the limit, but that resulted in combinatorial explosions, so I added a "max width" limit on the total size of the BFS search queue. If there are too many potential paths in the queue, you can only add one extension of every path you pop off to consider.

This seems to work fine, speed-wise. The accuracy results have improved over my last PR:

```
BRCA1   snp1kg  0.9544
BRCA2   snp1kg  0.8779
SMA     snp1kg  0.8024
LRC-KIR snp1kg  0.7394
MHC     snp1kg  0.8677
```

I also see some large inserts being called now, when I use a graph that contains them and when I make sure to include the improperly paired/mate-not-in-subregion reads in my input. The next step after merging this is to assess and improve on their accuracy.

### DIFF
--- a/src/call2vcf.cpp
+++ b/src/call2vcf.cpp
@@ -603,20 +603,20 @@ vector<SnarlTraversal> Call2Vcf::find_best_traversals(AugmentedGraph& augmented,
     // TODO: generalize it
     assert(site.type() == ULTRABUBBLE);
 
-#define debug
+
 #ifdef debug
     cerr << "Site " << site << endl;
 #endif
-#undef debug
+
 
     // Get traversals of this Snarl, with Visits to child Snarls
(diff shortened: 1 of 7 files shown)


adamnovak:
Another tweak I made is to skip substitutions of "N" for things when augmenting the graph. It makes no sense to call a something -> N variant.

PR #777, opened 2017-04-21
avoid redundant clusters in MEM chain model traceback

ekg:
The MEMChainModel was spitting out duplicate clusters for those nodes that we had marked redundant in the construction phase of the model. By recording these redundant nodes in the model and excluding them from scoring we can make fewer attempts to get the correct alignment.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -975,12 +975,6 @@ pair<vector<Alignment>, vector<Alignment>> Mapper::align_paired_multi(
 
         ++multimaps;
 
-#ifdef debug_mapper
-#pragma omp critical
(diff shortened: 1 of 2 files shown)


ekg:
These improvements to the MEM chainer come due to an investigation into reads that were getting a lower mapping quality in the human pan genome versus the linear reference (e.g. pan:0 vs ref:60). Thanks to @mlin for tracking this down. Please let me know if this resolves the cases you found.

The changes result in an improvement in runtime at no apparent cost in ROC AUC:

human_sim 1m-4c12438 roc
yeast_sim 1m-4c12438 roc

PR #782, opened 2017-04-26
Resolve a few problems that @mlin ran into in human testing

ekg:
We've been working through anecdotes featuring strange alignments.

The changes here should resolve most of what @mlin has run into these past few days.

### DIFF
--- a/src/gssw_aligner.cpp
+++ b/src/gssw_aligner.cpp
@@ -683,8 +683,6 @@ void BaseAligner::compute_mapping_quality(vector<Alignment>& alignments,
         mapping_quality = prob_to_phred(sqrt(phred_to_prob(cluster_mq + mapping_quality)));
     }
 
-    mapping_quality += quality_scale_factor * log(1 - alignments[max_idx].uniqueness());
-
     double identity = (double)alignments[max_idx].score() / (alignments[max_idx].sequence().size() * match);
     mapping_quality *= identity;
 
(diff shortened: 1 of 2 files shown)


mlin:
@ekg cool -- the example I found with mapQ=0 but no reported secondary mapping, is the explanation that the mapQ had been penalized by overlap?

ekg:
It was a mathematical error. I was taking the log of 0 for cases with
perfect uniqueness. This yielded an MQ of 0 in this case. I just got rid of
the whole idea of uniqueness in the MQ calculation and it seems to actually
have helped the yeast tests. I haven't been able to do a human sim test
today because our systems are not working. Let me know if this PR fixes
other examples you've found.

ekg:
@glennhickey when trying to view the jenkins report I get this (after logging in)

image

ekg:
Here are the results in the 1M sim. Seems fine.

# yeast

[yeast_sim.1M-f443b47.roc.pdf](https://github.com/vgteam/vg/files/961418/yeast_sim.1M-f443b47.roc.pdf)

# human

[human_sim.1M-f443b47.roc.pdf](https://github.com/vgteam/vg/files/961417/human_sim.1M-f443b47.roc.pdf)

The human results feature an error, which is that we were ignoring the mtDNA and dummy sequences that bwa mem is using.

glennhickey:
thanks for letting me know.  vgteam's supposed to be whitelisted but I'm
not surprised that didn't work on the first try.  will try to get it sorted
out (but the entire server seems down atm).

Issue #781, opened 2017-04-25
How does vg handle phased and unphased Genotypes in VCF?


subwaystation:
Hi vgteam!

I am trying to understand how `vg` handles different phenotypes of a diploid organism like `1|0`and `1|2` (phased) or  `1/0`and `1/2` (unphased). Therefore I took a look at your tiniest example data set:

```
vg construct -v tiny.vcf.gz -r tiny.fa > tiny.vg
```
```
vg view -j tiny.vg > tiny.json
```
But only path `x` appears in the JSON. What am I missing?

Even the following 
```
vg view -jpnw tiny.vg > tiny.json
```
Does not solve the issue. I would expect to see at least 3 paths there, right?
[tiny_SeqTubeMap.pdf](https://github.com/vgteam/vg/files/954300/tiny_SeqTubeMap.pdf)

Viewing the graph in `dot` format:
```
vg view -dpnw tiny.vg | dot -Tsvg -o tiny.svg
```
This looks much better in the browser. But still only one real path is drawn.
[tiny_Dot.pdf](https://github.com/vgteam/vg/files/954301/tiny_Dot.pdf)

Thanks!

edawson:
Hey @subwaystation ,

Sorry this has sat a few days. There short answer is that we don't do anything with haplotype info. 

Paths in the graph come from the reference ('x' in this example) or the alleles (all those 'alt_[a-z0-9]*_[0-9]* paths). We currently construct graphs so that all *possible* haplotypes (not just those that are *observed*) are present in the graph's structure.

We don't differentiate between phased or unphased genotypes, and if you put two adjacent phased SNPs in you'd get the same structure as with unphased SNPs.

Does that make sense? It's a bit different in design from how a lot of pangenomes / cDBGs do things.

subwaystation:
Hey @edawson,

thanks, that helped me a lot. Now I understand what is going on.

However, is there any planning to support a more differentiated haplotype model when constructing a graph? Or could you point out to me, why this may not be necessary any more?

edawson:
Technically, we support labeling the haplotypes as Paths, though it's kind of a nontrivial workflow. You could for example label all phased/unphased haplotypes as such, or (one day) phase using the graph. You could then look at how many reads support a given haplotype and quantify how much of a given sample is present in the reads (e.g. for metagenomics). We do support gPBWT and gBWT queries for indexing paths but that's something either @ekg or @adamnovak would know more about than I.

[...]

PR #787, opened 2017-05-02
Yohei's vg trace implementation

adamnovak:
We're getting people trying to use the tube map (#781), which means we need `vg trace` to be in vg master.

@yoheirosen, do you want to resolve the merge conflicts that are happening here?


yoheirosen:
Hi Adam,

Sorry, I had been traveling. I will get on this.

yoheirosen:
I'm actually also going to remove the "weighted haplotype list"
functionality from vg trace, since it is the only component which depends
on haplotypes.*pp; which I would like to merge in separately at a later
time. I don't think Wolfgang was really using this anyway but I'll check.

yoheirosen:
Let's drop this pull request in favor of (https://github.com/vgteam/vg/pull/793) as to leave out the `vg haplo` stuff

PR #789, opened 2017-05-04
Update vg.proto

benedictpaten:
Added side, added sides to snarl, added pointer for expressing chains in snarls.

I'm sure this is all misinformed, I just want to have a conversation about expressing chains in snarls and clarify my understanding of the existing snarl DS.


ekg:
@benedictpaten this has gone stale. Would you like to resurrect it?

benedictpaten:
This is stale and superseded by later PRs to bring in cactus sites

PR #791, opened 2017-05-05
Change Filtering

adamnovak:
I've made some changes to filtering, to try and improve SV calling performance.

Most of the filters are off by default, except for one on the likelihood of the binomial assignment of support to alleles. OIf the likelihood of minor-allele-depth successes in total-depth trials is less than `e^-9`, by default, the site will be filtered.

I'm also reporting a bunch of extra depth metadata in the VCF, to try and inform filtering choices. Each call now has the global average depth and the local average depth for its bin noted on it.

For SV calling I wanted to use a filter that scraps a variant if its depth is more than 2.5 times the global average depth for the primary path it's on, because there are some false positive SVs with depths in the hundreds that I want to filter out. However, SMA in the bake-off regions has a really low global average depth, so that filter was counterproductive there and it defaults to off.

I've also wired the min AD filter's value into the second-best allele selection logic, so the caller should no longer call a site as best/second best, and then immediately filter it out because the second best allele's depth is too low.

### DIFF
--- a/src/call2vcf.cpp
+++ b/src/call2vcf.cpp
@@ -597,7 +597,8 @@ vector<SnarlTraversal> Call2Vcf::find_best_traversals(AugmentedGraph& augmented,
 #endif
 
 
-    // Get traversals of this Snarl, with Visits to child Snarls
+    // Get traversals of this Snarl, with Visits to child Snarls.
+    // The 0th is always the reference traversal if we're on a primary path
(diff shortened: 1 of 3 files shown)


adamnovak:
The relevant bakeoff statistics are:

```
BRCA1	snp1kg	0.9426
BRCA2	snp1kg	0.8779
SMA	snp1kg	0.7659
LRC-KIR	snp1kg	0.7575
MHC	snp1kg	0.8634
```

PR #795, opened 2017-05-08
resolve failure to transmit unaligned read sequence and quality through

ekg:
@mlin caught this. If this resolves his test case then I'll merge. Unfortunately, I can't seem to find an equivalent test case for a small example (his involves the 1000G+h37 ref).


mlin:
looks like it worked; thanks!

PR #803, opened 2017-05-15
Dramatically improve the long read aligner

ekg:
Before this PR the long read aligner was barely working for reads with high error rates. We were seeing 40% reproduction coming out of `vg map --compare` with simulated reads at 10% SNP and 5% indel error.

By resolving a number of problems with the banded alignment algorithm I am now able to achieve higher correspondence between the alignment and the simulated truth

Here are things I'm doing to test the method:

```
vg construct -r 1mb1kgp/z.fa -v 1mb1kgp/z.vcf.gz -m 32 -p >z.vg  
vg index -k 16 -x z.xg -g z.gcsa -p z.vg
vg sim -x z.xg -n 1000 -l 10000 -e 0.1 -i 0.05 -a >simpb.gam
vg map -d z -G simpb.gam -w 256 -W 128 -B 2 --compare
# shows correctness > identity for almost all reads!
```

We can also apply this improvement to get better long read progressive assemblies. Try:

```
vg msga -f GRCh38_alts/FASTA/HLA/DQB1-3119.fa -w 1024 -B 2 -W 64 -E 3 -J 32 -t 4 -D >DQB1-3119.vg
```

Visualizing this graph with the [sequence graph tube map demo](https://wolfib.github.io/sequenceTubeMap/) shows some beautiful structures being pulled out by the alignment in a concise (not "shattered" or "exploded") kind of way:

image

[The Bandage assembly viewer](https://github.com/rrwick/Bandage) shows a single, compact representation of these sequences, which is to be expected:

image

I was previously unable to get this kind of coherent structure, and that matches the kinds of results I was getting from the long read simulations.

The improvement hinges on @jeizenga's banded global alignment algorithm. Without that, the patching wouldn't work reasonably and this would be impossible. :cheers:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -493,7 +493,7 @@ $(UNITTEST_OBJ_DIR)/genotypekit.o: $(UNITTEST_SRC_DIR)/genotypekit.cpp $(UNITTES
 
 $(UNITTEST_OBJ_DIR)/readfilter.o: $(UNITTEST_SRC_DIR)/readfilter.cpp $(UNITTEST_SRC_DIR)/catch.hpp $(SRC_DIR)/readfilter.hpp $(DEPS)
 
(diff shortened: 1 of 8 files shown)


ekg:
As usual, it turns out many of the "beautiful structures" seem to be resulting from various problems with the aligner. With these fixed, things seem to be going well for difficult HLA genes like DRB1-3123. However, it's still not clear if there is a good general parameter set we can apply to MSGA. To record what I used:

```
vg msga -f GRCh38_alts/FASTA/HLA/DRB1-3123.fa -w 1024 -B 4 -u 16 -W 256 -E 3 -J 64 -D >DRB1-3123.vg
```

image

PR #809, opened 2017-05-30
Fixes for vg call

glennhickey:
The whole genome run on HG002 brought some calling problems to the forefront.  This is a patch for the two biggest issues found so far:
* calling homozygous alts for the sole reason that a het would fail the min-ad filter
* missing snps that are adjacent to homozygous alt snps, because no augmented path to explain them


glennhickey:
Jenkins, test this please.

PR #810, opened 2017-05-31
Surject Good

adamnovak:
On top of Glenn's changes from #809, this includes some fixes and new tests for surject. Surjection should now properly convert the `/1` and `/2` notation we use for paired-end reads to a single shared name with end1/end2 flags in SAM format. The original alignment's mapping quality should also be preserved through surjection.

Also, `vg surject` is now a subcommand.


glennhickey:
I'm still working on #809, which seems to increase indel false positives in
1kg graph even thouhg overall accuracy goes up.

adamnovak:
Here it is without #809.

PR #811, opened 2017-05-31
Count support for shared nodes as half

adamnovak:
This may help with the problems @glennhickey was encountering.

### DIFF
--- a/src/call2vcf.cpp
+++ b/src/call2vcf.cpp
@@ -429,38 +429,39 @@ void trace_traversal(const SnarlTraversal& traversal, const Snarl& site, functio
 
 /**
  * Get the min support, total support, bp size (to divide total by for average
- * support), and min likelihood for a traversal, optionally excluding the
- * material used by another traversal.
+ * support), and min likelihood for a traversal, optionally special-casing the
+ * material used by another traversal. Material used by another traversal only
+ * makes half its coverage available to this traversal.
  */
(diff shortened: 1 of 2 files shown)


adamnovak:
@glennhickey Does this make our calling better or worse? Jenkins is supposed to tell me that, right? Where do I find its conclusions?

glennhickey:
Sorry, I got lost down a git hole chasing noise, and don't have anything
conclusive for this yet.  I'll have results for the cases I was looking at
on chr21 by first thing tomorrow.

The jenkins test results are here

http://jenkins.cgcloud.info/job/vg/43/testReport/jenkins.vgci/VGCITest/

Click any test and look at the standard output section,  ex mhc snp1kg is
here:

http://jenkins.cgcloud.info/job/vg/43/testReport/jenkins.vgci/VGCITest/test_map_mhc_snp1kg/
Standard out is this:

F1: 0.8608  Baseline: 0.8607  Threshold: 0.02

So it's saying a 0.0001 F1 boost over last successful build.  Most tests
are identical.  The threshold of 0.02 is what's used to see if the test
passed, and really needs to be tightened now.

glennhickey:
This change fixes the example I sent, but is having surprisingly little
impact on my chr21 tests.  Digging into it now.

Freebayes/samtools/platypus calls are just about ready for surjected bams,
which will be quite interesting too.

adamnovak:
It looks like it has almost no effect on the results in the tests, so I am going to merge this in.

PR #816, opened 2017-06-06
fix sam pairing flags in surject

glennhickey:


glennhickey:
Still need to set PNEXT and TLEN fields.  Not easy to do in single scan / parallel / unsorted input.

glennhickey:
Just learned samtools fixmate is a thing.  This means
`vg surject gam -x xg -b | samtools sort -n - | samtools fixmate - surject.bam`
should work

glennhickey:
Jenkins, test this please

Issue #818, opened 2017-06-09
Rocksdb alignment from node lookup returns random stuff


glennhickey:
I noticed this while extracting a gam from chr21 from @mlin's latest vg alignment against the linear reference:
`vg chunk -x index.xg -a HG002-NA24385-30x-ref.gam.index -p 21:282160186-286392791 -t 1 -b rechunk -i -c 0`
This command just calls Index::for_alignment_to_nodes() on each id in the range and dumps the results.. https://github.com/vgteam/vg/blob/master/src/chunker.cpp#L137-L149
But the output contains Alignments such as below, whose ids are all well outside the range (over in chromosome 10).

I haven't found a case where an alignment that should be returned isn't yet. 

Is this by design where we expect a certain number of hash collisions, or is there something wrong? 

[...]

PR #824, opened 2017-06-14
Filter out-of-range alignments  in vg chunk

glennhickey:
Make sure at least one node from each alignment returned from RocksDB is actually in the specified id set.  This is a workaround for #818


glennhickey:
Jenkins, test this please

PR #825, opened 2017-06-18
Named thread support

ekg:
This updates xg to a version that supports named threads and queries of them. Thread names are stored in a CSA, allowing for efficient storage of repetitive collections. Thread starts can be queried using a wavelet tree. `vg find` is updated to take `-q` parameter, which is a thread name start pattern. `vg construct` generates threads with names that reflect the sample names and haplotypes. This allows querying a subset of threads corresponding to the first or second haplotype, which is essential for our simulation experiments.


ekg:
@glennhickey do you know what is going on with Jenkins? I can't see the failing tests.

ekg:
It's impossible for me to know what's going on in jenkins. We need to move forward and so I'm going to merge this in.

glennhickey:
@ekg We gotta get you able to see the Jenkins output.  What happens when
you click on it?  Can you not see this page?
http://jenkins.cgcloud.info/job/vg/
Anyone on vgteam should be able to get there with a git login.

ekg:
I see "access denied"

glennhickey:
How about now?

ekg:
That works! I can see what's up.

glennhickey:
Good news, the Jenkins config options are a little cryptic.  Still, I'll
make a wiki and fix it up so that most errors found can be reproduced on
the command line independently.   The only problems right now are related
to the xg format change... Still need to fix up a couple tests to deal with
that, it seems.

PR #829, opened 2017-06-21
correct a library name for clean target in Makefile

byoo:
correct gfakluge library name in Makefile


glennhickey:
Jenkins, ok to test

Issue #830, opened 2017-06-21
alignment score not stable between forward and reverse complement of a set of alignments


ekg:
This simulation demonstrates a handful of alignments which get different scores when we alignment the forward vs reverse complement of a set of simulated reads.

```shell
time vg sim -s 1337 -x z.xg -n 100000 -l 100 -e 0.01 -i 0.002 -a | tee zsim.gam | vg view -a - | pv -l >/dev/null
```
Put this in a script called "revcomp":

```python
#!/usr/bin/env python

import sys
complement = {'A': 'T', 'C': 'G', 'G': 'C', 'T': 'A', 'N': 'N'}
for seq in sys.stdin:
    print "".join(complement.get(base, base) for base in reversed(seq.strip()))
```

From vg/test, make forward and reverse versions of the simulated reads:

```shell
paste <(vg view -a zsim.gam | jq .name) <(vg view -a zsim.gam | jq .sequence ) | sed 's/\"//g' | awk '{ print "{", "\"name\":", "\""$1"\",", "\"sequence\":", "\""$2"\"", "}" }' | vg view -JaG - >zsim.fwd.gam
```

```shell
paste <(vg view -a zsim.gam | jq .name) <(vg view -a zsim.gam | jq .sequence | revcomp ) | sed 's/\"//g' | awk '{ print "{", "\"name\":", "\""$1"\",", "\"sequence\":", "\""$2"\"", "}" }' | vg view -JaG - >zsim.rev.gam
```

Now align and sort by the difference in score between the two versions:

```shell
join <(vg map -d z -G zsim.fwd.gam -j  | jq -r '[.name, .score ] | @tsv' | sort) <(vg map -d z -G zsim.rev.gam -j  | jq -r '[.name, .score ] | @tsv' | sort) | awk '{ print $0, $2 - $3; }' | tr ' ' '\t' >tee zsim.scorediff.tsv
```

We see a number of alignments that get different scores:

```shell
cat zsim.scorediff.tsv| awk '{ if ($4 != 0) print }' | sort -n -k 4
```

```text
42da7c1844915c18        71      85      -14
5f88d0c0c7e4dbd8        80      88      -8
db12d7b55180c97f        82      90      -8
93db53ac27402b6c        90      95      -5
394cd65c77085870        78      81      -3
af5574384a6e88b0        83      86      -3
ecbcada23edeb083        85      88      -3
5acf323403b67a03        90      92      -2
24f88e2b941287a1        86      85      1
[...]

PR #831, opened 2017-06-22
polish the mapper

ekg:
Mitigate #830 by getting more subgraph in cluster_subgraph.

This doesn't completely resolve any discrepancies, even in my small test, but it is a step in the right direction. Here for testing and further exploration.


ekg:
This is the hold-one-out yeast simulation, holding out the strain BC187 from the yeast pangenome and measuring our ability to correctly align reads simulated from it. Previously the vg.ref and bwa mem curves did not exactly overlap, but here they do.

yeast-bc187 panlight sim-results-0d9799f roc

This is a significant improvement over past iterations. We're now spot-on bwa mem for reference-only mapping.

With the changes in this branch (including a more realistic simulation) it does appear that the addition of all the variants in these yeast strains is not helping us in the aggregate. Previous results presented on github and elsewhere were simulating from the pangenome (treating it like a markov model and sampling reads from it). This ignored haplotype frequency, which is obviously a problem for real applications, especially where LD decay is slow. @lh3 brought this up often as a criticism of our approach, and I am reasonably convinced that he was correct in his concerns.

The natural solution is to score alignments based on their agreement with a known panel of haplotypes. Practically, we would penalize novel recombination in the same way that we penalize novel variation in the alignment, and this penalty would propagate into the alignment score and thus mapping quality.

There has also been some work to implement haplotype-based indexing in clusters of variation. While this would improve performance and indexing time, it is not enough to remove recombinant haplotypes from the index as ultimately the local alignment algorithm scores alignments relative to the memoryless graph.

Unfortunately, we can't yet scale the haplotype index to the whole genome for the 1000G, so a complete solution to this problem is not readily achievable. We could do this with a hierarchical index as construction is the major bottleneck, and that would be sufficient to provide the recombination scoring. Alternatively, we could wait for @jltsiren's [gbwt](https://github.com/jltsiren/gbwt) to be ready and build from that.

ekg:
This ROC shows our state on the same holdout test as above. I've made a number of improvements to reseeding (reseeding whenever our MEM length, not LCP max is above the threshold) and also to mapping quality calculation (scaling the MQ estimate by the number of second-best mappings).

The first change ensures we don't get trapped by very-long exact matches that spuriously happen when we're aligning against the pangenome vs. the linear reference.

The second incorporates another feature of the alignment score distribution into the mapping quality.

I also have tweaked a parameter that scales the mapping quality by the score of the best alignment to the maximum possible score. This is cribbed from bwa mem, which this branch is now matching.

yeast-bc187 sim-results-9fedebe roc

ekg:
Current state of things on 1M read tests.

For 1M 150bp reads simulated from HG002_NA24385's haplotypes:

hg002 1m wg 6-results-f553b22 roc

1M 150bp reads simulated from BC187:

bc187 1m 2-results-f553b22 roc

Also, I wanted to check 100bp reads for the yeast. Here are 1M of them simulated from BC187.

bc187 1m 3-100bp-results-f553b22 roc

My objective is that vg performs equivalently to bwa mem for both pan and ref. I think we're basically there, at least in contrast to previous performance.

adamnovak:
retest this please

adamnovak:
Jenkins is a bit flaky with the plugin I am trying to make it use. It should build properly this time.

adamnovak:
It looks like this PR slows the mapper down a bit; the Jenkins build and test times go from about an hour to about 2 and a half hours: http://jenkins.cgcloud.info/job/vg/buildTimeTrend

That's fine, but we might want to look at speeding it up again somehow as our next project.

ekg:
@adamnovak  `make` is now recompiling everything, every time, even when I have only modified one file. Is that expected behavior?

adamnovak:
On Jenkins? I think so; it builds in Docker and doesn't do any caching.

On Travis, we definitely should not see that happening.

ekg:
@adamnovak the slowdown is largely related to bugfixes. For example, we had to slightly increase the amount of graph we picked up when aligning in order to avoid the bug where we saw different scores for forward and reverse alignments. Another example is the increase in the minimum number of mapping attempts, which was 4 but is now set at 16 (this makes a big difference). So we are trading accuracy for runtime. I'm sure we can improve things, but at this point we should focus on making sure the method works.

glennhickey:
Awesome, @adamnovak!  Will merge in some content for this shortly.

ekg:
@adamnovak the loss of `-ltcmalloc_minimal` from the linux compile flags meant a huge reduction in alignment speed. I see us going from 2k reads/s before I added it back to 3.4k reads/s as of cdeaefd.

adamnovak:
It looks like the test_sim_mhc_snp1kg test on Jenkins timed out for taking
more than 3600 seconds. You can up the timeout here:
https://github.com/vgteam/vg/blob/master/jenkins/vgci.py#L660

ekg:
3600 seconds is long, but I guess I'll up the time.

glennhickey:
test_sim_mhc_snp1kg used to take 7 minutes.

http://jenkins.cgcloud.info/job/vg/206/testReport/jenkins.vgci/VGCITest/

maybe it's especially struggling with the cactus graph.

ekg:
@glennhickey this seems to suggest that it isn't running it? http://jenkins.cgcloud.info/job/vg/206/testReport/jenkins.vgci/VGCITest/test_map_mhc_cactus/

adamnovak:
No, I think Glenn means the Cactus graph that's used in `test_sim_mhc_snp1kg`:

```
    @timeout_decorator.timeout(3600)
    def test_sim_mhc_snp1kg(self):
        """ Mapping and calling bakeoff F1 test for MHC primary graph """        
        self._test_mapeval(50000, 'MHC', 'snp1kg',
                           ['primary', 'snp1kg', 'cactus'],
                           score_baseline_graph='primary')
```

ekg:
I've downloaded the cactus graph, simulated some read pairs, and tested alignment.

```
vg sim -s 432 -n 1000 -e 0.01 -i 0.002 -l 150 -p 500 -v 50 -x cactus-MHC.xg -a | pv >mhc.sim
# runs out of memory on my laptop and is killed:
vg map -d cactus-MHC -iG mhc.sim --compare | pv -l >/dev/null
# single ended mapping runs fine:
vg map -d cactus-MHC -G mhc.sim --compare | pv -l >/dev/null
# paired end mapping is fine when we don't try to rescue
vg map -d cactus-MHC -iG mhc.sim -O 0 --compare | pv -l >/dev/null
```

So the problem seems to be that mate rescue is grabbing most of the graph in some cases. I can work around this with some kind of bound on the size of the graph that we get or we can disable the pair rescue for this test.

ekg:
The pair rescue is based off of an approximate distance metric that requires the graph to be partially sorted, at least in the region in which the rescue is being attempted. (NB: The indexing generates warnings suggesting that the graph paths aren't correctly sorted, so this may be part of the problem.)

Shifting to a path-based model isn't impossible, but I'm not sure it's going to solve the problem.

I'm not convinced that we should be trying to do pair rescue in degenerate assembly graphs. Doing this in a generic way seems like a recipe for problems.

Can we turn off the rescue for this test? I don't see a clear way to do so in the harness.

ekg:
I'm disabling the cactus MHC test. Using path positions for the rescue would be good, but it is complex and will affect the other results. I would like to make sure this is the only problem before merging, and I would like to merge back into master so that we don't have continued synchronization problems.

ekg:
@glennhickey @adamnovak do you understand what's failing? I would have expected that removing one failing test would reduce the total number of failures, but it seems to have increased?

glennhickey:
The problem is in the test_sim_**mhc**_snp1k test.  You removed cactus from the test_sim_**brca1**_snp1kg test.  Removing a given graph from one of these tests will not affect the total number of tests.

adamnovak:
It looks like the timing-out test still failed (since it wasn't changed), the test that *was* changed failed (because it checks all the graphs from the last test run against the current test run and it couldn't find the removed Cactus graph), and an unrelated test failed due to getting HTTP 500 errors from Quay when trying to download Docker containers (and so probably won't fail again next time).

If we do want to pull Cactus out of that test, we're going to have to change the comparison logic not to immediately fail if a graph that used to be there goes missing.

ekg:
I'm working on a fix for the underlying issue. And I'll leave the tests as
they were.

glennhickey:
Don't think it's timing out fwiw.  Jenkins is saying it ran 38 min and
shows stuff like this

CalledProcessError: Command '['docker', 'run', '--rm', '--log-driver',
'none', '--entrypoint', 'vg', '--env', 'TMPDIR=.', '-v',
'/mnt/ephemeral/tmp/toil-d049b061-3090-4b56-8e95-473347a75e01-dd1bac9cb23792a632bd6ee157719517/tmpL_ChX1/ae5ba0f9-2174-48f6-9825-213556542f31/tgM6apr:/data',
'-w', '/data', '--name', 'runchunkalignment--uTarYn37tUUO',
'jenkins-docker-vg-local', 'map', '-t', '8', '-G', 'reads_chunk_0_0.gam',
'--include-bonuses', '-i', '-x', 'graph.vg.xg', '-g', 'graph.vg.gcsa']'
returned non-zero exit status 139

E INFO:toil.lib.docker:Calling docker with ['docker', 'run', '--rm',
'--log-driver', 'none', '--entrypoint', 'vg', '--env', 'TMPDIR=.', '-v',
'/mnt/ephemeral/tmp/toil-d049b061-3090-4b56-8e95-473347a75e01-dd1bac9cb23792a632bd6ee157719517/tmpL_ChX1/ae5ba0f9-2174-48f6-9825-213556542f31/tgM6apr:/data',
'-w', '/data', '--name', 'runchunkalignment--uTarYn37tUUO',
'jenkins-docker-vg-local', 'map', '-t', '8', '-G', 'reads_chunk_0_0.gam',
'--include-bonuses', '-i', '-x', 'graph.vg.xg', '-g', 'graph.vg.gcsa'

E src/central_freelist.cc:333] tcmalloc: allocation failed 16384
E src/central_freelist.cc:333] tcmalloc: allocation failed 16384
E src/central_freelist.cc:333] tcmalloc: allocation failed 16384
E src/central_freelist.cc:333] tcmalloc: allocation failed 16384
E src/central_freelist.cc:333] tcmalloc: allocation failed 16384

ekg:
I've dug into my simulated example a bit further. I'm finding that the pair rescue induces fragment lengths that are very long, and these in turn throw a wrench in the paired end alignment by making every possible pair of MEMs in the graph "pairable". There may be simpler solutions than I expected.

So, I think the tests are failing due to out of memory-like conditions.

ekg:
I found a specific failure in the way that "perfect" pairs were computed that allowed very-long fragments to be considered concordant when they were presented in the reverse direction. Removing this seems to resolve the problem and hopefully will get us passing the jenkins regression tests.

adamnovak:
Looks like this time one of the tests [got an empty string from S3 instead of whatever test input data it wanted](http://jenkins.cgcloud.info/job/vg/257/testReport/junit/jenkins.vgci/VGCITest/test_full_brca2_primary/).

Jenkins retest this please.

adamnovak:
Also, did you mean to touch the xg and gssw submodules? It looks like this RP rolls back to an older xg version.

ekg:
I want to move forward with the submodules, so the xg GSSW rollback is
unintentional. I'll see what I can do to fix this.

adamnovak:
We probably should merge this one in since it's first.

What do we think of this BRCA1 ROC plot from Jenkins though:

plot

ekg:
@adamnovak that is concerning. I can dig into it.

ekg:
@adamnovak I wonder if the problems with the tests are due to there being multiple reference sequences in the graphs. This would lead to there being multiple positions for each read, which isn't really handled by the comparison scripts.

ekg:
10M reads, yeast SNP SGRP2 pangenome:

bc187 10m 3-results-96d719f roc

10M reads, human 1000G pangenome:

hg002 10m wg 1-results-96d719f roc

ekg:
I accidentally closed this rather than merging it.

PR #833, opened 2017-06-23
add min-end-match filter (-m) to vg filter

glennhickey:
This is similar to the overhang filter -o, but works similarly when mapping without -L 0


glennhickey:
Jenkins, test this please

PR #842, opened 2017-07-10
Make include guards be VG-prefixed

### LINKED ISSUE #42: Make include guards unique
I find that include guards like "[`INDEX_H`](https://github.com/ekg/vg/blob/4bbbd251aae0e753e6e326d14ea6c8b11be9de7f/index.hpp#L1)" and "[`PATH_H`](https://github.com/ekg/vg/blob/0f4e94613213e983ef3e030321efd20f5036b3e1/path.hpp#L1)" are too short for the safe reuse of your header files (when they belong to an application programming interface).

adamnovak:
Fixes #42

### DIFF
--- a/src/alignment.hpp
+++ b/src/alignment.hpp
@@ -1,5 +1,5 @@
-#ifndef ALIGNMENT_H
-#define ALIGNMENT_H
+#ifndef VG_ALIGNMENT_H
+#define VG_ALIGNMENT_H
 
 #include <iostream>
 #include <functional>
--- a/src/bin2ascii.h
+++ b/src/bin2ascii.h
@@ -5,8 +5,8 @@
  * it under the terms of the MIT license. See LICENSE for details.
  */
 
-#ifndef __BIN2ASCII_H__
-#define __BIN2ASCII_H__
+#ifndef VG__BIN2ASCII_H__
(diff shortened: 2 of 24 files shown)


elfring:
Thanks for your small source code improvement.

I assume that a probability will remain for name clashes because of short identifiers like “`VG_PATH_H`”.

adamnovak:
Unless someone else has a project called "vg" that they want to pull our
headers into, I'm not sure we need to worry about anyone else using the
macro VG_PATH_H.

If it does come up, we can change the include guards again.

elfring:
>If it does come up, we can change the include guards again.

I would prefer the application of an identifier selection algorithm where the probability for unwanted name clashes will be reasonably reduced further (besides [software development concerns because of standard compliance](https://github.com/vgteam/vg/issues/854 "reserved identifier violation")).

adamnovak:
[StackOverflow](https://stackoverflow.com/q/4867559) notes that the Google style is to have a trailing `_`, and the Boost style is an `_INCLUDED` suffix. Some people do put random "unique" strings in there, but IMO that's probably excessive.

I'm going to go with Boost style, since if people are going around defining `VG_WHATEVER_HPP_INCLUDED` macros then name collisions really are their problem.

PR #850, opened 2017-07-11
Test Jenkins does *not* update baseline from PR

glennhickey:
This is just a test.  Don't merge me.


glennhickey:
OK, baseline update finally looks like it's working again.

PR #856, opened 2017-07-12
Sync refactored mapper

jeizenga:
As discussed at the VG meeting, this has my refactored Mapper. I think it's important to get these synced up early so we don't diverge too much and create more work for ourselves later. I also had to pull in a refactor I did of the Aligner classes. There are also small changes in a lot of files to match new method call signatures, etc.

@ekg I know some of your changes are in another branch. If you want, I can try to sync up to your changes in that branch too. I'll just send a PR directly to it. First I want to get this merged in master though.


jeizenga:
Even if this passes the tests, hold off on it for a minute. I need to make sure I didn't squash some newer commits.

PR #861, opened 2017-07-14
Use new toil-vg style for mapeval and scan for worsening reads

adamnovak:


adamnovak:
Looks like we're getting per-read warnings in the Jenkins standard error when reads don't go up in score like we want them to, so I'm going to merge this in.

PR #866, opened 2017-07-14
Sync Mapper refactor to mapper branch

jeizenga:
Note: this is a PR to the mapper branch, not to master. As such, I'm not expecting to pass every test, since the mapper branch PR doesn't pass all of them yet either. In particular, it should fail tests 24 and 26 in the 07_vg_map.t.

The point of this PR is to get @ekg 's mapper synced with my refactor, which I made in the multipath-alignments branch. I was careful to preserve all of the changes @ekg made to Mapper methods. They are just rearranged according to my refactor. This PR also pulls in some of the changes to the Aligner, etc. that the refactor depended on.


jeizenga:
This only failed the things I expected it to fail, so I think it's good to merge (again, into the mapper branch, not master)

ekg:
So to be clear, this is to merge into mapper.

Unfortunately, I have a big change (31kb patch) that I hadn't pushed up.

I'll probably commit that and then try to sync.

jeizenga:
The sync is probably going to be pretty difficult. I rearranged things in my refactor, so the git diff was basically useless.

ekg:
I have things working to a particular standard in my branch. So I'll just
have to merge this in as soon as I have a passing build. I can do it
manually maybe with your help.

jeizenga:
I did the merge, but this is going to fail the Travis build. You seem to have added a dependency on the GSL, but that isn't standard on macOS. If we want to include it we're probably going to need to do something with the source_me script or the makefile or something.

ekg:
Thanks! I'll sync here and work on getting gsl and blas working.

I could avoid these dependencies if I can find a standalone regression
implementation.

ekg:
I've been able to sync this in locally and build. I've removed the gsl/blas dependency as it is causing nothing but pain. I'll push into the mapper branch all of these commits and we should be level.

ekg:
Merging this in has caused a performance (sensitivity) regression.

I'm trying to figure out what might have changed relative to 9dc5cffe. Maybe some of the mapper defaults were not ported through?

ekg:
I think I've mostly sorted this out, so no worries. Looks like a mixture of confusion and unresolved problems on my end.

jeizenga:
Hey, maybe you already worked this out, but linear regression isn't all that hard to code up. It's a little harder to code up _well_ because some of the matrices involved are poorly conditioned (which makes it require some fancier numerical technology to avoid large floating point errors), but the closed form solution is just (X'X)^-1 X'y. You could do it all with simple implementations of matrix multiplication and inversion like I have here:
https://github.com/jeizenga/hdp_mixture/blob/master/impl/hdp_math_utils.c#L810

ekg:
Yup. I did this and now have to guard against overflow. But it is easier
than dealing with the additional dependencies!

Issue #863, opened 2017-07-14
Map and align disagree on how to use the full length bonus


adamnovak:
Here's a graph where map and align give different scores with the default full length bonus of 5, but the same score when it's set to 0. I think the problem is that map is not applying the bonus at each end (requiring both ends to be full length to get any bonus), while align is applying it per-end.

```
[anovak@kolossus bonus-problem]$ vg index -x primary-part.xg -g primary-part.gcsa primary-part.vg -k 16
[anovak@kolossus bonus-problem]$ vg map -s $(cat read.txt) -x primary-part.xg -g primary-part.gcsa -j | jq '.score'
89
[anovak@kolossus bonus-problem]$ vg align -s $(cat read.txt) primary-part.vg -j | jq '.score'
94
[anovak@kolossus bonus-problem]$ vg map -s $(cat read.txt) -x primary-part.xg -g primary-part.gcsa -j --full-l-bonus 0 | jq '.score'
89
[anovak@kolossus bonus-problem]$ vg align -s $(cat read.txt) primary-part.vg -j --full-l-bonus 0 | jq '.score'
89
[anovak@kolossus bonus-problem]$ 
```

[bonus-problem.tar.gz](https://github.com/vgteam/vg/files/1149830/bonus-problem.tar.gz)

jeizenga:
In my refactor, I made some changes to the full length bonus. This was probably caused by something I did.

Now, instead of having the full length bonus as an argument to alignment methods, I made it a member variable given to the constructor. Aligners should automatically apply the full length bonus whenever it's meaningful (both ends on GSSW alignment, one end on pinned alignment, neither for global alignment).

Just an intuition, it might be that the defaults on the arguments are interacting poorly. The constructor has default arguments for score parameters, and the align methods have a default argument for printing matrices. I've seen it get confused by this previously and just case arguments intended for one parameter into the other.

adamnovak:
It looks like this difference is a deliberate choice by @ekg not to use the full length bonus in the scores output by the mapper (even though it's included by the aligner).

See the commits around https://github.com/vgteam/vg/commits/master?before=258cfdf74c2830c75c1af2a80e91a165bba5bd50+35 and in particular https://github.com/vgteam/vg/commit/7344458b2269a1277ce6b7639cde620c28369750

[...]

Issue #839, opened 2017-07-07
vg help reports `vg homogenize` twice


adamnovak:
I have two entries for `vg homogenize` in the output when I just run `vg`. There should only be one.

PR #874, opened 2017-07-20
Report full-length bonuses in scores from vg map again

### LINKED ISSUE #865: Alignment scores get worse when turning on full length bonus
This is probably related to #863.

When you turn on `--full-l-bonus 5` for this read, the score from `vg map` actually drops, because it's unwilling to use a softclip.

I think the full length bonus is not being added to the score, but is being accounted for in the dynamic programming.

```
[anovak@kolossus got-worse]$ vg map -s $(cat read.txt) -x snp1kg-part.xg -g snp1kg-part.gcsa -j --full-l-bonus 0 | jq .score
89
[anovak@kolossus got-worse]$ vg map -s $(cat read.txt) -x snp1kg-part.xg -g snp1kg-part.gcsa -j --full-l-bonus 5 | jq .score
86
```

[got-worse.tar.gz](https://github.com/vgteam/vg/files/1149866/got-worse.tar.gz)

This came up in my quest to enforce that reads aligned against larger graphs have their scores go up, because this read was showing as getting a worse score in snp1kg (where it won't take the softclip) than in primary (where it does take the softclip).

@ekg did you do something with end bonuses when changing the aligner around?

adamnovak:
This restores the old score reporting behavior, which is to report the score that the DP actually used to select the alignment.

(There are also some other fixes in the branch, like a patch for #839, conversion of vg align to a real subcommand, and fixes for mapper tests that were throwing errors but succeeding.)

If you want scores without bonuses (maybe because you want score to conveniently equal number of matches, or because some downstream tool works better with those scores), there's a new `--strip-bonuses` option.

I'm still not quite happy with the design here (@jeizenga is right that we probably should consolidate all the scoring stuff in the aligner, but the mapper needs to do re-scoring to account for distances when jumping between nodes). I could also swap the `--strip-bonuses` option for an `--include-bonuses` option and change the default.

This fixes #865 and #863.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,7 @@ SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/trace_main.o
 SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/gamsort_main.o
 SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/msga_main.o
 SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/map_main.o
+SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/align_main.o
 SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/surject_main.o
 SUBCOMMAND_OBJ += $(SUBCOMMAND_OBJ_DIR)/find_main.o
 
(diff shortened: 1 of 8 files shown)


adamnovak:
It looks like we do indeed see worse performance at the variant calling stage when outputting the bonused scores.

I think I'm going to change the default behavior, and only use the bonused scores in the "no read goes down in score" test.

Long term we might or might not want to copy BWA's practice of always reporting the score for the alignment that would have been optimal without the bonuses.

adamnovak:
I've merged this in. The default behavior should be the same as before, but we now have a `--include-bonuses` option to keep the bonuses in the score.

PR #890, opened 2017-07-24
Add basic framework for a log miner

### LINKED ISSUE #889: Make a null report
We need a stub of the report generator. I will make something that makes a null report file, so we can upload it and post it to Github.

Should make an HTML full version, and a markdown format summary to be sent to Github.

adamnovak:
We're going to have one script crunch the test logs and make a report.
Another script will upload the report to the Web somewhere and post
about it on the relevant PR or commit.

The upshot of all this is that we will be able to have nice reports
with performance statistics and maybe even plots in the Github
comments for PRs.

Fixes #889


adamnovak:
This cool bot script is going to be posting comments on PRs with Jenkins test info

PR #894, opened 2017-07-25
Add barebones Jenkins log miner

glennhickey:
This parses the JUnit xml and dumps some barely-formatted info the MarkDown and HTML output. 

Still working on getting the output be nicer/more useful.


adamnovak:
Looks like you missed a space after the first `##`, which Github requires now. But other than that this seems to be working.

adamnovak:
I'm going to merge this in so we get non-useless test reports on our other PRs. We can fix the formatting after.

PR #906, opened 2017-07-26
Activate haplotype simulator

glennhickey:
(on top of #899, which I think failed due to s3 idiocy on my part)

Relies on updated snp1kg graphs and indexes in S3.  Using temporary bucket for now, will move to cgl-pipeline-inputs when it passes.

### DIFF
--- a/jenkins/mine-logs.py
+++ b/jenkins/mine-logs.py
@@ -107,7 +107,7 @@ def md_summary(xml_root):
         md += '{} tests passed, {} tests failed and {} tests skipped in {} seconds\n\n'.format(
             ts['passes'], ts['fails'], ts['skips'], ts['time'])
 
-        if ts['fails'] is not None and int(ts['fails']) > 1:
+        if ts['fails'] is not None and int(ts['fails']) >= 1:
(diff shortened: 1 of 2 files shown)


adamnovak:
XG looks a bit upset about the gPBWT:

```
ip-172-31-21-11 2017-07-27 20:34:12,788 MainThread WARNING toil.leader: 4/s/job2ydBVf    vg: deps/xg/src/xg.cpp:3158: std::map<std::__cxx11::basic_string<char>, std::__cxx11::list<std::vector<xg::XG::ThreadMapping> > > xg::XG::extract_threads_matching(const string&, bool) const: Assertion `edge_index != BS_SEPARATOR' failed.
```

Did you make sure the graphs had IDs assigned in topological order before building the gPBWT? `vg construct` doesn't do it by default, and the gPBWT won't build correctly otherwise.

glennhickey:
Thanks!! These graphs are coming right out of construct so that would
explain it.  I presume vg ids -s would be enough to fix?

adamnovak:
I think that should do it.

We should add a check in the gPBWT construction code to see if it can catch this problem earlier.

PR #910, opened 2017-07-27
add link back to jenkins in reports

glennhickey:


glennhickey:
quay.io isn't happy http://status.quay.io/

PR #913, opened 2017-07-28
New mapper with current submodules

adamnovak:
This is #831 with the xg and gssw submodules bumped to their current master versions.


adamnovak:
@ekg What was the plan in https://github.com/vgteam/vg/commit/9ce8d3d87fd54615f481262b85f4c0c78040e0d3?

Did you intentionally change the behavior of xg? Or did you just change the test to match the behavior you observed? The XG master AFAIK will report 4 orientations rather than 2 threads.

ekg:
What is the use case that requires us to extract 4 threads for 2
haplotypes? Not understanding that it might be, I changed the behavior of
xg so that we only get the "forward" orientation of each thread when
extracting them.

adamnovak:
It's not really that there's a use case for it, it's more that I didn't see a good way to make XG match up the forward and reverse threads when emitting them. It's possible that they both start on the same side of the same node, for example; all I could think of was banging every thread against every previous thread to see if you'd emitted it or not.

Where are the xg commits that you made to make this change? I "fixed" what I thought was a bug in xg that caused it to completely ignore any threads starting on the final side in the graph, and it started actually spitting out both orientations, in https://github.com/vgteam/xg/pull/61 and I don't see anything by you since then. Was ignoring the last side in the graph your solution? Because if so I don't think it will really work if your graph has more than one chromosome in it.

ekg:
In the VCF->DAG case the threads starting on the final side of the graph
are going to be the reversed versions of the forward ones. Note that the
behavior that you fixed had us looking only at the threads starting at the
forward side of each node, so it didn't break for multi-chromosome graphs.
It would be weird to me if it did as my testing showed it emitted the
correct number of threads in multi-chromosome graphs.

Maybe we should change the thread extraction to work using the name
storage. This would remove the duplication. I see no reason why the
extraction should double the number of threads.

adamnovak:
I'm uncomfortable looking at only the forward side of each node, because that relies on the graph being made of nodes that are all oriented forwards. If you have a 2-node graph with both the forward sides on the outside, it's perfectly legal by the acyclicity and ordering rules for the gPBWT construction, but you'll get 4 threads out when you were expecting 2. And if you turn the nodes around the other way you'll get 0 threads.

The right way to do this probably is to look at the thread names. Do we want to do that before merging in the mapper changes?

ekg:
I can change extraction to an iteration over the thread names. It's not
essential for the mapper polishing though, but it does matter for building
the haplotype-based tests.

adamnovak:
I think what we should do is merge the mapper changes on top of the current xg master, suffering the extra extracted threads (as currently in vg master) and then fix xg thread extraction to use proper deduplication as its own issue (https://github.com/vgteam/xg/issues/65).

adamnovak:
#831 has all this now.

PR #914, opened 2017-07-28
Fix Jenkins to have a region for SimpleDB

adamnovak:
Without this Jenkins builds of master just loop forever due to not being able to talk to the SimpleDB mutex.


adamnovak:
I'm going to merge this in right away so I can see if it works in the resulting master build.

PR #915, opened 2017-07-31
Remove -march=native and =mtune=native from Mac builds

edawson:
I'm getting broken builds due to the dreaded `no instruction found: vzeroupper` error we saw a few months back. It is apparently a weird Mac OS quirk where the compiler can handle AVX but the assembler can't: https://stackoverflow.com/questions/10327939/error-no-such-instruction-while-assembling-project-on-mac-os-x

This should fix it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ CPP_DIR:=cpp
 
 EXE:=vg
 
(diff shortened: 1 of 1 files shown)


adamnovak:
Travis has been confused by Github vanishing on it. One of the Travis builds worked fine and the failed one doesn't seem to want to restart.

I'm going to merge this in.

PR #928, opened 2017-07-31
Improve remapping to simulation source graphs

### LINKED ISSUE #917: Split clusters happen when orientation isn't known due to too-low connection limit
Look at read pair `1cf6d11ad788744d` in the Jenkins BRCA1 test.

```
kolossus 2017-07-28 11:10:56,434 MainThread WARNING vgci: Read 1cf6d11ad788744d_1 has a negative score increase of -102 on graph snp1kg-pe vs. input
```

Basically one end is not mapping when we re-map against the graph it was generated from with paired-end mode on.

This is because, without a known value for the relative orientations of the two ends, we're considering too many links, and we're hitting the limit before we hit the link that connects the one end of the read to the other.

The short-term solution might be to raise the max links to consider. #916 might also help.

adamnovak:
I've upped the `max_connections` in the MEM chainer from 10 to 20, which should improve performance substantially on some of the Jenkins read score tests.

I'd still like to have a MEM chainer that knows more about orientations, but this should help for now.

This should fix #917.

### DIFF
--- a/jenkins/jenkins.sh
+++ b/jenkins/jenkins.sh
@@ -24,7 +24,7 @@ KEEP_OUTPUT=0
 # Should we show stdout and stderr from tests? If so, set to "-s".
 SHOW_OPT=""
 # What toil-vg should we install?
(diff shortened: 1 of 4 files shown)


ekg:
Note that the MEM chain model can know about orientation. This information
is provided via the transition weight function.

PR #932, opened 2017-08-02
Control graphs in Jenkins plots

glennhickey:
Adding these graphs to the simulation tests:
* snp1kg_HG00096, which contains just the sample graph we're simulating from plus all reference positions, as a positive control
* shifted1kg, our negative control from the bakeoff which is like 1kg but all variants are moved around

A second set of plots is produced for these controls to not have too many lines in one plot.  Also doubled the simulation reads to 100000, but there's still work to be done to make these plots more meaningful.


glennhickey:
Jenkins never ran.  This is bizarre.

glennhickey:
Jenkins, test this please.

glennhickey:
The MHC control plot looks like we'd expect based on previous results.  With the positive control being the best, then primary, then snp1kg, then the negative control.  I think it's still an open question (that shouldn't be too hard to answer) of how much a factor the edge-max limit during indexing plays in snp1kg being worse than primary, as opposed to ambiguity inherent in the graph (or bugs). 

The  sim BRCA1 test (which runs as long as MHC) seems to be too easy to give meaningful plots.  Still getting the score warnings though, so it may be useful yet.

glennhickey:
Boosting the edge crossing limit from 5 to 1000 only moves the plot a hair, if at all.

ekg:
I do find it strange that we are not killing the brca1 result with the positive control. This is a graph built from just the variants in hg0096 right? And the simulation is based on the haplotypes of hg0096?

glennhickey:
Yeah, the reads were simulated from the 2 HG00096 threads following your
procedure, and the positive control vcf was build like:

bcftools view 1kg_brca1.vcf -s HG00096 | grep -v '0|0'

So for BRCA1, snp1kg.vg has the full 2048 variants, while snp1kg_HG00096.vg
has just 155.  Of those 155, only 2 are not heterozygous, and just by
eyeball, the variants are mostly a few hundred bp apart, so I expect this
graph to be pretty close to an ideal control.

For MHC, those numbers are 165583 variants in snp1kg.vg,  27534 in
snp1kg_HG00096.vg, 9965 of which are alt/alt.  There are surely many more
places where variants are less than a read-length apart as well.

PR #936, opened 2017-08-02
re-enable tuning for linux

ekg:


edawson:
Sorry about that - I think I misread the conditional as if-equals.

ekg:
Not a problem. I was having some problems with getting the conditional to
work there and so I had to write a new one for the tcmalloc stuff. Make is
weird.

Issue #881, opened 2017-07-24
Examine remaining reads with lower scores on snp1kg than refonly


adamnovak:

adamnovak:
Glenn says this is actually important; when we filter these reads out, variant calling improves substantially.

glennhickey:
Just to clarify, those ROC's were for mapping accuracy, not variant
calling.

adamnovak:
So then what's happening is that reads that map worse to snp1kg than to
refonly are responsible for much of the mismapping that occurs when mapping
to snp1kg.

glennhickey:
That's it.  Which means we cannot blame these mismappings on finding
better/equivalent alignments elsewhere in the graph by chance.

MHC is obviously a bit special, but will run same experiment on at least
one chromosome.

adamnovak:
Glenn suggests breaking up the reads that are in the right place vs. not by whether they are getting lower scores than we expect or not.

faithokamoto:
I think whatever this was tweaking is long past the stage where we are still actively tweaking it :) Closing as stale.

PR #943, opened 2017-08-04
Fix worse mapping

adamnovak:
I found a read that was mapping worse to snp1kg than to primary in the BRCA1 Jenkins test, running locally on kolossus.

This was because the read had a MEM stareting at the beginning of the read and the end of a node (it aligned on the reverse strand), and the cluster subgraph extractor was not looking before that MEM to find any more graph material. The optimal alignment didn't actually use that MEM but instead went a bit further out and matched more repeat units for a better score.

I turned the problem into a unit test, and solved it by cribbing off @jeizenga's cluster graph extractor.

We should pay close attention to how many CI warnings about misbehaving reads Jenkins throws. I want them to come down from 72 (on Jenkins) for BRCA1 and 390 (on Jenkins) for MHC.

Oddly, I get different sets of warnings when I run the Jenkins tests locally versus on Jenkins. I think `toil-vg` is generating different reads with `vg sim`. Maybe @cmarkello knows if there's a difference in how the reads are seeded somehow.

This is related to #881 and may fix it if it's good enough.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -396,7 +396,7 @@ $(OBJ_DIR)/ssw_aligner.o: $(SRC_DIR)/ssw_aligner.cpp $(SRC_DIR)/ssw_aligner.hpp 
 
 $(OBJ_DIR)/vg_set.o: $(SRC_DIR)/vg_set.cpp $(SRC_DIR)/vg_set.hpp $(SRC_DIR)/vg.hpp $(SRC_DIR)/progressive.hpp $(SRC_DIR)/index.hpp $(DEPS)
 
-$(OBJ_DIR)/mapper.o: $(SRC_DIR)/mapper.cpp $(SRC_DIR)/mapper.hpp $(SRC_DIR)/mem.hpp $(SRC_DIR)/vg.hpp $(DEPS)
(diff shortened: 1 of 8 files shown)


jeizenga:
Aside: while I was showing this to you I found a mistake with how I calculated the distance bounds. The code is corrected on my branch here:

https://github.com/jeizenga/vg/blob/multipath-alignments/src/multipath_mapper.cpp#L137

adamnovak:
Looks like this is going to take some more work, since it lowers performance the way I have it now, especially remapping against the input graph. Even with the corrected distance bounds, this might be a case of more-principled-but-worse.

adamnovak:
This is reducing the number of reads which disobey our score invariants.

The Mac Travis test is failing because @mlin cleared the cache, and the build is taking too long (and getting killed by Travis) because it's trying to rebuild its empty cache. Our OSX builds are likely to fail forever unless we reduce the amount of code being compiled or adopt a clever partial-build solution to fill up the cache. I'm going to merge this PR anyway.

The other concern here is, because of the way I am now doing graph extraction and alignment deduplication, we can get low mapping qualities when two different alignments to the same local region are generated. We will probably want to add a bit of context expansion to the graph extraction, and/or port over the idea of merging overlapping clusters from the multipath mapper.

adamnovak:
@jeizenga @glennhickey It looks like this PR is the one that messed up paired end mapping. The AUC for snp1kg-pe went from ~0.80 to ~0.65, and we didn't get a test failure.

PR #962, opened 2017-08-18
Functional multipath mapper

jeizenga:
This is the first pass at a mapper that can produce MultipathAlignments. It's slow. It also only really does single ended alignment right now. But I do think the output it's producing is at least technically valid. Now I want to get a baseline assessment of the mapping accuracy so that I can compare optimizations to it.


jeizenga:
@adamnovak @glennhickey @ekg 
The ROC curves look notably funky on Jenkins, but I they also do on the previous PR:
http://cgl-pipeline-inputs.s3.amazonaws.com/vg_cgl/vg_ci/jenkins_reports/commit/958c0443998259aaa5a5a2f0ed7a2daffaa29e02/index.html

Any idea why the TPR would be so low on the paired end?

glennhickey:
It seems this PR is indeed the culprit

https://github.com/vgteam/vg/pull/943

PR #964, opened 2017-08-18
Make Jenkins report with mpmap (do not merge!)

glennhickey:
Make a test report for #962 using mpmap for everything.


adamnovak:
The mpmap reports are in master, so I will close this.

PR #969, opened 2017-08-21
Add small simulation test for multipath aligner

glennhickey:
This is like the other simulation tests but smaller.  It's wrapped in a try/except so as to let mpmap fail away without alarming people.   We can change this whenever it's ready.  @jeizenga  hopefully this is useful for your debugging.   To run locally, do

`jenkins/jenkins.sh -l -r -k -s -t jenkins/vgci.py::VGCITest::test_sim_brca2_snp1kg_mpmap`
 
then look in vgci-work/ for all the output data. You'll find the various tables and plots comparing vg mpmap to vg map, along with the reads that don't map to the right place or have a good enough score.   You'll need pigz and samtools/bwa installed to run....


adamnovak:
Jenkins test this please

PR #970, opened 2017-08-21
Undo more changes that decreased paired-end performance

### LINKED ISSUE #968: Read pair can be forced to the wrong place with a reasonable-looking fragment length distribution
Even with the old cluster method and added padding, we still get terrible results on e.g. Cactus paired-end.

What caused it? Was it the padding? The change in position measurement? Some other change?

adamnovak:
It turns out that my alternative actual-alignment-based deduplication mechanism resulted in worse performance than the old end-node-ID-based "signature" deduplication that was being used, even though different alignments might sometimes have the same signature.

Also, it turns out that my change to place the reads at the actual median position, instead of the position of the lowest-ID node, made BWA look really bad, because BWA reads are just getting placed at the read's start position. I still think placing reads at the real median position is better; I just need to fix the bwa position determination in toil-vg to also use a median or middle position.

Fixes #968.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -1003,135 +1003,33 @@ map<id_t, int> alignment_quality_per_node(const Alignment& aln) {
     return quals;
 }
 
-bool SameReadAlignmentOrder::operator()(const Alignment& a, const Alignment& b) const {
-    // Compare two alignments of the same sequence to potentially different
-    // graph paths. We want to order them so we can deduplicate them in a
-    // std::set.
-    
-    // First look at the scores
-    if (a.score() < b.score()) {
-        return true;
-    }
-    if (b.score() < a.score()) {
-        return false;
-    }
-    
-    // If they are the same, look at the paths.
-    if (a.path().mapping_size() < b.path().mapping_size()) {
-        return true;
-    }
-    if (b.path().mapping_size() < a.path().mapping_size()) {
-        return false;
-    }
-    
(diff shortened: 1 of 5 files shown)


adamnovak:
I'm still seeing a higher minimum FP count on brca1 for snp1kg-pe than in my [known good report](http://cgl-pipeline-inputs.s3.amazonaws.com/vg_cgl/vg_ci/jenkins_reports/commit/b437e1a1988426aec8dcd6929d8732dd5139a653/index.html).

adamnovak:
Jenkins test this please

adamnovak:
My padding of 20 increased false positives. I turned it down to a padding of 1, which is sufficient to pass the unit test I made from the case that motivated having padding in the first place, and which doesn't make large numbers of new false positives.

adamnovak:
The 0.870839667173 AUC for snp1kg-pe that Jenkins reports here for this commit does not match the 0.997624904996 AUC I have recorded in my notes (I believe from a local run of the test). It also does not match the 0.997457398296 AUC from my known good report. I'm not sure why I didn't see that this morning, but it means that even with this change there's *still* something missing.

adamnovak:
I just reran the commit tested here (minus some of the controls) and got these results, which disagree with Jenkins and say snp1kg-pe is doing fine.

image

I am rerunning with all the controls in now, in case that somehow changed something. My local results and the Jenkins results really ought to agree; they're using the same reads, the same graphs, and the same vg commit.

PR #979, opened 2017-08-22
Up padding (again)

### LINKED ISSUE #975: Pulled-out graphs for alignment can be too small if a MEM at the beginning of a read is not used
image

Here are the starts of two alignments of the same read (`ATATATATATGTATGTAATCCCAGCACTTTGGGATATATGTGGATATATGTTTGTTTTTTTTGAGATTCAATTTTGCTCTGTTCCCAGGCTTAGAGTGCAGTGGCGTGAACTCCGCACACTGCAACCTCCACCTCCCTGGTTCAGGTTGA`), one done with vg map and one done with vg align.

The alignment done with vg align is better; the read wants to start at the end of that other node. But the vg map alignment squashes the read into one fewer repeat units. It looks like this is due to a spurious `ATATATATATGTA` MEM at the beginning of the read, which is cause by that first G in the read, which is an error and gets mistaken for the reference G that produced the second G in the read.

This is similar to the problem I tried to address in https://github.com/vgteam/vg/pull/943, where a MEM that started at the start of a node would not expand out in that direction at all, causing problems if that MEM was not actually a part of the optimal alignment and instead more material before it needed to be included. I added a padding value to try and extract more graph, but when I turn that up too high my false positive rate in the mapeval tests starts to go up. This could be because I end up extracting enough graph material to find spurious mappings.

I've attached the read (with its simulated alignment) and the target graph. Aligning with `vg map --include-bonuses` will produce the worse alignment with the score shown in the image, while extracting the region near node 5219 and aligning with `vg align --include-bonuses` will produce the better alignment.

[problem.tar.gz](https://github.com/vgteam/vg/files/1243125/problem.tar.gz)

adamnovak:
Not having enough padding increases the chances that unused MEMs near the ends of reads and nodes will cause an alignment to too little graph.

This should fix #975, but really we either want to have a large amount of padding (and not let that cause false positives on the mapeval tests) or have some other way of knowing if a better alignment would or could happen if we pulled in more graph material.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -2558,6 +2558,7 @@ Mapper::align_mem_multi(const Alignment& aln,
     double mq_cap = max_mapping_quality;
 
     {
+        // Estimate the maximum mapping quality we can get if the alignments based on the good MEMs are the best ones.
         int mem_max_length = 0;
         for (auto& mem : mems) if (mem.primary && mem.match_count) mem_max_length = max(mem_max_length, (int)mem.length());
         double maybe_max = estimate_max_possible_mapping_quality(aln.sequence().size(),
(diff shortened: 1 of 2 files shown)


adamnovak:
The chr21 test takes ~2.5 hours and throws ~58,000 warnings, both of which are way out of proportion to the read count. @glennhickey suggested working from hs37d5 for this test, which doesn't have the centromeres in it and which would be directly comparable to Erik's manual testing runs.

adamnovak:
Jenkins, test this please.

adamnovak:
I want to see the warning counts.

Jenkins, test this please.

PR #981, opened 2017-08-23
Improve alignment speed

ekg:
Attempt to optionally disable traceback (and matrix unswizzling and saving) in order to speed up the scoring of candidate alignments. The idea is that this will improve alignment speed by not polluting our cache with the heavyweight traceback matrixes, but it is not possible to test this without a fair bit of hacking. For many operations, all we need are the scores. We only need tracebacks for those alignments that we emit.

To start this PR, I've added a flag that lets us turn off traceback in gssw. Now it should be possible to obtain score-only alignments from gssw without allocating space for the score matrixes. However, the semantics for this are unclear--- I need to work out how the score-only alignments are represented. It will be necessary to maintain cluster-to-alignment relationships longer in order to generate the full alignments up to the number of multimaps we are writing out.


glennhickey:
Looks like you'll need to get rid of --include-bonuses from jenkins/vgci.py
for the tests to run

ekg:
Aye, I'll do that.

ekg:
Here's the results of valgrind's kcachegrind output visualized. vg has been run in single-ended mode on some simulated reads from yeast. I've used the old-style simulator, not @jeizenga 's new one.

This call graph shows everything below align_mem_multi. I'll refer to it in this post:

vg-map-7422a7cf3-kcachegrind

The area plot is a nice overview of the relative times of things:

image

I've run vg map in a funny mode that I just added that allows it to avoid checking if the graph is acyclic (it assumes it is when this flag is set). This consumes about 10% of the runtime in the default mapper, but is something that could be precomputed or known at the time of indexing. Asking the user to flip a switch seems inconvenient. I'm thinking I'll make changes to XG as part of this PR, and will add the appropriate flags to the database and indexer then, so this command line flag is a hack.

The highlighted portion is the part that gets the cluster subgraph. The part on the left that isn't highlighted is the local alignment. The part on the right that isn't highlighted is the MEM finder.

This portion seems redundant. Ideally our cluster subgraph should already be available in memory and usable by gssw directly, which would make this portion incredibly small. However, it's not so simple because we would like to bound the region that we're aligning to in order to encourage traversal of our particular MEM cluster by the local aligner. This is implemented by graph walking, which is expensive given the interface choices with XG. We have to instantiate new vg::Node and vg::Edge objects for each node we encounter. Caching mitigates some of the pain caused by this walking pattern, but does so only partially because we now have to spend compute on caching logic. In XG the nodes are accessed via a series of select queries (that show up as the big blue box in the bottom middle of the above plot), which is a clear potential win; when our nodes are short we can just walk the bitvector to find the next node start. However, these feel like weak solutions and hacks on top of the XG format. Probably it would be a better idea to extend or modify the format to more directly match the access pattern we need in cluster_subgraph.

Looking at the call graph for the cluster_subgraph function gives a more precise readout of the time contribution of the various components. Virtually all the time is spent in this function, which consumes 30% of vg map's runtime:

```c++
void Mapper::cached_graph_context(VG& graph,
                                  const pos_t& pos,
                                  int length,
                                  LRUCache<id_t, Node>& node_cache,
                                  LRUCache<id_t, vector<Edge> >& edge_cache) {
```

I'd like to implement an alternative graph index (presumably an extension or ideally a deep modification of XG) that provides this same function and returns a sorted graph that can be loaded directly into gssw without further transformation.

The graph encoding should allow us to avoid select operations and to make the edge storage local to the node in the index. Also, it would be best if it allowed us to directly walk the graph as is now done in the cached_graph_context function, because otherwise we've got to think about mapping positions into cluster regions.

We can handle the check for acyclicity by marking positions we've touched in our traversal of the graph. If we haven't touched one twice then we know the graph (as far as we've expanded it) is acyclic. This requires us to run our cluster extraction mostly in the scope of the xg index.

Alternatively, we could imagine an index that maps from positions (node ids, more simply) into graph chunks. The problem here is edge effects. If we have an expected cluster size then we can preprocess the chunks so they overlap by an appropriate amount, and we can select the cluster as the one with the most target positions in it. This would also avoid the computation required to walk the graph.

As a final note, we still burn a lot of new/delete allocations on protobuf objects. There is a clear way to avoid this by using protobuf Arenas in certain areas, but it would only be a big deal if implemented in the right contexts and could conceivably cause problems if implemented broadly. Reworking the vg::VG object to optionally use such an arena could be very interesting.

ekg:
This is failing. It looks like an empty alignment is getting selected as the best alignment. Changing `-w` to another value can make it work. I want to fix this but it's not blocking so I'm noting it here.

```
vg msga -f msgas/cycle.fa -b s1 -w 21 -t 1 | vg mod -U 10 - | vg view -dp - | dot2browser
```

ekg:
Jenkins, skip this please

PR #985, opened 2017-08-24
Make the correctness window 200bp

### LINKED ISSUE #977: Make correctness determination for read mapping more lax
We can only really localize reads to node (up to 1000 bases in a general graph, or 100 in primary) or read-length ([150 bases](https://github.com/vgteam/vg/blob/master/jenkins/vgci.py#L447)) resolution in the mapeval tests right now, buw we're using a window of [100 bases](https://github.com/vgteam/toil-vg/blob/master/src/toil_vg/vg_mapeval.py#L92) for determining correctness. This is going to cause things that shouldn't reduce performance to look like they reduce performance, and inhibit the tests' ability to evaluate the broad placement of reads.

At the least, we should widen that window.

adamnovak:
We can only really localize reads to node resolution, and we localize a read at an arbitrary point along its length. With a 100bp threshold for the correct position for a read in the tests, we risk not detecting some of our true positives. 200bp ought to cover all the slop we could have at reasonably small node sizes.

Fixes #977

### DIFF
--- a/jenkins/vgci.py
+++ b/jenkins/vgci.py
@@ -482,6 +482,10 @@ class VGCITest(TestCase):
         mapeval_options.index_bases = [make_url(x) for x in test_index_bases]
         mapeval_options.gam_names = test_names
         mapeval_options.gam_input_reads = make_url(os.path.join(out_store, 'sim.gam'))
+        # We have 150 bp reads reduced to a point position, at a resolution of
+        # only the nearest 100 bp (on the primary graph). How close do two such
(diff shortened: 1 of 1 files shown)


ekg:
This makes sense to me.

adamnovak:
It looks like the test report is truncated; only the first test appears, and there's no closing HTML tag.

Jenkins, test this please.

adamnovak:
Once again, we have one read that's intermittently mismapping in one of our controls, so I am merging this over Jenkins' objections.

PR #988, opened 2017-08-24
Use smoother mapeval fragment length distribution estimation in the tests

adamnovak:
This should stop the BRCA1 mapeval test from randomly and intermittently failing until #968 is fixed.

The right solution might be to improve the learning model somehow; the learned fragment length distribution pasted in here was learned on data simulated with the vg sim parameters used for mapeval, but it seems to skew much shorter than the length requested from vg sim.

@jeizenga didn't like this idea, but I think we have to do it or our PRs will randomly fail the BRCA1 AUC assert.

### DIFF
--- a/jenkins/vgci.py
+++ b/jenkins/vgci.py
@@ -463,7 +463,8 @@ class VGCITest(TestCase):
             # Make sure we have the actual scores used to decide on alignments
             # Also force a fixed fragment length distribution to make alignment
             # deterministic for the test.
-            map_opts = ['--include-bonuses', '-I', '816:345.165:47.0999:0:1'], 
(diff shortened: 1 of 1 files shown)


adamnovak:
It turns out that using a fixed distribution (or at leas the fixed distribution I wanted to use) works fine for the snp1kg graphs, but lowers the performance of the Cactus graphs. I'm not sure there's a *single* fixed distribution we can use for all tests.

What I've done now is, instead of fixing the distribution, I've upped the reestimation window significantly. This should make the different estimates less variable, and, while not really forcing them to be deterministic, should at least make them consistent enough that the tests don't decide to vary in their outcomes quite so much.

adamnovak:
The AUC still goes down a bit, but hopefully it will be more stable at its new lower value. I'm going to merge this in.

PR #990, opened 2017-08-24
Fix sim bonus

### LINKED ISSUE #987: Option to include full length bonus in vg sim scores is broken
`vg sim -m` still removes full length bonuses despite the following documentation:

```
    -m, --include-bonuses include bonuses in reported scores
```

adamnovak:
Fixes #987 and makes vg sim have a nonzero full length bonus.

Still TODO is letting the user set the scoring parameters for vg sim.


adamnovak:
The AUC drop here is caused by spurious flapping, so I'm going to merge this in anyway.

Issue #933, opened 2017-08-02
Pruning step of indexing instructions on wiki generates lots of warnings for re-used node IDs we are supposed to have


cartoonist:
I am following this [wiki page](https://github.com/vgteam/vg/wiki/working-with-a-whole-genome-variation-graph) for constructing human whole gnome variation graph. During pruning step, i.e. masking out the high-complexity regions for each chromosome, I get millions of this warning message for all chromosomes (1-22):

    [vg] warning: node ID ##### appears multiple times. Skipping.

I do not expect to get this warning message, because `vg ids` should have generated a disjoint ID space. Might it be a bug? Is the resulting GCSA reliable to work with?

ekg:
Those warnings are just reminders that the graphs that are combined with
cat contain duplicate nodes. We expect them to though because of the way we
constructed them. In fact there isn't any problem as long as the node label
is identical. Maybe we should add a check that this is the case and these
warnings won't confuse us anymore.

adamnovak:
Yeah, that's expected during the pruning step, which overlays two nondisjoint pruned subgraphs.

I think the right solution probably is a command-line flag to acknowledge and suppress the warning. Even if the labels are identical, it still might be weird enough to alert the user by default.

adamnovak:
OK, an option to suppress those warnings for vg view (`-D` or `--expect-duplicates`) has been added, and the wiki page has been updated. This should be fixed.

PR #1000, opened 2017-08-29
Suppress duplicate warnings

adamnovak:
Now you can pass `-D` or `--expect-duplicates` to `vg view` when reading VG or JSON to tell it not to throw warnings when the same node or edge exists multiple times. This is useful when combining overlapping subgraphs during the pruning step.

After I merge this I am going to update https://github.com/vgteam/vg/wiki/working-with-a-whole-genome-variation-graph to include the option, which will address #933, but then you will need to be using master vg (and not the last release) in order for the instructions to work.

I've also attached some updates to the variant filter script, to save on Jenkins runs.


jeizenga:
PR 1000!
🎈🎉🎊🎉🎈

PR #1001, opened 2017-08-29
Add common1kg to chr21 test

### LINKED ISSUE #993: Add common1kg for chr21 test
We want a graph without the rare variants in repeats

adamnovak:
Now we should see the common1kg graph (generated with variants filtered based on frequency, expected read error rate, and local repetitiveness to exclude variants we would see more via errors than via real reads) in our chr21 test.

Fix #993


adamnovak:
Everything looks more or less how we expect it to look here.

PR #1003, opened 2017-08-30
Make warnings show up in CI reports that don't run Toil in-process

adamnovak:
~The warnings disappeared around when I merged https://github.com/vgteam/vg/pull/998, but those commits don't seem to actually be the cause.~

~Now the logger needs to be manually configured for some reason, and the message format is slightly different. But this makes it work again.~

### DIFF
--- a/jenkins/mine-logs.py
+++ b/jenkins/mine-logs.py
@@ -237,7 +237,7 @@ def get_vgci_warnings(tc):
     if tc['stderr']:
         for line in tc['stderr'].split('\n'):
             # For each line
-            if "WARNING vgci" in line:
+            if "WARNING" in line and "vgci" in line:
                 # If it is a CI warning, keep it
                 warnings.append(line.rstrip())
     return warnings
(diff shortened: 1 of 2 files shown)


glennhickey:
The warnings were lost due to this copypaste-o:

https://github.com/vgteam/vg/pull/998/files#diff-e49ccbeaca6d6c530c3fa11683f0cf05R209

There's a simulfix in #1002

adamnovak:
Hm. Looks like I added some kind of testing line and forgot to remove it.

I think I got even more confused when developing *this* patch; I started with the commit before all the Unicode stuff and added a test warning line at the beginning of one of the short non-mapeval tests. It didn't show up in the output.

I think we've been relying on running Toil in-process in order to get logging set up. We only ever used warnings after doing mapeval, which we do by running a Toil workflow in the current Python, which results in Toil configuring logging for us. That explains why it wasn't set up for the other tests, and why when I run basicConfig I end up with a different format.

So this PR addresses *a* case when warnings won't be displayed, but *not* the case that's actually been giving us trouble (which is fixed in #1002).

Assuming pre-configuring the logger doesn't give us trouble when Toil tries to configure it later, we probably actually want both sets of changes.

Issue #1005, opened 2017-08-30
Jenkins error rate is too high


jeizenga:
It looks to me like we're simulating reads with a 5% substitution rate:
https://github.com/vgteam/vg/blob/master/jenkins/vgci.py#L453

That's way too high for short reads--it's getting into nanopore territory. I think that's going to skew our results a lot.

glennhickey:
Good point!  I would have thought those numbers were taken from
scripts/map-sim, but that has -e 0.01 -i 0.002.

PR #1006, opened 2017-08-30
Lower simulation error rates to match with map-sim example

glennhickey:
Current rates are unrealistically high, as pointed out in #1005.


glennhickey:
I'm not sure how useful it would be to merge this.  On one hand, we
definitely want to use the lower rates to evaluate vg.  But on the other,
having the higher rates helps us find problem cases faster, which is
important as the tests are already slow.

glennhickey:
will do this when switching to trained simulator

PR #1010, opened 2017-08-30
Allow more control over tag names in Jenkins tests

glennhickey:
This will allow multiple tests to be made for the same region/baseline graph pair using the tag_ext parameter (see test_sim_brca2_snp1kg_mpmap for an example).  @jeizenga is this the issue you were referring to?  On a related note, do you need paired end tests activated for mpmap?  This is turned off inside toil-vg.


jeizenga:
Yeah, that should work. I guess I still don't understand why I'm not getting plots for this PR: https://github.com/vgteam/vg/pull/989

glennhickey:
That PR seems to have the broken log miner (fixed here #1003), so we can't
see the error message we want in the Jenkins report.  Can you rebase off
master?

PR #1012, opened 2017-08-31
Note decreases in stats

### LINKED ISSUE #995: Add emoji for test value changes

adamnovak:
This should fix #995 even though it doesn't use actual emoji.

Anything outside the basic multilingual plane shows up as busted XML entities for surrogate pairs in the actual output. I blame (my?) Python's internal reliance on 16-bit characters.

Anyway, we'll now have down arrows in front of the values for statistics that dropped, to make them easier to spot.


adamnovak:
Something is (still?) breaking the plotting of Jordan's PR plot. I think it's trying to put text at NAN positions. It may be because the plot ends up having 0 extent somehow. I don't think it's caused by this PR, so I'll open an issue for it and see if I can solve it.

PR #1015, opened 2017-09-01
Stop pr plots from breaking for single points

### LINKED ISSUE #1014: mpmap PR plot doesn't render
We get non-TRUE/FALSE values out of conditions, because I think we're comparing against NANs when planning text positions, because I think the plot might have a 0 dimension somewhere.

adamnovak:
This should fix #1014 by manually setting the PR plot bounds to reasonable values.


adamnovak:
This looks nice now, and Travis is just waiting on an OS X build that shouldn't have broken.

PR #1022, opened 2017-09-01
Activate trained simulator in Jenkins

glennhickey:
This is just to see what happens when switching over to @jeizenga's fancier simulation model.  I think a better fastq (longer reads?), as well as more thought on the parameters is required before merging.


adamnovak:
Jenkins, test this please

adamnovak:
Looks like Jenkins encountered https://github.com/vgteam/toil-vg/issues/313

We should update to the latest toil-vg to solve this problem.

glennhickey:
Jenkins, test this please.

PR #1025, opened 2017-09-06
Add script that can run toil-vg on AWS and compare graphs

### LINKED ISSUE #1019: Create a big AWS-based script to run for comparing performance on various graphs and seeing when we're done
It should do some of the logic that the Jenkins tests do in terms of figuring out what files to pull from S3 for a graph, and it should expose plots and tables from VGCI (possibly in HTML reports).

It should run against autoscaling Toil on AWS instead of on a single machine, and it should have a nice command-line (or inline variable-based) interface,

adamnovak:
You run something like:

```
scripts/compare-graphs.sh ./out anovak BRCA1 snp1kg primary
```

Although you need to replace `anovak` with the ID of your SSH key in AWS.

It will only work if you have access to UCSC's AWS buckets. It also sometimes gets stuck if Toil gets stuck and refuses to run jobs; I think this is Toil's fault.

Closes #1019.

### DIFF
--- a/scripts/compare-graphs.sh
+++ b/scripts/compare-graphs.sh
@@ -10,7 +10,7 @@ GRAPH_NAMES=( snp1kg primary )
 REGION_NAME="BRCA1"
 
 # What cluster should we use?
-CLUSTER_NAME="novakcluster"
+CLUSTER_NAME="novakcluster2"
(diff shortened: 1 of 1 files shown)


adamnovak:
OS X Travis is hosed so I am not going to wait on it.

PR #1026, opened 2017-09-06
Use latest toil-vg and latest released toil and make customizeable

adamnovak:
Should help with the missing Toil ordering constraints problem in #1022.


adamnovak:
Travis succeeded and just never got around to telling Github.

https://travis-ci.org/vgteam/vg/builds/272690449

PR #1027, opened 2017-09-09
Update README.md

konradjk:
google-perftools seems to be obsolete (`Error: google-perftools has been replaced by gperftools; please install that instead.`) and `md5sha1-sum` is not found (but there is a `md5sha1sum`)


konradjk:
Incidentally, it doesn't look like `autogen` is actually required (mine is broken, but it just built fine without it)

adamnovak:
Jenkins, test this please.

PR #1040, opened 2017-09-14
Plot recall correctly

adamnovak:


adamnovak:
These plots look much less zaggy. That makes more sense.

PR #1043, opened 2017-09-14
Switch to min-position gam lookup in vg chunk

glennhickey:
In most cases in toil-vg, we don't need to check every position in an alignment when pulling mappings by node id, because we're querying by range.   

So we switch from Index::for_alignment_to_nodes() to Index::for_alignment_in_range() (which is getting fixed to use min instead of first position in #1041).  Old logic preserved with -A option.


adamnovak:
Travis is happy on Linux.

PR #1044, opened 2017-09-14
Clean xg

adamnovak:
I've refactored MEMs out away from XG, and set out what changes I want to make to make everything happen in terms of the new internal representation.


adamnovak:
Travis has run out of mac potatoes or something. The Linux build worked.

PR #1048, opened 2017-09-14
Make xg smaller again by removing duplicate storage of sequences

### LINKED ISSUE #1046: xg::XG::g_iv should contain an offset in xg::XG::s_iv, rather than the actual sequence
Storing the sequence in g_iv is a bad idea and is extremely wasteful in terms of RAM. In theory it would help us in terms of locality of memory access, but this also increases the memory used by the sequence storage because we are only bitcompressing the bases.

ekg:
Remove sequence storage from g_iv (resolves #1046)


adamnovak:
Travis is happy on Linux.

adamnovak:
@ekg says he already did this

PR #1052, opened 2017-09-15
size_t is unsigned and therefore does not need an call to abs. 

JervenBolleman:
If min_value is also a size_t then they are the same type and we can use '<' directly


adamnovak:
Travis is happy with this, but Mac builds are super slow and timing out.

PR #1059, opened 2017-09-16
Use a trivial paired-end mapper

adamnovak:
This will let us get Jenkins results that at least get as high precision
as the single-end mapper.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -1680,6 +1680,34 @@ bool Mapper::pair_consistent(const Alignment& aln1,
     return length_ok && orientation_ok;
 }
 
+pair<vector<Alignment>, vector<Alignment>> Mapper::align_paired_multi_trivial(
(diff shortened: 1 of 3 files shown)


ekg:
Is there any reason not to merge this?

ekg:
Oh, is this something we can enable via a command line flag?

adamnovak:
This should not be merged really, because it hardcodes the mapper to this trivial implementation.

I'm concerned that this passes the tests, because that indicates that our tests are not very stringent. But there's also a question of why this does even vaguely well.

adamnovak:
OK it turns out that I hadn't hit all the calls to the old mapper, so the old mapper was still running.

*Now* this should actually be running the mapper I want it to.

adamnovak:
The trivial test mapper now matches the single-end mapper exactly. There's nothing weird going on here anymore.

PR #1061, opened 2017-09-16
Easy paired end

adamnovak:
A slightly better version of my super easy paired end mapper.


adamnovak:
It looks like with this mapper implementation we can get more or less perfect precision. But it comes at a cost in recall, and also apparently it silently fails to run the mapper for some tests.

adamnovak:
OK, merging in master I think made my primary-pe do worse than before. But it's still doing better than master is currently I think.

PR #1062, opened 2017-09-16
Optimize mpmap

jeizenga:
Optimized mpmap


jeizenga:
@adamnovak @glennhickey  I'm failing one test here on chromosome 21, but the difference looks pretty negligible to me. You think it would be OK to merge over it?

adamnovak:
Somehow https://github.com/vgteam/vg/pull/1086 doesn't fail this test, even though the only commit that's here and not there is 96d534b which doesn't touch the test that is failing here.

Unfortunately, #1086 fails an mpmap test, so I'm worried about merging it.

We should probably just merge both of them.

glennhickey:
I may have made thresholds too tight for the non-determinism in vg map.

adamnovak:
Yes, it does look like maybe we should widen them a bit. We can decide at the meeting.

PR #1067, opened 2017-09-18
Drop some graphs now covered by the comparison runner

### LINKED ISSUE #1017: Remove some graphs from the Jenkins tests
We're wasting time running comparisons between graphs in the guise of regression tests. We should compare our graphs with separate experiments.

adamnovak:
Closes #1017.

We need to look out for missing graphs that were supposed to run manually, now, since I turned off the assertion that we didn't lose any.

### DIFF
--- a/jenkins/vgci.py
+++ b/jenkins/vgci.py
@@ -751,9 +751,13 @@ class VGCITest(TestCase):
                 if len(stats_dict[key]) > 0:
                     self.assertTrue(stats_dict[key][0] == reads)
                 if len(stats_dict[key]) > 1 and len(val) > 1:
+                    # Compare accuracy stats
                     self.assertTrue(stats_dict[key][1] >= val[1] - acc_threshold)
                 if len(stats_dict[key]) > 2 and len(val) > 2:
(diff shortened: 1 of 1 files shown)


adamnovak:
@jeizenga Should I care that the mpmap tests failed?

jeizenga:
I don't think so. One of them seems to be that weirdness where perfect classification gives 0 AUC, the other one doesn't even look like it has a baseline.

adamnovak:
@jeizenga How did the test get into master without getting a baseline?

jeizenga:
Beats me

adamnovak:
OK, I will try and fix these tests; they're failing for every PR and upsetting Jenkins.

adamnovak:
OK it looks like the baseline is never updated if the tests fail. So you need to make sure the tests pass in your PR, without a baseline, because if they don't pass in the PR they're *not* going to pass later on master.

adamnovak:
Travis macs are still hosed, but everything else seems happy with this change. I'm going to merge it in.

PR #1070, opened 2017-09-19
Don't crash when trying to compare to baseline file that doesn't exist

glennhickey:
Hopefully this will get the new mpmap tests passing.  

Presently, the baseline is never getting updated (which is bad) as this only happens when all tests pass.  It may make more sense to do this on every merge to master instead.


glennhickey:
The useful part of this (not defaulting to 0 on score test) was scooped by a parallel PR.

PR #1071, opened 2017-09-19
Slightly harder paired end

adamnovak:


adamnovak:
I think that this is not the approach to take.

PR #1072, opened 2017-09-19
Always update baseline on merge

glennhickey:
This will keep the baseline data level with master.   But it will sweep away traces of regressions that get merged in.  @adamnovak what do you think about this?

### DIFF
--- a/jenkins/jenkins.sh
+++ b/jenkins/jenkins.sh
@@ -236,10 +236,10 @@ then
     tar czf "${VG_VERSION}_output.tar.gz" vgci-work test-report.xml jenkins/vgci.py jenkins/jenkins.sh vgci_cfg.tsv
     aws s3 cp --acl public-read "${VG_VERSION}_output.tar.gz" s3://cgl-pipeline-inputs/vg_cgl/vg_ci/jenkins_output_archives/
 
-    # if success and we're merging the PR (and not just testing it), we publish results to the baseline
-    if [ "$PYRET" -eq 0 ] && [ -z ${ghprbActualCommit} ]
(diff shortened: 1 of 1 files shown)


adamnovak:
I already pushed in a fix for the problem we had where the score worsening
check could fail if there was no baseline yet. So we hopefully wouldn't
need this for that. But it probably does make sense for the baseline to
always be the last master run, because we don't want master to fail forever
if we do end up lowering something below baseline.

jenkins/jenkins.sh
```diff
@@ -236,10 +236,10 @@ then
     tar czf "${VG_VERSION}_output.tar.gz" vgci-work test-report.xml jenkins/vgci.py jenkins/jenkins.sh vgci_cfg.tsv
     aws s3 cp --acl public-read "${VG_VERSION}_output.tar.gz" s3://cgl-pipeline-inputs/vg_cgl/vg_ci/jenkins_output_archives/
 
-    # if success and we're merging the PR (and not just testing it), we publish results to the baseline
-    if [ "$PYRET" -eq 0 ] && [ -z ${ghprbActualCommit} ]
+    # if we're merging the PR (and not just testing it), we publish results to the baseline
+    [ -z ${ghprbActualCommit} ]
```
adamnovak: @glennhickey Looks like you dropped the if; I think baseline might be getting updated from PRs now? Or maybe never updated?

glennhickey:
I blame jet lag.  It's been fixed though:
https://github.com/vgteam/vg/pull/1075

jenkins/jenkins.sh
adamnovak: Also looks like you fixed it yesterday.

PR #1076, opened 2017-09-20
Serious paired end

adamnovak:
I'm trying to turn #1061 into a serious mapper implementation, which means it needs to generate good MAPQs and emit multimappings in the expected way.


adamnovak:
OK, between the adjustment of the rescue thresholds, the deduplication logic, the enabling of proper multimap handling, and the addition of mapq recalculation, my mapper went from this:

image

To this:

image

I suspect the big drop in max precision is caused by the mapping quality recalculation. We have about the same recall when admitting all the reads, but we aren't able to actually filter out those mismappings and differentiate them from the true mappings. The MAPQ QQ plots go from this:

image

To this:

image

The new QQ plots just look like nonsense.

I think this is good news; I'm just calling out to the pair mapping quality routine that we use in the real mapper. So if that routine is bogus, it could also explain why we can't get to high precisions with the master mapper.

adamnovak:
Pulling out most of the uses of the paired-end mapping quality calculator did not solve things. So I suspect it must be one of the adjustment of the rescue thresholds, the deduplication logic, and the enabling of proper multimap handling that brought my performance down.

adamnovak:
I'm not actually proposing we use this or #1061 anymore. We have two other good mappers.

PR #1083, opened 2017-09-21
Improve Jenkins test reports

### LINKED ISSUE #1080: Make Jenkins PR plots log-log

adamnovak:
Now we will have log-scale recall and also PR titles.

Fixes #1080.

### DIFF
--- a/jenkins/mine-logs.py
+++ b/jenkins/mine-logs.py
@@ -139,6 +139,7 @@ def scrape_messages(text):
 
     messages = []
     msg = None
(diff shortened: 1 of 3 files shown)


adamnovak:
@jeizenga: It looks like this only failed because of the mpmap test that always fails. I'm going to merge.

Can we fix that test and/or pull it out of master? If it always fails it's not particularly useful.

glennhickey:
That test should be fixed since #1084

PR #1086, opened 2017-09-21
Start adding some tests for the multipath mapper

adamnovak:
Also refactors how we pull out aligners to use.


adamnovak:
The Mac compiler is unhappy:

```
src/multipath_mapper.hpp:100:21: note:   no known conversion for argument 3 from 'LRUCache<long long int, vg::Node>' to 'LRUCache<long int, vg::Node>&'
```

PR #1100, opened 2017-09-25
move mpmap testing into chr21 test

glennhickey:
There's the question of which simulation to use, but that'll be done in separate PR.


adamnovak:
It looks like this roughly doubled our Jenkins test time, from 2 to 4 hours.

glennhickey:
Yeah, we can reduce the number of reads to speed this up.  Sounds like
imminent mpmap improvements should help too.

This doesn't seem related to recent issues of all/most tests timing out
though.

PR #1103, opened 2017-09-25
✨🕴️Auto-magic Makefile!🔮🧙

adamnovak:
I'm deleting all the dependency and object list stuff from the Makefile. Now it will just build objects from whatever .cpp files it finds in the right places, and magically track dependencies (correctly, instead of what we did before where half the dependency links weren't present and things would get out of date).

I meant to be doing #1093, but then I decided I needed to split up vg_algorithms.cpp by algorithm, and *then* I decided I didn't want to make any more manual dependency lists.

If you have .cpp files laying around that aren't getting built now, they will get built and linked in, so you should probably delete or commit them.


adamnovak:
Jenkins is upset in the test uploader:

```
+ jenkins/mine-logs.py test-report.xml vgci-work/ report-html/ summary.md
Traceback (most recent call last):
  File "jenkins/mine-logs.py", line 544, in <module>
    sys.exit(main(sys.argv))
  File "jenkins/mine-logs.py", line 541, in main
    write_html_report(xml_root, options.work_dir, options.html_out_dir)
  File "jenkins/mine-logs.py", line 499, in write_html_report
    header = html_header(xml_root)
  File "jenkins/mine-logs.py", line 357, in html_header
    report += ' Error parsing Test Suite XML\n'
UnboundLocalError: local variable 'report' referenced before assignment
```

PR #1105, opened 2017-09-26
Try trained sim for chr21

glennhickey:
Just to see what happens.  Would like better training data before merging.


glennhickey:
Jenkins, test this please.

PR #1106, opened 2017-09-26
Refactor extract_extending_graph to use handles

adamnovak:
This will save a bunch of selects going from IDs to xg vector positions, and so *should* be faster, even though we just ignore all the caches now.

Still TODO: drop the xg and vg specializations and make all calls in vg use the `HandleGraph*` overload. Also don't bother making the caches.


adamnovak:
@jeizenga If you pulled from here, you should probably update. I fixed a segfault that would happen whenever nodes were larger than your graph extraction distance.

PR #1107, opened 2017-09-26
Mpmap gooder and gooder

jeizenga:
A place for my continuing optimizations of mpmap. So far, a small bug fix, better handling of repeats in the clustering step, and a more sophisticated mapping quality formula that takes fragment length into account. More changes probably to come.


jeizenga:
Jenkins, test this please.

jeizenga:
Jenkins, test this please.

jeizenga:
@adamnovak Well, it's faster now, but it's also not running any tests

jenkins/jenkins.sh
```diff
@@ -22,9 +22,9 @@ REUSE_VENV=0
 # Should we keep our test output around after uploading the new baseline?
 KEEP_OUTPUT=0
 # Should we show stdout and stderr from tests? If so, set to "-s".
-SHOW_OPT=""
+SHOW_OPT=""q
```
adamnovak: @jeizenga Someone seems to have written a bug here.

adamnovak:
```
============================= test session starts ==============================
platform linux2 -- Python 2.7.11, pytest-3.2.2, py-1.4.34, pluggy-0.4.0 -- /mnt/ephemeral/workspace/vg/.env/bin/python
cachedir: .cache
rootdir: /mnt/ephemeral/workspace/vg, inifile:
collecting ... 
------- generated xml file: /mnt/ephemeral/workspace/vg/test-report.xml --------
========================= no tests ran in 0.00 seconds =========================
ERROR: file not found: q
```

jeizenga:
@ekg I found a logical bug in my design of the find_sub_mems_fast algorithm, which might change the behavior of `vg map`. I figure I'd get a Jenkins run going so we can see how it's going to affect things. 

The bug caused the algorithm to sometimes find sub-MEMs that weren't actually "maximal", so that in general there were more hits, but the hits were artificially short. Getting rid of the extra hit should make things faster, but I'm not sure how it will affect accuracy. The bug gave us more hits, but it also makes it more difficult to distinguish between good hits and bad hits. 

If it ends up causing problems, I can either implement the old behavior as an alternative code path, or we can think about a recursive version of the algorithm.

jeizenga:
@ekg I think on balance this improved the map performance. Please advise.

PR #1114, opened 2017-09-28
Getting a handle of XG ha ha ha...

jeizenga:
I did some bug-fixes on the handle based traversal for XG and also switched to a handle-based traversal for the closest shared path oriented distance estimation function. This commit was cherry-picked from mpmap-good.


jeizenga:
Redundant with https://github.com/vgteam/vg/pull/1117 and https://github.com/vgteam/vg/pull/1107

PR #1115, opened 2017-09-28
Revise mpmap testing

adamnovak:
Rather than refactor everything to be public static methods, I'm making all the internal stuff we might want to test protected (instead of private) and defining a class in the tests that exposes it.
I think this makes sense: we need to test the internal API that derived classes might want to use, but in Catch we can't do friend declarations, so this is the next best way.

I'm also adding a couple of new tests that make sure topologically_ordered_subpaths isn't incredibly broken.

### DIFF
--- a/src/multipath_mapper.cpp
+++ b/src/multipath_mapper.cpp
@@ -569,17 +569,10 @@ namespace vg {
                                                const vector<MaximalExactMatch>& mems,
                                                const vector<memcluster_t>& clusters) -> vector<clustergraph_t> {
         
-        // Call the static implementation and point it at the parts of us it needs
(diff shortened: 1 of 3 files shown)


adamnovak:
`E   AssertionError: 0.981275507129 not greater than or equal to 0.981689777152`

This looks like nondeterminism to me.

Jenkins test this please.

adamnovak:
A bunch of little tests on Jenkins that I didn't touch are failing here, so I'm stopping the test.

glennhickey:
It's even happening to toil-vg
http://jenkins.cgcloud.info/job/toil-vg/604/console

adamnovak:
Jenkins, test this please.

adamnovak:
OK, Jenkins has been fixed, and the failure here looks mostly like AUC nondeterminism; we see the same thing on master.

PR #1116, opened 2017-09-29
Make simulated insertion sequences random

glennhickey:
Also increase test threshold by 50%.


adamnovak:
Looks like one of the vg mod tests failed:

```
not ok 37 - the expected graph translation is exported when the graph is edited
#          got: '1288'
#     expected: '1280'
```

I suspect the simulated reads it uses changed, and so the right answer changed.

PR #1117, opened 2017-09-29
Fix XG handles and refactor another algorithm

adamnovak:
It looks like the XG-based handle interface never actually worked; it was trying to pull sequence data from the wrong vector. But it was also pulling sequence length wrong, so it managed to even out and not crash.

I've fixed that, and also refactored another one of the vg::algorithms to use handles.

### DIFF
--- a/jenkins/jenkins.sh
+++ b/jenkins/jenkins.sh
@@ -234,7 +234,7 @@ then
 
     # we publish the results to the archive
     tar czf "${VG_VERSION}_output.tar.gz" vgci-work test-report.xml jenkins/vgci.py jenkins/jenkins.sh vgci_cfg.tsv
-    aws s3 cp --acl public-read "${VG_VERSION}_output.tar.gz" s3://cgl-pipeline-inputs/vg_cgl/vg_ci/jenkins_output_archives/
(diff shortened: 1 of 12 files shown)


adamnovak:
Jenkins, test this please.

adamnovak:
```
WARNING:toil.leader:P/U/jobv0E_Dz    CalledProcessError: Command '['docker', 'run', '--rm', '--log-driver', 'none', '--entrypoint', 'vg', '--env', 'TMPDIR=.', '--env', 'LC_ALL=C', '-v', '/mnt/ephemeral/tmp/toil-c3ceca9b-5c2b-42b5-9ad8-a9a652e696cd-dd1bac9cb23792a632bd6ee157719517/tmpFBMAV9/534ac06f-d12e-4f20-a54e-0ed3ba4ffe24/tbVqMdc:/data', '-w', '/data', '--name', u'runchunkalignment--Ci6ztUhYJXDN', u'jenkins-docker-vg-local', 'mpmap', '-S', '-A', '-G', 'reads_chunk_0_0.gam', '-t', '8', '-i', '-x', 'graph.vg.xg', '-g', 'graph.vg.gcsa']' returned non-zero exit status 139
```

I will pull the dumped failing files (or make my own) and see what's wrong.

adamnovak:
This passes the tests, but the new tests I add in #1126 found a bug that I think should block merging. I fixed it in #1126, so we should use that instead of this.

PR #1118, opened 2017-09-29
Fix Jenkins speed

adamnovak:
This PR is checking for an interaction between #1106 and #1113 which may have tanked Jenkins' performance.


adamnovak:
Actually it seems like Jenkins' speed was probably tanked by Lon's new AMI for the Jenkins instances, which updates Docker to something post-2015. I'm not sure why we wouldn't be compatible with new Docker at reasonable speeds.

I worked with Edwin to set us up with a special Jenkins instance type with the old AMI, so if that really is the cause then we should start seeing faster Jenkins builds for jobs starting with http://jenkins.cgcloud.info/job/vg/692/ building #1116.

glennhickey:
Thanks!  Seems like that's the issue.  Looking at the logs, it seems many
docker commands were running in reasonable times until something decided to
hang.  So either this AMI is making something unstable, or maybe we have
one problem image that's just not working with it..

adamnovak:
Jenkins has been running tests without them timing out, so I think the problem was fixed on the Jenkins end, and this commit I'm reverting here had nothing to do with it.

Why we can't run on modern Docker is still an open question.

PR #1128, opened 2017-10-03
re-activate mhc calling test

glennhickey:
This will give a better sense of where the calling is than the BRCA1 test.  @ekg, can you please pull this into the mapper branch?  There seems to be something up with map on that branch where this test fails with `vg: src/mapper.cpp:1782: bool vg::Mapper::pair_consistent(const vg::Alignment&, const vg::Alignment&, double): Assertion `aln1.fragment_size() == aln2.fragment_size()' failed.`


ekg:
Heh, I just did the same thing on the mapper branch.

Issue #1130, opened 2017-10-03
extract_connecting_graph step 2 mixes parts of different positions


adamnovak:
https://github.com/vgteam/vg/blob/6b0f4767392727cd0ac6cb315654cbd9b3f8d17b/src/algorithms/extract_connecting_graph.cpp#L263

This line mixes the ID of pos_2 and the orientation of pos_1 and that makes me super uncomfortable.

Changing it to be `id(pos_2), !is_rev(pos_2)` does not fail the unit tests, which also makes me uncomfortable.

@jeizenga Can you explain why it should be the way it is? Or should it be changed?

jeizenga:
That very much looks like a copy-paste bug to me, and I think you've identified the right way to correct it. Strange that it wouldn't be caught in the unit tests though. It must be that all of the tests that follow this code path have both positions on the same strand.

adamnovak:
I think this is fixed now.

PR #1131, opened 2017-10-03
Refactor extract_connecting_graph to use handles

### LINKED ISSUE #1093: Make algorithms.hpp algorithms be about handles, to speed up mpmap

adamnovak:
See also: #1130 

Fixes #1093

### DIFF
--- a/src/algorithms/extract_connecting_graph.cpp
+++ b/src/algorithms/extract_connecting_graph.cpp
@@ -10,13 +10,13 @@
 
 namespace vg {
 namespace algorithms {
-    unordered_map<id_t, id_t> extract_connecting_graph_internal(const HandleGraph* source, Graph& g, int64_t max_len,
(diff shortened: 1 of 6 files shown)


adamnovak:
I did some benchmarking, and the new implementation based on handles, against an xg, seems to be about twice as fast as the old one. I'm going to include my benchmarking code in another PR.

On the other hand, it looks like our overall runtime for the chr22 graphs is up by about 300 seconds between http://cgl-pipeline-inputs.s3.amazonaws.com/vg_cgl/vg_ci/jenkins_reports/commit/3eb53341f1780bbc0758381a5274f53e82d10b0f/index.html and http://cgl-pipeline-inputs.s3.amazonaws.com/vg_cgl/vg_ci/jenkins_reports/commit/9f21bdc589f5786132b39b2bee66d7b4652757ad/index.html

PR #1133, opened 2017-10-03
Benchmarks

adamnovak:
Adds some benchmarks on top of #1131. You can add new benchmarks in `src/subcommand/benchmark_main.cpp`. If we want more than a few I can try and set up an auto-registration system and a way to call individual benchmarks.

I want to see if the handle-based algorithm is really slow by running it again in this PR; my benchmarks say that my refactor is always faster (at least for the graph and query I am testing), while the Jenkins tests seemed to say that that I got a slowdown on tests that don't even call my function.

My benchmarks work by looping some code under test, interleaved with executions of some constant "control" code, and normalizing the runtime of the code under test by the runtime of the control. This produces a benchmark "score", which is essentially 1000 times the number of times you could run the code under test in the time it takes to run the control code.

I did it this way to try and normalize out variable load on the underlying system; it seems to be pretty necessary, because I am seeing basically the whole program running faster or slower by a factor of 2 on Kolossus between program runs.

### DIFF
--- a/src/algorithms/extract_connecting_graph.cpp
+++ b/src/algorithms/extract_connecting_graph.cpp
@@ -10,16 +10,13 @@
 
 namespace vg {
 namespace algorithms {
-    unordered_map<id_t, id_t> extract_connecting_graph_internal(Graph& g, int64_t max_len,
-                                                                pos_t pos_1, pos_t pos_2,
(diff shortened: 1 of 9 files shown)


adamnovak:
OK, it no longer looks slow. I must have had a slow Jenkins node.

PR #1140, opened 2017-10-05
Add a way to set the distribution for mpmap

### LINKED ISSUE #1111: Add argument(s) to set the mpmap fragment length distribution

adamnovak:
We also probably want to hint the distribution during the snp1kg test so that we get a runtime more comparable to the other cases.

Fixes #1111


adamnovak:
Since mpmap is now fast we maybe don't need to do the hinting to the tests.

PR #1129, opened 2017-10-03
Add mpmap calling test

glennhickey:
@jeizenga Let's get this test running, so we can make calls on mpmap output.  Currently, pileup is complaining:

     vg: src/pileup.cpp:256: void vg::Pileups::compute_from_alignment(vg::Alignment&): Assertion `alignment.sequence().empty() || alignment.path().mapping_size() == 0 || read_offset == alignment.sequence().length()' failed.

It wants the total to_lengths to be equal to the read length. 

You can see the read that's triggering it here (read length 101, total to_length 34)

     vg view vgci-work/outstore-BRCA1-snp1kg-mpmap/NA12878_default.gam -a | grep ERR194147.636304843/2

{"sequence": "TTTTGTATTTTTTTTTTTTTTTAGTTAATGGCGGAATTTCTCCGTTTTTAGTGGGAATGTTTTTTTTTTTCTACTTTTAAATTCAACACCCCGGGCCTCCC", "path": {"mapping": [{"rank": 1, "position": {"node_id": 7279, "offset": 5, "is_reverse": true}, "edit": [{"from_length": 1, "to_length": 1}]}, {"rank": 2, "position": {"node_id": 7278, "is_reverse": true}, "edit": [{"from_length": 1, "to_length": 1}]}, {"rank": 3, "position": {"node_id": 7276, "is_reverse": true}, "edit": [{"from_length": 14, "to_length": 14}]}, {"rank": 4, "position": {"node_id": 7273, "is_reverse": true}, "edit": [{"from_length": 9, "to_length": 9}]}, {"rank": 5, "position": {"node_id": 7268, "offset": 23, "is_reverse": true}, "edit": [{"from_length": 9, "to_length": 9}]}]}, "name": "ERR194147.636304843/2", "quality": "GxsbGyEbGyEWGyEbISEhISEhISEhGwYPDwYWDwYPBgYGBgYGDw8GBg8PDwYPDw8GBgYPDw8GDwYGBg8PBg8PFhsWDw8GBg8PBgYPDwYPBgYPDwYPDwYPDw8GDwYGBgYGDw8WBgY=", "mapping_quality": 60, "score": 30}

PR #1143, opened 2017-10-06
No more empty edits in MultipathAlignments from mpmap

jeizenga:
Fixed a bug in the banded global aligner that could cause empty edits when the graph had empty nodes. Resolves issue discussed in https://github.com/vgteam/vg/pull/1129


jeizenga:
I don't know why this is still showing up as waiting for Jenkins, which has clearly finished and passed. I'm merging.

PR #1144, opened 2017-10-06
Use algorithmic topological_sort and sort-based graph operations

adamnovak:
This changes everywhere we do topological sorts to use the new `vg::algorithms::topological_sort`. It also moves `sort()` and `orient_nodes_forward()` from the `VG` object to the `vg::algorithms` namespace.

According to my somewhat trivial benchmarks, this should be slightly faster:

```
# Benchmark results for vg v1.5.0-1466-geaa7d48
# runs	test(us)	stddev(us)	control(us)	stddev(us)	score	err	name
1000	1.55e+02	1.36e+01	2.62e+01	2.91e+00	169.68	24.00	vg::algorithms topological_sort
1000	1.88e+02	1.22e+01	2.56e+01	1.57e+00	136.34	12.13	VG direct topological_sort
1000	1.62e+02	2.69e+00	2.54e+01	8.49e-01	156.26	5.84	vg::algorithms sort
1000	1.98e+02	5.55e+00	2.54e+01	7.58e-01	128.34	5.26	VG direct sort
1000	1.54e+02	5.71e+00	2.51e+01	1.47e+00	163.07	11.27	vg::algorithms orient_nodes_forward
1000	2.77e+02	6.46e+00	2.54e+01	1.07e+00	91.77	4.41	VG direct orient_nodes_forward
1000	7.68e+01	1.71e+00	2.52e+01	7.47e-01	327.85	12.16	VG::get_node
1000	3.27e+02	9.57e+00	2.53e+01	6.47e-01	77.20	3.00	algorithms::extract_connecting_graph on xg
1000	2.81e+02	8.53e+00	2.53e+01	1.19e+00	90.09	5.03	algorithms::extract_connecting_graph on vg
1000	2.50e+01	5.58e-01	2.50e+01	5.54e-01	999.60	31.42	control
```


jeizenga:
Does this replicate the ordering heuristics and system-independent behavior in `VG::sort()`? I also implemented a linear time `VG::lazy_sort()` in https://github.com/vgteam/vg/pull/1107 that's only valid for DAGs and only guarantees topological ordering, not system independent behavior. I don't know how that would fit in with this new interface.

adamnovak:
I'm using the same algorithm for `algorithms::topological_sort` as we had for `VG::topological_sort`, adapted for handles.

I did ditch the new head node in plain `sort()`, although I didn't have to. Since we already are sorting between different heads by ID anyway, I don't think it matters if we have one head or many to start from, in terms of whether the sort is machine-independent or not.

In this new interface, `lazy_sort` might move into the `algorithms` namespace.

adamnovak:
My sort might not actually sort as much as it should sort. It looks like vg align is running into some trouble that could be explained by node ordering:

```
ok 12 - graphs where duplicated nodes need flipping can be used for alignment
error:[gssw] cannot align because node predecessors cannot provide seed
failing is node 12
not ok 13 - node flipping doesn't destroy the alignment
#          got: '1'
#     expected: '0'
error:[gssw] cannot align because node predecessors cannot provide seed
failing is node 12
not ok 14 - alignment correctly handles an inversion
#          got: '1'
#     expected: '0'
error:[gssw] cannot align because node predecessors cannot provide seed
failing is node 5
not ok 15 - the exploding graph doesn't blow up
#          got: '1'
#     expected: '0'
ok 16 - pinning doesn't cause problems for gssw
error:[gssw] cannot align because node predecessors cannot provide seed
failing is node 57
not ok 17 - unnamed test
#          got: '1'
#     expected: 'left pinning is correctly handled'
Failed 4/17 subtests 
```

Strangely Jenkins seems to have no trouble. But maybe those graphs are too simple and don't really need a good sort.

adamnovak:
The Mac build failed here because of the Travis Ruby/Homebrew buf I am trying to fix in another PR. I'm going to go ahead and merge this in.

PR #1146, opened 2017-10-08
Fix multithreading in mpmap paired end mapping

jeizenga:
Another premature PR from https://github.com/vgteam/vg/pull/1145. Edits the parallel functions for FASTQ and GAM so that they have the option to switch from single threaded to multithreaded mode. Before the multithreading on mpmap paired end mapping wasn't working, so it ran in single threaded the entire run.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -183,6 +183,13 @@ size_t fastq_unpaired_for_each_parallel(const string& filename, function<void(Al
 }
 
 size_t fastq_paired_interleaved_for_each_parallel(const string& filename, function<void(Alignment&, Alignment&)> lambda) {
+    return fastq_paired_interleaved_for_each_parallel_after_wait(filename, lambda, [](void) {return true;}, [](void) {});
+}
+    
+size_t fastq_paired_interleaved_for_each_parallel_after_wait(const string& filename,
+                                                             function<void(Alignment&, Alignment&)> lambda,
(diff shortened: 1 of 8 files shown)


jeizenga:
@glennhickey @adamnovak 
I'm not seeing any changes to the Travis script, but suddenly it's failing the Mac build? Tried restarting it already, but I don't think it's timing out. Looks to me like it has an out-of-date Ruby. How do we get an updated Ruby installed on the Travis potato?

adamnovak:
I think this was caused by https://github.com/Homebrew/brew/pull/3240 and specifically https://github.com/Homebrew/brew/commit/bcca2a7c6b80a6450bd8261af987a8da260b6b89

Homebrew no longer supports the Ruby 2.0.0 that we have been running on Travis. When we run `brew update`, we get the latest Homebrew, which doesn't run on the system-installed Ruby anymore.

Homebrew tries to be nice and install its own copy of Ruby 2.3.3 as it updates itself, but (I think because of the `rvm use` command that it looks like Travis is running for us at the top of the build) Homebrew continues to run under the existing Ruby 2.0.0 when invoked, and fails to work.

adamnovak:
Here's what looks like the corresponding Travis issue: https://github.com/travis-ci/travis-ci/issues/8552

jeizenga:
Pretty sure both of these failures aren't real.

PR #1150, opened 2017-10-09
More bucketing heuristics in mpmap

jeizenga:
Now tries to bucket hits that aren't included in any bucket using the nearest node(s). Hopefully this speed things up a bit.


jeizenga:
Only failing the bad Travis test, SMASHING THAT MERRGGGGE

Issue #1147, opened 2017-10-09
Make the Cactus test on Jenkins good


adamnovak:

adamnovak:
We could maybe do a simulation from ref and one of the alts on MHC, so that the Cactus graph will actually be useful.

adamnovak:
Evaluating correctness when mapping to the primary reference is tricky.

adamnovak:
We can use the Cactus alignment to define the correct part of the primary reference to map to for reads from an alt. If there's no aligned bases, we don't expect it to be possible to map that read correctly. We can mark it wrong, or we could remove it from the denominator.

adamnovak:
@ekg How are you planning to evaluate correctness for reads that don't actually have a true mapping location in the reference you're mapping to?

adamnovak:
Maybe we should be using alt-aware BWA as the control here. Which will probably win if we're just simulating from an alt.

adamnovak:
@benedictpaten thinks we should target MHC and yeast here.

adamnovak:
OK, here is my plan for how to get a better Cactus-based test:

- [ ] Modify vg sim to take a path name in the xg and simulate from it.
- [ ] Modify vgci.py to use the simulate-from-path functionality.
    - [ ] Alternately, extract individual names paths from the base graph as xg indexes and plug into the existing logic that lets us simulate from haplotypes represented as separate XGs.
    - [ ] Make sure simulated reads are annotated on all paths they touch.
- [X] Toil-vg should already annotate reads with positions on all the paths they touch in the main xg file used for alignment.
- [ ] Make mapeval handle multiple annotations cleverly: always take the reference path one, and ignore all the others. Right now it only takes the 0th annotation. Things with no reference path annotation can be treated as unmapped/censored for now.
- [ ] Create a Jenkins test that simulates reads from a couple of the alts, maps them to the graphs (including Cactus), and evaluates, of the reads that can be annotated with true reference positions, which are mapped correctly.
- [ ] Add alt-aware BWA-mem as a control. This will potentially be difficult because I haven't seen much documentation on setting up your own custom alt graph, and we would need to do one restricted to just MHC.
    - [ ] Make sure to surject/subset the alt alignments onto the ref path for comparison.
    - [ ] Alternately, add functionality to mapeval to compare positions on multiple paths and use that.

glennhickey:
[...]

PR #1151, opened 2017-10-09
Turn the main MHC test back on and add Cactus to it

adamnovak:
This means we have at least a basic cactus test, though in #1147 I want an actually-good Cactus test that doesn't just throw reads from ref-topology genomes at it.


adamnovak:
I'm going to overrule broken Mac travis and merge this.

PR #1152, opened 2017-10-09
Subdue travis

adamnovak:
This might fix @jeizenga's run-in with https://github.com/travis-ci/travis-ci/issues/8552 in #1146.


adamnovak:
Looks like we're stuck and we need to run the latest (or at least a newer)
Homebrew. We'll have to tinker with RVM and hope it doesn't upset Travis's
code.

PR #1156, opened 2017-10-11
Travis can't run out of time building protobuf if there's no protobuf!

adamnovak:
Checkmate Travis.

Everybody would have to get Protobuf 3 on their system if this merges in. It's easy for everyone except LTS Ubuntu users, who may need to use a PPA or build from source. We might need to touch the vg_docker Dockerfile too.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,9 +22,6 @@
 [submodule "sha1"]
 	path = deps/sha1
 	url = https://github.com/vog/sha1.git
-[submodule "protobuf"]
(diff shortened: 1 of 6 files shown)


adamnovak:
Travis really doesn't want to find the Protobuf libraries, and without an OS X Homebrew system and an Ubuntu 14.04 system I can't effectively debug it. Since we now have master running through the Travis tests successfully at least some of the time, I'm going to abandon this, at least for now.

PR #1157, opened 2017-10-12
Faster wavelet tree for XGPath node IDs

jeizenga:
A branch off of https://github.com/vgteam/vg/pull/1150 with a simplified wavelet tree representation for IDs in the XGPath. Turns out we were building the most expressive version of the wavelet tree but not using any of the additional functionality.


jeizenga:
Merging despite failure because of ongoing known Travis bug

PR #1164, opened 2017-10-16
don't accept xg version 3

glennhickey:
This gives

terminate called after throwing an instance of 'xg::XGFormatError'
  what():  Unimplemented XG format version: 3

instead of 

vg: /cluster/home/hickey/genomes/vg.new/include/sdsl/select_support_mcl.hpp:349: sdsl::select_support::size_type sdsl::select_support_mcl<t_bit_pattern, t_pattern_len>::select(sdsl::select_support::size_type) const [with unsigned char t_b = 1u; unsigned char t_pat_len = 1u; sdsl::select_support::size_type = long unsigned int]: Assertion `i > 0 and i <= m_arg_cnt' failed.

when loading previous xg version

### DIFF
--- a/src/xg.cpp
+++ b/src/xg.cpp
@@ -158,7 +158,6 @@ void XG::load(istream& in) {
         ////////////////////////////////////////////////////////////////////////
         switch (file_version) {
         
-        case 3:
         case 4:
             {
                 sdsl::read_member(seq_length, in);


adamnovak:
Really we ought to have a dedicated version-too-old check to go with the version-too-new check we have already. I originally intended the code to be backwards-compatible and fix up old XG versions when they were read, but nobody's been writing that conversion code, so we're hitting the "unimplemented conversion" error rather than a more informative "too old" error.

PR #1171, opened 2017-10-16
Sim from paths

### LINKED ISSUE #1167: Implement simulation from paths
Part of #1147, and also probably important for #1166.

adamnovak:
Closes #1167 

Allows you to specify `-P path_name` and restrict vg sim to reads that follow a path.

This isn't just the subset of the graph covered by the path; it correctly accounts for cycles.

Right now the changes to the trained simulator need the most testing (because I haven't tested them).

### DIFF
--- a/src/sampler.cpp
+++ b/src/sampler.cpp
@@ -5,17 +5,14 @@
 namespace vg {
 
 pos_t Sampler::position(void) {
-    // We sample either from the entire graph sequence, or the selected path.
-    size_t max_pos = source_path.empty() ? xgidx->seq_length : xgidx->path_length(source_path);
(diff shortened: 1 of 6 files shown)


arkal:
Jenkins, test this please.

@adamnovak i'm sorry i accidentally killed your instance running this test.

adamnovak:
The trained simulator seems to be working fine.

ekg:
Ideally we could take a list of paths or a path prefix so that we could simulate from all of them without bias. What do you think?

adamnovak:
That would be nice; we'd have to change the simulator to keep a bit of
state to know which path it's working on. But if we did that we could
weight the paths by length correctly.

PR #1174, opened 2017-10-17
Add a test to verify that path restriction works in FASTQ mode

adamnovak:


adamnovak:
This also includes an important bugfix for the sampler, so https://github.com/vgteam/toil-vg/pull/365 won't actually work correctly on graphs with paths that visit nodes in reverse until this is in.

adamnovak:
Our cactus performance isn't particularly *good*, but the results of the test make sense

PR #1183, opened 2017-10-19
Switch to vg annotate -i in simulation tests

glennhickey:
Need this for the cactus yeast test to run.  Shouldn't change anything in the current tests, but we'll see here.


glennhickey:
Jenkins, test this please.

PR #1184, opened 2017-10-20
Add Yeast test

glennhickey:
Yeast test based on the cactus graphs.  Reads are simulated from the SK1 path of the full graph.  The other graphs are made from this graph using vg mod:
* cactus_drop_SK1 : remove all elements that are on no other path but SK1 path
* cactus_SK1 : keep only SK1 path
* cactus_S288c : keep only S288c (reference) path
Position comparison looks for a match in all annotated paths.  

Still working on a few issues
* bwa coordinate extraction needs to be updated to behave like annotate -i.  Doesn't work at all, currently 
* pr plot window needs to be adjusted for this test
* need to spot check a few cases to make sure graphs and extracted positions make sense.  worried about cycles (will try the mgsa graph too).

### DIFF
--- a/jenkins/jenkins.sh
+++ b/jenkins/jenkins.sh
@@ -24,7 +24,7 @@ KEEP_OUTPUT=0
 # Should we show stdout and stderr from tests? If so, set to "-s".
 SHOW_OPT=""
 # What toil-vg should we install?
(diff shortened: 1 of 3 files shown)


glennhickey:
BWA issue should be largely fixed now that the fasta file its using has full sequence names (s3://cgl-pipeline-inputs/vg_cgl/cactus_yeast/YEAST.fa)

glennhickey:
Jenkins, test this please.

PR #1185, opened 2017-10-20
clear path before outputting input alignment with map -G

glennhickey:
As far as I can tell, this was only an issue for single-ended and never for -i.


ekg:
I've got the same fix but it is in the mapper itself so it catches all the cases. Mind if I push that instead?

glennhickey:
Done properly in #1186

PR #1187, opened 2017-10-22
Cactus pair

ekg:
Getting pair rescue to work on complex graphs.

### DIFF
--- a/scripts/map-sim
+++ b/scripts/map-sim
@@ -52,8 +52,10 @@ then
     vg annotate -n -x $sim_ref_xg -a $output/sim.gam | tail -n+2 | pv -l | sort >$output/novelty.tsv
     join $output/truth.tsv $output/novelty.tsv >$output/sim.gam.truth.tsv
     # split the file into the mates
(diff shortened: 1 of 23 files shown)


ekg:
Jenkins, test this please.

ekg:
@glennhickey any idea why Jenkins is failing here? Seems like it can't find blas?

glennhickey:
Yeah, weird.  It's funny, because I was just running into the same thing
today on my own instance outside of Jenkins when trying to reproduce some
genome-wide stuff.

Somehow  (from jenkins/jenkins.sh)

# Upgrade pip so that it can use the wheels for numpy & scipy, so that they
# don't try to build from source
pip install --upgrade pip
pip install scipy

seems to have stopped working for us all of a sudden.  It looks like scipy
just got changed in pip today:

https://pypi.python.org/pypi/scipy/1.0.0

which is maybe related?  Since it seems to reproduce on nodes made from
toil launch-cluster, hopefully it won't be too long to debug.  @adamnovak
any ideas?

glennhickey:
@ekg This looks like it's working:

https://github.com/vgteam/vg/pull/1198

ekg:
Travis is just erroring out while generating the build script. What on earth...

PR #1196, opened 2017-10-25
update toil-vg in jenkins tests

glennhickey:
This will turn on quality-adjusted alignment in mpmap on the trained sim data, and proportional path size simulation in the yeast test.


glennhickey:
Looks like turning on quality-adjusted alignment in mpmap caused an AUC regression in the mhc test.   Now that that's duly noted, I'm going to merge it in anyway.

PR #1198, opened 2017-10-25
try rolling back scipy version

glennhickey:


glennhickey:
Failure comes from previous pr which turned on qual-adjusted alignment for mpmap, but I guess hasn't had time to percolate to the baseline.

PR #1210, opened 2017-10-31
Detailing for real data

ekg:
When running on real data I found a problem with the fastq parallel parser akin to what @mlin had previously resolved for stream.hpp. I've monkey-patched his fix into alignment.cpp and am testing different buffer sizes to see where the blocking resolves on 32 core systems.

There are some mapper performance/parameter tweaks in here, as well as all of the cactus-pair branch.


ekg:
Jenkins, test this please.

ekg:
I've merged despite two persistent and very marginal failures in jenkins.

PR #1222, opened 2017-11-07
Add identity field to Alignments from mpmap

jeizenga:
`vg call` likes Alignments with the identity field filled out, so I added it into `vg mpmap`.

### DIFF
--- a/src/multipath_mapper.cpp
+++ b/src/multipath_mapper.cpp
@@ -705,6 +705,15 @@ namespace vg {
                         multipath_aln_pairs_out.back().first.clear_start();
                     }
                 }
+                
+                if (!multipath_aln_pairs_out.empty()) {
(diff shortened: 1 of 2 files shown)


glennhickey:
The test failures seem completely unrelated.  Is this just vg map being
non-deterministic?

ekg:
It could be something I pushed into master.

jeizenga:
Hmm. I don't see how that would have happened if you didn't change something in vg call. How are the identities used in vg call?

glennhickey:
There's an identity filter that's on by default in toil-vg that affects
calling.  It's turned off explicitly in test_map_brca1_snp1kg_mpmap now.

But it seems that Erik modified vg map directly in master (ie without going
through a PR) in a way that makes this test fail.  If this is the case,
then the next PR (ie this one) would be guaranteed to fail this test
regardless of the PR's contents.

PR #1224, opened 2017-11-07
reduce variance in estimate for minimum mem length

ekg:
Previously we estimated based on the size of the gcsa2 index. However, this results in lower performance on pan graphs at the benefit of lower runtime. As this confuses a number of analyses that are important for the paper I'd like to test fixing the value at 19 as it is in bwa mem.


ekg:
Looks like mpmap is now timing out. I guess this has something to do with the changes to master? @jeizenga ?

PR #1229, opened 2017-11-09
Add a path-to-component index to XG

jeizenga:
Most of the changes here are aimed toward improving the speed of `mpmap` on multi-component graphs. This PR adds two small indexes to XG that keep track of which paths are on separate connected components. The algorithm to construct them is linear, and it is executed automatically when constructing the XG. The serialization and deserialization code has been updated, and there is a conversion from version 5 to 6 implemented. The APIs that this index make possible are now integrated into the clustering code in `mpmap`.


jeizenga:
Jenkins, test this please.

ekg:
Would it make sense to encode this index into succinct data structures? I guess if we are going to rework things to handle larger numbers of paths it might, but if the number is small it makes sense to keep it in an map.

PR #1231, opened 2017-11-10
Sim bells and whistles

jeizenga:
Added the option to train separate distributions for the quality strings of the two reads in a pair, with a limited amount of dependence between the two strings.


jeizenga:
The new simulation seems to have hurt performance across the board on the trained sim tests, but I think this is appropriate. The new simulation is strictly more expressive than the last one, so as long as we still have enough data to estimate the additional degrees of freedom (I think we do) this should mimic real data better.

PR #1234, opened 2017-11-10
Yes, it's really important to be correct

ekg:
Here I'm investigating why we have such poor allele balance at hets that aren't in the graph, much worse than bwa mem.


ekg:
@jeizenga do you have a commit that's not far from master where this mpmap test doesn't fail?

jeizenga:
I think the furthest along commit I have that isn't showing this failure is here (but it does have some other failure in map):
https://github.com/vgteam/vg/pull/1222

ekg:
Uh am I misreading this or did the jobs fail because they couldn't find ggrepel??? Weird!??! I'll do a local test.

ekg:
I'm closing this branch as I don't see any improvement over master.

PR #1251, opened 2017-11-15
Refactor

tanessav:
Refactored the distance_to_head and distance_to_tail methods to use HandleGraphs


adamnovak:
@tanessav It looks like changes to sdsl-lite made it into this PR after all, and I suspect they may be what's causing the test to fail.

The right sdsl-lite commit is `3ea78d025eb823ca12ac6e6c6e1287750d238b71`. I think you can do something like:

```
cd deps/sdsl-lite
git checkout 3ea78d025eb823ca12ac6e6c6e1287750d238b71
cd ../..
git add deps/sdsl-lite
git commit
```

And then push that commit up to fix the sdsl-lite submodule pointer.

adamnovak:
It looks like you've committed a lot of extra files e.g. https://github.com/vgteam/vg/pull/1251/files#diff-879746b78464d2493ddc858d1ac7673d

You also committed deps/superbubbles, which is a now-removed submodule and which apparently confuses Git by its very presence, causing the build to fail.

Can you do a `git rebase -i`, edit (and maybe squash) the commits, and use `git reset` and `git add` to just commit the files/submodules you actually want to modify? I can help you out with that today.

adamnovak:
@tanessav It looks like we missed something; your unit tests are failing. I guess we should have run those this morning...

It looks like you might be measuring from the wrong side of the node or something.

adamnovak:
The Jenkins test failure looks like one of the servers we need to talk to timed out.

Jenkins, test this please.

adamnovak:
This is the bug Jordan is fixing, still.

Jenkins, test this please.

adamnovak:
The Mac failure is some new nonsense I am trying to fix in #1281. Neither of these problems is the fault of this commit.

Issue #1242, opened 2017-11-13
Make the cool DP-based bottom-to-top tree traversing version of genotype


adamnovak:

adamnovak:
We need to work out how we want this to work.

@glennhickey, are your changes to genotype to make it work with pileups ready to be built on?

@jeizenga You seem to have the clearest idea of exactly what it is we are building. Does it go like this?

1. Make the snarl tree for the graph augmented with plausible variants.
2. Go down to the leaves and compute the most likely genotypes at ploidies 1 through n, where n is some maximum ploidy limit we set.
3. After a snarl's children have their most likely genotypes computed, compute the most likely genotypes of the parent snarl at ploidies 1 through n, assigning actual ploidy values to the child snarls.
4. Repeat up the tree until we have genotyped all the snarls.
5. Emit the results as Loci, converting from SnarlTraversals to fully specified Paths.
6. Optionally convert the Loci to VCF.

If that's what we're doing, I think we have the Loci-to-VCF part more or less present already inside the Genotyper. But the Genotyper could stand to be cleaned up a bit; there's a "run" method, a bunch of public helper functions, and an entire "Experimental alternative genotyper" all inside of it. Some of this is stuff @edawson wrote; @edawson, can you describe what you've been doing with the Genotyper and how you have been designing it?

I think what we want to do is:

- [ ] Break out all the extra modes of `Genotyper` (like re-calling) into their own classes.
- [ ] Drop the existing `run()` method.
- [ ] Replace it with something that
    - [ ] Implements this DP-based bottom-up calling framework
    - [ ] Uses `Snarl`s instead of `Site`s
    - [ ] Draws on GenotypeKit where appropriate (like for traversal finding).

If we end up having to dig into the traversal finding code, I think it might make sense to rework its internal traversal logic to use NetGraphs, rather than having multiple implementations of traversing through the inside of a Snarl.

glennhickey:
Agree with all points.  Especially the cleanup.  I didn't do enough (though
I do have the re-calling stuff pulled out on my branch), and have gotten a
bit bogged down as a result.  I will do an intermediate PR to snyc up as
soon as I have something that's not broken.

adamnovak:
We need to:

- [ ] Make `get_genotype_log_likelihood` work for more ploidies (0, 1, maybe more than 2).
[...]

Issue #1206, opened 2017-10-30
Do an overall refactor of the caller


adamnovak:

adamnovak:
We want to represent a consistent set of assertions based on the bubble structure, and be able to propose and evaluate changes to it.

adamnovak:
We also just want to rename the objects we have and decouple from the pileups as much as we can.

adamnovak:
Have a look at the way PhasedGenome works.

faithokamoto:
`vg genotype` is deprecated https://github.com/vgteam/vg/pull/5019

PR #1257, opened 2017-11-17
Use Snarls instead of Sites in the Genotyper

adamnovak:
Part of #1206 and #1242.


adamnovak:
As Glenn noted, these changes make genotyping performance drop. I shall investigate,

PR #1259, opened 2017-11-17
Make the MEM reseeding algorithm recursive

jeizenga:
I implemented a fully recursive version of the sub-MEM finding algorithm. Instead of only reseeding SMEMs, you now designate a maximum recursion depth. The old behavior can be mimicked by setting the maximum recursion depth to 1.


jeizenga:
I re-did this better in https://github.com/vgteam/vg/pull/1465

PR #1265, opened 2017-11-22
Improve bluntify operation on falcon graphs

ekg:
The overlap description provided by falcon is approximate, so we need to rest more on computational reduction of the overlaps. We do this with the SSW aligner, and in parallel (depending on the given number of threads in vg mod).


ekg:
@glennhickey how do I ask Jenkins to update its submodules? I've changed deps/ssw.

glennhickey:
Are you sure everything's checked in properly to github?  Jenkins should be
working with the equivalent of a fresh clone recursive each time.

ekg:
Well, the file diff listing here on github looks right, and I've been able to build it on other systems. So I'm confused. Considering going back and trying to build this PR without the ssw library.

PR #1266, opened 2017-11-23
Haplotype scoring

yoheirosen:
For Jordan to use for read scoring. This is the minimal implementation for that purpose. Won't be as fast as it can be for long sequence, but should be fast for < 20 nodes

To integrate it, you should just need to call two functions

RRMemo(double recombination_penalty, size_t population_size);
static double haplo_DP::score(const vg::Path& path, xg::XG& graph, RRMemo& memo);


adamnovak:
Travis on Linux hung during the unit tests for 10 minutes. It could just be Travis runnign out of potato juice, or it could be that this PR introduces a very slow unit test.

yoheirosen:
We will want to leave this open until I have the GBWT version of these functions fully integrated. Should be soon

jeizenga:
@yoheirosen 
I could use a little more documentation on a couple of these functions. Particularly the public facing API from haplo_DP. Specifically:

- What do the bool/double return value from haplo_DP::score mean semantically?
- Just looking at the constructors, it seems like there's a hierarchy haplo_DP -> hDP_graph_accessor -> XG::ThreadMapping. What are these object's relationship to each other? 
- Which of these are static data structures and which should be constructed once per call?
- What ThreadMappings should I use to construct the hDP_graph_accessor? Is the algorithm valid for any choice of ThreadMappings, or do they need to follow certain restrictions?
- How doe the ThreadMappings relate to the vg::Path that gets passed to haplo_DP::score?

As a matter of style, I'd recommend putting the most public-facing APIs and objects at the top of the .hpp. Also, although I can't say I fully understand the factoring yet, my suspicion is that some of this would benefit from some refactoring to hide the details behind a cleaner public API.

yoheirosen:
Here is my proposed integration

Shared objects
xg::XG xg_index; or
gbwt::GBWT gbwt_index; or
gbwt::DynamicGBWT d_gbwt_index;
size_t n_haplotypes = 5008;
haplo::RRMemo haplo_memo(6, 9, 5008);

...

Single function to call for the score. Does everything internally

given a vg::Path your_path;

pair<double, bool> haploscore_xg = haplo::haploDP::score(your_path,
xg_index, haplo_memo); or

pair<double, bool> haploscore_gbwt = haplo::haploDP::score(your_path,
gbwt_index, haplo_memo); or

pair<double, bool> double haploscore_d_gbwt = haplo::haploDP::score(your_path,
d_gbwt_index, haplo_memo);

double is log-probability, bool is true if all edges in the path were in
the index

I need to commit some things to be usable again. Not tested yet, can do
after my 4 pm meeting.

On Dec 1, 2017 3:53 PM, "Jordan Eizenga" <notifications@github.com> wrote:

@yoheirosen <https://github.com/yoheirosen>
I could use a little more documentation on a couple of these functions.
Particularly the public facing API from haplo_DP. Specifically:

   - What do the bool/double return value from haplo_DP::score mean
   semantically?
   - Just looking at the constructors, it seems like there's a hierarchy
   haplo_DP -> hDP_graph_accessor -> XG::ThreadMapping. What are these
   object's relationship to each other?
   - Which of these are static data structures and which should be
   constructed once per call?
   - What ThreadMappings should I use to construct the hDP_graph_accessor?
   Is the algorithm valid for any choice of ThreadMappings, or do they need to
   follow certain restrictions?
   - How doe the ThreadMappings relate to the vg::Path that gets passed to
   haplo_DP::score?

As a matter of style, I'd recommend putting the most public-facing APIs and
objects at the top of the .hpp. Also, although I can't say I fully
understand the factoring yet, my suspicion is that some of this would
benefit from some refactoring to hide the details behind a cleaner public
API.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/vgteam/vg/pull/1266#issuecomment-348645099>, or mute
the thread

jeizenga:
@adamnovak 
I'm having trouble finding the function in XG where we would query the number of haplotypes in the gPBWT. Does that exist? If not, it should.

ekg:
I'd like to integrate this into the vanilla mapper as well. Let me know if
I can help. The interface looks good to me.

adamnovak:
As we discussed on Friday, I don't think there's a function to do this,
unfortunately.

jeizenga:
I've implemented in a new XG format in this branch: https://github.com/jeizenga/vg/tree/pop-mapq

ekg:
> As we discussed on Friday, I don't think there's a function to do this, unfortunately.

As long as the path names are retained, then you can get this from that structure. It may not be directly exposed. You would need to find the number of counts of the name delimiter in the bwt of the names.

yoheirosen:
Wouldn't the best thing to do to be just to store this in an additional
field when we build the xg? Since we know how many threads we're putting in
when we build it

ekg:
> Wouldn't the best thing to do to be just to store this in an additional field when we build the xg? Since we know how many threads we're putting in when we build it

I believe this should also be available in GBWT via the same method (the walks should have start markers). Is there a reason to add a separate field if it's something you can efficiently query from the data structure you're interested in?

adamnovak:
I forgot about the name storage. Are we storing names for all the
haplotypes? If so I think you probably can count those efficiently.

jeizenga:
I also forgot the name storage, but fwiw I did implement an (untested) new XG version in a branch that includes the number of threads as a field:

https://github.com/jeizenga/vg/tree/pop-mapq/src

PR #1268, opened 2017-11-24
Divide and conquer count index generation

ekg:
Enables parallelism when generating allele count indexes. These can be generated in parallel and then merged.


adamnovak:
I like it. I'm going to go ahead and merge it.

Issue #1258, opened 2017-11-17
Make get_genotype_log_likelihood work for more ploidies (0, 1, maybe more than 2)


adamnovak:
Part of #1242

PR #1274, opened 2017-11-28
Likelihoods and priors for non-diploid genotypes

adamnovak:
This mostly completes #1258, although I want to add some more unit tests before merging. This should allow 0-ploid and 1-ploid genotypes to be assigned likelihoods and priors.

Space in the prior is reserved for polyploid genotypes, but the likelihood function still can't really handle them, because it doesn't know how to do an arbitrarily censored/ambiguous multinomial likelihood yet.

### DIFF
--- a/src/distributions.hpp
+++ b/src/distributions.hpp
@@ -168,6 +168,7 @@ real_t geometric_sampling_prob_ln(ProbIn success_logprob, size_t trials) {
     return logprob_invert(success_logprob) * (trials - 1) + success_logprob;
 }
 
+#define debug
 /**
  * Given a split of items across a certain number of categories, as ints between
  * the two given bidirectional iterators, advance to the next split and return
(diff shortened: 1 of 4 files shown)


adamnovak:
The failing test doesn't even use the genotyper. It complains:

```
vg: src/mapper.cpp:489: std::vector<vg::MaximalExactMatch> vg::BaseMapper::find_mems_deep(std::__cxx11::basic_string<char>::const_iterator, std::__cxx11::basic_string<char>::const_iterator, double&, double&, int, int, int, bool, bool, bool, bool, int): Assertion `lcp_maxima.size() == mems.size()' failed.
vg: src/mapper.cpp:489: std::vector<vg::MaximalExactMatch> vg::BaseMapper::find_mems_deep(std::__cxx11::basic_string<char>::const_iterator, std::__cxx11::basic_string<char>::const_iterator, double&, double&, int, int, int, bool, bool, bool, bool, int): Assertion `lcp_maxima.size() == mems.size()' failed.
```

@jeizenga Do you have any idea what's up here? Do we have intermittent failures in the MEM/LCP code?

glennhickey:
The intermittent failures in this test have been going on for some time.
It would definitely be nice to figure out.

adamnovak:
Jenkins, test this please

ekg:
This check should be made dependent on the flags provided to the find_mems_deep function. If the test failure is intermittent it's going to be a bit tricky to test if this fixes it, but I guess we'll find out. I'll drop a relevant change into a subsequent PR.

ekg:
> Space in the prior is reserved for polyploid genotypes, but the likelihood function still can't really handle them, because it doesn't know how to do an arbitrarily censored/ambiguous multinomial likelihood yet.

@adamnovak You can crib code from freebayes to do this. The genotyper/bayesian stuff could be pulled out relatively easily. Let me know if I can help.

adamnovak:
I'm looking through the Freebayes code and it looks like it only handles normal multinomials. We're going to have a multinomial where we have options A, B, C and observations A, B, {either A or B}, A, C, {either B or C}, and we want to be able to calculate likelihood including those censored, ambiguous observations. Allowing, ideally, for the fact that some options may be more likely to produce censored observations than others and we don't know the distribution on that.

Does Freebayes do anything like that? If so, where in the code should I look?

ekg:
I'm not sure I understand what censored means. It does seem like freebayes handles this case to an extent. There is a concept of partial observation support--- where an observation supports multiple candidate alleles.

adamnovak:
So in Freebayes, if a read could support either allele 1 or allele 2, you
add 1 to the support counts for both of the alleles and feed that set of
support counts (which sums up to more than the total number of reads) into
the multinomial likelihood function?

Censored means basically that you have an observation form a distribution
but you don't know what it is, or you only know some information about it
(like that it is in a range, or is at least a certain number). It comes up
a lot in modeling cancer survival: people might just stop reporting in, or
they might die of some unrelated cause, so while you know they didn't die
of cancer for X months, you don't know at what point after that they would
have died of cancer. Nonetheless, you can still feed that X value into your
model and get information out of it.

In this case, I want to treat ambiguous reads as censored; we know they
come from one of some set of alleles, but we don't know which of the
alleles in that set they came from. For example, a read GATAAA could have
come from an allele GATAAATT, or from GATAAAAATT, but not from GATATT. I
think censored-ness is maybe the right way to look at it.

Right now I'm working on a model that will essentially sum up over all the
read assignments that are possible, given the data you have. So it would
calculate the likelihood assuming the read came from the first allele and
the likelihood assuming the read came from the second allele and sum them
up to get the likelihood of the ambiguous description overall. This could
be very slow, but I think if I break it down by ambiguity category (and
treat all reads coming from the same set of possible source alleles the
same) it might be tractable.

adamnovak:
I've tested this on my Mac and it works, modulo all the problems I'm having cleaning up after switching to Macports Clang to make the builtin assembler flag Jordan added work.

I'm going to merge this in, despite Travis not being able to run our Mac builds currently.

jeizenga:
Yeah, sorry about that. It's updated in the readme, but I guess it will probably be a surprise to current Mac users.

adamnovak:
I couldn't get it to build without manually sourcing source-me (which I never had to do before) and prepending `AS_INTEGRATED_ASSEMBLER=1` to my make command. I also have a bunch of tests failing now because of differing behavior for e.g. `wc`; did we always require the GNU versions of those? Maybe they moved since I originally installed them or something.

adamnovak:
OK, I put `/opt/local/libexec/gnubin` on my PATH and the tests are much happier. We should maybe add that to the README.

PR #1275, opened 2017-11-28
GBWT construction

jltsiren:
Builds GBWT instead of gPBWT with `vg index -x -v -G`. Stores thread names in XG and the threads themselves in GBWT.

The intended usage is to launch a separate `vg index` process for each chromosome. This builds separate XG and GBWT indexes for all chromosomes. The GBWT indexes can be merged with the `merge_gbwt -f` tool in the GBWT package, or with the `GBWT(const std::vector<GBWT>& sources)` constructor. I'm not sure what to do with the XG indexes, though.

See the [GBWT wiki](https://github.com/jltsiren/gbwt/wiki/Construction-Benchmarks) for performance benchmarks with 1KGP data.


ekg:
We need to add a method to merge (at least concatenate) the XG indexes. It's straightforward to do, and probably we should consider using this pattern to build large graph indexes, however no one has attempted it yet.

glennhickey:
(as also discussed in #1274) I think we can disregard this test failure as
unrelated to the PR in question.

jltsiren:
Travis fails on Mac because Homebrew can't find `/usr/local/Cellar/gcc`, but otherwise the tests are passing. I fixed the issues that caused the earlier Travis tests to fail on both Linux and Mac, so I guess this should be ok to merge.

yoheirosen:
Could we keep the old xg-index construction in as an option as well? Will need it for some benchmarking

jltsiren:
The old XG/gPBWT construction is still the default with `vg index -x -v`. GBWT built instead of gPBWT if you also add `-G`.

PR #1276, opened 2017-11-28
Compressed chunked count indexes

ekg:
This is work to improve the chunked count index handling of edits. These are now broken up into 1MB bins and the CSA is computed for these independently.


adamnovak:
@ekg, any idea why Travis doesn't like this? It looks like it might have something to do with `vector<ofstream>& out` and `ofstream` maybe not being copyable?

adamnovak:
I guess it doesn't matter much, it seems to be happy-ish with master, except for more OS X nonsense with Homebrew (I think) changing something.

PR #1279, opened 2017-11-29
Pack it

ekg:
Rename `vg count` to `vg pack` and make construction partially run in parallel.

The edit storage is still slow to access.

On this branch I will implement a query interface for getting allele tallies and improve the tabular output.


ekg:
I'm getting this error on the OSX build: Error: No such keg: /usr/local/Cellar/gcc

Could someone with a Mac look into this? Is it affecting all the builds now?

jeizenga:
I think I let a non-deterministic bug slip through the CI on mpmap. I wouldn't let it keep you from merging if everything else is working.

PR #1285, opened 2017-12-01
Groundwork for DP-based genotyper

adamnovak:
This sets up `Genotyper` to pass around `SnarlTraversal`s, although it still assumes all the `Visit`s are to nodes.

I'm also renaming the `visits` field on the `SnarlTraversal` protobuf, since they should be singular. I want to get this in before I write the real refactor.

### DIFF
--- a/jenkins/jenkins.sh
+++ b/jenkins/jenkins.sh
@@ -24,7 +24,7 @@ KEEP_OUTPUT=0
 # Should we show stdout and stderr from tests? If so, set to "-s".
 SHOW_OPT=""
 # What toil-vg should we install?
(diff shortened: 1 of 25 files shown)


adamnovak:
The problem we're having now is that the genotyper, with the more complex augmented graph, is taking more memory than the Jenkins node has. I think that this is because embedding all the reads in the graph is very expensive. I'm going to try and tweak it so that it doesn't do that.

glennhickey:
From what I was seeing before, it's the augmentation process itself (ie
edit()) that is the bottleneck.

adamnovak:
I thought so too, but these tests definitely are logging failures during genotype.

adamnovak:
We're passing the tests and increasing the Genotype F1s. Time to merge!

PR #1281, opened 2017-11-30
Generalized Travis and OS X improvements

adamnovak:
This should fix Travis on OS X, and filter out the assembler warnings that everyone has been getting.

Closes #1288

PR #1290, opened 2017-12-04
Use global alignment to improve surjection and allow mapper to directly output SAM/BAM/CRAM

ekg:
Using global alignment to project the alignment onto the trimmed reference subgraph yields softclip-free surjection. Also, this fixes problems with deletions that occurred at the beginning of a node mapping. These broke surjection as reported before.


ekg:
`Error: No such keg: /usr/local/Cellar/gcc`

Anyone have some insight into this? @adamnovak this doesn't look like a failure related to the packer.

adamnovak:
I believe I fixed that in #1281, along with some fixes that make the Packer work. But I'm still fighting with that PR; now I'm trying to get it to build within the allotted Travis time limit, since I had to discard the old cache.

PR #1292, opened 2017-12-04
turn off quality adjust alignment in jenkins

glennhickey:
This is just a test, not to merge


jeizenga:
Kind of what I thought. The AUC goes up on the trained tests because we can't mismap those false reads, but everything else hurts a bit. I'm going to take this as justification that I shouldn't keep working on optimizing accuracy here.

PR #1293, opened 2017-12-04
Fixed slower move constructor

tanessav:
The default move constructor turns out to be faster than the copy constructor and the original move constructor. Since the default move constructor worked properly, we replaced the move assignment operator with the default as well. 

Benchmark results for the original move constructor:
move_original

Benchmark results for the default move constructor: 
move_default


adamnovak:
This is failing some tests, even though the default move constructor and assignment operator really should work for the vg class.

@ekg Is there a reason you know of why they shouldn't work, or why we haven't been using them all along?

adamnovak:
I'm going to close this for now; I think @tanessav has moved on to other projects in the codebase. If we ever figure out why there's such a weird dependence on the way the move constructor works, we can fix it and update this PR.

PR #1294, opened 2017-12-04
Fix annotate bug

jeizenga:
Fixes a bug that would put the forward strand offset on reverse strand positions in vg annotate.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -2489,7 +2489,7 @@ void Mapper::annotate_with_initial_path_positions(vector<Alignment>& alns) {
 }
 
 void Mapper::annotate_with_initial_path_positions(Alignment& aln) {
-    aln.clear_refpos();
+    if (aln.refpos_size()) return;
     auto init_path_positions = alignment_path_offsets(aln);
     for (const pair<string, vector<pair<size_t, bool> > >& pos_record : init_path_positions) {
         for (auto& pos : pos_record.second) {
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -2489,7 +2489,7 @@ void Mapper::annotate_with_initial_path_positions(vector<Alignment>& alns) {
 }
 
 void Mapper::annotate_with_initial_path_positions(Alignment& aln) {
-    if (aln.refpos_size()) return; // refpos already cached in the alignment
+    aln.clear_refpos();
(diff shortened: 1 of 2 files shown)


jeizenga:
Jenkins, test this please.

jeizenga:
Jenkins, test this please.

jeizenga:
Jenkins, test this please.

glennhickey:
Actually, I *can* reproduce errors locally....  looking into it now.

glennhickey:
I think changing the semantics of offsets_in_paths() without any compensatory changes where it's used in mapper.cpp may be causing vg map to

glennhickey:
For the first test, we're getting a reduction of calling accuracy from 90 to 20:

jenkins/vgci.py:398: in _verify_f1
    self.assertGreaterEqual(f1_score, baseline_f1 - threshold)
E   AssertionError: 0.2692 not greater than or equal to 0.9136</failure><system-out>2017-12-08 15:29:30,833 - toil.toilState - INFO - (Re)building internal scheduler state

where the results table is this:

<VGCI name = "vcfeval Results" tsv = "True">
Threshold	True-pos-baseline	True-pos-call	False-pos	False-neg	Precision	Sensitivity	F1	Baseline F1	Test Threshold
142.000	21	21	8	106	0.7241	0.1654	0.2692	0.9186	0.005
None	21	21	8	106	0.7241	0.1654	0.2692	N/A	N/A
</VGCI>

glennhickey:
Hypothesis:  this same problem is causing one of the other tests to run long.  And the timeout is causing Jenkins to not make a useful report. 

One thing we can do is to ensure that the individual test timeouts don't total to more than the 10 hour jenkins timeout.  This would hopefully avoid a single test failure bringing down the whole report.

jeizenga:
Update: I think I might have made a similar change in the other branch. I don't remember when/why, but the code is there. So at least that's consistent.

adamnovak:
Are we giving up on this PR?

jeizenga:
The fact that this is persisting really irritates me, but I also don't know if I'll have the time to figure out what interactions are messing it up anytime soon. I don't know. It's not that much code, really. It wouldn't be hard to rewrite it, but I'd at least want it to move over to the issues if we close it without merging.

ekg:
This is super stale, so I'm closing.

PR #1299, opened 2017-12-07
Better edit

adamnovak:
This modifies `VG::edit()` not to add paths to the graphs by default, but to only modify the graph itself. You can tell edit to still put the paths in the graph if you need them, or you can tell it to modify the paths you fed it in place, so you can get the perfect-match paths back without any modification to the graph's paths.

Also by default it will not break nodes at the boundaries of paths that start/end with perfect match mappings, unless it needs to in order to add the path to the graph.

I need this for my work on vg augment, which I want to not weirdly half-embed and then immediately re-extract the read paths.


adamnovak:
I also have my packer fixes in here, cherry-picked from #1281, because I needed them for the packer tests to pass locally.

PR #1300, opened 2017-12-07
Unstranded clustering in mpmap

jeizenga:
This adds the option in mpmap's OrientedDistanceClusterer to cluster MEMs by unstranded paths initially, leaving it to later parts of the algorithm to enforce strand consistency. It also adds support functions in the MultipathMapper that handle strand consistency.


jeizenga:
@glennhickey 
We can merge this PR without it, but eventually I want to add a test for this into at least some of the Jenkins tests. Basically, one run of mpmap with the -n flag and one without it. Is that going to be a pain in the ass?

glennhickey:
Shouldn't be a problem.  We have 5 tests with mpmap getting run now.   Can
we just turn on -n in one (or more) of these tests?

test_sim_mhc_cactus
test_sim_chr21_snp1kg
test_sim_chr21_snp1kg_trained
test_sim_mhc_snp1kg_mpmap
test_map_brca1_snp1kg_mpmap

jeizenga:
I think initially we could do that on the MHC test, since this is supposed to make mpmap not suck on more complicated graphs.  For the other ones I'm most interested in the -n/no -n comparison, so we'd have to do two mpmap runs.

glennhickey:
I see.  That'll take some minor hacking in toil-vg (
https://github.com/vgteam/toil-vg/issues/406) but I should be able to do it
today or tomorrow.  The only question is if there are enough colours in the
universe for the resulting plots?

jeizenga:
How nasty would it be to do some separate intra-mpmap tests, since that's really what I care about?

glennhickey:
Sure.  Maybe we can convert one of the existing sim tests to be this way.

PR #1301, opened 2017-12-08
tighten runtime bounds of jenkins tests

glennhickey:
As discussed in #1294, tripping the global timeout causes there to be to report at all, so try to catch timeouts at the test level instead.


glennhickey:
As discussed in #1302, tightening the test timeout bounds isn't bulletproof.  May need a better replacement for timeout_decorator.  Still, better bounds are better than worse bounds for what we have, so merging anyway.

PR #1302, opened 2017-12-08
Test timeouts -- dont merge

glennhickey:
This is just #1301 stacked on #1294, to see if we can get some sensible errors out of Jenkins.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -2489,7 +2489,7 @@ void Mapper::annotate_with_initial_path_positions(vector<Alignment>& alns) {
 }
 
 void Mapper::annotate_with_initial_path_positions(Alignment& aln) {
-    aln.clear_refpos();
+    if (aln.refpos_size()) return;
     auto init_path_positions = alignment_path_offsets(aln);
     for (const pair<string, vector<pair<size_t, bool> > >& pos_record : init_path_positions) {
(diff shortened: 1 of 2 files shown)


glennhickey:
Looks like the timeout_decorator doesn't have enough muscle to stop the yeast test.  When running locally, I eventually get a timeout error 

     >           raise exception()
     E           TimeoutError: u'Timed Out'

but only after 46828 seconds: well above the 1800 second limit.  Maybe a docker process goes long and can't be stopped by the timeout_decorator?  It's hard to see from the logs.  

A global timeout due to the yeast test seemingly caused by the change to XG::offsets_in_paths(pos_t pos) in #1294 is why we're getting no useful logs in that PR. 

@jeizenga it looks like your other PRs that are producing not logs are failing due to timeouts as well.  If you see something like 

     jenkins/vgci.py::VGCITest::test_sim_yeast_cactus <- .env/local/lib/python2.7/site-packages/timeout_decorator/timeout_decorator.py Build timed out (after 600 minutes). Marking the build as failed.
     Build was aborted
     Recording test results
     ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?

In the Console Output on the Jenkins website, it's probably safe to say that the test in question went overlong due to a code change causing a catastrophic slowdown in vg.

PR #1312, opened 2017-12-11
option to run multipe mpmaps each test

glennhickey:
@jeizenga I applied this to test_sim_mhc_snp1kg_mpmap what now runs mpmap with '-n' in addition to its default options (results come out as -mp2).  You can change as you see fit or apply to a different test.  You can run as many different mpmaps as you want by specifying multiple strings in the more_mpmap_opts list.


glennhickey:
@jeizenga merging this now before it gets too stale.  You can look at test_sim_mhc_snp1kg_mpmap for an example of how to run multiple mpmaps in the same test with different options.

PR #1317, opened 2017-12-12
Improve mpmap's performance on whole genome mapping

jeizenga:
Some changes to the algorithm that help resolve problems I was seeing on large, multicomponent graphs with high repeat content. Main changes are more aggressive rescue logic and more precise MAPQ estimation, which together are less likely to overestimate MAPQs.


jeizenga:
I'm pretty sure the only problem here is that there wasn't a baseline yet. Going ahead with merge.

PR #1319, opened 2017-12-12
Speed up count operations in sub MEMs algorithm

jeizenga:
Implements an option to "thin" count operations to reduce the total number of calls to GCSA::count at the expense of more GCSA::LF operations, which are cheaper. The tradeoff is pretty favorable. I might be able to push it even further, but this already brings count to less than 2% of mpmap's run time. If everything is working correctly, the algorithmic guarantees should be unaffected.


jeizenga:
Jenkins, test this please.

jeizenga:
Jenkins, test this please.

jeizenga:
For some reason the run time in master has shot up a whole hell of a lot, but this PR does speed it up relative to that.

PR #1321, opened 2017-12-13
Fix rescue logic in mpmap

jeizenga:
I think this is responsible for the speed regression in master. It also might be responsible for the improved performance on the untrained simulator, we'll see I guess.


jeizenga:
Looks like this is about a 20% speed hit, but it was indeed the source of the improved performance on the untrained simulation.

PR #1329, opened 2017-12-18
Add chrom21 calling test

glennhickey:
This runs in about 15 minutes locally, not including downloading.  It's only running vg call and freebayes though, with genotype activation commented out for now. Will add it in shortly (at least as a one-off for a baseline as it's slow), along with plots that take into account the GIAB bedfiles (currently doing whole chromosome)


glennhickey:
I'm going to merge this new test in, even though at 2 hours it's pretty slow.  Once the baseline is uploaded I'll deactivate it until it gets sped up, though it should be used locally/in prs to evaluate calling changes in the meantime...

PR #1336, opened 2017-12-20
Add xz as a dependency on Homebrew

### LINKED ISSUE #1333: Missing dependency in homebrew setup
homebrew package "xz" should be in the list to install, in order to provide lzma.h

adamnovak:
Fixes #1333


adamnovak:
Glenn decided to ignore that test. I'm going to merge this in.

PR #1342, opened 2017-12-21
use Simon's suggestion to build low-memory CSA

ekg:
For edit-including vg pack.


adamnovak:
This is failing because we changed the underlying test data for that test in a way that reduced the final F1 score. There's a fix in master, and the baseline should update soon.

But maybe it will work if we mash rerun? Jenkins, test this please.

PR #1344, opened 2017-12-25
Dramatically improve kmer generation algorithm (+ pangenome oriented improvements)

ekg:
I am working on building pangenomes from fully assembled genomes with collections of unplaced contigs. I will do this using vg as a workaround. Existing approaches ([progressiveCactus](https://github.com/glennhickey/progressiveCactus), [REVEAL](https://github.com/jasperlinthorst/reveal), [TwoPacCo](https://github.com/medvedevgroup/TwoPaCo)) have specific problems for my use that may be resolvable, but are beyond the scope of what I can immediately work on, specifically:

- progressiveCactus produces an output that cannot be directly read into vg, and the conversion utility is pegged to an old version of vg
- REVEAL produces DAG/MSA outputs with a specific trick used to detect inversions. Genomes I have been working with tend to have CNVs, and these won't be integrated directly.
- TwoPaCo proposes a compacted deBruijn graph, but it is not clear if this is appropriate. Also, the graph encoding does not seem to match vg's.

This branch will aim to improve the runtime of msga on large genomes. Rather than following the "full" approach as in progressiveCactus, where we might align all chromosomes against all others, we can reduce the complexity and improve parallelism by aligning together only those sequences that are likely to be the same chromosome.

Unplaced contigs may be clustered (somehow, I'm not sure yet) and aligned together in a similar way as applicable. Maybe using minhashing would be the way to go.


ekg:
This change does seem to help on my small test graphs, which definitely run faster through msga, but it does not seem to help when I'm working on larger chromosomes. There is something that scales in a decidedly non-linear way in the "chunked"/"banded" alignment pipeline. It could be alignment simplification, or perhaps the alignment merging step. I'll investigate.

ekg:
The problem is almost certainly here, where we seem to be filling out half of the full matrix of possible transitions between every pair of alignment candidates!

```c++
    for (vector<AlignmentChainModelVertex>::iterator v = model.begin(); v != model.end(); ++v) {
        for (auto u = v+1; u != model.end(); ++u) {
            if (v->next_cost.size() < max_connections && u->prev_cost.size() < max_connections) {
                if (v->band_idx + vertex_band_width <= u->band_idx) {
                    double weight = transition_weight(*v->aln, *u->aln, v->positions, u->positions);
                    if (weight > -std::numeric_limits<double>::max()) {
                        v->next_cost.push_back(make_pair(&*u, weight));
                        u->prev_cost.push_back(make_pair(&*v, weight));
                    }
                } else if (u->band_idx + vertex_band_width <= v->band_idx) {
                    double weight = transition_weight(*u->aln, *v->aln, u->positions, v->positions);
                    if (weight > -std::numeric_limits<double>::max()) {
                        u->next_cost.push_back(make_pair(&*v, weight));
                        v->prev_cost.push_back(make_pair(&*u, weight));
                    }
                }
            }
        }
    }
```

PR #1349, opened 2018-01-05
Scale MAPQ in mpmap

jeizenga:
Some of the sensitivity gap between mpmap and map seems to be that map has some additional features in its MAPQ formula that produce lower raw MAPQs on the whole. This is an experiment with arbitrarily scaling raw MAPQs in mpmap.


jeizenga:
May revert this later, but I want to get this in the main tree so I can make a docker and try out the whole genome pipeline.

PR #1363, opened 2018-01-10
Dont hardcode fragment in thread extraction when simulating haplotypes

glennhickey:
Should really be using the equivalent toil-vg code https://github.com/vgteam/toil-vg/blob/master/src/toil_vg/vg_construct.py#L688-L731 which was fixed a while back.


glennhickey:
This Jenkins failure seems to be due to wobble in the mapping, which caused the baseline f1 to randomly get bumped up for this test on the last merge-to-master.

PR #1364, opened 2018-01-10
Improve GBWT use in vg map

adamnovak:
Now we can get the total haplotype count from the XG, which is more or less the number Yohei's code wants.

I also added a flag to reads so you can tell if their scores are corrected.

I am also now allowing score correction even when some mappings have no paths, because it turns out that happens a lot.

With these changes I am able to correct (and see that I can correct) about 93% of read pairs and 96% of single reads, in my testing on snp1kg MHC.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -1379,13 +1379,16 @@ void BaseMapper::apply_haplotype_consistency_scores(const vector<Alignment*>& al
         // On a first pass through, compute all the scores and make sure they all can be computed
         
         if (aln->path().mapping_size() == 0) {
-            // Alignments with no actual mappings don't need scoring.
-            // But I wouldn't call them successfully scored.
-            // So treat it as a scoring failure.
-            if (debug) {
-                cerr << "Not applying haplotype consistency due to empty path" << endl;
-            }
-            return;
+            // Alignments with no actual mappings don't need scoring. But we
+            // don't want to treat them as scoring failures, because we expect
+            // some due to e.g. read pair mapping locations where one read maps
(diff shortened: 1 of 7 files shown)


adamnovak:
@jeizenga This includes my XG changes to allow it to hold a haplotype count; right now it just assumes all the samples are diploid and counts 2 * the number of samples in the input VCF. You also need to build your GBWT and XG in the same vg index run.

glennhickey:
Jenkins, test this please.

adamnovak:
It looks like the last merge to master got an anomalously high F1 of 0.9302 and now it won't accept the 0.9225 that we used to get.

I can't replicate the higher F1 when running the master branch myself. I'm going to merge this in and delete the existing baseline for that test so it makes a new one:

```
aws s3 rm s3://cgl-pipeline-inputs/vg_cgl/vg_ci/jenkins_regression_baseline/outstore-BRCA2-primary/NA12878_vcfeval_output_f1.txt
```

adamnovak:
Actually, I am just going to push the baseline back:

```
echo 0.9225 > baseline.txt

aws s3 cp --acl public-read ./baseline.txt s3://cgl-pipeline-inputs/vg_cgl/vg_ci/jenkins_regression_baseline/outstore-BRCA2-primary/NA12878_vcfeval_output_f1.txt 
```

adamnovak:
Hm. Looks like it got the higher value this time. Did you do something @glennhickey ?

glennhickey:
Not sure what's going on.  As far as I can tell, it's randomly bouncing
back and forth.  I'll assume due to minor differences in the GAM output.
Weird that it just started happening.

PR #1374, opened 2018-01-12
Use fragment length distribution parameters from chrom21.bam for chrom21 sim tests

glennhickey:
This doesn't take away the need for a more realistic distribution in vg sim, but it should at least help the existing one be more realistic.   Values are from `samtools stats` on the bwa-mem output for chrom 21 in HG002


glennhickey:
Jenkins, test this please.

PR #1383, opened 2018-01-19
option to toggle traversal finders in genotype

glennhickey:
Switch between existing traversal finders. Still a few things to do before merging
* fix affinity functions, also figure out why indel realignment doesn't seem to be helping
* determine (at least as option) traversal finder from size of snarl


glennhickey:
Jenkins, test this please.

PR #1036, opened 2017-09-13
Add an interface to convert bed to gam

6br:
In order to annotate a path with genes or repeats as a subpath, I propose to add an interface to convert BED format to GAM format.

It intends to work with commands below.

```
vg annotate -p -x input.xg -b input.bed > output.gam
vg mod -P --include-aln output.gam input.vg > output.vg
```

PR #1386, opened 2018-01-20
Fix an interface to convert bed to gam

6br:
I found some unexpected behaviors on bed2gam(https://github.com/vgteam/vg/pull/1036).

1. Although `target_alignment` requires BED-styled coordinates as arguments, `parse_bed_regions` calls this function with 0-based inclusive coordinates.
2. When the first segment should be trimmed, sometimes `path_from_length(aln.path()) - trim_end` is overflowed since trim_start is subtracted.  (I checked on an additional test.)
3. The path doesn't exist, and we'll get a segfault if we go look for positions in it.

I fixed these problems.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -1241,9 +1241,16 @@ void parse_bed_regions(istream& bedstream,
         } else {
             ss >> name;
             assert(sbuf < ebuf);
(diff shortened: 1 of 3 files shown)


glennhickey:
Jenkins, ok to test.

ekg:
Thank you!

PR #1389, opened 2018-01-23
allow incomplete paths when extracting path edges

ekg:
We don't need to freak out when we can't find all the edges in a given path. We might be applying `vg mod -N` to a subgraph of a larger graph and should expect not to always have the full paths embedded in our graph.


ekg:
Looks like the OSX build will never start. Merging as everything else is fine.

PR #1390, opened 2018-01-24
Use a smaller cluster subgraph expansion

ekg:
This should speed up vg map by about 50%. I'll only merge if tests work and I can add a test that demonstrates we can still detect indels under this condition.

Another thing to consider is changing the node length of the graph (via `vg mod -X`), as this might also interact with this parameter.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -3111,7 +3111,7 @@ Alignment Mapper::align_cluster(const Alignment& aln, const vector<MaximalExactM
         }
     }
     // get the graph with cluster.hpp's cluster_subgraph
-    Graph graph = cluster_subgraph(*xindex, aln, mems);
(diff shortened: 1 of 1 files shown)


ekg:
I'm seeing some minor indications of problems with this, so I don't think it's worth merging.

Here's the test:

```
vg msga -f ~/vg/test/GRCh38_alts/FASTA/HLA/DRB1-3123.fa -D >DRB1-3123.vg
```

Running this with and without the restriction on cluster size results in a 40bp-larger graph in the case that we restrict the expansion factor to 1 (it defaults to ~1.6). This is a small change (the graph is 20kb) so maybe there's nothing to worry about, but I'm leaning towards not merging this. I suspect there will be a few isolated problems due to this change. However, they're going to be at places where there are variants. I'm not sure it's worth the risk. Better performance gains can be had by integrating libgaba.

PR #1392, opened 2018-01-24
Enable thread extraction from the GBWT

adamnovak:


ekg:
@adamnovak can we query/extract a given thread by name using this interface?

adamnovak:
Not with what I have hooked up here. I think @jltsiren's intention is that we keep using the same indexes we were using to do that with the gPBWT, but we just go get the actual haplotype data from the new haplotype index.

PR #1397, opened 2018-01-26
Change backtraces

adamnovak:
This will use backward-cpp and libdw to provide backtraces with debug info even with static builds. It even does source listings if available.

This makes backtraces even longer and less user-facing (if source is available).

There's also a "vg crash" for on-demand crashing.


adamnovak:
Jenkins test this please.

PR #1404, opened 2018-01-26
Use mapping qualities in genotype

glennhickey:
The remaining gap here

http://cgl-pipeline-inputs.s3.amazonaws.com/vg_cgl/vg_ci/jenkins_reports/commit/a5431275fc75b30319a06ff0f7c39e13b7caeabc/test_call_chr21_snp1kg-roc-weighted.svg

doesn't seem to be related to snarl traversing or augmentation, but rather qualities assigned to small indels.  Using the mapping qualities (which I'm turning on by default) should help.

### DIFF
--- a/src/genotyper.cpp
+++ b/src/genotyper.cpp
@@ -1529,33 +1529,31 @@ double Genotyper::get_genotype_log_likelihood(VG& graph, const Snarl* snarl, con
         // pull it out of the sum.
         double log_atom_weight = prob_to_logprob(0.5) * total_reads;
 
-        // We calculate the probability of each atom, then sum, and then weight.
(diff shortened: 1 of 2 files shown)


ekg:
How are these being integrated? I would suggest either as a cap on the base quality or using the "effective base depth" metric, which is basically `1-(1-MQ * 1-BQ)`.

glennhickey:
It's doing the 1 - (1-MQ * 1-BQ)

https://github.com/vgteam/vg/blob/master/src/genotyper.cpp#L1417-L1418

I sometimes worry about the numerical stability of all the one minuses
involved, but it doesn't seem to have come up in practice yet.

Issue #1337, opened 2017-12-20
Make a command that converts a .xg into a .vg


jeizenga:
Maybe make it live in `vg xg`?

ekg:
I am also interested in this. It should be noted that you can work around this using the `-r` node range flag and the context expansion option `-c`. You just need to specify a range that's bigger than that of the graph.

```shell
vg construct -r 1mb1kgp/z.fa -m 32 -p >z.vg
vg index -x z.xg z.vg

vg find -c 1 -x z.xg -r 1:10000000 | vg stats -lz -
#nodes   31250
#edges   31249
#length  1000000
<z.vg vg stats -lz -
#nodes   31250
#edges   31249
#length  1000000
```

Hacks. :fork_and_knife:

PR #1407, opened 2018-01-29
XG to VG Converter

tanessav:
Fixes issue #1337. The converter function takes in a pointer to a HandleGraph and converts it to a VG graph. It doesn't convert the paths from XG to VG because the HandleGraph interface doesn't expose them. The function is called `converter()` and should maybe be renamed in the future. 
Usage: `vg xg -i x.xg -X y.vg`


adamnovak:
This passed everything except for Mac Travis, where it failed with:

```
t/26_deconstruct.t ... 
1..2
ok 1 - vg deconstruct produces the expected number of superbubbles in a simple graph.
ok 2 - vg deconstruct produces the expected superbubble format and the right bubbles on a small graph.
ok
Converted 100 alignments to embedded paths
Statistics:
Number of Non-Degenerate Snarls: 4
Snarls traversed by reads: 4
Snarls on reference: 0
Snarl length distribution: 
Converted 100 alignments to embedded paths
Statistics:
Number of Non-Degenerate Snarls: 4
Snarls traversed by reads: 4
Snarls on reference: 4
Snarl length distribution: 
2	1
1	3
error:[vg map] A sequence or read file is required when mapping.
The command "if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then git submodule update --recursive && timeout 1500 make deps -j4 && make -j4 && echo Testing && travis_wait 40 make test; fi" exited with 137.
```

This is nowhere near anything this PR touches, from what I can tell, and it's not clear how vg map (I guess being called from `27_vg_genotype.t `) is getting into trouble, or how we're getting that exit code from that error.

I guess I'm just going to mash rerun on Travis?

glennhickey:
Not sure why that popped up but this seems to have fixed it

https://github.com/vgteam/vg/commit/c2315cb984e01d32c4051986fdd0f73dfeb70a4e#diff-52ae0e12f83a6430f942b056157e6712

so rebasing on master should fix.

PR #1420, opened 2018-02-02
Haplotype-aware indexing

jltsiren:
This PR includes the `vg prune` subcommand and other changes required for building haplotype-aware (unfolded) GCSA2 indexes. It also updates GCSA2 to build 256-mer indexes with 4 prefix-doubling steps.

## Pruning

By default, `vg prune` is equivalent to `vg mod -p -l 16 -e 4` followed by `vg mod -S -l 33`, with the addition that all embedded paths are removed. The numerical parameters can be changed with `-k`, `-e`, and `-s`, respectively.

There are several ways to add paths to the pruned graph:

* `--restore-paths -x xg_index` restores the nodes and the edges on XG paths.
* `--preserve-paths` only deletes alt paths from the graph. The pruning does not delete edges on the remaining paths. As a result, small subgraphs bordered by complex regions remain in the graph as well.
* `--unfold-paths -x xg_index -g gbwt_index` unfolds the pruned regions into simplified subgraphs that contain all distinct paths supported by XG paths or GBWT threads (see the next section).

The various options affect the 32-mer contents of the GCSA2 index for 1000GP chromosomes 20, 21, and 22 in the following way:

* There are 658,564,666 kmers common to all indexes.
* `--restore-paths` adds 2,217,694 kmers (0.34%).
* `--preserve-paths` adds 3,512,617 kmers (0.53%) over `--restore-paths`. It also loses 1,948 kmers (this is probably a bug somewhere).
* `--unfold-paths` adds 7,310,544 kmers (1.1%) over `--restore-paths`.
* `--unfold-paths` has 5,550,964 kmers (0.84%) and `--preserve-paths` 1,751,089 kmers (0.27%) not in the other index.

## Unfolding

Unfolding considers all paths in the XG index and all threads in the GBWT index and adds them back to the pruned graph.

The algorithm is the following:

* Build a complement graph containing the edges that exist on XG paths or in the GBWT index but not in the pruned graph.
* For each connected component in the complement graph:
  * Determine the border nodes that exist both in the complement graph and in the pruned graph.
  * Generate all border-to-border paths supported by XG paths or GBWT threads.
  * Build an unfolded component that contains exactly the paths generated in the previous step, merging shared prefixes and suffixes when possible.
* Add the unfolded components into the pruned graph.

## Indexing

The unfolded graph contains nodes that are duplicates of the original nodes. In order to build a GCSA2 index that maps to the original graph, we need a node mapping.

* Adding `-m mapping_file` to `vg ids -j` creates an empty node mapping. The empty mapping tells where the ids for duplicate nodes should start from.
* Adding `-m mapping_file -a` to `vg prune` updates the node mapping with the duplicate nodes in the current graph.
* Adding `-f mapping_file` to `vg index -g` uses the node mapping when building the GCSA2 index.

## Pruning parameters

The default parameters work well most of the time. More aggressive pruning (e.g. with `-e 3`) should be used for building a 256-mer index with `--unfold-paths`. Otherwise duplicated nodes cause the number of kmers to grow uncontrollably in the fourth doubling step.

## Tests

* `phase_unfolder.cpp` has unit tests.
* There are no tests for `vg prune` and `vg index -G`.
* With parameter `-v`, `vg prune` checks that the pruned graph contains all the paths it is supposed to contain. This can be very expensive with `--unfold-paths` when there are many samples.
* The unfolded index is a valid index of the input kmers, but I have not tested whether it is a valid index of the original graph.


ekg:
Looks like the OSX build timed out. I've poked it in hopes it'll make it through on a second try.

ekg:
Is it just me or is `make test` taking a really long time as of late? It seems to be rebuilding vg twice?

jltsiren:
I'm going to test this with #1421 before merging.

ekg:
@jltsiren I think that's a good idea. For clarity, the new pruning algorithm implements what I believed I had implemented in the previous version. However, their outputs are not the same. The new one tends to pick slightly more edges to remove than the old one. The relative simplicity of the new version and the fact that an algorithm written with the same structure produces the correct kmer reduction leads me to believe I made a mistake in the old version. It would be good to know that your results hold with the new pruning method.

jltsiren:
The new pruning algorithm is way more aggressive than the old one. I had to optimize the unfolding code to handle it. This introduced some new bugs I had to hunt down. I also implemented a faster and more space-efficient `VG::prune_short_subgraphs()` that doesn't traverse or duplicate large components.

I'm merging this now. Next I'm going to try building unfolded whole-genome indexes.

PR #1421, opened 2018-02-06
Fast prune

ekg:
This speeds up the prune operation using the same kmer enumeration algorithm as in my previous patch to kmer generation. I remove the VG object "kpaths" functions and as a result slightly reduce the functionality of vg paths by removing features that weren't used.


ekg:
I'm getting "Error: homebrew/science was deprecated. This tap is now empty as all its formulae were migrated."

Any ideas OSX-users?

jltsiren:
I think you can just remove the homebrew/science line. Samtools is now in homebrew/core.

ekg:
This seems to have suffered a timeout again but otherwise the tests are OK. Moving to merge.

Issue #1424, opened 2018-02-07
Malformed BAMs from VG Surject


cmarkello:
@ekg @adamnovak @glennhickey 
So I've been attempting to use `vg surject` to produce BAM files on a set of samples for the purposes of doing joint genotype calling on a cohort and have found that it produces malformed BAM files.

When running a sample BAM file through Picard tools' `ValidateSamFile` it produces the following errors:
(out of a total of ~1182156685 records)
```
Error Type      Count
ERROR:INVALID_ALIGNMENT_START   781
ERROR:INVALID_MAPPING_QUALITY   365824
ERROR:INVALID_MATE_REF_INDEX    1182155904
ERROR:MISSING_READ_GROUP        1
WARNING:MISSING_TAG_NM  1181765146
WARNING:RECORD_MISSING_READ_GROUP       1182156685
```

The BAM file is produced by merging chromosomal BAM files that were each the product of vg surject on their respective chromosomal GAM file.

Descriptions for the various errors output from `ValidateSamFile` can be found here: https://gatkforums.broadinstitute.org/gatk/discussion/7571/errors-in-sam-bam-files-can-be-diagnosed-with-validatesamfile

Of note, the `ERROR:INVALID_MATE_REF_INDEX` errors seem to indicate unpaired reads. It looks like this was noticed in a previous issue? https://github.com/vgteam/vg/issues/369
What's the status on that issue?

adamnovak:
For the read group, it appears that GAM can hold read groups, and the `--read-group` option to `vg map` ought to make them get set in the emitted mappings.

As for the other issues, I shall investigate.

adamnovak:
I just closed #369 because in general paired reads come through paired and that read was for the initial implementation. If some particular reads are coming unpaired, that's a new bug.

cmarkello:
@adamnovak @glennhickey  BAMs are much improved, I'm no longer getting `ERROR:INVALID_MATE_REF_INDEX`, `ERROR:MATE_NOT_FOUND`, and a majority of `ERROR:INVALID_MAPPING_QUALITY`. However, I'm still getting some residual invalid mapping qualities on the surjected BAMs.

samtools `ValidateSamFile` still returns a number of `ERROR:INVALID_MAPPING_QUALITY` type errors on these BAMs.

They're formatted like the following:
`$QNAME   4   *   0   60  151M    *   0   0 $SEQ $QUAL`

[...]

PR #1427, opened 2018-02-08
Fix indels and short softclips in surjection

ekg:
This resolves two major problems with surject. The first is that we mishandled deletions, resulting in errors in the surjection cigar where they were omitted. The second is that surjection as implemented generates a lot of short soft clips, as if the alignment ends in non-reference sides of bubbles it will be aligned against a reference sequence that is shorter than it. To resolve the problem we can force the softclips to be "matches" in the cigar sense and adjust the position of the alignment as needed. The reference bounds are checked to make sure we don't fall outside them.

PR #1432, opened 2018-02-09
Improve bam

### LINKED ISSUE #369: vg surject loses pairing information
Would be super handy to preserve this (and our eventual mapping qualities), to be able to see how other callers perform on vg alignments.

adamnovak:
This should fix #369 and will make some headway on #1424.

Now we crossreference paired reads in surjected GAM to BAM when `-i` is passed, and we leave their /1 and /2 on. We don't flag them paired  or rename them when `-i` isn't passed.

I also moved where the URL encoding is happening to just SAM that goes into URLs, and moved the old surject `-i`, for a file of paths, to `-F`.

Also, I stopped using uninitialized memory for read positions for unmapped reads.


adamnovak:
We probably still want to look at how the direct output to SAM?BAM works. I made separate paired/unpaired Alignment conversion functions, and the existing code might call the paired functions (which match the old signatures) always.

ekg:
I would *really* love to use SeqLib to manage the htslib format IO.

ekg:
I moved to start merging through this conflict, but I'm worried that this doesn't also fix the problems in map_main, which is also capable of producing cram/bam/sam output. Could we talk about what needs to change, or if it's easier would you please match master and fix the mapper output in the same way as well?

adamnovak:
I've just fixed up map to care about whether the reads are paired or unpaired when surjecting. Now I'm going to rebase on master and fix up the conflicts.

adamnovak:
OK, master is now merged into this branch, and I fixed whai I was complaining about at https://github.com/vgteam/vg/pull/1427#issuecomment-364539524

PR #1447, opened 2018-02-13
Enable mpmap to use the GBWT for haplotype scoring

adamnovak:
I had to add dispatch on the GBWT as opposed to the XG for scoring, but mostly this activates Jordan's existing code.


ejacox:
Jenkins, test this please.

PR #1453, opened 2018-02-14
Disable repeat rescue in mpmap

jeizenga:
It doesn't seem to be helping much and I've seen multiple cases where it messes up MAPQ estimation.

### DIFF
--- a/src/multipath_mapper.cpp
+++ b/src/multipath_mapper.cpp
@@ -1150,6 +1150,10 @@ namespace vg {
         
         bool rescued_order_length_runs_1 = false, rescued_order_length_runs_2 = false;
         
+#ifdef debug_multipath_mapper_mapping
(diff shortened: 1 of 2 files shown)


jeizenga:
The mac Travis is timing out on the integration tests and the Jenkins doesn't seem to be communicating well with GitHub, but I don't really care. Everything is working correctly.

ekg:
Interesting. We should discuss this as it's probably relevant to map as
well.

jeizenga:
Sure. Quick summary:

We're slower on real reads than on simulated reads. I think I've figured out that the reason is because real reads are more likely to map to repeats because repetitive regions are preferentially missing from the reference assembly, so when we simulate from the assembly we're depleting for repetitive content. I tried to implement a feature where `mpmap` would try to figure out when one read pair had unique mappings and the other was very repetitive so that I could avoid doing the alignments on the repetitive end and just go straight to the rescue code. The results were not great.

PR #1456, opened 2018-02-16
mapper improvement

ekg:
This is mostly focused on improving mapping on assembly graphs. I will also use the automated testing to check some basic changes to the mapper, possibly incorporating some of the MEM finding functionality used in mpmap.


ekg:
Jenkins, test this please.

ekg:
I'm getting a runaway memory allocation in:

```
#12 haplo::haploMath::RRMemo::RRMemo (this=0x7fffffffc290, recombination_penalty=20.699999999999999, population_size=140726686235344) at src/haplotypes.cpp:451
#13 0x0000000000cada73 in vg::BaseMapper::apply_haplotype_consistency_scores (this=this@entry=0x290e000, alns=std::vector of length 1, capacity 1 = {...}) at src/mapper.
cpp:1521
```

It's happening via these steps:

```
vg construct -r small/x.fa -v small/x.vcf.gz >x.vg
vg index -x x.xg -g x.gcsa -k 11 x.vg
vg map -s ATGGCTAATATGCGGCAAAGTGAGGACTTGAACCCAGATTTGTATGGCTCCAAATCTTTGTATTCTTTCCATTTCCCCACATTACCCCTAGTGATAGATT -d x -j
```

I think it can be reproduced at commit 613eb0d4.

ekg:
The failure I mentioned above is apparently because there was an "x.gbwt" hanging around from another index generation. Removing that resolves the problem.

ekg:
66d2c17 is a example of a configuration that works perfectly on small graphs that we're testing in Jenkins but completely falls when we run on the whole human genome.

Here is the map-sim output for it on HG002/1kgp:

human-1m-results-66d2c17 1 roc

The paired end results here are *really* bad. But Jenkins is happy. In fact we seem to be improving.

In 66d2c17 I attempted to prune the paired end alignment search space when it didn't seem like things were getting better as we work through the cluster list. This made things much faster on yeast, which I tested on my laptop, and as we see Jenkins was all smiles about the result.

ekg:
Jenkins, test this please.

ekg:
Jenkins, test this please.

ekg:
Merging through the OSX build failures because they are timeout related.

Also this merges in Jordan's improved multithreading code. This seems OK but there seems to be a related problem in some alignment workloads (short single-ended reads, many cores) and I will perhaps need to re-do or un-do that.

adamnovak:
We seem to have acquired some non-timeout-related build failures in master now, and they are affecting my PR #1484 . Any idea what's wrong with:

```
t/16_vg_msga.t ....... 
1..13
not ok 1 - graph for GRCh38 HLA-V is unaffected by the number of alignment threads
#          got: 'd41d8cd98f00b204e9800998ecf8427e'
#     expected: '16e56f0090b310d2b1479d49cf790324'
```

adamnovak:
In general, if the OS X builds time out, you can mash rerun until the cache fills enough that they finish.

PR #1463, opened 2018-02-21
GFA2 + SVs

edawson:
I've updated both the GFA parser to read write GFA1 and GFA2 without any fiddling. I've hooked that in to vg's output, but looks like I still need to do the input side.

This also brings in changes to vcflib that properly "canonicalize" SVs using the SVTYPE/SVLEN/END for deletions and insertions this isn't a huge hack. For inversions this is a huge and terrible hack.


ekg:
We can encode inversions in the graph faithfully. Any reason not to link that into construction?

edawson:
Mostly just an uncertainty on my part about the merits of really messing with the construct code. This way keeps it consistent with the way SNPs are handled. But yes, it would be good to embed inversions properly.

Aside: it might be wise to test if embedding them faithfully works well downstream using [svaha](https://github.com/edawson/svaha) before committing to modifying construct a bunch.

edawson:
Welcome to the future. Merging.

PR #1464, opened 2018-02-21
Better integer hash function and vg prune parameters

### LINKED ISSUE #1457: Loading VG graphs is sometimes very slow
It takes a lot of time to load VG graphs from disk. Loading a single human chromosome takes several minutes, meaning that the reading speed is 1 MB/s or less. Recently I have encountered even worse graphs that take tens of minutes to load. This annoyed me enough that I started investigating.

Loading a VG graph essentially consists of three kinds of steps (assuming that there are no paths):

1. Read the next `Graph` chunk from disk.
2. Add the nodes and the edges from the chunk to the `Graph` object in the VG graph.
3. Insert the loaded nodes and edges into various hash maps.

The first two steps are fast enough, so the issue is with the hash maps. By loading a problematic graph with various hash maps disabled, I narrowed the problem down to the following three hash maps:

```
hash_map<id_t, Node*> node_by_id;
hash_map<id_t, vector<pair<id_t, bool>>> edges_on_start;
hash_map<id_t, vector<pair<id_t, bool>>> edges_on_end;
```

All three have an integer key. In the GCC implementation, the `std::hash` of an integer is the integer itself. If the node identifiers form a contiguous range, they will occupy a contiguous range of buckets in the hash table as well.

The problematic graphs all come from haplotype-aware pruning. Their node identifiers form two contiguous ranges: one for the original nodes and another for duplicates. Sometimes these ranges overlap modulo bucket count. If the overlap is bad enough, searching the hash table will take about sqrt(n) probes for many keys.

We need a better hash function for integer keys to solve this problem.

jltsiren:
* Replace `std::hash` with Thomas Wang's hash function for integer keys in `dense_hash` and `sparse_hash` to resolve #1457.

* Use `--kmer-length 24` and `--edge-max 3` as defaults in `vg prune` in order to build 256-mer indexes. `--kmer-length 16` is appropriate for building 128-mer indexes with `--prune` and `--restore-paths`, but it fails when trying to build a 256-mer index for 1000GP chrX.

* One of the `vg augment` tests fails with this PR. As far as I understand the output, the resulting files have the right nodes with wrong identifiers, possibly because we now get things in different order from the hash tables.


jltsiren:
The first test in `test/t/17_vg_augment.t` now outputs the same data in different order. I updated the truth file to take this into account. However, it would be better to have a test that does not depend on the order of the elements in a hash table.

ekg:
Often I have worked around ID ordering problems by checking the label space
of the graph, which should be stable. In GFA this would be the second
column of the S records.

PR #1466, opened 2018-02-22
Improved implementation of streaming/multithreading code

jeizenga:
In the old implementation we have one designated thread that does the streaming in. It built a buffer of tasks and then would wait if the buffer got too big. In my observation, we were basically always bound by compute, not disc reading, so the buffer would always fill up and then this thread would sit idle most of the time. I've changed things around so that this thread now does useful work instead of waiting. I also refactored to get rid of some duplicate code.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -141,55 +141,55 @@ bool get_next_alignment_pair_from_fastqs(gzFile fp1, gzFile fp2, char* buffer, s
     return get_next_alignment_from_fastq(fp1, buffer, len, mate1) && get_next_alignment_from_fastq(fp2, buffer, len, mate2);
 }
 
-
-size_t fastq_unpaired_for_each_parallel(const string& filename, function<void(Alignment&)> lambda) {
-    gzFile fp = (filename != "-") ? gzopen(filename.c_str(), "r") : gzdopen(fileno(stdin), "r");
-    if (!fp) {
-        cerr << "[vg::alignment.cpp] couldn't open " << filename << endl; exit(1);
-    }
-    size_t len = 2 << 18; // 256k
+size_t unpaired_for_each_parallel(function<bool(Alignment&)> get_read_if_available, function<void(Alignment&)> lambda) {
+    
     size_t nLines = 0;
-    int thread_count = get_thread_count();
     vector<Alignment> *batch = nullptr;
-    char* buf = new char[len];
-    const uint64_t batch_size = 2 << 8;;
-    // max # of such batches to be holding in memory
-    const uint64_t max_batches_outstanding = 2 << 8;
     // number of batches currently being processed
     uint64_t batches_outstanding = 0;
-    bool more_data = true;
-#pragma omp parallel default(none) shared(fp, more_data, batches_outstanding, batch, len, buf, nLines, lambda)
+#pragma omp parallel default(none) shared(batches_outstanding, batch, nLines, get_read_if_available, lambda)
 #pragma omp single
     {
+        // max # of such batches to be holding in memory
+        const uint64_t max_batches_outstanding = 1 << 9; // 512
+        // number of pairs in each batch
+        const uint64_t batch_size = 1 << 9; // 512
+        
+        // alignments to hold the incoming data
+        Alignment aln;
+        // did we find the end of the file yet?
+        bool more_data = true;
+        
         while (more_data) {
-            if (!batch) {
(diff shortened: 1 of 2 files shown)


ekg:
Do you have benchmarking results showing how much this improves things?

jeizenga:
I was hoping Jenkins would give me some idea, but he seems to be laying down on the job. On big compute servers it probably won't help a ton because it will basically be like switching from, e.g., 31 to 32 threads. Local runs on single CPUs could speed up substantially.

jeizenga:
Jenkins, test this please.

adamnovak:
Jenkins, test this please

jeizenga:
Jenkins, test this please.

ekg:
@jeizenga I see the behavior you're trying to correct on my laptop, but I notice it tends to subside after a little while. In other words I go from using 3 cores (of 4) to all 4 after about 30 seconds of operation. If this resolves this part and doesn't introduce problems on high-core systems then I'm all for it.

ekg:
@mlin would you please review this?

mlin:
One concern- if the master thread gets unlucky with a task that takes extra long (e.g. a batch of difficult-to-map reads) then there is a risk of starving the whole thread pool of work to do. The risk of this depends on the distribution of task difficulty vs the queue size limit vs the number of threads.

I do see the problem you're addressing where only 3 of the 4 cores would be utilized. I'm typically running `vg map` with 16 or 32 threads alongside streaming decryption of the incoming reads, which has masked the effect for me.

Would setting ```OMP_NUM_THREADS=$((`nproc`+1)) vg map ...``` address that problem? (But the PR does some good refactoring we should use regardless)

ekg:
I've tested this and it's looking good!

adamnovak:
On Tue, Feb 27, 2018 at 3:30 PM, Mike Lin <notifications@github.com> wrote:

> One concern- if the master thread gets unlucky with a task that takes
> extra long (e.g. a batch of difficult-to-map reads) then there is a risk of
> starving the whole thread pool of work to do. The risk of this depends on
> the distribution of task difficulty vs the queue size limit vs the number
> of threads.
>

I complained about that too; Jordan's response was basically that our
buffers are big enough, and that the mapping takes so much longer than the
loading that basically they are always full, so it shouldn't be a problem
with the task size distribution we have.

I guess that's a problem to monitor for (somehow), but I suspect it won't
happen a lot, and if it does happen the offending long-running reads would
themselves count as a mapper bug.

mlin:
But the one loader thread has to keep up with N worker threads. For N=4 this may be easy but N can be 32, 64, 128...

Sent from my iPhone

On Feb 28, 2018, at 11:10 AM, Adam Novak <notifications@github.com<mailto:notifications@github.com>> wrote:

On Tue, Feb 27, 2018 at 3:30 PM, Mike Lin <notifications@github.com<mailto:notifications@github.com>> wrote:

> One concern- if the master thread gets unlucky with a task that takes
> extra long (e.g. a batch of difficult-to-map reads) then there is a risk of
> starving the whole thread pool of work to do. The risk of this depends on
> the distribution of task difficulty vs the queue size limit vs the number
> of threads.
>

I complained about that too; Jordan's response was basically that our
buffers are big enough, and that the mapping takes so much longer than the
loading that basically they are always full, so it shouldn't be a problem
with the task size distribution we have.

I guess that's a problem to monitor for (somehow), but I suspect it won't
happen a lot, and if it does happen the offending long-running reads would
themselves count as a mapper bug.

jeizenga:
I was watching it on a 32 core machine not so long ago and never saw the 512-batch buffer dip below 510. That said, I do think it would make sense to have the buffer size adapt to the number of threads if we're going to take this approach, rather than be a fixed constant.

ekg:
It took a few iterations but I'm seeing pretty poor performance with some mapping workloads. Basically, total resource starvation and only a handful of threads running. I'm looking to verify the cause is this PR.

@mlin had addressed this before and I think we might have regressed to the previous state of things with this patch.

ekg:
It looks like the problem I ran into isn't related to this PR. It may instead be related to how surjection works as I'm getting it when running --surject-to bam.

jeizenga:
This PR hasn't been merged yet, so it really couldn't be

ekg:
It has been merged into the mappering branch.

ekg:
As noted in #1480 and #1481 I haven't found the multithreading to be a problem. Disregard my comments here.

Issue #1475, opened 2018-02-28
Simplify indexing


jltsiren:
The current indexing workflow is complicated and poorly documented. I'm going to simplify it, set reasonable default parameters, and document it in the wiki.

- [X] Build per-chromosome XG indexes in `vg prune` when needed.
- [X] Take the default values for `--kmer-size` and `--doubling-steps` in `vg index` from GCSA.
- [X] Write the temporary kmer files to the specified temp directory in `vg index`.
- [X] Reorganize the command line help for `vg index`.
- [X] Separate GBWT construction from XG construction in `vg index`.
- [x] Remove unnecessary parts of RocksDB indexing from `vg index`.
- [x] Pass thread names and sample counts from GBWT construction to XG construction.
- [x] Merge per-chromosome indexes in `vg gbwt` (see #1474).
- [X] Parameter `--size-limit` is confusing in `vg index`. We should use it as an actual limit for temporary disk space usage.
- [x] Automatically remove temporary files.
- [x] Document the indexing process.

jltsiren:
The documentation is now [in the wiki](https://github.com/vgteam/vg/wiki/Index-Construction).

PR #1478, opened 2018-03-01
Try to fix Mac tests

jltsiren:
I may have figured out why the Travis tests often time out on Mac.

The first tests in a freshly cloned repository work fine. They do not clean up properly afterwards. One of the files left behind is `x.gbwt`. On the next time, in `06_vg_index.t`, we run the following line:

```
vg map -T <(vg sim -s 1337 -n 100 -x x.xg) -d x > x1337.gam
```

This loads `x.gbwt`, and then things get out of control. For some reason, the `vg map` command uses tens of gigabytes of memory. Macs don't have that much memory, so they use swap instead.

The tests now delete all temporary files that would show up on `git status`. This hopefully fixes the issue.

Also, because I was really working on #1475, this PR includes other changes:

- `vg prune` builds the XG indexes it needs. We should not have to store XG indexes for individual files when indexing multi-file graphs.
- `vg index -g` takes the default values for `--kmer-size` and `--doubling-steps` from GCSA. As a result, it now builds 256-mer indexes by default.
- Temporary files are no longer hidden, and they will be written to the appropriate temp directory instead of the working directory. This is particularly important for the kmer files in `vg index -g`.

### DIFF
--- a/src/build_index.hpp
+++ b/src/build_index.hpp
@@ -25,7 +25,7 @@ void build_gcsa_lcp(VG& graph,
                     int kmer_size,
                     size_t doubling_steps = 3,
                     size_t size_limit = 500,
-                    const string& base_file_name = ".vg-kmers-tmp-");
+                    const string& base_file_name = "vg-kmers-tmp-");
 
 }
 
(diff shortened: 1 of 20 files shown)


ekg:
I believed I resolved this on the mappering branch, but I still see the timeouts.

Sorry that I have diverged for two weeks so we are duplicating effort.

PR #1480, opened 2018-03-02
Wrapping up loose ends on multithreading

jeizenga:
https://github.com/vgteam/vg/pull/1466 was mostly merged via https://github.com/vgteam/vg/pull/1456, but I wanted to address some last minute concerns about starving the buffer of loaded alignments if the IO thread got a particularly unbalanced and costly batch. Now it will try to detect when this might happen and increase the buffer size accordingly. Specifically, if more than 1/4 of the buffer gets eaten up while it's trying to process a batch, the buffer doubles (up to a point). It's a little hack-y, but I think it will circumvent this problem most of the time. I'm open to more robust options if anyone has them.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -150,10 +150,13 @@ size_t unpaired_for_each_parallel(function<bool(Alignment&)> get_read_if_availab
 #pragma omp parallel default(none) shared(batches_outstanding, batch, nLines, get_read_if_available, lambda)
 #pragma omp single
     {
-        // max # of such batches to be holding in memory
-        const uint64_t max_batches_outstanding = 1 << 9; // 512
-        // number of pairs in each batch
+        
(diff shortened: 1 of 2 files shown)


ekg:
So to be clear, I have *not* found the previous changes from #1466 to be a problem.

I just discovered the actual issue that was driving my alignment performance down is that I completely unbounded the number of MEM hits that we could consider in `vg map`. This uncovered a problem with the MEM chainer related to some kind of block against huge memory allocations. Eventually all threads would run into a MEM with millions of hits, and then get stuck. This was the actual problem, not some thread exhaustion thing.

That noted, I think that this makes sense as a response to the concerns raised by @mlin.

jeizenga:
Gotcha

jeizenga:
Merging through a timeout failure on the Mac potato

PR #1482, opened 2018-03-02
Add Yohei's Li and Stephens code and a new haplotype scoring facade

adamnovak:
Now we can use a polymorphic `haplo::ScoreProivider` to score haplotypes instead of having to pick the right template instantiation based on what the mapper has available.

Right now everything is limited to GBWT-based scoring, but @yoheirosen's sublinearLS code *should* just need to be hooked up to the command line in order to work.

The polymorphic dispatch could be marginally slower, but probably not enough to notice.


adamnovak:
OK maybe it's not a facade so much as it is a strategy setup.

adamnovak:
The Travis Mac tests are failing here again because of #1490. All the other Travis tests have passed.

We will see about the Jenkins tests.

PR #1484, opened 2018-03-02
 Add a read name prefix filter to vg filter 

adamnovak:


adamnovak:
The test failure here is #1490 which is present in master. I am going to merge this anyway.

PR #1485, opened 2018-03-02
Know your limits

ekg:
There are two goals I have here.

First, the handling of the `--hit-max` condition is improved so that we filter down the list of candidate MEMs that is higher than the threshold we filter it down deterministically to just under our `--try-at-least` or `--max-multimaps` limit, whichever is higher.

Second, I want to debug surjection, and this PR includes some changes to help make that easier.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -4869,6 +4869,7 @@ Alignment Mapper::surject_alignment(const Alignment& source,
     Alignment surjection = source;
     // Leave the original mapping quality in place (because that's the quality
     // on the placement of this read in this region at all)
+    surjection.clear_mapping_quality();
     surjection.clear_score();
     surjection.clear_identity();
     surjection.clear_path();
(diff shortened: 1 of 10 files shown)


ekg:
Would someone on a mac please help me understand why the msga test is returning different results when we run with more than one thread or not?

I'd be curious to see the differences in the graphs that are produced from these two runs:

```
vg msga -f GRCh38_alts/FASTA/HLA/V-352962.fa -t 4 -k 16 | vg mod -U 10 - | vg mod -c - >a.vg
vg msga -f GRCh38_alts/FASTA/HLA/V-352962.fa -t 1 -k 16 | vg mod -U 10 - | vg mod -c - >b.vg
```

I've disabled the test for now as I have no ability to debug it, but am trying to get travis to print out the different results in JSON.

ekg:
It looks like there is a crash on the OSX build when doing the multithreaded msga. This can be reproduced on master... If you have a Mac. Any insights?

jltsiren:
The multithreaded msga test works on my MacBook Pro, on both master and know-your-limits. (On the other hand, my local runs of tests 19 and 21 of `t/14_vg_mod.t` always produce a wrong md5, which is probably a locale issue.)

ekg:
Jenkins takes a long time these days...

PR #1498, opened 2018-03-06
More progress on haplotype aware mapping evaluation, and restored XG upgrade code

adamnovak:
I wrote the code to upgrade old XG formats to version 8 by converting all the paths.

Also I'm hooking up Yohei's sublinearLS library to try and compare against the GBWT. I don't actually have it working yet (it seems to run out of memory, and it crashes when the XG index was built alongside a GBWT), but it is hooked up in mpmap.


adamnovak:
Jenkins had some kind of internal failure in pip for no apparent reason.

Jenkins, test this please.

PR #1499, opened 2018-03-07
Minimized stack traces for Mac and Linux systems

tanessav:
Modified crash system to print stack trace to a file when a crash occurs. If an environment variable is set (VG_FULL_TRACEBACK=1), then the stack trace will output to standard error like normal. This works on Mac and Linux systems.


ekg:
Thank you!

Issue #1508, opened 2018-03-08
Memory issues with vg mpmap


cmarkello:
@jeizenga @adamnovak @ekg 

So I've been attempting to run the latest version of vg (v1.6.0-654-g1c819fff-t142-run) against my dataset and found that it takes up an unreasonable amount of memory during a run of `vg mpmap`. This didn't use to happen with previous versions of vg.

Using the vg container version `v1.6.0-621-g766ecbcb-t141-run` (vg commit hash 766ecbcbbe9d685c7cee908b9ef498f6cacdacdb) the run of `vg mpmap` only took up a maximum of about 74 GB of memory per subset read dataset.

When attempting to use vg container version `v1.6.0-654-g1c819fff-t142-run` (vg commit hash 1c819fffaefdc3ae7573f998cb6e27f0bac52600) the exact same runs take up more than 200 GB of memory. Memory during this run grows in the following intervals during a `vg mpmap` run:
```
1 min: 37 GB
2 min: 59.3 GB
3 min: 71.8 GB
7 min: 80.4 GB
23 min: 151.3 GB
33 min: 185.4 GB
36 min FAILURE: 198.4 GB
```

adamnovak:
This looks a lot like a memory leak to me; memory usage increases quickly to ~70-80 GB, and then slowly and linearly from there until the process is killed.

Relevant changes that were made over that commit range include:

* Jordan's new parallel driver code to let the input thread work
* Changes to MEM finding
* Changes to surject

adamnovak:
I doubt we've fixed it, but you could try.

Can you share the exact `vg mpmap` command you are using? Even without the reads themselves, I want to be able to replicate exactly what you are doing (including the types of indexes and input and output formats in use).

cmarkello:
Sure, it's pretty basic. No particular parameter tuning. It uses vg container `quay.io/vgteam/vg:v1.6.0-621-g766ecbcb-t141-run`
```
vg mpmap -S -f /home/markellocj/split_fastqs_[SAMPLE#]/fq_chunk_1.part.aa.fq.gz -f /home/markellocj/split_fastqs_[SAMPLE#]/fq_chunk_2.part.aa.fq.gz -x /mnt/1kg_ref_hs37d5.xg -g /mnt/1kg_ref_hs37d5.gcsa -t 32
```

cmarkello:
[...]

PR #1519, opened 2018-03-09
Fix new[]/delete mismatch

adamnovak:
This may or may not help with #1508.

If you make an array with new, you need to delete it with `delete[]`.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -347,7 +347,7 @@ size_t fastq_unpaired_for_each_parallel(const string& filename, function<void(Al
     
     size_t nLines = unpaired_for_each_parallel(get_read, lambda);
     
-    delete buf;
+    delete[] buf;
     gzclose(fp);
     return nLines;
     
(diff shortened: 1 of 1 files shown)


adamnovak:
I'm actually pretty sure that this won't fix #1508; the problem was present in the last known good commit for that issue. But undefined behavior is undefined, which includes potentially changing in behavior for no apparent reason. So maybe this is somehow the cause anyway.

ekg:
@adamnovak do you see a huge slowdown on your own system from this? Could Jenkins be sad because of competing processes on the same node?

adamnovak:
@ekg Jenkins has been thrown out of whack because Edwin on our end changed the Jenkins worker image to one that uses the current version of Toil (because Jenkins updated and stopped running properly on the older images). We found that our tests didn't work right with current Toil as of a few months ago, but we just decided to use the old Toil indefinitely because we couldn't work out how to fix it.

I will have to see about getting Jenkins back in working order.

adamnovak:
Jenkins has been fixed!

Jenkins, test this please.

PR #1521, opened 2018-03-12
Put a hard upper limit on the hit count of locate-able MEMs

### LINKED ISSUE #1508: Memory issues with vg mpmap

adamnovak:
This limit is hopefully "big enough".

The real solution I think would be to roll this into locate, and have locate
limit the result count itself and return some kind of flag and a truncated
result set if there are too many hits. But I'm going to implement this instead
because it's easier and will get @cmarkello going faster.

Should fix #1508.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -200,7 +200,7 @@ BaseMapper::find_mems_simple(string::const_iterator seq_begin,
     for (auto& mem : mems) {
         if (mem.length() >= min_mem_length) {
             mem.match_count = gcsa->count(mem.range);
-            if (mem.match_count > 0) {
+            if (mem.match_count > 0 && mem.match_count <= absolute_locate_maximum) {
                 gcsa->locate(mem.range, mem.nodes);
                 if (hit_max > 0 && mem.nodes.size() > hit_max) {
                     if (hit_limit) {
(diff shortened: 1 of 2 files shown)


ekg:
@adamnovak you've implemented exactly what I was considering implementing.

I found this problem in my testing and didn't appreciate that it might also be causing issues for mpmap. This perfectly matches the bug reports.

This is relevant: https://github.com/jltsiren/gcsa2/issues/34.

ekg:
Also, from my tests 16k is probably enough.

Sorry Jenkins, I'm merging through you.

PR #1529, opened 2018-03-14
use GCSA2's new random subsampling of matches

ekg:
https://github.com/jltsiren/gcsa2/issues/34

We downsample to BaseMapper::hit_max hits using @jltsiren's new in-GCSA match downsampling functionality.


ekg:
I can't wait for Jenkins to not finish, so I'm merging through.

We need to figure out what's going on there :worried:

PR #1547, opened 2018-03-20
Long read alignment and lower vg::VG memory usage

ekg:
This implements a bunch of changes that make the AlignmentChainModel do the right thing irrespective of bandwidth. The result is that we can build linear graphs for all the HLA genes in the test directory:

joined gfa

It also works on the memory usage of the vg::VG object. This is done by using [sparsepp](https://github.com/greg7mdp/sparsepp) rather than google::dense_hash_map. This part isn't done as of yet.


ekg:
I have spent hours trying to figure out what's wrong with the cactus MHC paired-end mapping. Finally I realized that the actual problem is that there is something deeply wrong with the AUC calculation. In the [latest test](http://cgl-pipeline-inputs.s3.amazonaws.com/vg_cgl/vg_ci/jenkins_reports/commit/25eb85b935317ca8c01419388101d3b748976bcf/index.html) we see a lower AUC for cactus-pe than cactus-se, but this is clearly not the case in the plots. The same is true for bwa mem and other mappers. What is the AUC actually calculating?

ekg:
Unfortunately, I don't think I can wait for the OSX build to start. It's been >12 hours. It's time to close this and move on.

adamnovak:
Sadly, while Travis was broken, this PR seems to have introduced a problem (#1580) that prevents building on Mac.

PR #1551, opened 2018-03-21
Adding custom matrix file argument including unit tests

kevyin:
Hi vgteam,

@koadman added an argument to accept a custom score-matrix file and we're hoping this will be useful to the community.

There are some tests in the PR which hopefully is sufficient. It's passing "make test" and "vg test" atm.

Please let me know if you have any suggestions
Thanks


kevyin:
The osx test seems to have timed out?
> ESC[31;1mTimeout (40 minutes) reached. Terminating "make test"ESC[0m

adamnovak:
Jenkins, OK to test.

adamnovak:
Is the `OPT_SCORE_MATRIX` define the Right Way to assign numbers to long options? Not that I have a better way to do it; I think on other places we've just hardcoded small integers for the options that don't have short option letters.

kevyin:
I'm no expert either, 
All the alphabet letters and single digit ints are used up except for 'h' and '0' in the map subcommand
so I followed this suggestion from SO
https://stackoverflow.com/questions/22464891/specify-long-command-line-arguments-without-the-short-format-getopt

koadman:
Thanks guys!

Issue #1544, opened 2018-03-20
Add features into vg annotate during RDF summit 3


6br:
I will add some new features into VG during RDF summit 3 as discussed with @ekg.

`vg annotate` accepts

* BED format with strands(+/-) -- gene/exon annotations include the direction of strands, which should not discard.
* not only BED format but also GFF3/GTF format

6br:
In addition to the first idea, it seem to be needed.

* Inverted surject command: converter from bam to gam for the compatibility between existing *linear* alignments and alignments on the graph

6br:
Currently, if we want to merge the paths implied by annotations, we have to run 2 commands; `vg annotate` to generate gam and `vg mod -i` to merge the paths.
It seems to be good if there is a subcommand like below.

input: xg and a set of bed/gff3/gtf(annotations)
output: vg file merged the paths implied by annotations

6br:
@JervenBolleman  Now #1546 is merged, so I think BED file with strands(+/-) can handle accurately in the master branch.

6br:
All discussion on RDF summit 3 was extremely meaningful for me. I appreciate @ekg and all developers!

PR #1546, opened 2018-03-20
Add is_reversed option into target_alignment

6br:
Refer: #1544 
Now vg annotate accepts BED format with strands(+/-) -- gene/exon annotations include the direction of strands, which should not discard.

PR #1552, opened 2018-03-21
Add an interface to convert gtf/gff3 to gam

6br:
Refer: #1544


6br:
MEMO:
-[ ] GFF/GTF is 1-based
-[x] GFF/GTF has no `name` field without using attributes. Current implementation discards any ids or attributes to generate a GAM file.

adamnovak:
Jenkins, ok to test.

6br:
It is ready to be merged after #1546 is merged.

Issue #1470, opened 2018-02-23
Possible route forward with surject


jeizenga:
@ekg @adamnovak 
I ran into a lot of the same problems that I think we're running into with the surject algorithm when I was working out how to construct MultipathAlignments. I think the code I wrote could probably be repurposed to make a pretty good surjection algorithm.

The MultipathAlignment workflow is kind of like this:
1. Extract a subgraph around some MEMs (https://github.com/vgteam/vg/blob/master/src/multipath_mapper.cpp#L1929)
2. Walk out MEMs into Paths (https://github.com/vgteam/vg/blob/master/src/multipath_mapper.cpp#L3641)
3. Compute the connectivity between the Paths (https://github.com/vgteam/vg/blob/master/src/multipath_mapper.cpp#L3666)
4. Do alignments between Paths that could be connected into longer alignments (https://github.com/vgteam/vg/blob/master/src/multipath_mapper.cpp#L2276)

I can imagine a similar workflow for surjection:
1. Extract the reference path graph around an Alignment
2. Get the sub-Paths from the Alignment that are on the reference path
3. Compute the connectivity between the sub-Paths
4. Do alignments between the sub-Paths

It would take a little refactoring, but the algorithms are all already there and I'm quite confident in them.

ekg:
This would be ideal. It would provide maximum control over the surjection
with minimum realignment cost.

I haven't done so but perhaps now that I'm rewriting surject again I should
refactor it into an object where these different approaches can live.

I'll be in touch about the performance of the "heavy full length bonus" way
of handling surjection. Thus far it looks like it is working very well but
this doesn't mean there won't be low frequency problems with it.

ekg:
Here's a test harness we can use:

```
vg construct -r 1mb1kgp/z.fa -v 1mb1kgp/z.vcf.gz -m 32 -p >z.vg
vg index -g z.gcsa -k 16 -p z.vg
vg sim -x z.xg -l 100 -n 1000 -e 0.01 -i 0.005 -a >z.sim
[...]

PR #1571, opened 2018-03-27
New and improved surject algorithm

jeizenga:
Finishes the previously merged code that included the new path-anchored surject algorithm. It is now debugged and functional, and the results summarized in https://github.com/vgteam/vg/issues/1470 suggest that it is an improvement in both accuracy and speed. I've replaced invocations of the current surject algorithm with the new one, so there aren't any CLI-accessible code paths through the old surject anymore. I didn't remove the code though.


ekg:
It'll be worth checking the alignments manually to see if there are any funny artifacts. Also, we aren't testing variant calling on the output BAMs and previously we've found up errors that way. That all noted I'm going to merge so we can go forward from here.

Issue #1563, opened 2018-03-22
vg constuct maf import: `aln.second.size() == aln_len'


ChriKub:
Hi,

I am trying to import a maf file that was constructed using progressiveCactus and hal2maf into vg but it fails. The file has passed the validation with maftools mafValidator. The maf file can be downloaded [here](https://owncloud.tuebingen.mpg.de/index.php/s/2vrdMI5ZtxSvmBe).

Thanks

`hal2maf 15.2.hal 15.2.maf`
`vg construct -M 15.2.maf > 15.2.vg`

> vg: src/msa_converter.cpp:135: vg::MSAConverter::MSAConverter(std::istream&, std::__cxx11::string, size_t): Assertion `aln.second.size() == aln_len' failed.

> Stack trace (most recent call last):
> #10   Object "/ebio/abt6_projects9/abt6_software/bin/vg/bin/vg", at 0x4b9ce8, in _start
> #9    Object "/ebio/abt6_projects9/abt6_software/bin/vg/bin/vg", at 0x113a5b9, in __libc_start_main
> #8    Object "/ebio/abt6_projects9/abt6_software/bin/vg/bin/vg", at 0x113a3c5, in generic_start_main
> #7    Object "/ebio/abt6_projects9/abt6_software/bin/vg/bin/vg", at 0x404ee6, in main
>       Source "src/main.cpp", line 61, in main [0x404ee6]
> #6    Object "/ebio/abt6_projects9/abt6_software/bin/vg/bin/vg", at 0x7e4f97, in vg::subcommand::Subcommand::operator()(int, char**) const
>     | Source "src/subcommand/subcommand.cpp", line 72, in operator()
>       Source "/usr/include/c++/5/functional", line 2267, in operator() [0x7e4f97]
>        2264:     {
>        2265:       if (_M_empty())
>        2266:    __throw_bad_function_call();
>       >2267:       return _M_invoker(_M_functor, std::forward<_ArgTypes>(__args)...);
>        2268:     }
>        2269:
>        2270: #if __cpp_rtti
> #5    Object "/ebio/abt6_projects9/abt6_software/bin/vg/bin/vg", at 0x81707c, in main_construct(int, char**)
>       Source "src/subcommand/construct_main.cpp", line 334, in main_construct [0x81707c]
[...]

PR #1573, opened 2018-03-27
Fix MAF parsing in MSA constructor

jeizenga:
The MAF parser would break on files with multiple contigs (https://github.com/vgteam/vg/issues/1563). This fixes the bug and also pretties-up some of the factoring and the user friendliness.

### DIFF
--- a/src/msa_converter.cpp
+++ b/src/msa_converter.cpp
@@ -13,9 +13,18 @@
 namespace vg {
 
 using namespace std;
+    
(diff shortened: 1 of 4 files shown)


fbemm:
I tested #1573  in the morning on 506 gene loci. Works! Users should supply --noDupes and --noAncestors during hal2maf. vg constructe does not like duplicates contigs in MAF blocks and vg index -x does not like the ancestor nodes.

PR #1575, opened 2018-03-27
Factor the updateable priority queue out into the new Structures

adamnovak:
I'm also fiddling with the indentation, which is why I touched so many lines.

We can't agree on whether namespaces should indent or not.


ekg:
It looks like we're blocked on the OSX builds on travis not starting. I'm going to merge anyway.

PR #1597, opened 2018-04-02
Update mpmap to use 256-mer indexes better

jeizenga:
This might not be totally final, but I want to get Jenkins working on it. I've been investigating bad reads on whole genome simulated data. I have one bug-fix and some parameter adjustments that collectively give me better performance on many of them. Hoping to merge this soon so I can get it into the quay -> toil-vg testing pipeline.


jeizenga:
Jenkins, test this please.

jeizenga:
I'm okay with this failure, I think. The mapping all seems more accurate, and I don't want to confuse my current optimizations for mpmap with optimizations for variant calling.

PR #1602, opened 2018-04-04
Flip reads around to be in the "good" order for mpmap

adamnovak:
mpmap apparently cares what order the input reads are in.

Also, my supposedly snarl-cutting-free controls had snarl cutting in them after all. This fixes that.


ekg:
Should we move to merge this?

adamnovak:
I think not, still. It did not have the effect I thought it would in my
testing. I want to be sure that this actually improves performance before
merging it; right now it looks like it might be making things worse, but I
haven't run enough experiments to really pin it down.

adamnovak:
I don't think we really want to touch order right now. I may come back to this eventually but for now I will just leave everything in the input order.

PR #1606, opened 2018-04-05
Safe svs

edawson:
This fixes a bunch of bugs I've introduced.

- GFA output is now correct, and doesn't eat memory building a full gfakluge object.
- vcflib gets a bunch of checks to make sure you can't get an SV length or SV end that doesn't exist (but probably not enough checks).
- I ripped out the deconstruct tests, because they cause build issues and don't reflect the behavior we want anyway.


edawson:
I'd like to flush the Travis cache, because I think it's choking this build. This works locally, everything is up to date, and it shouldn't have any problems. I'll wait for a go ahead.

edawson:
@adamnovak @ekg The Jenkins test that fails returns this error: 
`URLError: <urlopen error [Errno 110] Connection timed out>`

The last commit shouldn't have broken Jenkins. Is this a random Jenkins failure? Can I restart the job somehow?

adamnovak:
That looks like random Internet weather to me. If you want to rerun the tests you can say the magic words like I did here: https://github.com/vgteam/vg/pull/1687#issuecomment-390366249

Issue #1604, opened 2018-04-05
record time to align a read in the alignment object


ekg:
This goes totally against the reduction in alignment size I wanted to achieve. I'd like to record statistics about the complexity of a given alignment in the output GAM so that the reads that are problematic can be reviewed. Maybe I should emit a second stream of this, but it could be generally useful for people applying vg to complex graphs.

Issue #1587, opened 2018-03-30
cleanup Alignment object


ekg:
The Alignment object is extremely bloated by annotations that are only relevant to SAM compatibility. I suggest we pack these into the same kind of integer field that is used in the SAM record and remove the numerous boolean and annotation fields that we've added.

The current object requires 256 bytes in memory. This is way too much for what the object is doing.

adamnovak:
What do we do about bools that we want that aren't in SAM? Or about SAM flags that might not be applicable here?

Do we want to remove all the non-bool annotations? Were they only added as temporary hacks? Or do they represent things that we still sometimes want to annotate alignments with? If so, how will we handle that annotation?

I am thinking about an annotation system for tagging Alignments and MultipathAlignments with features in a more sustainable way, because https://github.com/vgteam/vg/issues/1603 and some of the machine learning stuff I am looking into kind of want to be able to attach their own data to Alignments.

jeizenga:
While we're thinking about redesigning Alignments, my vote is that the fragment_next and fragment_prev fields be merged into field that is just a string instead of another Alignment (i.e. the way I have it in MultipathAlignments). Molecularly, the next/prev distinction is not meaningful. Since we sequence the reverse complement of the other side of the fragment, the opposite read is forward along the genome from the perspective of both reads. I know there are subtle differences between the read1's and read2's from Illumina machines, but I feel like that is such a narrow platform-dependent feature that it's best just left in the read-naming convention if people want to use it.

adamnovak:
I think we have the next/prev setup so we can do multi-fragment templates.
SAM/BAM has support for this; you can express that more than 2 reads are
part of the same template molecule. Are we sure we will never want to
support a technology that does that?

jeizenga:
Are we considering this settled as closeable?

adamnovak:
I think we have an idea of what to do: refactor things that use the in-Alignment annotations into things that use the new annotation system. But we haven't moved over very much stuff yet.

PR #1608, opened 2018-04-06
Machine Learning Groundwork: Features

adamnovak:
This addresses #1587, at least partly, by creating a system of `Feature`s which can be used to annotate `Alignment`s and `MultipathAlignment`s. It supports flags ("tags") as well as single-valued and multi-valued `double`s (into which you can put integral values if you want those too).

It doesn't (yet?) support annotating with extra Protobuf data, so things like `refpos` are still in the base Alignment. It also is a bit inefficient for things that we *always* want to include on `Alignment`s, so `identity`, and the standard-ish SAM-alike flags are still in the `Alignment`.

I've pulled out a bunch of the `Alignment` fields that were little-used annotations and made them into `Feature`s. I think this is the Right Way to implement things like #1604 as well.

Since this is intended to lay the groundwork for MAPQ recalibration, I am also pulling in `dlib` as a dependency, which Kishwar in our lab recommended. It seems to be a header-only library, so it's just getting checked out and copied into include right now.


ekg:
I'm not a big fan of this. I understand the goal is to keep memory usage down but it does really break the regularity of the JSON version of these annotations. In fact it makes it positively difficult to parse out important features from the alignment, such as the fragment distribution model, which previously was something you could copy and paste directly into the map command line in order to rerun with exactly the same model as the alignment was down with.

Issue #1603, opened 2018-04-05
vg map --compare using reference positions


ekg:
I'd like to measure when vg map works without external scripts in cases where we don't have a matching graph node id space between the simulated and test graphs.

The obvious solution is to do what's been done in the various mapping simulation tests and use the reference position of the simulated alignment as truth. I'd guess this different `--compare` function would
take a tolerance distance within which the mapping is considered correct.

adamnovak:
This is something I want integrated into vg as well; I think it would dramatically speed up my experiments on haplotype-aware mapping to not have to constantly be converting things to and parsing JSON.

The tolerance distance is exactly what we're using now, AFAIK. It would work fine for acyclic graphs.

adamnovak:
So the goal here is to change the existing `--compare` to use positions, pulling from `Alignment.refpos()`? And your GAM input would define `refpos()`?

adamnovak:
Is it important to preserve the existing `--compare` behavior of calculating overlaps? It doesn't seem to have tests constraining it to work the way it does now.

ekg:
Looks like we implemented the same thing. Thank you for doing this. I guess I have to merge to you given the extension of this into the feature stuff. I implemented another function called vg gamcompare in the long-read-alignment branch. The idea with that is to take in the alignments from another process (like bwa mem) read them through inject and then do the comparison against another set.

faithokamoto:
Seems this was done at the time? If it wasn't then it would be stale anyhow as we're not testing vg map and haven't been for a hot minute.

PR #1609, opened 2018-04-06
Position comparison in map

adamnovak:
This implements #1603 on top of the features system, as a `--compare-range` flag on `vg map`.

I also tweaked `pb2json` to output feature names instead of opaque numbers, since `json2pb` can parse both.


ekg:
I think we should hold off on this. I implemented an alternative way of encoding these that I prefer. It annotates the alignments with the distance to the correct alignment and the specific path on which the measurement is done. So I'd like to rebase some of the changes here on the long-read-mapping branch.

adamnovak:
We don't need this anymore now that we have `vg gamcompare`.

PR #1616, opened 2018-04-09
How should we do features and annotations?

adamnovak:
This PR is not really for merging and will definitely fail the tests.

Instead I'm sketching out how annotations and the alignments would look if be broke out a `CoreAnnotations` object to hold the ones we use all the time, and we left another place for other annotations.

Questions for discussion:

1. What should go in the core annotations vs. the object itself? I think maybe the read group stuff is a candidate for promoting to real alignment fields, even though it would be duplicated, because people want to use that all the time.

2. What should be a core annotation vs. an extra annotation? I think the Locus is a candidate for demotion.

3. Is `repeated Any` a reasonable way to do extra annotations? Can pb2json be induced to serialize/deserialize them as JSON objects when the types are known to the program? Should it be a `repeated string`, or maybe a message with a string-string map in it?

@ekg @jeizenga @edawson Discuss.


ekg:
What about using a repeated Struct for the "feature"? The alternative is a repeated Any, which has advantages and disadvantages, among which each message must retain a URI to the schema of the object and there is not forward compatibility if you change this schema. [Both options and an interesting hybrid are discussed here.](https://blog.envoyproxy.io/dynamic-extensibility-and-protocol-buffers-dcd0bf0b8801)

The [Struct](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/struct) looks like the safest thing for our context. You can round trip it through JSON. It won't add weight to the semantics nor object. In fact, it might help to lighten up our in-memory representation. This looks like a great option to me.

adamnovak:
Struct does look like a pretty great idea.

If the struct is basically a big key/value store, maybe we can use just a
single `Struct annotations`, without having to have it repeated and search
all the copies. It might be as big as a map<string, string> in the
generated code, but we have the benefit of non-string values.

We'll have to see how pb2json/json2pb deals with this. The JSON
representation is specified, but I don't recall seeing much support for any
of the "well known" types in that library, so it may not be fully compliant.

adamnovak:
OK, I think Struct is the way to go.

`pb2json`/`json2pb` doesn't know how to deal with Struct at all, because it doesn't actually know anything about maps (which is how Struct is implemented).

However, it turns out that Protobuf (now?) has its own JSON serialization/deserialization code built in. And *that* code not only knows about Struct but is also clever enough to represent it as idiomatic JSON, with the field names as keys and the values as values in a normal JSON object.

So now this PR is mostly swapping over to that JSON code instead of the guts of the implementation we were using. There are some warts, though:

* Protobuf refuses to serialize `int64` as anything other than a string, because if it were unquoted and read by Javascript it would be loaded into a double, and some 64-bit int values wouldn't come through unmodified. I had to change some of the tests that look at the JSON to accept quoted `node_id` values.
* Protobuf doesn't know how to do anything with a JSON stream other than turn it into a Protobuf binary stream in a blocking way. So right now I'm still using Jansson and some fragments of `json2pb` to parse out a self-delimited JSON message from a stream, and then re-serialize it to a string for Protobuf to read. Maybe instead we really want something with a FIFO and a thread, or our own custom JSON half-parser that can count braces and handle string escapes to find where each input message ends.

But I think once we've handled all that, and maybe out a slightly nicer API over it on the C++ side, a Struct is the Right Way to handle annotations.

adamnovak:
I'm going to merge this in.

The actual annotations stuff is not ready; I want to put a get/set API on top of the struct, like my `features.hpp`, because working directly with the oneofs in `Value` is a bit of a pain. But the stuff in this PR is mostly groundwork tinkering with JSON serialization (and fixing up some test warnings), which I think we want no matter how the final API looks.

PR #1618, opened 2018-04-09
Add correctness flag in GAM comparison

adamnovak:
This is on top of #1612.


ekg:
There is a simple typo here (`id` instead of `is` in the bash tap tests). I would push to this PR's base branch but I can't because it's attached to your repo.

adamnovak:
#1620 took care of this.

PR #1621, opened 2018-04-10
Initial implementation of MAPQ recalibration

adamnovak:
This adds a Vowpal-Wabbit-based MAPQ recalibration method. You can train it on the test set, and it sort of works:

image

You can use it like this (although you may want a bigger graph than `x.vg`, or reads longer than 10bp):

```
vg construct -r test/small/x.fa -v test/small/x.vcf.gz > x.vg
vg index -x x.xg -g x.gcsa x.vg
vg sim -x x.xg -n 100000 -l 10 --sub-rate 0.1 --random-seed 1 -a >sim.gam
vg map -x x.xg -g x.gcsa -G sim.gam -k 5 >mapped.gam
vg gamcompare -r 5 mapped.gam sim.gam | vg recalibrate --model recal.model --train -
vg recalibrate --model recal.model mapped.gam | vg gamcompare -r 5 - sim.gam --tsv --aligner recal > stats.tsv
vg gamcompare -r 5 mapped.gam sim.gam --tsv --aligner orig | sed 1d >>stats.tsv

Rscript scripts/plot-qq.R stats.tsv qq.svg
```

(I've also added TSV output to `vg gamcompare`, so we can get away with not using `toil-vg mapeval` or `jq`.)

Still TODO on this is:

* VW needs Boost as a dependency. If we're going to link in VW we also need to pull in Boost as a submodule, and build `boost_program_options`. Right now this PR just hopes that boost_program_options is already available, which is true for me, but which is likely to break our tests.
* VW's C++ API... isn't. You feed a command line to a magic init function, all the types are dropped in the global namespace, the things you can and can't do aren't really documented at all, and I'm not sure how to get it to stop printing to stderr. Do we really need VW's out-of-core abilities on our relatively small data? Are we sure we like VW better than the way more polished [dlib](http://dlib.net/ml.html#krr_trainer)? Or should we be calling out to the `vw` binary instead of trying to use `libvw.a`?
* I haven't actually tested this on bigger data sets or with real-looking reads. I've just trained on this test set of 10bp reads on the x.vg graph and gotten a sort of OK QQ plot. We can throw this in, but we shouldn't push it for users unless it really works and we can validate it with an experiment.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -85,3 +85,6 @@
 [submodule "deps/sparsepp"]
 	path = deps/sparsepp
 	url = https://github.com/greg7mdp/sparsepp.git
+[submodule "deps/vowpal_wabbit"]
+	path = deps/vowpal_wabbit
+	url = https://github.com/JohnLangford/vowpal_wabbit.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -88,3 +88,6 @@
 [submodule "deps/vowpal_wabbit"]
 	path = deps/vowpal_wabbit
 	url = https://github.com/JohnLangford/vowpal_wabbit.git
+[submodule "deps/boost-subset"]
+	path = deps/boost-subset
+	url = https://github.com/vgteam/boost-subset.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ dist: trusty
 before_install:
   - if [ -e deps ]; then mv deps deps_cached; fi
   - git submodule update --init --recursive
-  - rsync -rv --checksum deps/ deps_cached/
+  - rsync -rv --links --checksum deps/ deps_cached/
   - rm -Rf deps
   # Keep the cached deps if the right compiler version was used.
   # Otherwise start fresk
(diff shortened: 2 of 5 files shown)


ekg:
VW is a little awkward because of the boost dependency. If I can I will try to figure out how to build it in. We should only need a few header files I think? Although the interface may not be polished in the same way as dlib it is absolutely polished in terms of learning algorithms and performance. Could we compare the two or, given that we are merging them in, use both optionally in recalibrate.

adamnovak:
I've pulled out the necessary parts of Boost using the `bcp` tool, and put those in particular as a submodule. They're not too big and they don't take too long to build.

We could try and put in both vw and dlib, and have some way to switch them out. They have pretty different interfaces; VW wants to stream things, while dlib I think wants to work on batches and matrices. I'm also not sure what exactly we would compare about them; I guess we would be comparing the real kernel ridge regression model from dlib to whatever the closest approximation of that algorithm vw can deliver is?

ekg:
That's the thought. We can compare the best of each. But it's true that the interface differences are awkward.

The streaming model in vw makes it very useful in a few different parts of vg. For instance, we could apply it to learning the fragment distribution model dynamically, or particular sequences that are basically unmappable (thus requiring a lot of runtime to map). We could also directly apply it to genotyping or graph filtering. There are a lot of things that it can be used for and the more I think the more it makes sense to have it available internally.

PR #1622, opened 2018-04-11
Refactoring i_iv in xg.cpp

emikoifish:
- Refactored xg.cpp
   - store i_iv locally only 
   - change i_iv.size() to node_count
   - xg now makes i_iv and r_iv and compress, then make s_bv and s_iv and compress to decrease memory usage
- added better error statement when piping stdin into vg_set


jeizenga:
This has failed the automatic tests, so I guess it's time for some introduction to how those work. Every time you PR, it should kick off two tests: Travis and Jenkins. Travis mostly checks to make sure it builds and that the integration tests work. Jenkins runs a more complex pipeline that lets us check if the accuracy of the mapper and variant caller is declining.

It looks like you passed the Linux Travis tests, but not the Mac Travis tests. I wouldn't worry about this. The Mac that Travis lets us use is super slow and shitty and sometimes it times out. Since you're developing on Mac and your integration tests are passing, I'm not too concerned.

The Jenkins tests all timed out. I'm not sure why that would have happened, but it's something to look into. It's possible that it's a bug in our cloud system, so I'm going to restart the tests again.

jeizenga:
Jenkins, test this please.

src/xg.hpp
```diff
@@ -630,7 +630,7 @@ class XG : public HandleGraph {
     ////////////////////////////////////////////////////////////////////////////
 
     // maintain old ids from input, ranked as in s_iv and s_bv
-    int_vector<> i_iv;
+    //int_vector<> i_iv;
```
jeizenga: Better to clean this up

src/xg.cpp
```diff
@@ -1016,7 +951,7 @@ void XG::build(vector<pair<id_t, string> >& node_label,
     cerr << "|g_bv| = " << size_in_mega_bytes(g_bv) << endl;
     cerr << "|s_iv| = " << size_in_mega_bytes(s_iv) << endl;
 
-    cerr << "|i_iv| = " << size_in_mega_bytes(i_iv) << endl;
+    //cerr << "|i_iv| = " << size_in_mega_bytes(i_iv) << endl;
```
jeizenga: These lines could just be deleted

src/xg.cpp
```diff
@@ -1047,8 +982,10 @@ void XG::build(vector<pair<id_t, string> >& node_label,
     cerr << "total size [MB] = " << (
         size_in_mega_bytes(s_iv)
         //+ size_in_mega_bytes(s_bv)
-        + size_in_mega_bytes(i_iv)
+
+        //+ size_in_mega_bytes(i_iv)
```
jeizenga: These can be cleaned up too. Also, not that this is your code, but isn't s_bv still part of the XG? Seems to me that it shouldn't be commented out.

src/xg.cpp
```diff
@@ -1585,11 +1416,19 @@ int64_t XG::rank_to_id(size_t rank) const {
...
         assert(false);
     }
-    if(rank > i_iv.size()) {
-        cerr << "[xg] error: Request for id of rank " << rank << "/" << i_iv.size() << endl;
+    if(rank > node_count) {
+        cerr << "[xg] error: Request for id of rank " << rank << "/" << node_count << endl;
         assert(false);
     }
-    return i_iv[rank-1];
+    int i = 0;
```
jeizenga: I think this is the source of the time outs. Won't this require you to iterate over the entire vector to get IDs at the end of it? That will give this algorithm linear time performance. Using the rank and select operations on g_bv, there should be a way to do it in constant time.

ekg:
Is it possible that this is slowing down access to xg? It's hard to draw
many conclusions from the test time on Jenkins because we share the server
AFAIK. But, it does look like this is a few thousand seconds slower than
previous.

ekg:
Sorry, disregard this. I see that the test times have arrived here in other
PRs!

I had some time yesterday to read through the diff and aside from the O(N)
iteration you fixed in the last commit it looks good!

PR #1627, opened 2018-04-13
Sketch out an annotation API over structs

adamnovak:
Here's a follow-on to #1616.

We can put annotations in Structs in the alignments, and read/write them with some templates:

```
set_annotation(&aln, "exploding", true);
get_annotation<double>(aln, "dogs") == 1000.0;
```

The JSON looks like:

```
{"annotation":{"cats":null,"dogs":1000,"exploding":true}}
```

Annotations are identified by strings, which isn't as safe (because you can typo them), but which may be sufficiently good.

If we like this I can extend the implementation to storing lists of things and entire protobuf messages (so we can annotate with Paths and so on), and I can start swapping out this API for some of the lesser-used annotation fields.

One caveat is that a Struct can't hold an int64 as a number; we'll get strings for node IDs inside Paths, just like in JSON. But converting back to Protobuf Path objects ought to take care of that.

### DIFF
--- a/src/annotation.hpp
+++ b/src/annotation.hpp
@@ -150,7 +150,10 @@ void set_annotation(Annotated* annotated, const string& name, const AnnotationTy
 
 template<typename Annotated>
 void clear_annotation(Annotated* annotated, const string& name) {
-    Annotation<Annotated>::clear(annotated, name);
+    // Get ahold of the struct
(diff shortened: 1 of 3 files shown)


adamnovak:
I'm going to merge this in. It's not fully implemented, but this is the design I want to go with, and it has some tests for what is there.

PR #1628, opened 2018-04-16
Path names

ekg:
This resurrects `vg paths` and repurposes it for interacting with the thread and path sets of graphs and their xg/GBWT indexes. My goal here is to make the diploid simulation experiments reproducible. @adamnovak should I be simulating in a different way that's simpler to only simulate reads from the embedded threads or paths in the graph?


ekg:
Jenkins, test this please.

ekg:
Merging because Jenkins appears to be non-responsive.

adamnovak:
Right now the way we do it is we make two single-haplotype subset graphs for our test sample, [using vg find](https://github.com/vgteam/toil-vg/blob/master/src/toil_vg/vg_construct.py#L739), and simulate from them with vg sim. This is currently pretty awkward; if we could do it in a simpler way that would be good.

PR #1638, opened 2018-04-19
Use distinct-endpoint tracebacks to examine positional uncertainty in multipath alignments

### LINKED ISSUE #1595: mpmap performance regression
Since about the 22nd, my GBWT tests have started returning more mismapped reads, and more mismapped MAPQ 60 reads, in the mpmap positive control.

It's not snarl cutting or VCF filtering, since the positive control isn't using those. Something else in the mapping code has changed in a deleterious way. I need to find and fix it.

adamnovak:
This is another attempt to fix #1595, this time by looking inside multipath mappings to see if we can get nonoverlapping linear alignments out of them.

I don't want to merge it until I've confirmed it works in my tests, but I want to throw it at Jenkins.


adamnovak:
Here's the performance on my 10m read pair test:


Condition | Wrong reads total | At MAPQ 60 | At MAPQ 0 | At MAPQ >0 | Avg. Correct MAPQ | Correct MAPQ 0
-- | -- | -- | -- | -- | -- | --
snp1kg-mp-gbwt-traceback-snarlcut-unfiltered | 1853 | 60 | 971 | 882 | 59.8339 | 1483


I am going to merge this over Jenkins' objections about a couple variants comming out worse, because I am cutting out a bunch of MAPQ 60 mismappings. Not as many as are cut out by using the wrong graph in the reachability edge reconstruction, but enough that I think it is worth it to merge this, sync up with master, and then start addressing individual reads on their own instead of as a diff from the previous, wrong behavior.

Issue #1642, opened 2018-04-23
XG conversion code from version 7 to version 8 crashes


jeizenga:
I tried to load a version 7 XG I had laying around, but it crashed. I have replicated the bug using an XG I created from commit f8be42c3e84be489e6a1df58f9054d148b2186e4, which is right before the version 8 branch got merged in. It complains about a bad_alloc, so that could be a lot of things.

To make the version 7 XG, I did the command
```
vg index -x snp1kg-BRCA2_7.xg snp1kg-BRCA2.vg
```
And then to crash it with current master I did
```
vg xg -i snp1kg-BRCA2_7.xg -o snp1kg-BRCA2_8.xg
```

The error text is
```
vg(93271,0x7fff7aac7000) malloc: *** mach_vm_map(size=17591840239616) failed (error code=3)
*** error: can't allocate region
*** set a breakpoint in malloc_error_break to debug
terminate called after throwing an instance of 'xg::XGFormatError'
  what():  XG input data not in XG version 8 format (std::bad_alloc)
ERROR: Signal 6 occurred. VG has crashed. Run 'vg bugs --new' to report a bug.
Stack trace path: /tmp/vg_crash_pmYjxf/stacktrace.txt
```

I'm on macOS, so the stactrace is all mangled and unhelpful. The data is all here:
https://www.dropbox.com/s/xml344clsvzzaep/snp1kg-BRCA2_7.xg?dl=0

@emikoifish I think this'll go to you

jeizenga:
Resolved in https://github.com/vgteam/vg/pull/1643

PR #1643, opened 2018-04-24
increasing the version number seems to fix issue #1642

emikoifish:
Conversion code for xg changed from version 8 to 9 so that i_iv is not loaded in version 9

### DIFF
--- a/src/xg.cpp
+++ b/src/xg.cpp
@@ -145,11 +145,12 @@ void XG::load(istream& in) {
         case 5: // Fall through
         case 6:
         case 7:
+        case 8:
             cerr << "warning:[XG] Loading an out-of-date XG format. In-memory conversion between versions can be time-consuming. "
                  << "For better performance over repeated loads, consider recreating this XG with 'vg index' "
                  << "or upgrading it with 'vg xg'." << endl;
             // Fall through
-        case 8:
+        case 9:
             {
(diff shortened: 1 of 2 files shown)


jeizenga:
Do you have a clear idea why that would be? It might work as a short-term fix just so I can keep development moving, but I'd also like to make sure this isn't a case of one bug hiding another.

One test that might be nice to verify that this is correct would be to try updating an even earlier XG. For instance, you could try indexing that .vg file using 3ed68800b6b65f01c255797c75f7f809551edcd6, which was the previous time we merged a new XG version into master. You should be able to build with the old commit and then update with master like I did in https://github.com/vgteam/vg/issues/1642.

emikoifish:
The xg version was increased to version 8 at master before my pull request was merged. So, the file in commit f8be42c had xg version 8, and the i_iv was skipped even though that file still had the i_iv in it. 
I will confirm that it works with 3ed6880.

jeizenga:
Ah gotcha, that makes sense. Cool. Well, it's still probably a good test to do, but that reasoning seems solid to me. Do it if you have time, but I'll merge this either way if it passes the tests.

jeizenga:
Confirmed: works on my end too. I guess if there's a lesson here it's that it's always smart to merge with master and re-run the tests you did locally before you PR. Anyway, nice work, and thanks for getting to this so quickly!

PR #1656, opened 2018-05-03
Change default behavior for snarl filtering

jeizenga:
I found some default I didn't like and some weird selective filtering when I was addressing another issue in the `vg snarls` command.


jeizenga:
@adamnovak @glennhickey @cmarkello This PR will switch the default behavior in `vg snarls` to filter trivial snarls. I've been told that there might be pipelines out there (maybe in toil-vg?) that use the snarls interface. If so, I should update those accordingly alongside this PR. Do you guys know of anywhere I should be looking?

adamnovak:
For toil-vg, you probably want to touch this line, where --filter-trivial
is passed:

https://github.com/vgteam/toil-vg/blob/f94113eda1cfbf2a6ccc552b5eee59193416ee42/src/toil_vg/vg_index.py#L495

-Adam

PR #1663, opened 2018-05-08
Allow plots to be in a plots directory for the CI

adamnovak:


adamnovak:
Jenkins looks to be picking up the plots in their current locations just fine. Not sure how it will like the new locations yet, but it *should* work, so I will merge it and fix it in the PR upgrading toil-vg if it doesn't.

Issue #1658, opened 2018-05-06
raptor build fails with missing yycleanup on ubuntu 18.04


ekg:
It seems that yycleanup no longer exists in the library distributed with ubuntu 18.04. I have a hack to fix this, which seems to work for me, which involves just removing the reference to yycleanup.

```
[ 63%] Linking C executable raptor_sort_r_test
libraptor2.a(turtle_lexer.c.o): In function `turtle_lexer_lex_destroy':
turtle_lexer.c:(.text+0x3850): undefined reference to `yycleanup'
collect2: error: ld returned 1 exit status
src/CMakeFiles/raptor_sort_r_test.dir/build.make:98: recipe for target 'src/raptor_sort_r_test' failed
make[2]: *** [src/raptor_sort_r_test] Error 1
CMakeFiles/Makefile2:131: recipe for target 'src/CMakeFiles/raptor_sort_r_test.dir/all' failed
make[1]: *** [src/CMakeFiles/raptor_sort_r_test.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
```

glennhickey:
https://github.com/vgteam/vg/commit/7a294c105cc69d3cb4f921c777985d447d0e339e was a bit of a build breaker, and seemed to cause errors on anything building a fresh clone of vg from scratch (including jenkins, vg_docker, and an ubuntu 16.04 and 18.04 system I tried).   I've reverted it and pushed to master and now jenkins and vg_docker are happy again.  Jenkins will now send messages on build problems, which should help finding another patch for this issue...  I didn't have much luck trying @JervenBolleman's apt raptor patch due to a bunch of linker errors, but there maybe a simpe fix there that I missed

PR #1684, opened 2018-05-18
Switch to apt raptor

glennhickey:


glennhickey:
I think apt raptor is a lost cause as far as static linking goes.

adamnovak:
The alternative is to fix #1658.

I tried to resurrect this PR, but I've eventually run into a problem I don't really think I can solve.

libraptor2-dev ships a libraptor.a that needs libcurl.a. In turn, libcurl.a pulls in a bunch more libraries, because curl is a bit of a do-everything downloader. One of the things Ubuntu's libcurl.a is built against is krb5 Kerberos. [That Kerberos implementation is built around a plugin system that loads shared libraries dynamically at runtime, and can't be built statically.](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=439039) No static library for it is available in Ubuntu.

So you can link statically against libcurl.a on Ubuntu, but you can't produce a fully static binary linked against Ubuntu's libcurl, because you still need to link dynamically against Kerberos. In theory we could statically link against everything *except* the Kerberos libraries, and dynamically link those at runtime, but explaining that to the linker is a pain, especially when we're trying to tell it about dependencies 3 levels down the transitive dependency tree which aren't even required when building on some systems.

PR #1685, opened 2018-05-18
Sv del off by one

apregier:
I am honestly not at all confident in this code change, but it fixed my test case and didn't seem to break any other tests...careful review is appreciated!  Also I have not tried other SV types.


apregier:
Oops, I included some commits in here unintentionally.  I'll make a clean one.

PR #1686, opened 2018-05-18
Fix an off-by one error for deletion end

apregier:
I am not at all confident in this code change - but it fixed my test case and didn't break any other tests...careful review is appreciated!  Also note I haven't tested any other sv types, just DEL.


edawson:
I think this should work, but vcflib might still choke. I have a PR in to fix that but I've been distracted by other things; I'm hoping to merge in a bunch of SV-related changes in the next week. I'm super appreciative of the test case. Thanks for the patch!

ekg:
Awesome! Thank you!

PR #1687, opened 2018-05-18
Use a heuristic instead of all-against-all tip distance finding

### LINKED ISSUE #1653: Vg snarls does not finish in 5 days
Here is the graph:

https://transfer.sh/2jdnK/graph.bclam.bluntified.vg

I am running vg from https://github.com/vgteam/vg/releases/tag/v1.6.0

I am using the following command:

./vg snarls -l -t -r graph.bclam.bluntified.trans graph.bclam.bluntified.vg > graph.bclam.bluntified.snarls

It didn't get any output.

Looks there is a bug. How to fix it?

adamnovak:
Should fix #1653

### DIFF
--- a/src/cactus.cpp
+++ b/src/cactus.cpp
@@ -451,11 +451,20 @@ pair<stCactusGraph*, stList*> vg_to_cactus(VG& graph, const unordered_set<string
             
         }
         
-        // Otherwise, compute tip reachability and distance by Dijkstra's
-        // Algorithm and pick the pair of reachable tips with the longest shortest
-        // paths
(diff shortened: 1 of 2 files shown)


adamnovak:
`vg snarls` on @shilpagarg's graph is still pretty slow with this (I haven't waited around for it to finish), but it gets through Cactus graph rooting and moves on to actually looking for snarls.

If the snarls aren't that great, we can revisit the rooting tip finding and try to improve the heuristics.

adamnovak:
Looks like there was a timeout downloading something.

Jenkins, test this please.

PR #1692, opened 2018-05-21
New stats format

adamnovak:
Allow plotting scripts to take a "count" column, so we can condense multiple correct/wrong reads into one line in the file.


adamnovak:
Mac Travis is failing because libvw isn't building right. I definitely didn't touch it. I will rerun the build to see if maybe we get a better error?

adamnovak:
I think it was just cache-filling timeouts. Mashing rerun seems to have fixed everything.

PR #1696, opened 2018-05-22
Fix our Raptor on Ubuntu 18.04.

### LINKED ISSUE #1658: raptor build fails with missing yycleanup on ubuntu 18.04

adamnovak:
This reverts commit f1f10e1613fcde9dabdb169c4bcb818227cdeb71.

This ought to fix #1658 and I have successfully built it on 18.04 (from a dirty build). I am not sure what caused @glennhickey trouble but maybe the tests will reveal it.


adamnovak:
OK, this should work now, even from a clean build.

The underlying problem was that the file we were trying to modify is dynamically generated, and we were modifying it by copying it, changing it, and putting it back, *and* we were doing this before Raptor had a chance to do any building.

On a fresh build, the file didn't exist yet, so were cating a nonexistent file, and replacing it with the modified, still-empty file, and then raptor wasn't actually generating the file (since we made it and it was new enough to fool make into thinking it didn't need to be built.

Now I am making sure that the file we want to modify is built before we try to modify it.

PR #1697, opened 2018-05-22
Upgrade toil-vg to the one that does condensed stats

adamnovak:


adamnovak:
It looks like two of the tests time out on this new toil-vg, and one crashes because of nonsense data in the .tsv.gz files that RTG plots.

@glennhickey Do you have any insight into the RTG plotting problem? It looks like you touched that most recently.

I'm not sure what to do about the timeouts. Those conditions are throwing a bunch of new warnings about missing alt paths when constructing a haplotype index, so I think the problem is related to this:

https://github.com/vgteam/toil-vg/blob/1f2d99243b1de3d8b6b9ddbbb48ab855213ebdd3/src/toil_vg/vg_toil.py#L191-L217

We're collecting a bunch of phasing VCFs and then discarding them and just passing generic input VCFs as the phasing VCF. And earlier, the baseline truth VCF got added as a generic VCF. So we're trying to build a haplotype index from the truth VCF, which wasn't the VCF used to make the graph, and it is failing.

I don't know why we weren't seeing this problem before, because it looks like it predates  but I'm pretty sure it is a problem.

glennhickey:
@adamnovak  I'll take a look.  I'm sure I've made refactors to toil-vg in
the past months that need compensatory changes in vgci.py.

The link you sent in vg_toil.py is definitely a bug, but I don't think
vgci.py is doing any haplotype indexing via toil-vg run.

glennhickey:
#1702 (now in master) switched to the latest vg.  So I think this one can be closed...

PR #1698, opened 2018-05-23
index GAMs with GBWT

ekg:
This enables GAM input to GBWT. It is also possible to query by read name. The non-graph edits are not handled, this is just for the node traversals as per GBWT's format.


ekg:
OSX build timeout on travis. Merging ahead.

PR #1699, opened 2018-05-24
add libffi-dev to dependencies on linux

ekg:


ekg:
I see that @jeizenga already did this and it's been through the tests.

PR #1700, opened 2018-05-24
RNAseq: Enable the construction and use of a transcript model graph

ekg:


ekg:
This works. Here, for 100k RNAseq reads from S288C, against a gene model graph (x=.vg) or the linear reference (y=id.ref).

image

Yeast has very few introns, so we don't expect much of an affect from adding in the gene model. However, there are a significant subset of the RNAseq reads that appear to overlap a splice site, and we see them represented on the right hand border of the plot.

ekg:
Here are the steps I took to build the transcript model graph:

```
# download and unpack the reference and gene model
wget https://downloads.yeastgenome.org/sequence/S288C_reference/genome_releases/S288C_reference_genome_R64-2-1_20150113.tgz
tar xzvf S288C_reference_genome_R64-2-1_20150113.tgz
cd S288C_reference_genome_R64-2-1_20150113

# build the base graph
vg construct -r S288C_reference_sequence_R64-2-1_20150113.fasta -m 32 -p >S288C.vg

# build the .xg index we'll need to project the .gff into the graph
vg index -x S288C.xg S288C.vg

# convert the .gff to .gam
vg annotate -x S288C.xg -f <(awk '$3 == "CDS" { print }' saccharomyces_cerevisiae_R64-2-1_20150113.gff) -p >saccharomyces_cerevisiae_R64-2-1_20150113.gff.CDS.gam

# augment the graph with the transcript model
vg mod -i saccharomyces_cerevisiae_R64-2-1_20150113.gff.CDS.gam S288C.vg >S288C+CDS.vg

# at this point you index the graph as normal
```

The last step takes about an hour, which is not ideal. Also, it makes a big graph... 30x bigger. I assume this is because of path storage, but I haven't confirmed. There is an alternative arrangement we can do here and put the paths for the transcripts into GBWT, but it does help to keep them in the graph as they can drive correct positional inference there.

Looking at the previous plot I wonder if I have an off by one error in the .gff conversion :confused:

PR #1712, opened 2018-06-01
Use a Raptor compatible with Bison 3.0.5 which Mac Travis picked up

### LINKED ISSUE #1705: Mac and raptor still can't play nicely
1. What you were trying to do
Build on Mac OS / Travis.

2. What you wanted to happen
A clean recursive clone / push to a PR builds a working vg executable.

3. What actually happened
I get the dreaded "yycleanup not found" error (#1658 and elsewhere). What have we done / can we do to prevent this? I end up patching the raptor install with Adam's lexer.zip fix every time I do a fresh clone.

4. What data and command line to use to make the problem recur, if applicable
Both of these fail on Mac OS. They work on Linux now.
`git clone --recursive https://github.com/vgteam/vg && cd vg && make -j 4`
`git clone --recursive https://github.com/vgteam/vg && cd vg && git checkout safe-svs && git submodule update --init --recursive && make -j 4`

This is the last block to the `safe-svs` pull request. It passes all the tests on Linux and Jenkins.

adamnovak:
This ought to fix #1705.

It turns out Bison cut a new release, Homebrew automatically started installing it, and it confused the Perl scripts that Raptor uses to munge the Bison generated code, because it fixed a bug that the code munging was trying to work around.


edawson:
I see. MacPorts has also updated but hasn't forced me to upgrade yet. Good catch.

I'm going to merge this since it seems innocuous and passes tests.

PR #1734, opened 2018-06-13
tweaks that help when building graphs from SV VCFs

ekg:


ekg:
I'm using this script to build a single VCF file from the HGSVC haplotypes.

First I convert each pair of haplotypes into a diploid phased VCF:

```
ls *hap0.vcf.gz | cut -f 1 -d\. | while read base; do echo $base; time vcfcombine <(zcat $base.hap0.vcf.gz | sed s/$base/$base.a/ | sed s%\\./\\.%1% | vcf-sort -c ) <(zcat $base.hap1.vcf.gz | sed s/$base/$base.b/ | sed s%\\./\\.%1% | vcf-sort -c) | vcf-sort -c | vcfnull2ref - | awk 'BEGIN { OFS="\t" } /#CHROM/ { $11="" } /#/ { print } !/#/ { $10=$10"|"$11; $11=""; print }' | sed s/$base.a/$base/g | cut -f 10 | bgziptabix $base.haps.vcf.gz; done
```

Then I merge these together into one final VCF:

```
vcfcombine HG00514.haps.vcf.gz HG005733.haps.vcf.gz NA19240.haps.vcf.gz | awk 'BEGIN { OFS = "\t"} $10 == "." { $10 = "0|0" } $11 == "." { $11 = "0|0" } $12 == "." { $12 = "0|0" } { print }' | bgziptabix HGSVC.haps.vcf.gz
```

ekg:
@edawson we have to work around some problem with SV-based construction by removing tags in the VCF file that indicate an allele is an SV. Then things work:

```
vcfkeepinfo HGSVC.haps.vcf.gz NA | vcffixup - | bgziptabix HGSVC.haps.noinfo.vcf.gz
```

ekg:
@edawson and I are talking about how to deal with long variants in the input.

We're using vcflib to parse the allelic primitives out of the VCF file. This is fine for short variants but falls down when we have big variants because of the quadratic cost of the simplistic smith waterman alignment that vcflib does to find the alleles.

The solution to this is the `vg construct --flat-alts` flag. This prevents us from running out of memory while parsing the variant into its components. However, it also leaves us with weird 1bp nodes for all insertions and deletions (the REF or ALT allele will always be at least 1bp in VCF).

One hack is to check if this base is the same as the reference, and if so parse it out of the variant. I'm inclined to follow this pattern for any allele above a certain length. Then we can avoid having users decide if they need --flat-alts or not.

PR #1739, opened 2018-06-14
Change the way vg handles structural variants

edawson:
We've run into a bunch of SV handling issues now that we're trying real data; these changes should help fix most of these and further harden the code to malformed SVs.
- [x] When clumping variants, use the right get_bounds function (i.e. one that protects against SVLEN/END errors and which ignores SVs when desired).
- [x] Default to using the REF/ALT fields (over SVTYPE/SVLEN/END tags)
    - [x] When #1734 gets merged in, call `allATGCN` rather than `allATGC`
    - [x] Remove the is_sv() checks from the code path when a variant contains only [ATGCN]*
- [x] Force vg to only handle symbolic SVs when `-S` is passed.
    - [x] call canonicalize_sv() only when a variant has non-ATGCN alt alleles
- [x] If a variant's END comes before its position, fail gracefully and warn the user.
- [x] If the variant is an insertion, find its sequence in this order:
    - [X] Check the ALT fields for ATGCN. If they're valid, non-symbolic, and longer than 1 N, use that.
    - [X] Check for a SEQ tag. If one is present, non-symbolic, and allATGCN, use that.
    - [x] Check if there is an insertions fasta file and corresponding fasta index. Check if the alt allele(s) match  '<[A-Za-z0-9]*>'. If any do, see if they exist in the fasta index, and if they're found, replace the alt allele tag with the sequence in the fasta.
- [x] Ensure the ordering of where to pull variants from is tested.
- [x] Verify deletion test cases, which look to be off by one.
- [X] Add tests for INS types.
- [x] Add tests for INV types.
- [X] Fix quadratic GFA normalization on input.


edawson:
@adamnovak I'd like to go ahead and merge this, but am way of the Jenkins oddities going on. I assume the comment is more up-to-date than the tests widget, and that the initial test run just got lost in the cloud?

adamnovak:
It looks like in this case the tests themselves passed, but the code that uploads the final build status ran out of memory. We can probably merge it, unless vg somehow caused that to happen.

PR #1743, opened 2018-06-15
Shuffle alignment hits before sorts

### LINKED ISSUE #1725: vg map reverse strand preference - clarification request
1. What you were trying to do

I am simulating error free reads using vg sim from forward and reverse strand (without the -f option), converting them into a fastq file and mapping them back to the same reference using vg and bwa. Afterwards I compare the mapping positions to the positions the reads where simulated from.

2. What you wanted to happen

I want to have mostly similar mapping positions for bwa and vg map

3. What actually happened

The reads mapped using vg map almost always map the rerverse strand if there is a reverse strand representation. Which in most cases is the position the read was simulated from. bwa mem reports a equally good forward strand position for those reads, which don't overlap with the positions they where simulated from. 

vg seems to prefer revese strand records whereas bwa reports the forward strand. Is this a wanted behavior? It can't be the solution to simulate reads from the forward strand only to obtain comparable results for bwa and vg on the same variant free reference sequence.

adamnovak:
This will hopefully fix #1725 by deterministically (based on read sequence) randomizing the order of equivalent alignments, so we don't always pick e.g. one contig or earlier positions or one strand over any other when equivalent mappings exist.

### DIFF
--- a/jenkins/vgci.py
+++ b/jenkins/vgci.py
@@ -1473,7 +1473,7 @@ class VGCITest(TestCase):
         log.info("Test start at {}".format(datetime.now()))
         self._test_bakeoff('MHC', 'primary', True, tag_ext='-genotype', genotype=True)
         
-    @timeout_decorator.timeout(1600)        
+    @timeout_decorator.timeout(1800)        
     def test_map_mhc_snp1kg_genotype(self):
         """ Mapping and calling (with vg genotype) bakeoff F1 test for MHC snp1kg graph 
         """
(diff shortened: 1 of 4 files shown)


adamnovak:
At the meeting today, we decided we ought to take a different approach, since this PR seems to slow things down. Rather than shuffling all the mappings, we only want to shuffle the mappings tied for first place, and we want to do it after the sort.

So we need to make the shuffle know a bit about the sorting criteria in each case, so we can find the top n tied alignments and shuffle them.

PR #1744, opened 2018-06-18
Faster GBWT construction

### LINKED ISSUE #1703: Faster GBWT construction
GBWT construction is unnecessary slow. There are two main bottlenecks:

1. Parsing the entire phased VCF file once for every 200 samples.
2. Finding the path corresponding to each haplotype separately.

I'm going to reimplement `vg index -G` to deal with these issues. The first bottleneck is easy: we can simply parse the VCF file once and write run-length encoded phasing information to temporary files, using a separate file for each batch of 200 samples.

The second issue is a bit more complicated, because I'm not sure about the fine details of VCF files and VG graph construction.

* **Can we ignore sequence offsets?** Let the reference path R[0,n-1] be a sequence of (node, orientation) pairs. Can we assume that for every variant site, there are coordinates (i,j) such that we can simply replace R[i,j] with the path X corresponding to an alternate allele?
* **Can the haplotypes use edges not in the VG graph?** Under some circumstances, the graph does not have an edge from R[i-1] to X[0]. Such situations are currently treated as phase breaks, but we could choose to keep the paths connected instead. Checking for the existence of edges is expensive, and I would like to get rid of both the VG graph and the XG graph after parsing the VCF file.
* **What should we do with phase breaks?** If there is no phasing information at a site, the current approach is to append reference nodes until that site and then start a new path with the reference nodes between the previous site and the current site. This creates a bit of overlap between adjacent paths.
* **What should we do with adjacent phase breaks?** Should there be a separate path for each site without phasing information? Or should we connect the paths randomly?

### LINKED ISSUE #1731: Improve GBWT construction to resolve repeated nodes
@jltsiren is working on an improved GBWT haplotype construction algorithm that doesn't spend so much time parsing and re-parsing the VCF, and which can better handle the case of adjacent variants. This issue tracks that project.

jltsiren:
Parse the VCF file only once when building GBWT for haplotypes. Store the reference/alt paths in a GBWT support structure, write the phasings to temporary files, get rid of unnecessary in-memory structures during construction, and use GBWT structures to generate the haplotypes.

Some notes:

* 1000GP GBWT construction time was reduced from 29 hours to 12.5 hours.
* Adjacent/overlapping variants cause less phase breaks now. The number of paths in the 1000GP GBWT was reduced from 29.3 million to 18.8 million. Without any phase breaks, there would be around 0.25 million paths.
* Memory usage is a bit lower now. We can run 14 parallel jobs instead of 12 when building the 1000GP GBWT. Overall construction time is now determined by chr2.
* The temporary files use roughly the same amount of space as the `.vcf.gz` file. 
* GBWT now uses `gbwt::vector_type` (`std::vector<std::uint32_t>`) instead of `std::vector<gbwt::node_type>` (`std::vector<std::uint64_t>`).
* Option `-P` of `vg index` replaces unphased genotypes with randomly phased ones.

Resolves #1703. Resolves #1731.


jltsiren:
It looks like gPBWT construction is failing in some Jenkins tests, because the threads use edges that do not exist in the XG index. I guess this happens when we move from an alt path to another alt path without visiting a reference node.

glennhickey:
The Jenkins tests no longer construct GPBWT's.  Rebasing from master will hopefully fix any problems they're giving you.

jltsiren:
Mac tests time out again, this time at `t/27_vg_genotype.t`. When I run it locally, test 4 sometimes freezes. When I run the relevant commands:
```
vg construct -v tiny/tiny.vcf.gz -r tiny/tiny.fa > tiny.vg
vg index -x tiny.vg.xg -g tiny.vg.gcsa -k 16 tiny.vg
printf "" > tiny.gam
vg index -d tiny.gam.index -N tiny.gam
vg genotype tiny.vg tiny.gam.index -vSp
```
the last command either freezes or crashes.

In either case, we get various stack traces. These include:

Crash:
```
Caught signal 11 raised at address 0x1067f31ea

Next ip: 0x1067f31ea Next bp: 0x108160f00
Address 0x1067f31ea in demangled symbol vg::Snarl::Snarl(vg::Snarl const&) at offset 0x4a, in library /Users/jltsiren/bin/vg at offset 0x6b91ea

Next ip: 0 Next bp: 0x108160f10
Stack leaves code at ip=0
Stack trace complete
```

Freeze:
```
Caught signal 11 raised at address 0x10e0ef103

Next ip: 0x10e0ef103 Next bp: 0x10e833a500000
Address 0x10e0ef103 in demangled symbol vg::Visit::Visit(vg::Visit const&) at offset 0x23, in library /Users/jltsiren/bin/vg at offset 0x6b9103
```

Freeze:
```
Caught signal 11 raised at address 0x10a0ff536

Next ip: 0x10a0ff536 Next bp: 0x7fc56cc0b578
Address 0x10a0ff536 in demangled symbol google::protobuf::UnknownFieldSet::MergeFrom(google::protobuf::UnknownFieldSet const&) at offset 0x16, in library /Users/jltsiren/work/vg/lib/libprotobuf.14.dylib at offset 0xcc536

Next ip: 0x10a004441 Next bp: 0x2
Address 0x10a004441 in mangled symbol count at offset 0xab5ebd, in library /Users/jltsiren/bin/vg at offset 0x1dfd441
```

Freeze:
```
Caught signal 11 raised at address 0x1046931c3

Next ip: 0x1046931c3 Next bp: 0xd000000000000002
Address 0x1046931c3 in demangled symbol vg::Snarl::Snarl(vg::Snarl const&) at offset 0x23, in library /Users/jltsiren/bin/vg at offset 0x6b91c3
```

Crash:
```
Caught signal 11 raised at address 0x7fff4febff49

Next ip: 0x7fff4febff49 Next bp: 0x7ffee784bfd0
Address 0x7fff4febff49 in mangled symbol _platform_memmove$VARIANT$Haswell at offset 0x29, in library /usr/lib/system/libsystem_platform.dylib at offset 0x4f49

Next ip: 0x1083b296a Next bp: 0x7fc607e4ce00
Address 0x1083b296a in demangled symbol google::protobuf::internal::ArenaStringPtr::CreateInstanceNoArena(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const*) at offset 0xba, in library /Users/jltsiren/bin/vg at offset 0x296a

Next ip: 0x300007fc607e4c41 Next bp: 0x112301000
Stack leaves code at ip=0x300007fc607e4c41
Stack trace complete
```

PR #1745, opened 2018-06-19
Feature/emily to from

emikoifish:
Changed edge encoding from to/from to start/end.
- Edge types are now 0 (go to start side of node) and 1 (go to end side of node).
- Build and load turn old type encoding into new type encoding.
- Doubly reversing edges are only stored once.
- Removed some functions that used old edge type encoding (edge_type, edge_filter).


src/xg.cpp
```diff
@@ -258,6 +272,163 @@ void XG::load(istream& in) {
     }
 
 }
+    
+void XG::convert_old_edge_to_new(int_vector<> g_iv_old, bit_vector g_bv_old, rank_support_v<1> g_bv_rank_old, bit_vector::select_1_type g_bv_select_old) {
```
jeizenga: These arguments should be passed in by reference to avoid deep copying the vectors

src/xg.cpp
```diff
@@ -258,6 +272,163 @@ void XG::load(istream& in) {
...
+    size_t g_iv_old_size =
+        node_count * G_NODE_HEADER_LENGTH // record headers
+        + edge_count * 2 * G_EDGE_LENGTH; // edges (stored twice)
+
```
jeizenga: Has this conversion code been tested? You would have had to do it yourself, because there are no tests in `test` directory or in `vg test` that do it.

src/xg.cpp
```diff
@@ -258,6 +272,163 @@ void XG::load(istream& in) {
...
+    int64_t header = G_NODE_HEADER_LENGTH;
+    int64_t headerfrom = G_NODE_HEADER_LENGTH;
```
jeizenga: Why create these new header variables when you already have `G_NODE_HEADER_LENGTH`?

src/xg.cpp
```diff
@@ -258,6 +272,163 @@ void XG::load(istream& in) {
...
+    int doubly_reversing_sides = 0;
```
jeizenga: I don't think you're using the "doubly reversing" terminology the same way it's used in the rest of the code base. I believe you're looking for reversing self-edges here, right? Elsewhere, doubly reversing refers to edges that have both `from_start == true` and `to_end == true`.

src/xg.cpp
```diff
@@ -258,6 +272,163 @@ void XG::load(istream& in) {
...
+        int64_t on_node = g_iv_old[i];
+        int64_t g = i;
+        int edges_to_count = g_iv_old[g+G_NODE_TO_COUNT_OFFSET];
+        int edges_from_count = g_iv_old[g+G_NODE_FROM_COUNT_OFFSET];
+        int sequence_size = g_iv_old[g+G_NODE_LENGTH_OFFSET];
+        int64_t t = g + G_NODE_HEADER_LENGTH;
+        int64_t f = g + G_NODE_HEADER_LENGTH + G_EDGE_LENGTH * edges_to_count;
+        // to side
+        for (int64_t j = t; j < f; ) {
+            int64_t length = g_iv_old[j];
```
jeizenga: Not super important, but I'd call this an 'offset' rather than a 'length'.

src/xg.cpp
```diff
@@ -258,6 +272,163 @@ void XG::load(istream& in) {
...
+    // Now find size of the new g_iv and remove doubly reversing sides
+    size_t g_iv_size =
+        node_count * G_NODE_HEADER_LENGTH // record headers
+        + edge_count * 2 * G_EDGE_LENGTH - doubly_reversing_sides/2 * G_EDGE_LENGTH; // edges (stored twice)
+    int z = 0;
+    util::assign(g_iv, int_vector<>(g_iv_size));
+    util::assign(g_bv, bit_vector(g_iv_size));
+    int ss_length_change = 0;
+    for (int64_t i = 0; i < g_iv_old_size; i += header) {
+        g_bv[z] = 1;
```
jeizenga: I'd prefer comments saying what you're copying over here

src/xg.cpp
```diff
@@ -258,6 +272,163 @@ void XG::load(istream& in) {
...
+        g_iv[z++] = g_iv_old[i];
+        g_iv[z++] = g_iv_old[i+1];
+        g_iv[z++] = g_iv_old[i+2];
```
jeizenga: I'd prefer comments saying what you're copying over in this block

src/xg.cpp
```diff
@@ -258,6 +272,163 @@ void XG::load(istream& in) {
...
+                }
+                else{
+                    temp.push_back(1);
+                    end_side.push_back(temp);
+                }
+            }
+            else if (edge_type == 3){
+                if (g_iv_old[g] == g_iv_old[g+length]){
+                    ss_length_change ++;
+                    cerr << (int64_t)g_iv_old[g]<< endl;
```
jeizenga: This is debugging detritus

src/xg.cpp
```diff
@@ -773,29 +977,38 @@ void XG::build(vector<pair<id_t, string> >& node_label,
...
         g_iv[to_edge_count_idx] = to_edge_count;
         for (auto end : { false, true }) {
-            auto& from_sides = from_to[make_side(n.id(), end)];
-            for (auto& e : from_sides) {
+            auto& e_sides = end_side[make_side(n.id(), end)];
+            for (auto& e : e_sides) {
                 g_iv[g++] = side_id(e);
-                g_iv[g++] = edge_type(end, side_is_end(e));
+                if (e<0){
+                    g_iv[g++] = 0;
```
jeizenga: Rather than having these 0/1 constants floating around and counting on future coders to just understand their semantics, it would be better to define the edge types in the header file. You could add something along lines of 
```
const static int64_t G_TO_START_EDGE_TYPE = 0;
const static int64_t G_TO_END_EDGE_TYPE = 1;
```
into the `xg::XG` namespace. You can look at the `G_EDGE_LENGTH`, etc. constants as an example.

src/xg.cpp
```diff
@@ -1718,12 +1874,13 @@ bool XG::follow_edges(const handle_t& handle, bool go_left, const function<bool(
     size_t to_start = g + G_NODE_HEADER_LENGTH;
     size_t from_start = g + G_NODE_HEADER_LENGTH + G_EDGE_LENGTH * edges_to_count;
     
-    // We will look for all the edges on the appropriate side, which means we have to check the from and to edges
-    if (do_edges(g, to_start, edges_to_count, true, go_left, is_reverse, iteratee)) {
-        // All the edges where we're to were accepted, so do the edges where we're from
+    if (!is_reverse == go_left || is_reverse == !go_left) {
```
jeizenga: These two boolean statements are equivalent

src/xg.cpp
```diff
@@ -1718,12 +1874,13 @@ bool XG::follow_edges(const handle_t& handle, bool go_left, const function<bool(
     size_t to_start = g + G_NODE_HEADER_LENGTH;
```
jeizenga: You still have the to/from terminology here. It should be changed to start/end.

src/xg.cpp
```diff
@@ -1801,7 +1958,7 @@ vector<Edge> XG::edges_to(int64_t id) const {
     for (int64_t j = t; j < f; ) {
```
jeizenga: This whole function (as well as `edges_from`) should be deleted and replaced with something along the lines of `edges_on_start` and `edges_on_end`.

src/xg.cpp
```diff
@@ -2221,7 +2379,23 @@ void XG::expand_context_by_steps(Graph& g, size_t steps, bool add_paths,
...
-        }       
+        }
+        for (auto& edge : edges_to(n)) {
+            if(last_step_nodes.count(edge.from())) {
+                // This edge connects two nodes that were added on the last
+                // step, and so wouldn't have been found by the main loop.
+                
+                // Determine if it's been seen already (somehow).
+                // TODO: it probably shouldn't have been, unless it's a self loop or something.
+                auto sides = make_pair(make_side(edge.from(), edge.from_start()),
```
jeizenga: First, the first invocation of `make_side` looks wrong to me. I think it should be `!edge.from_start()`.

Second, this isn't a canonical representation. The edge could equally validly be represented as 
```
make_pair(edge.to(), edge.to_end())
          edge.from(), !edge.from_start()))
```
That means you can't count on the `edges` set to identify duplicates, because you might have represented the same edge in two different ways (so the `==` operator would return false). You need to break the symmetry and add a method to always choose the same canonical way of representing an edge. Typically, this involves choosing the representation that's lowest among some sort order. For instance, so that the lower node ID is always the first one in the pair (but be careful about reversing and non-reversing self-edges).

Issue #1647, opened 2018-04-26
vg construction, Segmentation fault 


evanbiederstedt:
It's possible this is an older version of vg---I'll update with the new installation which I'm trying

```
$ vg construct -r hg38.fa -v myfile.vcf.gz  > output.vg

Stack trace (most recent call last):
#9    Object "", at 0xffffffffffffffff, in 
#8    Object "/nfs/sw/vg/vg-1.6.1/vg", at 0x6b63ed, in _start
#7    Object "/usr/lib64/libc-2.17.so", at 0x7f4995da0b14, in __libc_start_main
#6    Source "src/main.cpp", line 61, in main [0x5dd264]
#5    Source "src/subcommand/construct_main.cpp", line 287, in main_construct [0xa55e88]
#4    Source "src/constructor.cpp", line 1471, in construct_graph [0xac8fa4]
#3    Source "src/constructor.cpp", line 1306, in construct_graph [0xac813f]
#2    Source "src/constructor.cpp", line 437, in construct_chunk [0xac619a]
#1    Object "/nfs/sw/vg/vg-1.6.1/vg", at 0xd834b3, in vcflib::Variant::parsedAlternates(bool, bool, bool, float, float, float, float, float, std::string, std::string)
#0    Object "/nfs/sw/vg/vg-1.6.1/vg", at 0xd9407c, in CSmithWatermanGotoh::Align(unsigned int&, std::string&, std::string const&, std::string const&)
Segmentation fault (Address not mapped to object [0x590246000])
Segmentation fault
```

Is this an error seen previously which has been fixed circa March 2018?

adamnovak:
I've never seen this before. It looks like an internal crash in vcflib, when trying to interpret the alts for one of your VCF records (and align them to each other and the reference in a sensible way).

Is your VCF file unusual? Maybe it is doing something with the alts that vcflib can't deal with.

evanbiederstedt:
@adamnovak 

> Is your VCF file unusual? Maybe it is doing something with the alts that vcflib can't deal with.

The ALTs are quite long. I'm filtering the VCF based on line size to check whether this is why vcflib crashes. 

You're happy to have a look of course

evanbiederstedt:
@edawson Thanks for the help. 

     -f, --flat-alts N     don't chop up alternate alleles from input vcf

I'm a bit confused how this flag should work...How do I use this with 

    $ vg construct -r hg38.fa -v myfile.vcf.gz  > output.vg

[...]

PR #1755, opened 2018-06-21
Wrap Graph with VG object before serialization

apregier:
This allows the messages to be split up in a
way that is valid for protobuf (not too large)

This fixed the error that I was seeing in a graph that I had constructed (https://github.com/vgteam/vg/issues/1647), and it doesn't seem _worse_ in terms of memory management than what was previously there - but I might be wrong about that.

### DIFF
--- a/src/subcommand/construct_main.cpp
+++ b/src/subcommand/construct_main.cpp
@@ -198,12 +198,19 @@ int main_construct(int argc, char** argv) {
     auto callback = [&](Graph& big_chunk) {
         // TODO: these chunks may be too big to (de)serialize directly. For now,
         // just serialize them directly anyway.
+        bool sent = false;
+        function<bool(Graph&)> get_graph = [&](Graph& graph){
+            if (sent) {
+                return false;
+            }
+            graph = Graph();
+            graph = big_chunk;
+            sent = true;
+            return true;
+        };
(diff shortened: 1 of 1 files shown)


glennhickey:
Jenkins, OK to test.

apregier:
I'm not sure what happened to the TravisCI tests?  When I looked yesterday it seemed that one had failed because of a timeout but the other has passed.  Now when I came back today they aren't linked anymore?

src/subcommand/construct_main.cpp
```diff
@@ -198,12 +198,19 @@ int main_construct(int argc, char** argv) {
     auto callback = [&](Graph& big_chunk) {
         // TODO: these chunks may be too big to (de)serialize directly. For now,
         // just serialize them directly anyway.
+        bool sent = false;
+        function<bool(Graph&)> get_graph = [&](Graph& graph){
+            if (sent) {
+                return false;
+            }
+            graph = Graph();
```
adamnovak: Why construct an empty Graph here?

adamnovak:
I think we ended up working around #1647 by just making the maximum chunk size very large with https://github.com/vgteam/vg/pull/1718, but this ought to fix it altogether.

adamnovak:
I'm going to merge this, and do some stuff on top of it.

PR #1765, opened 2018-06-26
WIP: GFA import refactor to fix #1607

### LINKED ISSUE #1607: Handling inexact overlaps in string graph
Here is one test example, that we want to fix in terms of handing overlaps. We don't want to skip this edge because the overlap doesn’t represent exact match, and it’s set based on 00195046 (the successor) which has a length of 16669 bp.

https://transfer.sh/tAGhU/testexample.gfa

Also, we want to create bubbles from these inexact matches, which is not handled in the current bluntification implementation.

adamnovak:
This is (finally) the start of a serious attempt to rework GFA I/O.

So far I have just refactored the existing algorithms out of the vg class and into their own functions, and modified them slightly to use more-external vg interfaces. As part of this I have const-ified or provided const versions of some more vg functions.

Still to do is adapting the import algorithm to actually provide the algorithm/semantics I have documented for it.


adamnovak:
This is (almost) ready. I still want to test it on some bigger GFA files, but if anyone wants to review it, go right ahead.

Issue #1775, opened 2018-07-09
VG annotate can't add BED records across the junctions of circular genomes


JervenBolleman:
Please describe:

1. What you were trying to do
vg annotate a 126 pangenome e. coli graph using bed files from EnsemblGeneomes.
2. What you wanted to happen
Have an humongous vg gam file
3. What actually happened
```
vg: src/alignment.cpp:1450: void vg::parse_bed_regions(std::istream&, xg::XG*, std::vector<vg::Alignment>*): Assertion `sbuf < ebuf' failed.
ERROR: Signal 6 occurred. VG has crashed. Run 'vg bugs --new' to report a bug.
```
```
Stack trace (most recent call last):
#9    Object "/scratch/uuw_sparql/jerventest/vg", at 0x4cc079, in _start
#8    Object "/scratch/uuw_sparql/jerventest/vg", at 0x16ab208, in __libc_start_main
#7    Object "/scratch/uuw_sparql/jerventest/vg", at 0x406510, in main
#6    Object "/scratch/uuw_sparql/jerventest/vg", at 0x8371c7, in vg::subcommand::Subcommand::operator()(int, char**) const
#5    Object "/scratch/uuw_sparql/jerventest/vg", at 0x7b4355, in main_annotate(int, char**)
#4    Object "/scratch/uuw_sparql/jerventest/vg", at 0xb427a9, in vg::parse_bed_regions(std::istream&, xg::XG*, std::vector<vg::Alignment, std::allocator<vg::Alignment> >*)
#3    Object "/scratch/uuw_sparql/jerventest/vg", at 0x16af121, in __assert_fail
#2    Object "/scratch/uuw_sparql/jerventest/vg", at 0x16af0ab, in __assert_fail_base
#1    Object "/scratch/uuw_sparql/jerventest/vg", at 0x16bba50, in abort
#0    Object "/scratch/uuw_sparql/jerventest/vg", at 0xdf9c07, in raise
```

4. What data and command line to use to make the problem recur, if applicable
These are the xz compressed vg file and bed file.
Please reindex the vg file for xg locally.
[vg file](https://drive.sib.swiss/index.php/s/zKMxaKSYd4QEqz9)
[bed file](https://drive.sib.swiss/index.php/s/u4IUH7HHZFP9aT9)
```
vg annotate -p -x ecoli.xg -b ecoli-all.bed > ecoli.all.gam
```

vg version is static compiled vg 1.8.0

ekg:
You should try:

[...]

PR #1779, opened 2018-07-10
Remove assertion on parsing bed format (#1775)

6br:
This is a naive solution to handle #1775.


ekg:
I am not sure this is enough to resolve #1775. Could you please construct a test that shows we can generate annotations that wrap around the origin of a circular path using this fix?

6br:
Yes, this is not a complete solution for #1775. 
Using this pull request, we can parse bed files including circular path without failing but it prints the warning the lines are not parsed.

Issue #1730, opened 2018-06-12
Use a new topology-based distance estimator


adamnovak:
@xchang1 is developing a new distance estimator that does not rely on paths so much. We would like to pull it in and use it in the next release.

faithokamoto:
Done at the time.

PR #1780, opened 2018-07-10
New distance index

xchang1:
Created new topology-based distance estimator
Added random graph generator for unit testing


adamnovak:
This is half of https://github.com/vgteam/vg/issues/1730, right? The other half is tying it into the mapper.

xchang1:
Yeah, this is just the distance calculation itself. It's not used for anything yet

PR #1788, opened 2018-07-12
Implement and test samtools-compatible read-name-based downsampling

adamnovak:
Now we can use our `-d` option on `vg filter` just like the `-s` option on `samtools view`, passing seed and probability specifiers that look like `44.3`. They get broken up into a seed and a probability, and we use the seed exactly the way that samtools does, combining it with a hash of the read QNAME using hash functions in htslib.

As a side effect, we now sample paired reads together. Either you get both ends of the paired read, or you get neither end (assuming the other filters in `vg filter` don't disqualify one end or the other).


adamnovak:
Samtools at one point changed its downsampling algorithm. My code matches the current version, and I know it works with samtools 1.8. It doesn't seem to match the version of Samtools that Travis Linux has installed. I will see if I can upgrade it.

Issue #1792, opened 2018-07-16
Index for seeking into GAM, and seekability improvements for stream.hpp


adamnovak:
We want to try a sorting-and-seeking-based alternative to RocksDB indexing for GAMs, in hopes of speeding up @wolfib's random access queries in the tube map. I'm going to prototype something and add it in to the commands the tube map is using (I think `vg find` and `vg index`, and possibly functioning as an implementation of `vg gamsort -a`), so we can see if it works out to be any faster.

adamnovak:
This should be fixed now. The sort could be faster, and `vg find` still doesn't support the sorted GAM/index combination, but `vg chuk` does and it should be ready for @wolfib to add to the tube map.

PR #1795, opened 2018-07-17
Block the stream

adamnovak:
This is my attempt to move vg over to blocked-gzip compression, instead of just gzip compression, to facilitate GAM indexing for #1792.

So far I have managed to get the output to be in blocked gzip format. I've also refactored things a bit so that stream.hpp isn't such a common transitive include, and so that we don't use the group lengths for progress reporting anymore (since they didn't represent the whole file).

I still have to run the input data through BGZF (probably along with format autodetection), and expose some sensible interface to allow for seeking and for recording offsets.


ekg:
Have you considered other non-zlib based compressors?

adamnovak:
No, not really. If we're willing to make breaking changes, that opens up a lot of options (like dropping the group length field). But if we use blocked GZIP, the resulting files are still readable by older versions of vg, since they're still GZIP files.

adamnovak:
I think that once I get this passing all the tests, it will be ready to merge.

I've had to point at an htslib fork, because being backwards-compatible depends on https://github.com/samtools/htslib/pull/744 being in htslib, to give it multi-member gzip read support.

adamnovak:
I'm going to go ahead and merge this.

As a side effect of the changes I had to make to htslib, we also have a newer htslib being used here. This gives us support for `libdeflate` as a compression library instead of zlib, for blocked gzip files. It uses the same format, but is supposed to be much faster. So we could potentially boost our IO speed by pulling `libdeflate` into the build process, either as a submodule or as a package dependency.

PR #1800, opened 2018-07-20
Distance

xchang1:
Fixed compiler specific error in distance calculation


adamnovak:
Jenkins seems to have lost its worker somehow.

Jenkins, test this please.

src/distance.cpp
```diff
@@ -15,7 +15,28 @@ DistanceIndex::DistanceIndex(VG* vg, SnarlManager* snarlManager){
         cerr << endl << "Creating distance index"<< endl;
     #endif
     const vector<const Snarl*> topSnarls = sm->top_level_snarls();
-    calculateIndex(&topSnarls);
+//              calculateIndex(&topSnarls);
```
adamnovak: This line should probably go away. But I'm going to merge this in anyway.

PR #1752, opened 2018-06-21
X-drop alignment

ekg:
This works towards merging in @ocxtal's X-drop DP algorithm. Thus far I've quashed a few bugs, but testing with vg msga suggests things aren't yet completely solid.

I have observed an improvement in performance of about 50% for longer reads at minimal loss in accuracy given the right parameterization. We can also use longer bands efficiently in the chunked/banded aligner, and this really helps accuracy on simple tests and presumably harder cases as well.

PR #1803, opened 2018-07-24
Xdrop Jenkins Investigation

glennhickey:
#1752 fails 3 Jenkins tests because vg genotype uses a bit more memory -- just enough to get crashes on the 8G nodes.  Will try some stuff here (so as not to mess up the original PR) to fix.  

I first thought it was the gam output of vg map, but the gams look identical.  

Looking at the diffs in the PR, my only other idea is in the indel realignment part of genotype:  it uses the simplest ailgner interface, which repeatedly constructs a temporary Aligner object.  Aligners now construct XdropAligners, so if there's a small memory leak in, say, dz_init(), that would be enough to explain it


glennhickey:
Looks like it's indeed a memory leak in the XdropAligner constructor.

PR #1805, opened 2018-07-28
Sorted GAM Index

adamnovak:
This finishes #1792.

I'm providing a way to sort GAMs by minimum node ID visited (with unmapped reads at node 0) in `vg gamsort`, and a "gai" index very similar to the BAM index to support seeking to the alignments that are relevant to a range or set of ranges of node IDs.

This is still probably WIP until I see how exactly the new sort implementation and the index perform on large GAM files. Also, since I changed `vg chunk` to use the new index, we might need to change toil-vg's use of it to get the CI tests to pass.


adamnovak:
I'm not happy with my sorting setup on large GAM files; it dumps way too many temp files in the temp directory, and is likely to run out of FDs when trying to merge them back together. I'm going to try something else.

adamnovak:
The sort-and-permute strategy seems worryingly slow, even with the supposed-to-be-faster libdeflate. Fundamentally we're wasting a lot of time seeking to individual reads in a block-compressed file, because we need to read and decompress the whole block for basically every read. The blocks are something like 65 kb uncompressed, and the reads are maybe hundreds of bytes uncompressed, so that's a lot of read amplification. During the shuffle, GDB shows the sort spending more time decompressing and reading blocks than doing anything else.

I'm going to see if I can do better with the temp-file-based sort, trying to keep the IO amplification to something like 2x or 3x instead of ~100x.

adamnovak:
OK, I've finally gotten this to build and pass the tests. I'm going to compare sort speed against RocksDB, and look at the size of the .gai file for a whole-genome read set, and then I'll either merge this as is or swap the actual sort logic out for a temporary rocksdb, depending on what wins and by how much.

glennhickey:
@adamnovak Any thoughts on using this framework to sort GAMs by read name instead of node id (as in `samtools sort -n`)?  This is useful for, say, making sure a gam is properly interleaved after having extracted it by region.

adamnovak:
OK, I've run a whole-genome GAM through this. It took 22.5 hours wall clock, started out as 168 GB, and came down to 109 GB after sorting. The sort uses a maximum of about 40 GB of memory; if that's too much, you can keep the memory consumption at around 5 GB if you run only a single thread in the initial temp file sort and split.

The resulting .gai index is only 17 MB (which might be artificially low, since this GAM was only aligned to the primary reference), and queries take a few tenths of a second, which should be sufficient for the tube map.

I'm still going to see how long the RocksDB sort takes, but after that I think this is done; the index itself definitely seems to work.

adamnovak:
OK, RocksDB can sort the same file in 17.5 hours wall clock, but it took ~225 GB of memory. So I'm going to leave the temp file sort in place for now.

adamnovak:
Some things still get slower, most notably:

```
test_call_chr21_snp1kg
Passed in 6653 seconds
            
test_call_chr21_snp1kg
Passed in 7488 seconds
```

But we need the fast standing-start random access for the tube map, so I am going to merge this in and we can work out ways to speed it up later.

PR #1810, opened 2018-08-03
add full-length bonus support in xdrop_aligner

ocxtal:
Implemented full-length bonus feature for use in `vg map -L / vg map --full-l-bonus`. I'm worried if it degrades its speed but I believe it would be small enough for most cases (~1-5%).


ekg:
Jenkins, test this please.

ekg:
I'm merging this despite possible cloud weather problems on travis.

PR #1826, opened 2018-08-14
Refactor and expand algorithms namespace

jeizenga:
Preparing for some algorithm development where I'll need a good number of graph algorithms. I'm taking the opportunity to refactor some of the VG functions into the algorithms namespace and to expand/clean up some of what's already there.


jeizenga:
I realized that the new version of `algorithms::is_acyclic()` has a bug, but I'm going to fix it in a PR I'm working on sometime in the next few days.

PR #1830, opened 2018-08-15
improving map performance

ekg:
This is to explore changing defaults in the mapper, perhaps using xdrop alignment by default and exploring some high costs that may not need to be paid to get high accuracy.


ekg:
@ocxtal just to let you know, I want to complete the integration of xdrop into vg map with this PR. I think it is actually very close to being done as the single-ended mapping performance is very similar to that of the default mapper. However, there is work to be done in paired end mapping and I wanted to ask what you thought about a few options.

This is from a simulation where I ran with `--max-gap-length 150` to try to emulate the behavior of the default aligner (GSSW). I wanted to check if degradation in performance were caused by xdrop itself or the bandwidth:

screenshot from 2018-08-19 14-28-45

It looks like single ended mapping is not affected much at all, and this is the same even when we use the default max gap length of 40 (see attached PDFs below), which is nice because that represents a significant improvement in runtime with minimal ROC AUC cost.

However, something is very wrong with the paired end mapping.

I understand that the MEM seeds are important for alignment as xdrop needs to be started in an exact match in order to ensure the banding is correct. This isn't available in the pair rescue use, as we simply align against a graph region that's expected to hold the mate which we did not align. If this is causing the problem then I could implement an alternative seeding strategy, such as by using k-mers, or perhaps by running an alignment just of the starting part of the sequence against the graph, and only extending hits with a certain score.

What I don't understand is if this might already be possible in dozeu. Could it be easily adjusted to return a multiple alignment for a given graph? Specifically, I'm thinking we could aligning the first SIMD-word sized band of the read along the reference until conditions for extension were met, then explore that particular alignment until X-drop termination, and finally look for the next non-overlapping alignment by re-setting the band to the beginning of the query sequence. Iterating this across the graph would give a good image of likely alignments. (If my prose isn't clear I can try to draw a picture.)

--------

These PDFs show results for a 1M read diploid simulation akin to that in the vg paper. Each has four pages, which in order are ROC overall, ROC for reads matching the reference, ROC for reads covering any variant, and a QQ-plot of mapping quality.

I think it is encouraging that the use of xdrop moves the single ended curves close to those of bwa mem, which is also using a similar banding technique.

Default mapper: [HG002.1M.wg.9-fdcd109.default.pdf](https://github.com/vgteam/vg/files/2300305/HG002.1M.wg.9-fdcd109.default.pdf)

`--xdrop-align` [HG002.1M.wg.9-fdcd109.xdrop.pdf](https://github.com/vgteam/vg/files/2300304/HG002.1M.wg.9-fdcd109.xdrop.pdf)

`--max-gap-length 150` [HG002.1M.wg.9-fdcd109.xdrop.gap150.pdf](https://github.com/vgteam/vg/files/2300303/HG002.1M.wg.9-fdcd109.xdrop.gap150.pdf)

ocxtal:
Sorry for my late reply.

It seems the easiest way to implement the "scan the first word" algorithm is setting the initial values on the first row (parallel to the reference sequence) to zeros. It makes the extend function not penalize left-end gap on the reference side. I'm not sure how long the scanned length should be (8 cells is not sufficient to discriminate spurious matching to real?) but it could be a parameter.

I've added `dz_scan` function for this purpose to dozeu: https://github.com/ocxtal/dozeu/commit/2ef28c29421225d408e45005f7a654aa4cef671c We further need a routine to aggregate scores on the bottom row and I think it will also better included in dozeu so I'll work on it (and combining it to the vg codebase) tomorrow. I'm also interested in the multiple / alternative path feature but it is too much for this purpose. (It's out of scope here but I think the alternative paths, especially local ones, are necessary to properly evaluate the alignments under specific set of haploids or paths. It would also simplify MSGA-like algorithms.)


Seeing the ROC curves I'm concerned the X-drop alignment generates too many false positives than I expected. I suspect banding should affect more strongly on the sensitivity than the specificity but the situation seems the contrary. Do you think it is caused by a bug (in xdrop_aligner.cpp or dozeu) or just a characteristic of the algorithm?

Hajime Suzuki

ekg:
Thanks Hajime, post if you have any questions about the implementation of the scan idea. If the seed is small then it will match several places. This is OK, I think. We really just want to make sure we can try a small subset of a larger graph via the banding and local extension where applicable.

>  (It's out of scope here but I think the alternative paths, especially local ones, are necessary to properly evaluate the alignments under specific set of haploids or paths. It would also simplify MSGA-like algorithms.)

I think this is very similar in spirit to what @jeizenga has implemented in mpmap. The key idea there is multipath mapping, so the alignment itself is represented as a DAG of sub-alignments. This has benefits for haplotype aware mapping to graphs. As he has explained to me that method doesn't spend a lot of time in DP during alignment, and he has implemented his own banded global alignment method. There could be a benefit to integrating dozeu into mpmap, and as well to extending dozeu to generate multipath alignments directly. Let me know if this matches your thinking.

> Seeing the ROC curves I'm concerned the X-drop alignment generates too many false positives than I expected. I suspect banding should affect more strongly on the sensitivity than the specificity but the situation seems the contrary. Do you think it is caused by a bug (in xdrop_aligner.cpp or dozeu) or just a characteristic of the algorithm?

Note that the ROC curves that do worse are only the paired-end ones. The problem is that I'm trying to use dozeu for the pair rescue, and it is failing as it can't find the correct seed position to start the alignment. Does this seem like a fair assessment? The read will be aligned against a large region. If a spurious alignment, or no alignment occurs within the X-drop parameter limit then we would expect to get back a failed alignment. It is possible that I am making a mistake, because I haven't yet checked failing examples, but this is my reading of the code currently.

ekg:
@ocxtal I'm picking this up again. I think we're close. I'd like to test use of the scan function, and so I'm testing updating dozeu to 55c9903, but I get errors of this type during the vg build: [err.txt](https://github.com/vgteam/vg/files/2420107/err.txt). The errors look like this generally:

```
In file included from src/xdrop_aligner.cpp:15:0:                                                   
src/../deps/dozeu/dozeu.h: In function ‘void dz_unused_772()’:                              
src/../deps/dozeu/dozeu.h:773:28: error: ‘DZ_UNITTEST_SCORE_PARAMS’ was not declared in this scope
  struct dz_s *dz = dz_init(DZ_UNITTEST_SCORE_PARAMS);                                              
                            ^~~~~~~~~~~~~~~~~~~~~~~~                           
src/../deps/dozeu/dozeu.h:773:28: note: suggested alternative: ‘DZ_UNITTEST_INDEX’
  struct dz_s *dz = dz_init(DZ_UNITTEST_SCORE_PARAMS);                                              
                            ^~~~~~~~~~~~~~~~~~~~~~~~              
                            DZ_UNITTEST_INDEX                                                      
src/../deps/dozeu/dozeu.h: In function ‘void dz_unused_999()’: 
src/../deps/dozeu/dozeu.h:1000:28: error: ‘DZ_UNITTEST_SCORE_PARAMS’ was not declared in this scope
  struct dz_s *dz = dz_init(DZ_UNITTEST_SCORE_PARAMS);         
                            ^~~~~~~~~~~~~~~~~~~~~~~~                                               
src/../deps/dozeu/dozeu.h:1000:28: note: suggested alternative: ‘DZ_UNITTEST_INDEX’
  struct dz_s *dz = dz_init(DZ_UNITTEST_SCORE_PARAMS);
                            ^~~~~~~~~~~~~~~~~~~~~~~~                               
                            DZ_UNITTEST_INDEX         
src/../deps/dozeu/dozeu.h:1003:43: error: ‘dz_unittest_query’ was not declared in this scope
  struct dz_query_s *q = dz_pack_query(dz, dz_unittest_query, dz_unittest_query_length);
                                           ^~~~~~~~~~~~~~~~~                                       
src/../deps/dozeu/dozeu.h:1003:43: note: suggested alternative: ‘dz_pack_query’
  struct dz_query_s *q = dz_pack_query(dz, dz_unittest_query, dz_unittest_query_length);
                                           ^~~~~~~~~~~~~~~~~                                
                                           dz_pack_query        
src/../deps/dozeu/dozeu.h:1003:62: error: ‘dz_unittest_query_length’ was not declared in this scope
  struct dz_query_s *q = dz_pack_query(dz, dz_unittest_query, dz_unittest_query_length);
                                                              ^~~~~~~~~~~~~~~~~~~~~~~~
src/../deps/dozeu/dozeu.h: In function ‘void dz_unused_1187()’:
src/../deps/dozeu/dozeu.h:1188:28: error: ‘DZ_UNITTEST_SCORE_PARAMS’ was not declared in this scope
  struct dz_s *dz = dz_init(DZ_UNITTEST_SCORE_PARAMS);                            
                            ^~~~~~~~~~~~~~~~~~~~~~~~
src/../deps/dozeu/dozeu.h:1188:28: note: suggested alternative: ‘DZ_UNITTEST_INDEX’
  struct dz_s *dz = dz_init(DZ_UNITTEST_SCORE_PARAMS);                  
                            ^~~~~~~~~~~~~~~~~~~~~~~~
                            DZ_UNITTEST_INDEX
```

Basically, the build can no longer find the unittest related defines. It's a little confusing to me, but I'll see if I can work through it.

ocxtal:
(sorry for confusion on the compilation error of dozeu library.) The scan function is now working. It scans the last 15 bases (or first 15 bases for reverse-complemented mapping) to determine one possible alignment end position, and align back toward the head of the sequence. On the `sample1.fa_{1,2}.fastq` dataset in the `test/alignment` directory, the scan function recalls 15 alignments out of 17 that are mapped by `pair_rescue` without X-drop alignment option. The missing two were due to improper end position by the scan function, which is roughly resulted from too short and too biased scanned range (15 bp at the end). So the the result is expected to be better if the scan length be adjustable and scan position chose at the middle (or even multiple positions) of the query sequence.

In testing, vg was built with `-Ddebug_rescue` and the number of rescued alignments was counted by `vg map --debug -ijt1 --xdrop-alignment -d graph -f sample1.fa_1.fastq sample1.fa_2.fastq >/dev/null 2>&1 | grep rescued | wc -l`.

Hajime Suzuki

ekg:
@ocxtal thank you for picking this up. I'm just now getting back to it. I like the idea of adjusting the scanning to use multiple windows, maybe one at each end, or three (one at the beginning, middle, and end). I'm concerned that the test you ran was a little strange due to the `-i` parameter (interleaved fastq) and the specification of only one fastq (the second was ignored). This is the right way to run the test:

```
vg map --debug -j -t 1 --xdrop-alignment -d g -f sample1.fa_1.fastq -f sample1.fa_2.fastq
```

You'll notice that it's much, much faster as the pair model is correctly established.

In fact, I get a different result, and it's better!

```
erik@foldy [03:47:00 PM] [~/vg/test/alignment] [map-perf *]
-> % vg map --debug -j -t 1 -d g -f sample1.fa_1.fastq -f sample1.fa_2.fastq 2>&1 | grep rescued | wc -l
300
erik@foldy [03:47:10 PM] [~/vg/test/alignment] [map-perf *]
-> % vg map --debug -j -t 1 --xdrop-alignment -d g -f sample1.fa_1.fastq -f sample1.fa_2.fastq 2>&1 | grep rescued | wc -l
301
```

I'm going to try this out in larger scale testing. Thus far, it looks great!

ekg:
@ocxtal good news! This fixes the problem we observed before. Here's the ROC for the same 1M simulated reads.

screenshot from 2018-11-29 14-05-13

I ran this with `--xdrop-align --max-gap-length 150`. I'll now check with the default max-gap-length (40). Maybe that should be made a fraction of the read length rather than a fixed value? Otherwise, I think this is ready to merge into master.

Cool!

ekg:
This is with the default max gap length (40):

screenshot from 2018-11-29 15-11-50

Note that the performance is slightly degraded. However, it's only visually obvious in the case of the pangenome curves. I wonder why this is. Perhaps 40 is too short for standard Illumina-type sequences.

PR #1831, opened 2018-08-16
Allow vg paths to list thread names, and refactor it

adamnovak:


adamnovak:
It looks like the failing Jenkins test failed due to a failure to import the input files into Toil. It works fine when I run it locally. I'm going to try it again.

Jenkins, test this please.

PR #1835, opened 2018-08-20
Use handles for subgraph extraction algorithms

jeizenga:
I refactored my subgraph extraction algorithms away from Protobuf to a purely handle-based implementation.


src/graph_synchronizer.cpp
```diff
@@ -122,37 +119,36 @@ void GraphSynchronizer::Lock::lock() {
             // Make them into pos_ts that point left to right, the way Jordan thinks.
             pos_t left_pos = make_pos_t(start_left.node, start_left.is_end, 0);
             pos_t right_pos = make_pos_t(end_right.node, !end_right.is_end,
-                synchronizer.graph.get_node(end_right.node)->sequence().size() - 1);
+                synchronizer.graph.get_node(end_right.node)->sequence().size());
             
             // Since these are already at node ends, we don't need to worry about node cuts.
             
             // Extract paths out to the length we need to connect the ends, or a bit further.
             // TODO: be sure to extract really big indels somehow...
```
jeizenga: @adamnovak 
I could use a little review on this. 

I eliminated one of the options you were using here because I decided it wasn't good. Before you could request that the source and destination positions would be the only tips in the extracted graph. I got rid of that option. You do still have the option of requiring that all material in the subgraph is reachable from the source position and can reach the destination position, which is a slightly stronger property. I switched you over to that for now.

I also no longer provide the guarantee that the node IDs will be the same as long as you don't duplicate. Also take a look out my translation solution to that mismatch.

adamnovak: I guess that's sufficient to pass whatever tests we have in there for vg add.

I'd have to go back and think about how vg add thinks to really evaluate what it means that the node IDs might not be "the same" anymore. I can proibably look at this tomorrow.

But the reachability criterion ought to be good enough for vg add; I don't think it is going to be trying to add lots of ugly unpaired reversing edges.

jeizenga: I wasn't super clear about this, but I believe the node translation trick I do after the graph extraction should make the node ID space the same as the full graph again.

jeizenga:
Jenkins, test this please.

src/algorithms/extract_containing_graph.hpp
```diff
@@ -19,21 +19,22 @@
...
-    /// the positions vector and all other nodes and edges that can be reached within a maximum
-    /// distance from any of these positions. Node IDs in the subgraph are retained from the full graph.
+    /// Fills graph 'into' with the subgraph of the handle graph 'source' that contains all of the
+    /// positions in the positions vector and all other nodes and edges that can be reached within
+    /// a maximum distance from any of these positions. Node IDs in the subgraph are retained from
+    /// the source graph.
     ///
     /// Args:
     ///  vg                         graph to extract subgraph from
     ///  g                          graph to extract into
```
adamnovak: The arguments have changed names, but this little table hasn't been updated.

Also, shouldn't it be `source` and `dest` and not `source` and `into`?

src/graph_synchronizer.cpp
```diff
@@ -122,37 +119,36 @@ void GraphSynchronizer::Lock::lock() {
...
-                true, // Keep the specified positions in the graph
                 false, // Disallow terminal node cycles, so we don't duplicate nodes
-                true, // Remove other tips
-                false, // We don't care about being on a path
-                false); // Or being strictly less than the specified length
+                true, // We don't want extraneous material that doesn't connect the positions
+                false); // But we don't care about being strictly less than the specified length
                 
 #ifdef debug
             cerr << "Extracted " << context_graph.node_size() << " nodes and " << context_graph.edge_size() << " edges between " << path_name << ":" << start << "-" << past_end << endl;
```
adamnovak: This line ought to change from context_graph to context, or things will go wrong if debug is defined.

src/graph_synchronizer.cpp
```diff
@@ -122,37 +119,36 @@ void GraphSynchronizer::Lock::lock() {
...
                 
-            // Any duplication is going to mess things up, since we need
+            // Any ID mismatch is going to mess things up, since we need
             // operations on the new graph to make sense in the original graph.
-            // We need all the entries in this duplication map to be no-ops.
-            for (auto& kv : duplications) {
-                assert(kv.first == kv.second);
+            // We need all the entries in this translation map to be no-ops, so
+            // we translate all IDs back to their original (which is possible
+            // because we chose extraction paramters that never duplicate nodes).
```
adamnovak: The documentation for extract_connecting_graph doesn't seem to specify when nodes may be duplicated. If we rely on the lack of duplication here, we ought to put it in the function spec that under these particular conditions there will be no duplication. This new swap code won't actually detect and abort if duplication *has* occurred when it wasn't supposed to.

jeizenga:
Merging without waiting for tests because all I changed was comments.

PR #1839, opened 2018-08-23
Faster VCF parsing

### LINKED ISSUE #1838: VCF parsing is slow with a huge number of samples
I am running `vg construct` from a VCF file with tens of thousands of samples. The construction is very slow, processing maybe 5 variants per second. I believe the bottleneck is VCF parsing, which builds heavy data structures for the samples.

The `vg` process runs in cycles that last for 3-4 minutes. First the memory usage grows slowly to 30-40 gigabytes, probably because VG is parsing the next chunk of ~1024 variants. When the memory usage peaks, VG writes tens of kilobytes to the graph file. Then it starts freeing memory, which happens much faster than the growth.

The `vcflib::VariantCallFile` object has a member variable `bool parseSamples`. If we set it `true`, vcflib should not attempt to parse the samples. In GBWT construction, we can then find the genotype strings from the original line, which is also stored in the `Variant` object.

jltsiren:
Disable sample parsing using vcflib in `vg construct` and `vg index -G`. Find the genotype strings using lightweight parsing in `vg index -G`.

Resolves #1838.

I am not sure if VCF files with a large number of samples are used in other subcommands.

### DIFF
--- a/src/subcommand/construct_main.cpp
+++ b/src/subcommand/construct_main.cpp
@@ -273,6 +273,7 @@ int main_construct(int argc, char** argv) {
                 return 1;
             }
             vcflib::VariantCallFile* variant_file = new vcflib::VariantCallFile();
+            variant_file->parseSamples = false; // Major speedup if there are many samples.
             variant_files.emplace_back(variant_file);
             variant_file->open(vcf_filename);
             if (!variant_file->is_open()) {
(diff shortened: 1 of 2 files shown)


jltsiren:
Travis Mac tests froze in test 4 of `27_vg_genotype.t` ("vg genotype finds few ultrabubbles for a subset of just the reference"). This is expected behavior. The command (`vg genotype tiny.vg tiny.gam.index -vSp` with an empty `.gam`) crashes on both Mac and Linux, but sometimes it freezes on Mac instead.

PR #1840, opened 2018-08-24
fix the condition of max nodes

hanzou666:
An equal sign isn't contained in the source code, although `vg snarls` help is as following:

```
    -m, --max-nodes N      only compute traversals for snarls with <= N nodes [10]
```

So, I propose to append an equal sign to the condition of max_nodes.


ekg:
Jenkins, OK to test.

PR #1843, opened 2018-08-24
Expose recombination rate to the command line for haplotype scoring

adamnovak:


adamnovak:
Looks like Microsoft's R package repo at https://mran.microsoft.com/snapshot/2018-07-02/src/contrib/PACKAGES was temporarily down again, causing a bunch of plotting commands to fail. I can reach it now.

Jenkins, test this please.

PR #1844, opened 2018-08-25
Add public methods to get edges count by id

cartoonist:
The `g_iv` vector is a private member variable; therefore cannot be
accessed by other or derived classes. And, there is no way to query
the number of forward/backward edges without getting the list of edges.
This commit adds two methods (`outdegree` and `indegree`) returning the
outdegree and indegree of nodes by their ids.


ekg:
Ok to test

PR #1847, opened 2018-08-28
Add input validation and limit dynamic range to fix #1846

### LINKED ISSUE #1846: mpmap dynamic_alt_alns feature can ask for 0 alignments altogether and crash
Now mpmap's `MultipathAlignmentGraph` has a `dynamic_alt_alns` setting which causes it to use `algorithms::count_walks` to work out how many alignment alternatives to ask for from the aligner.

Unfortunately, `count_walks` can return 0, and somewhere between `MultipathAlignmentGraph::align` and the actual GSSW traceback functions, that number of alternative alignments gets interpreted as a number of *total* alignments. Then the code that uses the GSSW alignments [tries to get the first alignment](https://github.com/vgteam/vg/blob/48f30e470aaef6f720851776cbac3e1bd04ab7df/src/gssw_aligner.cpp#L1379) and ends up wandering all over memory and segfaulting if you're lucky.

I'm not sure whether we should be always performing at least one alignment, even when `count_walks` returns 0, or if we should dutifully perform no alignments and just bail out on that case. I want to change the semantics overall so that "alt alignments" variable actually represents the alternate alignments to compute (i.e. in addition to the primary), but that would overall result in us computing one more alignment than we did before in most cases, unless we also changed how we compute the parameter.

adamnovak:
I'm going to try just stopping the dynamic code from going below 1, and I've added input checking code so the aligners can't be made to do 0 alignments (they'll bail out of the whole program instead, which is also pretty terrible for debugging).

### DIFF
--- a/src/gssw_aligner.cpp
+++ b/src/gssw_aligner.cpp
@@ -889,6 +889,10 @@ void Aligner::align_internal(Alignment& alignment, vector<Alignment>* multi_alig
         cerr << "error:[Aligner] cannot specify maximum number of alignments in single alignment" << endl;
         exit(EXIT_FAILURE);
     }
+    if (max_alt_alns <= 0) {
+        cerr << "error:[Aligner] cannot do less than 1 alignment" << endl;
+        exit(EXIT_FAILURE);
+    }
 
     // alignment pinning algorithm is based on pinning in bottom right corner, if pinning in top
     // left we need to reverse all the sequences first and translate the alignment back later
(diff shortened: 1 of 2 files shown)


adamnovak:
This fix seems to add on the order of a minure to each of the Jenkins runs, which is enough to cause this one to fail.

@jeizenga Any idea where I should cut in the mpmap parameters to make up the extra time? Should I just skip cases where it would be wanting to do 0 alignments altogether?

PR #1848, opened 2018-08-29
Do no alignments as a different way to address #1846

adamnovak:
I'm going to try letting the adaptive alignment count go down to 0, and just skip over the actual alignment code in those cases, to try and speed things up.

### DIFF
--- a/src/gssw_aligner.cpp
+++ b/src/gssw_aligner.cpp
@@ -889,6 +889,10 @@ void Aligner::align_internal(Alignment& alignment, vector<Alignment>* multi_alig
         cerr << "error:[Aligner] cannot specify maximum number of alignments in single alignment" << endl;
         exit(EXIT_FAILURE);
     }
+    if (max_alt_alns <= 0) {
+        cerr << "error:[Aligner] cannot do less than 1 alignment" << endl;
+        exit(EXIT_FAILURE);
+    }
 
     // alignment pinning algorithm is based on pinning in bottom right corner, if pinning in top
     // left we need to reverse all the sequences first and translate the alignment back later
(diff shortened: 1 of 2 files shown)


adamnovak:
I want @jeizenga to think about this when he can, but I am going to merge this now, because it passes the tests and it can't be worse than the undefined behavior we have in these cases now.

jeizenga:
@adamnovak Do you have examples of reads that display this behavior? I think your fix makes sense, but I also think that `count_walks` should only return 0 for empty graphs or graphs that don't meet its input requirements (i.e. single-stranded DAGs). I'd like to look into what's making it act this way.

adamnovak:
What I have is the S3 URLs for the problematic data files, and a command
that I ran under gdb to diagnose the problem:

```
aws s3 cp
s3://cgl-pipeline-inputs/vg_cgl/pop-map/alignments/wg38/v5/sim/graph.vg.snarls
.
aws s3 cp
s3://cgl-pipeline-inputs/vg_cgl/pop-map/alignments/wg38/v5/sim/reads_chunk_0_0.fq.gz
.
aws s3 cp
s3://cgl-pipeline-inputs/vg_cgl/pop-map/alignments/wg38/v5/sim/graph.vg.xg .
aws s3 cp
s3://cgl-pipeline-inputs/vg_cgl/pop-map/alignments/wg38/v5/sim/graph.vg.gcsa
.
aws s3 cp
s3://cgl-pipeline-inputs/vg_cgl/pop-map/alignments/wg38/v5/sim/graph.vg.gcsa.lcp
.
aws s3 cp
s3://cgl-pipeline-inputs/vg_cgl/pop-map/alignments/wg38/v5/sim/graph.vg.gbwt
.

gdb vg

run mpmap --single-path-mode --snarls graph.vg.snarls -f
reads_chunk_0_0.fq.gz -t 32 -i -x graph.vg.xg -g graph.vg.gcsa --gbwt-name
graph.vg.gbwt --recombination-penalty 22.0 >/dev/null
```

When I ran that with vg d7e8a63e I eventually had it crash (after like 20
minutes of loading in snarls and 20 minutes of mapping without crashing.


I also have a read sequence:
AACCTTGCAAAAGGCAAACTTATGCCATTGTGTCCTCTGACATACTGGGTGATGTACTGTATGTTACTGATGTGAGGGGAACTTCCTAAATTGGCTAGTAAATTATGCCAAATAAAAAGCAAAAAA

You might be able to search through the FASTQ for that to find an offending
pair more quickly.

PR #1855, opened 2018-09-04
Distance

xchang1:
Added distance index to vg index for testing @jeizenga


ekg:
Is this built by default?

src/subcommand/index_main.cpp
```diff
@@ -1016,6 +1035,33 @@ int main_index(int argc, char** argv) {
...
+        } else {
+            ifstream vg_stream(dist_graph);
+            if (!vg_stream) {
+                cerr << "error: [vg index] cannot open VG file" << endl;
+                exit(1);
+            }
+            VG vg(vg_stream);
+            vg_stream.close();
+
+            CactusSnarlFinder bubble_finder(vg);
```
jeizenga: We also have the `vg snarls` command that will serialize snarls to disk. I think it would make sense to also let your command load snarls from one of these files instead of detecting them itself (e.g. [here](https://github.com/vgteam/vg/blob/master/src/subcommand/mpmap_main.cpp#L836)). If you decide to keep both code paths, you should probably also let the snarl manager serialize the snarls it detects too.

Issue #1856, opened 2018-09-04
Incorrect Snarls lead to segfault in Genotyper, which causes Travis to periodically hang on the integration tests


jeizenga:
We've been experiencing periodic hangs from Travis in the genotyper integration tests. I tracked the segfault down to the [RepresentativeTraversalFinder::find_traversals](https://github.com/vgteam/vg/blob/master/src/traversal_finder.cpp#L903) algorithm, but the issue appears to be that it's working with incorrect Snarls. This in turn breaks some expectations in the SnarlManager::shallow_contents algorithm.

The Genotyper uses [CactusSnarlFinder::find_snarls](https://github.com/vgteam/vg/blob/master/src/snarls.cpp#L30) to find these snarls, which as far as I can tell is just every edge on the reference path:
```
{"type":"ULTRABUBBLE","start":{"node_id":"1"},"end":{"node_id":"3"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
{"type":"ULTRABUBBLE","start":{"node_id":"3"},"end":{"node_id":"5"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
{"type":"ULTRABUBBLE","start":{"node_id":"5"},"end":{"node_id":"6"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
{"type":"ULTRABUBBLE","start":{"node_id":"6"},"end":{"node_id":"8"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
{"type":"ULTRABUBBLE","start":{"node_id":"8"},"end":{"node_id":"9"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
{"type":"ULTRABUBBLE","start":{"node_id":"9"},"end":{"node_id":"11"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
{"type":"ULTRABUBBLE","start":{"node_id":"11"},"end":{"node_id":"12"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
{"type":"ULTRABUBBLE","start":{"node_id":"12"},"end":{"node_id":"14"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
{"type":"ULTRABUBBLE","start":{"node_id":"14"},"end":{"node_id":"15"},"start_end_reachable":true,"directed_acyclic_net_graph":true}
```

In this graph (tiny.vg):
[...]

PR #1861, opened 2018-09-06
Fix hang in Travis by removing subset graph option in Genotyper

jeizenga:
Resolves https://github.com/vgteam/vg/issues/1856

Not the most elegant solution, but I removed the subset_graph option in the Genotyper. We were only using it for the part of the code that did snarl detection, but snarls from a subset graph are not necessarily snarls in the full graph. That was breaking some of the algorithms' expectations and leading to segfaults. 

In the future, we could also have the Genotyper use the subset graph throughout the genotyping process, but this would take a significant refactor since much of the code base uses the AugmentedGraph class rather than the VG class, which is what the subset graph is.

### DIFF
--- a/src/genotyper.cpp
+++ b/src/genotyper.cpp
@@ -24,7 +24,6 @@ void Genotyper::run(AugmentedGraph& augmented_graph,
                     string contig_name,
                     string sample_name,
                     string augmented_file_name,
-                    bool subset_graph,
                     bool output_vcf,
                     bool output_json,
                     int length_override,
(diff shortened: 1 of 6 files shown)


src/genotyper.cpp
```diff
@@ -81,21 +80,26 @@ void Genotyper::run(AugmentedGraph& augmented_graph,
...
-    }
+    // TODO: I've removed this option because it has the chance to create incorrect Snarls from the subset
+    // graph that are not actually snarls in the full graph. If we are going to retain this feature, we need
+    // to use the subset graph throughout the genotyping process, but that's currently not possible because
+    // the rest of the code uses an AugmentedGraph instead of a VG.
+    
+//    // We need to decide if we want to work on the full graph or just on the subgraph that has any support.
+//    // We make the subset a local out here so it will stick around as long as we
+//    // use the SnarlManager.
+//    VG subset;
```
adamnovak: Shouldn't we just drop all this code instead of commenting it out?

jeizenga: I thought about that, but I figured that if someone came back in the future and wanted to re-implement this feature it might be handy to have a prototype to look at without digging into the git history. If you feel strongly about it, I can remove this.

PR #1866, opened 2018-09-10
Do away with most of the indexes in the SnarlManager

### LINKED ISSUE #1862: Snarl loading is slow
The snarls file at s3://cgl-pipeline-inputs/vg_cgl/pop-map/alignments/wg38/v5/sim/graph.vg.snarls takes about 20 minutes to load into a SnarlManager, despite being only about 300 MB.

Jordan thinks the problem is that we are wasting all our time maintaining multiple hash tables on the defining boundary nodes of the snarl. I think he's right; I tried swapping out the hash table implementation for the wang_hash based Google hash table we have, as well as for just a tree-based `std::map`, and didn't notice dramatic improvements. The `std::hash` we have for pairs isn't great, but it doesn't look to be producing oodles of collisions either.

I am going to rearchitect the SnarlManager to get rid of the whole key form concept, and all the parallel hash tables. This might mean keeping things in a bunch of vectors by snarl number, or in a big vector of fully-populated snarl-describing structs that also hold children, etc.

While I'm at it maybe I can do #1797.

adamnovak:
This should fix #1862. It's still not really great in terms of speed, but it is more than twice as fast:

```
time vg.old mpmap --single-path-mode --snarls ./trash/graph.vg.snarls -f empty.fq -t 1 -i -x x.xg -g x.gcsa >/dev/null
real	31m45.393s
user	22m15.431s
sys	9m29.165s

time vg mpmap --single-path-mode --snarls ./trash/graph.vg.snarls -f empty.fq -t 1 -i -x x.xg -g x.gcsa >/dev/null
real	13m17.208s
user	7m41.112s
sys	5m35.912s
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make get-deps;fi
 script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then git submodule update --recursive && make -j4 && echo Testing && make test && make static -j4; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then git submodule update --recursive && timeout 1500 make deps -j4 && make -j4 && echo Testing && make test; fi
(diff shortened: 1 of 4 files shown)


jeizenga:
Cool, thanks!

PR #1882, opened 2018-09-20
Allow the GBWT scoring to be forced to happen

adamnovak:


adamnovak:
Jenkins ran out of memory trying to send the test status to Github for some reason. I highly doubt that this is vg's fault; I think we can merge this anyway.

Issue #1879, opened 2018-09-20
pruning indexing complex graph


sanboissel:
I'm hoping you can give me a better understand of how pruning is affecting my graph. I am aligning multiple yeast genomes using progressiveCactus but the resulting graph is too large to index. I got it to work with the following 
`./bin/vg prune -k 12 -e 2 -p yeast/cactus_yeast_msa5.vg > yeast/cactus_yeast_msa5_norpruned.vg`

`./bin/vg index -p -t 16 yeast/cactus_yeast_msa5_norpruned.vg -g yeast/cactus_yeast_msa5.gcsa`
 
but I'm not entirely clear on how this will affect alignments. Will I not be able to align to regions of the map with high complexity?
I tried including `-r` in the prune step since I gave a `--refGenome` for the hal2vg step but this essentially brought the map back to the original size.

I also tried breaking my MSA into roughly equal-sized chunks based on the Anc0Chr# (I can't run alignments by reference chromosome) but the resultant graphs appear only slightly less complex than the full genome alignment and don't do any better during indexing.

sanboissel:
Thanks very much - 
When I go to do the mapping now I'm getting the following error 
```[libprotobuf FATAL /home/sboissel/vg/include/google/protobuf/repeated_field.h:1478] CHECK failed: (index) < (current_size_):
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: (index) < (current_size_):
ERROR: Signal 6 occurred. VG has crashed. Run 'vg bugs --new' to report a bug.
Stack trace path: /tmp/vg_crash_lgW8RT/stacktrace.txt```
with my full fasta file. If I break up the file into smaller sections there seem to be some sections that don't run correctly but I can't figure out why. For example - my entire FASTA file is 4000 reads. The first 1500 reads run no problem, but a file containing the next 50 reads gives me the error above. If I split this file into the top 25 reads and bottom 25 reads, both run fine. If I try concatenating the two files back together they again fail. If I concatenate the files together in reverse order (bottom 25 reads first then the top 25 reads) it works. This seems to be an issue with the index but as far as I can tell by eye the index file looks fine.

**Comment by jltsiren** [comment]
That error looks weird. Did you use the node mapping in `vg prune` and `vg index`, as instructed in the wiki:
```
vg prune -u -g graph.gbwt -m node_mapping graph.vg > graph.pruned.vg
vg index -g graph.gcsa -f node_mapping graph.pruned.vg
```
[...]

PR #1883, opened 2018-09-20
Prune and unfold paths without GBWT

jltsiren:
Enable unfolding paths in `vg prune` without a GBWT index. This can be useful with graphs created by aligning multiple sequences (see #1879).

Also remove obsolete `PhaseDuplicator`.


jltsiren:
Something weird is going on with Jenkins. The tests passed, but now it can't find the results.

PR #1885, opened 2018-09-21
Try building Doxygen docs on Travis and pushing the result to gh-pages

adamnovak:
There's a [cool way](https://gist.github.com/domenic/ec8b0fc8ab45f39403dd) to use deploy keys and Travis to auto-build Doxygen docs and push them to Github Pages.

Since ReadTheDocs builds have been failing forever, apparently because of the inadequacy of the Doxygen-to-Sphinx Breathe bridge stuff, I want to move the "official" vg internals documentation off of RtD and just use some Doxygen-generated HTML.

Once this works, we could also use something like http://mcss.mosra.cz/doxygen/ to replace the Doxygen output with something using a nicer theme.

We'll have to keep an eye on PRs for attempts to steal our encryption keys for the encrypted deploy key, which gives access to the vgteam/vg repo.


adamnovak:
Jenkins ran out of memory trying to post the test results again. I think the logs for successful tests are long now for some reason (Maybe these deprecation wanrings?):

```
/mnt/ephemeral/workspace/vg/.env/local/lib/python2.7/site-packages/toil/fileStore.py:1846: DeprecationWarning: object.__init__() takes no parameters
  super(FileID, self).__init__(fileStoreID)
```

I've turned off the option to retain long logs for passing tests for the vg project, which warns that it can take a lot of RAM (because I think it parses the whole log into memory at once???). Maybe that will stop this from happening.

Either way, it's not a problem with this PR, and the tests actually passed, so I'm going to merge this in and see if we get any Doxygen docs.

PR #1888, opened 2018-09-24
Move CI and scripts from the cgl-pipeline-inputs bucket to the vg-data bucket

adamnovak:
We have to delete the old bucket because UCSC now wants all data in a bucket belonging to a single project.


adamnovak:
Erich and I added what we think is the right policy to let Jenkins write to the new bucket.

Jenkins, test this please.

PR #1894, opened 2018-09-27
Coverage maps

ekg:
This is to improve the coverage maps and their interfaces. I hope to add coverage for edges, and also queries that let us pick up the coverage for a given part of the graph. It would also be nice to improve vg viz to obtain subgraphs with coverage tracks over them.


adamnovak:
I'm going to have to restart the Jenkins tests because we moved all the test data. Hopefully it will test the merge commit with master that picks that up.

Jenkins, test this please.

faithokamoto:
I think this is de facto not happening, and if we want to pick it back up again it should be worked on from a modern vg version.

PR #1903, opened 2018-10-01
Make make clean idempotent to fix #1852

### LINKED ISSUE #1852: make clean (after a failed build in a clean checkout) fails with an error 2 because of an absence of a clean rule in the vcflib dep.
This was reported at https://github.com/vgteam/vg/issues/1850#issuecomment-417187691

If vcflib doesn't have a clean, we shouldn't call it. If it needs to do some setup work before clean will work, we should do that during the vg clean.

adamnovak:
Depends on:
https://github.com/ekg/multichoose/pull/4
https://github.com/edawson/libVCFH/pull/3
https://github.com/vgteam/vcflib/pull/12


adamnovak:
All the dependency PRs are in. I'm not sure how Travis managed to pass without that. Maybe the cache let it use an old version of the submodules in question.

Jenkins, test this please.

adamnovak:
This looks liek another failure of the R package mirror.

Jenkins, test this please.

PR #1923, opened 2018-10-09
Remove date from Doxygen footer so all the files don't constantly change

adamnovak:


adamnovak:
That Jenkins test didn't time out because I changed the Doxygen template. I'm not sure why it did, though. It stopped in vg filter.

Maybe it's just cloud weather?

Jenkins, test this please.

PR #1924, opened 2018-10-09
Use snarl-based distance index for clustering instead of paths in mpmap

jeizenga:
I factored out the OrientedDistanceClusterer's distance measurement to an external object with an abstract API. I also made a second distance-measuring implementation based on @xchang1's snarl-based distance index.

A current limitation of this code is that the SnarlManager is used in two potentially conflicting ways: as backing data for the DistanceIndex and for snarl-cutting. I'm not dealing with the conflict at the moment, but it would help if Snarls were labeled as trivial/nontrivial.


jeizenga:
Right now you need to do create the distance index separately in `vg index`. You also need snarls from `vg snarls`. It might be worth rethinking how to do all of it in one go. My new development for this PR is mostly centered around `SnarlOrientedDistanceMeasurer:: oriented_distance` in `cluster.hpp`.

ekg:
We should get something in the tests to confirm this is helping and we
don't break it, like a small assembly graph. Ideas?

Issue #1926, opened 2018-10-09
vcflib SVLEN and END semantics displease me


adamnovak:
I'm not sure about this whole block:

https://github.com/vgteam/vcflib/blob/98f3096f559644c2bdd296bd915256bd2e8be3a8/src/Variant.cpp#L301-L341

SVLEN is supposed to hold the length *change* caused by the alt allele of the variant, according to the vcflib spec. If we have it, we populate `info_len` with the absolute value of the field.  But then we turn around and, if SVLEN isn't set, we populate `info_len` based on the difference between POS and END, which will produce a value with completely different semantics. Then at the end we compare `info_end` and the position plus `info_len`, and bail on canonicalization if they don't match.

It looks like for an insertion we would expect END not to be the same as POS but to be POS plus the length of the inserted sequence, which would be a really weird and I think incorrect way to write the variant. That's actually what our unit test VCFs are doing, as in https://github.com/vgteam/vg/blob/a5dee39f6d3c28c05e5a3bf5bc0af1fc864e69b5/src/unittest/constructor.cpp#L1452, but I am pretty sure it is wrong. But it's something we guarantee:

https://github.com/vgteam/vcflib/blob/98f3096f559644c2bdd296bd915256bd2e8be3a8/src/Variant.h#L241

I also think I have a problem with this guarantee from canonicalization:

https://github.com/vgteam/vcflib/blob/98f3096f559644c2bdd296bd915256bd2e8be3a8/src/Variant.h#L245

SVLEN is defined as a length change by the spec, so it should be negative for deletions, positive for insertions, and zero for inversions. Right now we're using it to try and pass inversion length as well:

https://github.com/vgteam/vg/blob/a5dee39f6d3c28c05e5a3bf5bc0af1fc864e69b5/src/constructor.cpp#L741

See also: https://github.com/vgteam/vg/issues/1922

eldariont:
I agree on all your points. The VCF specification defines SVLEN as the "Difference in length between REF and ALT alleles". This would mean:

For deletions, SVLEN<0 and the region between POS and END is deleted ==> POS - END = SVLEN
For insertions, SVLEN>0 and POS == END.
For inversions, SVLEN=0 and the region between POS and END is inverted.

edawson:
> For insertions, SVLEN>0 and POS == END
> For deletions, SVLEN<0 and the region between POS and END is deleted ==> POS - END = SVLEN

[...]

PR #1929, opened 2018-10-10
Fix construct

adamnovak:
@eldariont Try this branch.

To work around #1926 I recommend you only set one of SVLEN and END for a variant. Internally we're reading everything off the END now, because the VCF spec defines it well:

> For precise variants, END is POS + length of REF allele - 1, and the for imprecise variants the corresponding best estimate.

Our VCFs and vcflib and our tests didn't agree on how the base level semantics should be for INS, DEL, and INV SVs.

The VCF spec doesn't really say either, so I've defined some semantics myself:

* For INS, the POS base is anchoring and the insert happens after it. You ought to have an SVLEN set to the number of new bases.

* For DEL, the POS base is anchoring and not deleted. You ought to have an END set to the inclusive position of the last deleted base.

* For INV, the POS base **is** inverted. You ought to have an END set to the inclusive position of the last inverted base.

This is so that you can have symbolic and non-symbolic representations of the same variant with the same POS value.

I'm going to check this interpretation of the semantics with the VCF spec folks. If they like it I am going to fix up vcflib so it respects it and doesn't do weird things like want an END not right next to POS for an insertion.


adamnovak:
Apparently Jenkins' compiler doesn't like my initializer lists.

adamnovak:
Also, @eldariont found part of the spec (in the description of the REF field), where it says that:

> If any of the ALT alleles is a symbolic allele (an angle-bracketed ID String “`<ID>`”) then the padding base is required and POS denotes the coordinate of the base preceding the polymorphism.

That would cover the inversion case too, and say that it also needs an anchor. Although there's an example contradicting this in the breakend part of the spec, I think (and it seemed to be the consensus at lab meeting) that the anchor-required version is the one I should use.

I asked the VCF folks by commenting on https://github.com/samtools/hts-specs/issues/173, and they might say differently, but for now I am going to revise this PR for the anchor-requiring interpretation.

eldariont:
That looks really good so far. I'm experiencing only two problems with this branch:
1. If the VCF records contain the END field, vcflib's `canonicalize()` complains sometimes that `END and SVLEN do not agree`. Once we have decided on the VCF semantics, we should update that method too.
2. If the VCF record does not contain the END field (e.g. for a DEL), the following call to `variant->info.at("END").at(alt_index)` gives me a `std::out_of_range` exception:
https://github.com/adamnovak/vg/blame/385c4ffaec32ba8354d7fc1d8ddf3ab77aee74b9/src/constructor.cpp#L771
Can we maybe store the end of a DEL in a variable as either `variant->info.at("END").at(alt_index)` (if the field exists) or `POS-SVLEN` (if it doesn't)?
For an INV (https://github.com/adamnovak/vg/blame/385c4ffaec32ba8354d7fc1d8ddf3ab77aee74b9/src/constructor.cpp#L806), we always need the END tag in the VCF but if it is missing, the `std::out_of_range` exception might not be that helpful either.

adamnovak:
I'm chasing down what I think is a weird memory error in vcflib right now. Once I fix that in this PR, I am going to go back to re-doing the vcflib canonicalization logic, so that it correctly handles both END and SVLEN in the same record, and so that it supplies missing values for those fields as long as the variant is fully specified, basically as you describe. Then we can swap out for the improved vcflib and have your VCFs hopefully magically start working.

edawson:
> Internally we're reading everything off the END now

I'd like to add SVLEN back to the tests, because it's pretty common to not get an END tag. Is that what you meant @adamnovak by "correctly handles both END and SVLEN in the same record?"

> That would cover the inversion case too, and say that it also needs an anchor. Although there's an example contradicting this in the breakend part of the spec, I think (and it seemed to be the consensus at lab meeting) that the anchor-required version is the one I should use.

I agree with this. The anchoring base is useful for flat alleles and for checking ref allele / reference agreement as well.

Sorry about all the bugs around.

adamnovak:
This PR still has the canonicalization code from vcflib mostly unchanged, and only works right if you specify just whichever of SVLEN or END will make it through vcflib canonicalization correctly. I have #1931 based on this with some new canonicalization that ought to properly handle both SVLEN and END being in the same variant, and not bail out complaining that they don't agree when they do agree, although it doesn't yet try and cross-check them against each other.

PR #1930, opened 2018-10-11
Tolerate there being no changes to the docs

adamnovak:
Looks like when I changed the Doxygen template in #1923, I made the Doxygen builds fail when there are no doc changes and nothing to commit to the gh-pages branch. This should fix that.


adamnovak:
This passes CI for the Doxygen build. I am going to merge it into master right now so master stops being broken and yelling at me.

PR #1931, opened 2018-10-12
New SV handling

adamnovak:
This should address #1926


adamnovak:
OK, I have talked with @eldariont and it seems like vg with this patch can handle the SVs he is interested in correctly.

PR #1937, opened 2018-10-16
Edgepack

edawson:
This adds an edge coverage vector to the pack system, as well as utilities to fill them by normalizing edges using utilities in XG. There are probably more compact ways to do this, but we'll need to maintain an index of edges, which we don't do currently.

### DIFF
--- a/src/packer.cpp
+++ b/src/packer.cpp
@@ -7,8 +7,7 @@ Packer::Packer(void) : xgidx(nullptr) { }
 
 Packer::Packer(xg::XG* xidx, size_t binsz) : xgidx(xidx), bin_size(binsz) {
     coverage_dynamic = gcsa::CounterArray(xgidx->seq_length, 8);
-    size_t g_iv_sz = xgidx->get_g_iv_size();
-    edge_coverage_dynamic = gcsa::CounterArray(g_iv_sz, 8);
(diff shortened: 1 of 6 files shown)


glennhickey:
On my chromosome 21 test with a `1.6GB` GAM:
* old pack (no edges) : `50MB` on disk, `2m15s` to run (7 cores)
* new pack (inc edges): `80MB` on disk, `2m35s` to run (7 cores)

This seems like a small enough increase to leave edge packing on by default, though it's trivial to toggle it via cli if needed down the road.

ekg:
I think the edge coverage will be important for some types of variant
detection. We can always test going forward to see how important it is. The
difference is surprising to me though. With an edge every 10bp of graph I
would expect a smaller increase, maybe only 10%. Why is the cost to store
an edge coverage greater than that for a single basepair?

glennhickey:
I should have mentioned:  These numbers are from the *augmented* HGSVC graph/GAM.  Same comparison on the original graph (which is way less edgy): 
* old pack (no edges): `46MB` on disk, `45s` to run (7 cores)
*  new pack (inc edges) : `49MB` on disk, `45s` to run (7 cores)

ekg:
Oooohhh. That makes perfect sense!

Issue #1933, opened 2018-10-15
kpath code from vg::VG should be in terms of handle graphs


adamnovak:
There's no compelling reason that @jeizenga and I can see for leaving the kpath generation code inside the VG class. We ought to factor it out and run it over the handle graph interface instead.

PR #1938, opened 2018-10-17
Allow indexing XG to GCSA with a SourceSinkOverlay

adamnovak:
This properly finishes off #1933. Our kpath code was already in terms of handle graphs, but in practice we could only index vg files, since GCSA indexing needs single start and end nodes hooked up to every graph component, and the code to do that depended on inserting nodes into the vg graph.

I've replaced that code along all the indexing codepaths with a `SourceSinkOverlay`, which synthesizes source and sink nodes on top of any backing `HandleGraph` implementation. I've had to extend the `HandleGraph` interface to report a few more statistics than it did before to make this work (e.g. I need to be able to get ahold of the max node ID in the backing graph to make IDs for the new nodes).

Using this, you can now produce a GCSA index directly from an XG index with:

```
vg index -x input.xg -g output.gcsa
```

I also had to fix parallel mode in XG's `for_each_handle` implementation, which would just loop forever on the first node and never make any progress.

We still use the old `add_start_end_markers` in `vg mod`, because applying it to mutate a graph is a `vg mod` option, and in `VG::prune_complex_with_head_tail`, because again we're mutating the graph. It's also used in the unit tests and checked against `SourceSinkOverlay` behavior. The two approaches can vary in what nodes they arbitrarily select to break into connected components with no tips, but beyond that they produce the same results.

### DIFF
--- a/src/algorithms/topological_sort.cpp
+++ b/src/algorithms/topological_sort.cpp
@@ -286,7 +286,7 @@ vector<handle_t> lazy_topological_order_internal(const HandleGraph* g, bool lazi
         
         if (orientation.size() != g->node_size()) {
             cerr << "error:[algorithms] attempting to use lazy topological sort on unorientable graph" << endl;
-            exit(1);
+            assert(false);
         }
         
         // compute the degrees by following the edges backward
@@ -331,7 +331,7 @@ vector<handle_t> lazy_topological_order_internal(const HandleGraph* g, bool lazi
     
     if (order.size() != g->node_size()) {
         cerr << "error:[algorithms] lazy topological sort is invalid on non-DAG graph, cannot complete algorithm" << endl;
-        exit(1);
+        assert(false);
     }
     
     return order;
--- a/src/algorithms/weakly_connected_components.cpp
+++ b/src/algorithms/weakly_connected_components.cpp
@@ -50,5 +50,68 @@ vector<unordered_set<id_t>> weakly_connected_components(const HandleGraph* graph
     return to_return;
 }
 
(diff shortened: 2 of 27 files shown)


adamnovak:
I've managed to introduce an exciting new bug!

Grab these files: https://ipfs.io/ipfs/QmeyqhVVYmUjJSyx4LBJbkPb9FfUkQ4a2SbbkVciRi1npH/bad-sort.tar.gz

And run this:

```
vg mpmap -B -S -A -f reads_chunk_0_0.fq.gz -t 8 -x graph.vg.xg -g graph.vg.gcsa >/dev/null
```

You get this:

```
error:[algorithms] lazy topological sort is invalid on non-DAG graph, cannot complete algorithm
```

It's weird because, even though the indexes are generated with my new code, they only fail when mapped to with the new code, and I didn't touch the mapper as far as I can tell. When mapping with the old code everything is fine.

I could have messed up my new handle encoding in the xg, but why would that only break one of many Jenkins tests?

adamnovak:
OK, I've redone the handle bit-packing code and consolidated it into one place for anyone who needs to pack an int and a bool in a handle. I must have messed up some of the bit accesses in what I had before; replacing them all with this one correct implementation seems to fix the topological sort problem.

adamnovak:
I think test_sim_chr21_snp1kg_trained failed because it managed to get a truncated GCSA file, probably due to https://github.com/DataBiosphere/toil/issues/2375

I fixed the other bug, which was a failure of VG::dagify to deal with doubly-reversing edges participating in the cycles it was dagifying. I've no idea why that only came up in this PR.

PR #1941, opened 2018-10-18
VG::edit speedup

jonassibbesen:
Changed VG::edit so that paths indexes are only rebuild after all paths have been added. This makes editing and adding paths using *vg mod -i* significantly faster.

### DIFF
--- a/src/path.cpp
+++ b/src/path.cpp
@@ -152,9 +152,10 @@ void Paths::make_linear(const string& name) {
     circular.erase(name);
 }
 
-void Paths::extend(const Path& p, bool warn_on_duplicates) {
+void Paths::extend(const Path& p, bool warn_on_duplicates, bool rebuild_indexes) {
     const string& name = p.name();
-    auto& path = get_create_path(name);
+    // Make sure we preserve empty paths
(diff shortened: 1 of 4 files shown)


adamnovak:
Jenkins, OK to test.

Also, Jenkins, add to whitelist.

jonassibbesen:
There might be some special case that I am not considering where rebuilding the indexes after each path have been added matters. However all the tests I have done so far using *vg mod -i* have given the same results before and after the change.

ekg:
I think it is not necessary. How much faster does this make it for you?

jonassibbesen:
There seem to have been an issue with GitHub's data storage system the past day resulting in my comment first getting copied and then deleted. So here it is again.

> I think it is not necessary. How much faster does this make it for you?
> […](#)

It is pretty significant. Adding 8021 single-end mapped RNA-seq reads as paths to a graph build from chr21 (no variants) using *vg mod -i* goes from taking ~30 minutes to ~11 seconds.

adamnovak:
I think Erik was talking about *rebuilding the indexes after every path* not being necessary.

This is a pretty great speedup.

Issue #1940, opened 2018-10-17
Error with VG construct


letitiaismyname:
I was trying to run vg construct with /public/groups/cgl/graph-genomes/letitia/hg19.fa
and /public/groups/cgl/graph-genomes/letitia/finalFile.vcf 
and it gave me these error messages: (copied and pasted)

```
Warning: inversion END and SPAN do not agree [canonicalize] chrX	147662854	SRR359096.160043869-A	c	c	0	.	END=147662872;SPAN=0;SVLEN=0;SVTYPE=INV
END: 147662872  SPAN: 0
warning:[vg::Constructor] Lowercase characters found in chrY; coercing to uppercase.
```

PR #1942, opened 2018-10-18
Fix VCF misparsing when the reference is soft-masked

adamnovak:
VCFlib has neglected to upper-case sequence that it pulls from the reference and demands be upper case. This uses a VCFlib that fixes that, and also now produces correct ref and alt sequences in the variant for inversions.

I've also fixed (and added some protection against) double canonicalization, which (while it ought to be idempotent) was a factor in the other bug; variants were being canonicalized, and then being canonicalized again when lower-case sequence was found in them. Now VCFlib should protect against double-canonicalization (because it's a waste of time and maybe means you made a mistake) and VG shouldn't try to do it anymore.

### DIFF
--- a/src/constructor.cpp
+++ b/src/constructor.cpp
@@ -1729,66 +1729,70 @@ namespace vg {
             // We need to decide if we want to use this variant. By default we will use all variants.
             bool variant_acceptable = true;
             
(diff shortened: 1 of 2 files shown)


adamnovak:
This should fix #1940

PR #1951, opened 2018-10-23
Target value search

xchang1:
@jeizenga I think this is correct but it's probably going to be slow


adamnovak:
The Linux Travis jobs failed because they decided that g++-6 wasn't a package. I've restarted them and they seem to be reconsidering that conclusion.

PR #1954, opened 2018-10-23
Fix #1949 by implementing rare-variant-removing simplification

### LINKED ISSUE #1949: Implement graph subsetting to remove rare variants
In order to get anywhere with mapping to graphs the size of TOPmed, we need to reduce the size of the graph before we do the pruning and GCSA-indexing.

We still want to talk to @benedictpaten about this, but @jltsiren and I figured the right approach is to have a step where we filter the extremely rare variants and their ref and alt paths out of the constructed graph, leaving a subgraph that only covers the primary path and the relatively common variants that will be small enough to haplotype-prune and index. If we remove the variant ref and alt paths for the variants we skip, then we can use the normal GBWT construction method to build a GBWT without the rare variants, in order to support pruning (although we'll want to turn off the missing variant warnings for that step, which is probably its own issue).

So this issue tracks the implementation of the subsetting functionality we need. We need to be able to take in a vg graph with alt paths, and a VCF, both of which may be quite large, and remove all the nodes that aren't either on the primary path or part of a variant with a total AC at or above a user-specified threshold. We also need to remove the alt and ref paths tracing out variants below that AC threshold.

It's probably OK to leave edges that are involved in rare variants but connect nodes that are also involved in common variants. It's also probably OK to keep rare alleles of some VCF records, as long as the record's total AC across all alts meets the threshold. Really we just need to get rid of the singletons and super-rare variants, because we need to cut *something* in order to be able to build the GCSA index.

adamnovak:
This modifies `vg simplify` to have another simplification algorithm, which removes rare variants from the graph using a VCF file, leaving the subset that #1949 demands.

I wanted to do this through handle graphs, so I had to add a path-modification handle graph interface. It's about as empty as I could make it to make this work; all you can do is append to paths, create paths, and delete paths.

I also had to tinker with the handle graph path interface itself, because we had no way to know if a path name was valid for grabbing a handle besides trying and crashing the program if it wasn't. Similarly for trying to get the first/last occurrences, because a path can be empty and not have them.

Also, some of the logic to go between path IDs and path names didn't use the accessors, so paths were sometimes not getting assigned IDs, which caused all sorts of trouble.

I also had to fix vcflib to take string constants as info tag names in its accessors.

Still TODO:

- [x] Work out why the old simplifier (SmallSnarlSimplifier) doesn't pass its tests. I suspect it was broken before, since it never had any real unit tests, but it's possible this PR breaks it.
- [x] Fix `VG::has_next_occurrence`, which crashes my loop when I use it to loop over all the occurrences on a path. I'm using a comparison against the last occurrence right now instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -485,16 +485,21 @@ $(OBJ_DIR)/version.o: $(SRC_DIR)/version.cpp $(SRC_DIR)/version.hpp $(INC_DIR)/v
 
 # Define a default rule for building objects from CPP files
 # Depend on the .d file so we rebuild if dependency info is missing/deleted
+# Make sure to touch the .o file after the compiler finishes so it is always newer than the .d file
 # Use static pattern rules so the dependency files will not be ignored if the output exists
 # See <https://stackoverflow.com/a/34983297>
 $(OBJ) $(OBJ_DIR)/main.o: $(OBJ_DIR)/%.o : $(SRC_DIR)/%.cpp $(OBJ_DIR)/%.d $(DEPS)
 	. ./source_me.sh && $(CXX) $(CXXFLAGS) -c -o $@ $< $(LD_INCLUDE_FLAGS) $(FILTER)
+	@touch $@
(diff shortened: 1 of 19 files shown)


adamnovak:
It turns out that `has_next_occurrence` works fine, and I just had the wrong sense in my path emptiness check.

This PR ought to be ready now, if it passes tests.

PR #1961, opened 2018-10-26
Add a flowchart of the vg pipeline

adamnovak:
I'm adding a flowchart we can maybe use on the wiki.

Jenkins, please ignore. This really shouldn't break the build.


adamnovak:
Jenkins why don't you listen to me.

PR #1968, opened 2018-10-30
Allow RepresentativeTraversalFinder to find traversals through inversions

adamnovak:
The RepresentativeTraversalFinder has previously been restricted to ultrabubbles.

I've extended it so that it ought to work for snarls that are start-end reachable with a-directed-cyclic net graphs. So it should now be able to deal properly with reversing edges.

It now knows about which orientation it manages to find its way back to the reference path in, and can handle matching up partial traversals on the left and right so it finds a consistent orientation of the reference path.

We need this to support calling inversions.


adamnovak:
I've had to touch the way edge support is assigned to strand. In general we were looking at support in nodes' local strands and not the traversal's strands. For edges we don't really have a proper concept of strand. Right now I'm using low ID to high ID as the forward direction, and hoping that that persists through augmentation. Someone ought to go through the augmenter and make sure that if it changes the direction of an edge it flips the supports for the edge too.

What I have now beats what we had before where end-to-end edges had all support on the forward strand and start-to-start edges had all support on the reverse strand.

I'm also now smearing child snarl node support across both strands, because I don't have canonical orientations for nodes in child snarls.

PR #1969, opened 2018-11-01
Fix compile on my machine

jeizenga:
The clang version I'm using (below) doesn't like it when we use non-const references to lvalues, which showed up in some of the VCF library's code we're using. The silly, simple fix I did was just to convert the strings that it was complaining about with rvalues.

```
Apple LLVM version 10.0.0 (clang-1000.11.45.2)
Target: x86_64-apple-darwin18.0.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
``` 

It also looks like another commit from my TVS branch found it's way in here, but it won't hurt anything.


adamnovak:
I think you just forgot to update the vcflib submodule. I modified it specifically to use const references for those strings so we could pass constant strings.

jeizenga:
Yep. I think you're right. Nevermind then, my bad.

Issue #1975, opened 2018-11-05
Potential false negative SV calls by (toil-)vg call


eldariont:
Hi,

following up on today's call I open up this issue to discuss reasons for potential FN SVs in `vg call`. So this is what I did:

I have 2 PacBio yeast assemblies: S288C and Y12. I aligned Y12 against S288C with nucmer and used Assemblytics to find SVs between the two. Then I used the resulting VCF to generate a vg graph with `vg construct`. Now I mapped Illumina reads from the same Y12 sample to the graph and ran `toil-vg call` to call SVs:
```
toil-vg call --vg_docker quay.io/vgteam/vg:v1.11.0-74-gdab42acd-t242-run --nodeTypes c5.2xlarge:0.2,r4.large --minNodes 0,0 --maxNodes 1,1 --provisioner aws --batchSystem mesos --mesosMaster=${MASTER_IP}:5050 --metrics aws:us-west-2:vgcall-yeast-construct-jobstore $XG SRR4074383 aws:us-west-2:vgcall-yeast-construct-outstore --gams $GAM --recall --chroms chrI chrII chrIII chrIV chrV chrVI chrVII chrVIII chrIX chrX chrXI chrXII chrXIII chrXIV chrXV chrXVI 2> SRR4074383.construct.log
```

I now compared the resulting VCF file with the original VCF used to construct the graph and measured recall. I looked at the false negatives and noticed several examples where I don't understand why the variant wasn't at least called as heterozygous. I wouldn't be surprised if there were good reasons for not calling from the alignments but I don't understand `vg call` good enough to see these reasons yet.  Can anybody help?

Here are 3 of the examples:

**Example 1 (chrIII:148094, 104bp insertion)**
![grafik](https://user-images.githubusercontent.com/6477692/48026844-0ed4af00-e0fc-11e8-8bcc-c15fb4eb64d2.png)

**Example 2 (chrII: 197780, 333bp insertion)**
![grafik](https://user-images.githubusercontent.com/6477692/48027132-c8cc1b00-e0fc-11e8-9ab5-079744e88c49.png)

**Example 3 (chrXII: 9630121, 333bp insertion)**
![grafik](https://user-images.githubusercontent.com/6477692/48027568-ff566580-e0fd-11e8-977f-ab571906a878.png)

The data can be found here:
**graph**: /public/groups/cgl/users/daheller/yeast_graph/sequenceTubemapData/construct.vg
[...]

PR #1978, opened 2018-11-06
Allow better support for SV recall in vg call

glennhickey:
#1975 showed examples of toil-vg's --recall option missing some large insertions.  This was happening because the reads were not supporting the insertion sequence exactly as it is in the graph, presumably due to sequencing errors somewhere.  

Using this patch, along with caller options '-u -n 0' seems to work for the examples.  I'll fix toil-vg call --recall to use these options by default in a minute.


glennhickey:
This PR along with https://github.com/vgteam/toil-vg/pull/638 should resolve the three examples in #1975.

PR #1979, opened 2018-11-07
Add a timer to mpmap to warn about reads that are taking a long time to map

adamnovak:
Now mpmap should warn about reads that are taking more than 20 minutes to map, which will hopefully help with identifying reads that trigger shortcomings in the mapping algorithm.

The warning threshold time should probably be configurable from the command line, but I didn't want to add yet another CLI option that nearly all users will not care about, so right now it is hardcoded.


adamnovak:
I'm also tracking peak memory usage growth,while each read is being mapped, because some slow reads also drive huge increases in memory usage and we also want to identify those.

PR #1981, opened 2018-11-09
Indexing VG files for random access

adamnovak:
I've generalized the GAM indexing machinery to allow indexing ID-sorted VG files as well, and I'm working on a HandleGraph implementation backed by an indexed on-disk VG file.

Right now performance is going to be terrible, because it is re-opening the file for every access and deserializing loads of things repeatedly. But with a bit of pooling of cursors and cacheing of some deserialized graph chunks, and if we make the vg file chunks/runs small enough, I could see performance being reasonable even with constrained memory.

I have no tests yet, so I don't even know if this works, but I wanted to put it up to show that I am working on it.


adamnovak:
OK, I'm going to merge this in. We don't actually use the indexed VGs for anything yet, but I'm hoping to hook them up to e.g. vg find after I convert a bunch of the operations to use the handle interface.

ekg:
Great work.

PR #1991, opened 2018-11-19
Don't context-expand paths in vg chunk

glennhickey:
Expanding back from the start of a subgraph and hitting a deletion could put a hole in the reference path if there are not enough context steps to cover it.  Holes in the reference path can then throw off offset computations downstream.  This came up when doing whole-genome calling on a SV graph (https://github.com/vgteam/toil-vg/issues/649), as the larger the deletions are the more likely to run into this problem.


adamnovak:
So the new approach is to still do context expansion that can pull in some additional ref path material, but to only keep the contiguous subpath that was originally pulled out, and not to add path visits to the new nodes?

PR #2020, opened 2018-12-10
Plug download

adamnovak:
On the vg call today, the issue came up that nobody knows where our binary downloads are. So far they've just been a little "vg" link on each release, with nothing telling people about them.

I'm adding big buttons to the README and also the individual releases so that the Linux binaries are easier to find. This should hopefully reduce the number of people who need to build from source, and especially the number of people who need to build from source on really old HPC Linux.

I'm also making the vg logo smaller in the README because I got tired of it being so huge. Our logo is great, but probably not so great that everyone needs to see it that large.


adamnovak:
Jenkins, skip this please.

PR #2025, opened 2018-12-12
Haplotype-aware linearization for haplotype scoring, with MultipathAlignmentGraph refactoring and bugfixes

### LINKED ISSUE #2012: Make optimal_alignments slightly haplotype aware
Right now, when we do haplotype-aware mapping, we are trying to rescue reads that have errors in their true placements that happen to match SNPs in incorrect placements. We are operating on the theory that the SNP in the incorrect placement should cause the read there to become haplotype-inconsistent, or at least less haplotype consistent. Conversely, in the correct placement, the mismatch at best is just a mismatch, and at worst matches a SNP alt and makes the correct placement also haplotype-inconsistent in its optimal mapping.

We're generating a bunch of suboptimal tracebacks from MultipathAlignments, and trying to combine their alignment scores and haplotype consistency scores into a unified score. But we aren't biasing our selection of suboptimal alignments towards haplotype consistency at all, so what we see, especially in graphs like TOPmed, is one optimal-alignment-score traceback, and then a bunch of lower-alignment-score tracebacks that brute force every SNP mismatch from the read, and then a lower tranche trying all combinations of two SNP mismatches, and so on.

When we have a SNP every 8 bases or so in a 150 bp read, we need to make sure to try mismatches at the correct SNPs if we want to get a haplotype-consistent suboptimal traceback in the 10 or so we are asking for,

I think we need to change how optimal_alignments works so that it can take in info from the GBWT and keep a haplotype consistency count with each traceback in progress. Then we need to keep all haplotype-consistent partial tracebacks with a given alignment score in the queue, if any are in, and sort the queue by consistent haplotype count after sorting by alignment score.

That will let us guarantee that, if there is a haplotype-consistent traceback with a given alignment score, and we produce any tracebacks with that alignment score, we will produce the haplotype-consistent one first.

That solves the brute-force-all-the-single-SNPs problem, although we will probably run out of requested tracebacks before getting into those that require two base errors or an indel error.

We could also try and mix haplotype consistency scoring with support for recombinations into the multi-traceback logic. Maybe we need some form of A* or something?

adamnovak:
This is my attempt to let haplotype information guide traceback generation for haplotype consistency scoring of MultipathMappings.

I'm just generating all tracebacks that are haplotype-consistent. If the multipath has too many ways to spell the same linear alignment, this could explode. So I'm going to try testing it whole-genome before really merging it.

This should resolve #2012.

### DIFF
--- a/src/algorithms/extract_extending_graph.hpp
+++ b/src/algorithms/extract_extending_graph.hpp
@@ -21,9 +21,11 @@ namespace algorithms {
     /// Fills graph 'into' with the subgraph of the handle graph 'source' that extends in one direction from
     /// a given position, up to a maximum distance. The node containing the position will be "cut" so that only
     /// the portion that is forward in the search direction remains. Node IDs may be changed in the extracted
-    /// graph, but they can be translated back to node IDs in the original graph with the returned map. The
-    /// node containing the source node may optionally be duplicated to preserve cycles on it after its cut,
-    /// but no other nodes will will duplicated.
+    /// graph, but they can be translated back to node IDs in the original graph with the returned map, although
+    /// that translation procedure MUST handle the node that pos is on specially, as it may be cut.
+    /// translate_node_ids from path.hpp can do this as long as you pass along what part of the node was removed.
+    /// The node containing the source position may optionally be duplicated to preserve cycles on it after its
+    /// cut, but no other nodes will will duplicated.
     ///
     /// Args:
     ///  source                  graph to extract subgraph from
--- a/src/haplotypes.cpp
+++ b/src/haplotypes.cpp
@@ -829,6 +829,19 @@ int64_t ScoreProvider::get_haplotype_count() const {
   return -1;
 }
 
+bool ScoreProvider::has_incremental_search() const {
+  // By default, say that we lack incremental search support.
+  return false;
+}
+
+IncrementalSearchState ScoreProvider::incremental_find(const vg::Position& pos) const {
+  throw runtime_error("Incremental search not implemented");
+}
+
+IncrementalSearchState ScoreProvider::incremental_extend(const IncrementalSearchState& state, const vg::Position& pos) const {
+  throw runtime_error("Incremental search not implemented");
+}
+
 /*******************************************************************************
 XGScoreProvider
 *******************************************************************************/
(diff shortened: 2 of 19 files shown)


adamnovak:
Looks like sourceware.org went down and we couldn't clone the dependencies.

Jenkins, test this please.

adamnovak:
OK, I'm still seeing slightly worse performance with my haplotype-aware tracebacks and in-tail anchor synthesis. Not making the tail anchors (and hence not really considering all snarl combinations in the tails) does better:

Conditition | Wrong read count from ~63k difficult reads
--- | ---
Primary | 39375
TOPmed | 41515
TOPmed with Haplotypes and haplotype-aware traceback | 40029
Also with in-tail anchor synthesis | 40105

I've looked at some of the failures of the in-tail anchor synthesis. Mostly it seems to be making wrong things that had MAPQ 0 now have nonzero MAPQ, possibly also making them wrong, by promoting one formerly-equivalent alignment above another with haplotype scoring.

Also, I've had it cause the correct alignment to not get found by disabling secondary rescue, by lowering the MAPQ of an incorrect alignment so the mapper no longer thought it might be overconfident.

I'm also seeing MultipathAlignments get split up as having multiple connected components, when doing the in-tail anchors, when I don't think they should be. And I know I'm producing a lot of duplicated PathNodes (from snarl cutting on substantially overlapping tail alignments) and hence subpaths, and hence duplicated tracebacks. I want to cut that down before merging this.

adamnovak:
OK, I've worked out that my alignments are falling apart into multiple components because the alignments for the tails aren't being translated back from the graph from extract_extending_graph into the `align_graph`'s space when extract_extending_graph cuts nodes. This makes PathNodes that I generate as tail anchors appear to overlap in graph space with the PathNodes they are supposed to come after, so no connecting graph between them can be found and the edges linking them with the rest of the MultipathAlignmentGraph get deleted.

adamnovak:
OK, I've fixed the alignments-falling-apart bug I think, and because of that I'm getting better results:

Conditition | Wrong read count from ~63k difficult reads
--- | ---
Primary | 39375
TOPmed | 41515
TOPmed with Haplotypes and haplotype-aware traceback | 40029
Also with in-tail anchor synthesis | 39734

Still not better than primary, but I think it might be worth merging.

adamnovak:
Going back to the non-haplotype-aware traceback (just the top 10 alignments in alignment score) produces 40041 wrong reads, and doing that while also turning off tail anchor generation produces 40212 wrong reads. So I'm pretty confident that this is an improvement, and should be merged.

Still TODO is to work out what's up with the additional wrong reads we still have over mapping to primary, and what can be done about them.

Issue #2028, opened 2018-12-13
Default parameters for PacBio/Nanopore long reads and contigs in map and mpmap


ocxtal:
For now, we don't have default parameter features in vg map / mpmap commands. It seems the defaults in these commands looks like for Illumina reads, and not appropriate for mapping long reads nor contigs.

I'd like to propose the following:

* extracting current default parameters (`-q1 -z4 -o6 -y1 -L5`) as default for Illumina single and paired (just `illumina`).
* adding `pacbio` and `nanopore` defaults as `-q2 -z4 -o4 -y2 -L0`, which is minimap2's for them (equivalent to  `-q1 -z2 -o2 -y1 -L0`). It is also possible to use smaller `--min-mem` or larger `--mem-chance` to reduce possibily that we fail to find true anchor due to high error rate.
* tentatively setting `contig` (assuming assembled contigs of several kilobases to megabases with  >99% identity) the same as `illumina`

I think first two are reasonable while I'm not sure for the last one. Considering correctly assembled contigs keep good identity, it would be better to set mismatch and gap penalties higher to detect low-identity regions sensitively (which indicate misassemblies or structual variations).

For example, mismatch and gap-open penalties are set as large as 19 and 39 for 5%-div. mode in minimap2:
```
asm5    Long assembly to reference mapping (-k19 -w19 -A1 -B19 -O39,81 -E3,1 -s200 -z200 --min-occ-floor=100).  Typically, the alignment will not extend to regions with 5% or higher sequence divergence. Only use this
        preset if the average divergence is far below 5%.
asm10   Long assembly to reference mapping (-k19 -w19 -A1 -B9 -O16,41 -E2,1 -s200 -z200 --min-occ-floor=100).  Up to 10% sequence divergence.
asm20   Long assembly to reference mapping (-k19 -w10 -A1 -B4 -O6,26 -E2,1 -s200 -z200 --min-occ-floor=100).  Up to 20% sequence divergence.
```

Related issues:
https://github.com/vgteam/vg/issues/770
https://github.com/vgteam/vg/issues/1881
https://github.com/vgteam/vg/issues/2026

Thanks,

Hajime Suzuki

ocxtal:
It should be useful just importing `asm5`...`asm20` parameters from minimap2. A concern is that the gap penalty model is more complicated in minimap2 so we have to simplify them (choosing smaller for the gap-open and use the first one for gap-extend?)

ocxtal:
[...]

PR #2029, opened 2018-12-13
add presets in map/mpmap

ocxtal:
implementation for #2028


ocxtal:
@adamnovak Thanks for comment. I fully agree with you and fixed the behavior to give priority to individual parameters. Unrecognized argument is now treated as an error.

@jeizenga I've merged the parser for -E/--long-read-scoring option to --preset for simplicity. I also made the long_read_scoring flag set for PacBio and Nanopore presets.  Which do you think is better for contig / asm{5,10,20} presets (set or not)?

adamnovak:
I've updated this to agree with the latest master. We can probably go ahead and merge it if the tests pass.

ekg:
@ocxtal sorry, I should have merged this in because I just did something similar. It needs to be sync'ed with master.

fbemm:
What is the current preset / parameter set for mapping contigs against a graph?

jeizenga:
The codebase has diverged quite a bit since this PR and something similar now exists in both map and mpmap. Some of the presets might still be worthwhile additions, but I think it would be as easy to start this PR over again as to sync it up.

Issue #2017, opened 2018-12-10
Potential MEMChain distance bug when mapping near graph fork


kevyin:
Hi vgteam

![vg_memchain](https://user-images.githubusercontent.com/141234/49716201-1cef8f80-fca7-11e8-8cb9-9faca48df746.png)

When we map a read pair with  vg map --interleaved we notice only a single end maps, but when we map in single end mode both ends map.
Digging a little deeper 
At this line:
https://github.com/vgteam/vg/blob/6d6e41aa1a38870e70d67b1607bbd214ff75293b/src/cluster.cpp#L135
the `abs(p->first` - q->first) < band_width` becomes `18654 < 5000`
18654 seems to be adding the distance from 451,450,449..... etc

Code and graph and read files:
https://gist.github.com/kevyin/00d5fbcf29691fba770850665b29fbe8
```
prefix=subset_9
vg view -F ${prefix}.gfa -v > ${prefix}.vg
vg index ${prefix}.vg -x ${prefix}.vg.xg -g ${prefix}.vg.gcsa

vg map -t 8 --interleaved  -x ${prefix}.xg   -g ${prefix}.gcsa --min-ident 0.5   -f read_pair.fq  > read_pair.gam
vg stats -a read_pair.gam ${prefix}

```

Incidentally I tried mpmap and the read seems to have mapped, could this be a known issue and I should use mpmap instead?
`vg mpmap -b 1 -I 500 -D 500 -S `

Thanks

adamnovak:
Hm. I thought that this might be vg misusing the path-based distance heuristics, but it looks like your GFA file doesn't have any paths. I would have thought that map would be better in this case than mpmap, but you say mpmap is working and map isn't?

@jeizenga and @xchang1 have been working on better distance estimation logic in mpmap for cases like this. Perhaps that's helping here?

kevyin:
Would it be easy to insert the newer I'm guessing MEMClusterer? methods in vg map?

kevyin:
Looks like approx_position is using nodeids as an estimate of distance between MEMs?
https://github.com/vgteam/vg/blob/6d6e41aa1a38870e70d67b1607bbd214ff75293b/src/mapper.cpp#L4346

452, 13 would then be very "far apart"

Am wondering if there could be a combination of `vg ids or vg mod` that would "normalise" these situations?

ekg:
We should bring in the new distance estimation method (based on bubbles) to
[...]

PR #2038, opened 2018-12-17
#2017 pair_consistent orientation checks on same node, map fallback to ignore approx dist

kevyin:
…ignore approx dist
Some workarounds for issue #2017
* option args to pair_consistency to ignore dist and orientation
* during pair_consistency where there are no Paths, find a common node between the two alns and compare orientation. Will return a more accurate answer than previous, but relies on alns having a common node.
* for vg map: retry with pair_consistency ignoring approx dist, if the first check fails. Current implementation unfortunately will repeat the check for orientation on the retry


ekg:
Ok to test.
Jenkins, please test this please.

PR #2049, opened 2019-01-08
Synthetic anchors

adamnovak:
Here's a performance comparison vs #2025:

Conditition | Wrong read count from ~63k difficult reads
--- | ---
Primary before | 39375
Primary now | 39366
TOPmed before | 39734
TOPmed now | 38578
TOPmed now without GBWT | 40702

I've managed to improve TOPmed performance to beat primary, on this wrong read dataset, without adversely impacting alignment against the primary graph. (Weirdly, when I had a bug where I accidentally destroyed all the alignments on edges and returned empty vectors, I got even fewer wrong reads against TOPmed, but Primary was terrible.)

The overall approach is to scan out from bases in MEM anchors, reading through a certain number of mismatches (2, here) to find new matches between the graph and the read and create new perfect match anchors. The new created anchors can only touch single nodes. This gets better coverage of the different possible diagonals in the alignment.

I also then drop the anchors that have been formed within snarls, leaving only the anchors on the snarl boundary nodes.

Doing this can create combinatorial explosions in low-complexity regions, where multiple plausible anchors on either side of a snarl end up with reachability edges between all pairs across the snarl.

image

All the edges between PathNodes on node 18158962 and PathNodes on node 18158957 cross the same snarl, and we have more or less all-against-all connectivity across it.

Then each reachability edge can balloon out into multiple subpaths, because we look for multiple alignments across the snarl between each connected pair of anchors. So the MultipathAlignmentGraph gets really wide:

image

Things are actually fine until we look for all the haplotype-consistent tracebacks through the MultipathAlignmentGraph, at which point a combinatorial explosion happens between all these different subpaths that follow only a few actual possible walks through the snarl.

I tried to clamp down on this by deduplicating alignments, keeping just the best alignment for each walk of visited nodes. That helped a bit, but not enough.

What I eventually ended up doing was putting a hard cap on the queue size when generating haplotype-consistent tracebacks through the MultipathAlignmentGraph. So now we don't get *all* the haplotype-consistent alignments in cases where we have this sort of explosion. It's a hack, but it still lets us beat primary.

@jeizenga My approach is quite hacky, but I think it indicates that it will in fact be worth generating a better and more diverse set of anchors from the initial MEM positions. Do you have any idea how to handle the all-combinations-across-snarls problem in a principled way?

Also, without Jenkins, this is tough to benchmark from a speed perspective. I suspect it is much slower, since we're re-doing all the PathNodes a few times. Once we settle on exactly what anchors we want, we should just construct them from the start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,11 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-6 60 --slave /usr/bin/g++ g++ /usr/bin/g++-6; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then mkdir -p deps; touch deps/gcc6; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PKG_CONFIG_PATH="/usr/local/lib/pkgconfig:$PKG_CONFIG_PATH"; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" && "$INSTALL_GCC" == "1" ]]; then brew install gcc6 || true; fi
+  # Try installing gcc6 twice in case of errors like this:
+  # Error: HOMEBREW_LOGS was not exported!
+  # Please don't worry, you likely hit a bug auto-updating from an old version.
+  # Rerun your command, everything is up-to-date and fine now.
+  - if [[ "$TRAVIS_OS_NAME" == "osx" && "$INSTALL_GCC" == "1" ]]; then brew install gcc6 || brew install gcc6 || true; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" && "$INSTALL_GCC" == "1" ]]; then brew link --overwrite gcc@6; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" && "$INSTALL_GCC" == "1" ]]; then mkdir -p ./bin; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" && "$INSTALL_GCC" == "1" ]]; then ln -sf `which g++-6` ./bin/g++; fi
--- a/src/algorithms/extract_extending_graph.hpp
+++ b/src/algorithms/extract_extending_graph.hpp
@@ -21,9 +21,11 @@ namespace algorithms {
     /// Fills graph 'into' with the subgraph of the handle graph 'source' that extends in one direction from
     /// a given position, up to a maximum distance. The node containing the position will be "cut" so that only
     /// the portion that is forward in the search direction remains. Node IDs may be changed in the extracted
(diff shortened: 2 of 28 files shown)


adamnovak:
I merged in some bugfixes to the haplotype scoring code, and now I'm seeing a bit better performance:

Conditition | Wrong read count from ~63k difficult reads
--- | ---
Primary | 39366
TOPmed 38400

I'm going to investigate the remaining wrong reads now.

adamnovak:
OK, I've investigated a few wrong reads. Unlike last time, I'm not looking at reads that are correct in primary; this is just sampling from all wrong reads in my wrong reads dataset.

Read | Cause | Centromeric?
---|---|--
seed_102_fragment_102605_ | Missing variants in TOPmed at true location | N
seed_146_fragment_85345_ | A haplotype-consistent mapping better than the true mapping exists | Y
seed_97_fragment_93372_ | A haplotype-consistent mapping better than the true mapping exists | Y
seed_147_fragment_33909_ | Wrong with MAPQ 0; chose an equally-good consistent mapping | N
seed_150_fragment_101683_ | A haplotype-consistent mapping better than the true mapping exists | Y

It looks like we have a lot of wrong centromeric reads, where errors in the read produce a better mapping than the real source mapping. All centromeric mappings are going to be haplotype-consistent, because there are no variants there.

I also saw a read that looked to be mismapped because the variants used to generate it weren't present in the graph at the correct location, while there was a better mapping to another location. And one of my reads was just completely ambiguous, with multiple equally-good placements.

I'm going to try restricting to just the wrong reads that Primary gets right, to see if any are left. I'm also going to stratify the wrong read count by MAPQ. It could be that there just aren't many of the remaining wrong reads that we can plausibly fix.

adamnovak:
I found that this implementation gives me a bunch of MAPQ 60 wrong reads that primary doesn't get. I surveyed 5 of them, and 4 of them can be blamed on my anchor synthesis code and snarl cutting just not getting along (in that having snarl cutting on somehow prevents anchors that should be found from being found and used in the final alignment), while the remaining one looks like it should be blameable on anchor exploration not considering indel errors, but can also be fixed by turning off snarl cutting.

Read | Reason
---|---
seed_118_fragment_97418_ | Correct mapping became terrible due to anchor exploration not considering short indels
seed_123_fragment_27157_ | Buggy(?) interaction between anchor synthesis and snarl cutting
seed_132_fragment_55387_ | Buggy(?) interaction between anchor synthesis and snarl cutting
seed_145_fragment_12030_ | Buggy(?) interaction between anchor synthesis and snarl cutting
seed_145_fragment_28523_ | Buggy(?) interaction between anchor synthesis and snarl cutting

I still think better anchors is the right way to go, @jeizenga, but when you do your implementation you should carefully integrate it with (or just throw away) the snarl cutting algorithm, which is basically designed to do the same thing and pare back poorly-considered anchors.

adamnovak:
I believe that @jeizenga has written a better approach to this problem.

PR #2053, opened 2019-01-10
Dagification in handle graphs

jeizenga:
I implemented the dagify algorithm in a handle graph, making a few modifications to the original. Notably, I changed the semantics a bit so that this must now be called on a single-stranded graph. To execute it on a non-single-stranded graph, you will first need to call split_strands.

Along the way, I also re-implemented Eades' algorithm (again on single-stranded graphs) for handle graphs so I could use it in my dagify implementation. This can replace the old Eades algorithm, which seems to have been implemented with an unnecessary logarithmic factor on the run time.


jeizenga:
Oh yeah, another cool thing in here is the SubHandleGraph, which is a convenient HandleGraph interface that let's you treat a small subgraph of a larger HandleGraph as a HandleGraph in its own right.

adamnovak:
@jeizenga It looks like maybe you forgot to commit some new `structures` code?

Issue #2036, opened 2018-12-14
Refactor the ordering concept out of HandleGraph interface and subdivide the MutableHandleGraph


jeizenga:
I have some ideas about how to organize the inheritance tree of the HandleGraph better.

- I don't like that we have swap_handles as method of MutableHandleGraph. It bakes in the concept that graphs have an intrinsic ordering, which I don't like. Intrinsic orderings clash with some good backing data structures, namely hash tables. I propose to instead require that HandleGraphs provide an external Ordering class, which can optionally be provided to for_each_handle to ensure a specific order of iteration. This would allow a HandleGraph implementation to maintain responsibility for its internal ordering or export the responsibility to the Ordering class.
- I think that we should subdivide further the MutableHandleGraph interface. In particular, I think deleting nodes and edges is actually a pretty substantial increase in expressivity over creating nodes/edges and dividing nodes. Most importantly, there's no really good way to maintain paths in a graph that can delete. I think that there should be a NoDeleteMutableHandleGraph and a full MutableHandleGraph. Further, the MutablePathMutableHandleGraph should inherit from the NoDeleteMutableHandleGraph, and not MutableHandleGraph.

Any thoughts on this are welcome.

adamnovak:
I am +1 on dropping the intrinsic ordering and also on factoring out deletion into its own interface, since we don't always want to expose it.

I'm not sure how to factor ordering if it's not built in, though. Is there any kind of Ordering class implementation that makes sense on both a VG and an XG, without doing loads of expensive `get_handle` calls on the XG? Why pass the Ordering to handle iteration rather than letting the Ordering just give you the handles in order from the graph?

jeizenga:
Sure, I think that makes sense. Really, I just want it to be somewhat removed from the HandleGraph. I see how the factoring you're suggesting could be cleaner though.

ekg:
I agree that the ordering needs to be factored out of the handlegraph interface itself, or defined in a different way in the interface that's not implicit.

PR #2057, opened 2019-01-15
Separate "mutable" and "deletable" in handle graph interface

jeizenga:
This PR partially addresses the concerns discussed in https://github.com/vgteam/vg/issues/2036.

I've subdivided the MutableHandleGraph interface into two. One interface allows addition of new graph material and manipulation of existing graph material. The other interface inherits from the first and adds the ability to delete graph material.

The overriding motivation here is to provide an interface with no deletes, which can then guarantee semantic consistency of embedded paths. However, our implementation is currently lagging behind the specification. The only merged MutableHandleGraph implementation is VG, which doesn't work as hard to maintain path consistency as you might hope. That means there are currently no fully compliant implementations to this interface, but I think it's a good roadmap barring further discussion about what semantics we want to see.

### DIFF
--- a/src/algorithms/extract_connecting_graph.cpp
+++ b/src/algorithms/extract_connecting_graph.cpp
@@ -15,7 +15,7 @@ namespace algorithms {
 using namespace structures;
 
 unordered_map<id_t, id_t> extract_connecting_graph(const HandleGraph* source,
-                                                   MutableHandleGraph* into,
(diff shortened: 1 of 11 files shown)


ekg:
I said I was OK with this on the call. I guess it doesn't matter
practically as it's all composable, but I worry we are adding too much
complexity in further subdividing this. Not a big deal though.

PR #2060, opened 2019-01-17
Set up for Gitlab CI

adamnovak:
This should get our CI going on the new Gitlab infrastructure at https://ucsc-ci.com.

It *almost* probably works now. We'll see if it builds this PR.


adamnovak:
I think I can get this working for master, but actually it looks like it will never really work right for our workflow, because of https://gitlab.com/gitlab-org/gitlab-ee/issues/5667

Gitlab CI doesn't have a way to run CI on PRs from Github forks. When Gitlab CI is working on Gitlab projects, the CI pipeline just runs when the fork has commits pushed to it, in the context of the forked project, and the CI results on the merge request are pulled from that.

There's also definitely no Jenkins-like whitelist functionality to approve people to have their PRs run.

I should have caught this a week ago, instead of after spending a week writing us a Gitlab CI pipeline that now doesn't really do what we want.

We could rearrange our workflow to all be branch-in-vgteam/vg-based (I think @ekg does that already), or we could not use Gitlab for our CI. Or we could have different CI for external PRs and vgteam development.

adamnovak:
The good news is that just dumping the PR commit into a branch in the main repo lets CI run, and the status gets propagated back to the PR. Maybe we can work with that?

adamnovak:
Alright, despite the shortcomings vs. our old setup (no comment posts (yet), no way to build fork PRs without dragging the commit into the main repo, no reports when tests fail because of the way I've done the pipeline), I'm going to merge this in. The new setup is definitely way faster; because it's parallel it can run through in a couple hours, with one particular test being the bottleneck.

glennhickey:
Anyway to get travis to forward pr's to temporary vgteam branches to get CI
running?

adamnovak:
I hadn't thought of making Travis do it.

We could, but I don't think there'd be a way to protect it at all; any
PR would be able to come in and mess with the S3 stuff that the tests
need to write.

Maybe if I change around the way we do AWS authentication, so we can
use access keys that only get exposed to master builds?

PR #2067, opened 2019-01-22
Force path endpoints to be tips in chunk graphs

### LINKED ISSUE #2066: Problem snarling chunked SV graph
Context expanding over deletions in vg chunk can leave holes in the reference path.  vg chunk handles this by clipping the reference path at the first sign of discontinuity.  In certain cases, this can lead to a path whose last node is not on a graph tip in the chunk.  This in turn will cause the cactus snarl finder to not use the reference path when looking for telomeres.  Then the telomeres it does find (via pingpong) sometimes don't seem to make a terrific root for the cactus graph, as it then recurses itself into oblivion while enumerating snarls.  

tldr:  dangling ends from cutting in complex regions can lead to crashes when finding snarls in the resulting chunks.

glennhickey:
Resolves #2066

### DIFF
--- a/src/chunker.cpp
+++ b/src/chunker.cpp
@@ -111,23 +111,27 @@ void PathChunker::extract_subgraph(const Region& region, int context, int length
     // Cut our graph so that our reference path end points are graph tips.  This will let the
     // snarl finder use the path to find telomeres.
     Node* start_node = subgraph.get_node(mappings.begin()->node_id());
-    if (!mappings.begin()->is_reverse() && subgraph.start_degree(start_node) != 0) {
-        for (auto edge : subgraph.edges_to(start_node)) {
-            subgraph.destroy_edge(edge);
-        }
-    } else if (mappings.begin()->is_reverse() && subgraph.end_degree(start_node) != 0) {
-        for (auto edge : subgraph.edges_from(start_node)) {
-            subgraph.destroy_edge(edge);
+    if (rewrite_paths && xg->position_in_path(start_node->id(), region.seq).size() == 1) {
+        if (!mappings.begin()->is_reverse() && subgraph.start_degree(start_node) != 0) {
+            for (auto edge : subgraph.edges_to(start_node)) {
+                subgraph.destroy_edge(edge);
+            }
+        } else if (mappings.begin()->is_reverse() && subgraph.end_degree(start_node) != 0) {
+            for (auto edge : subgraph.edges_from(start_node)) {
+                subgraph.destroy_edge(edge);
+            }
         }
     }
     Node* end_node = subgraph.get_node(mappings.rbegin()->node_id());
-    if (!mappings.rbegin()->is_reverse() && subgraph.end_degree(end_node) != 0) {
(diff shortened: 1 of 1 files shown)


glennhickey:
@adamnovak Any idea what happened to the gitlab tests?  Is there a way to look at a report?  Or rerun the tests?

adamnovak:
@glennhickey Yeah, if you click on the details link, it takes you to the pipeline page for the whole pipeline. You can then click on the test node in the middle of the graph, and then select a particular test job to view its output. The one that is giving us trouble mostly is test 1: https://ucsc-ci.com/vgteam/vg/-/jobs/814

```
============================= test session starts ==============================
platform linux2 -- Python 2.7.15rc1, pytest-4.1.1, py-1.7.0, pluggy-0.8.1 -- /builds/vgteam/vg/.env/bin/python2
cachedir: .pytest_cache
rootdir: /builds/vgteam/vg, inifile:
collecting ... collected 1 item

vgci/vgci.py::VGCITest::test_call_chr21_snp1kg <- .env/local/lib/python2.7/site-packages/timeout_decorator/timeout_decorator.py Pulling docker image gitlab/gitlab-runner-helper:x86_64-8d829975 ...
ERROR: Job failed (system failure): Cannot connect to the Docker daemon at tcp://172.31.7.97:2376. Is the docker daemon running? (executor_docker.go:967:60s)
```

The other failed jobs in the pipeline failed when I restarted the runner system to make a config change to hopefully fix that first job.

There's no report partly because I haven't hooked up Github comment posting yet (I think I need a new account, or to get ahold of an access token for the old bot account), and partly because when a test job fails, the report job isn't run, because it waits for all the test jobs to complete successfully. I still need to look into how to allow the test jobs to fail and run the report step, but then still fail the whole pipeline.

The actual underlying problem for this test is still a mystery though. Something bad happens to Docker somehow, and the cleanup/artifact upload step seems to be running before pytest is actually done testing.

I've swapped the executor nodes back up to c5.4xlarge from c5.2xlarge, which I shrank them to when I set up the autoscaling runners, and that seems to have allowed test 1 to pass on master at least. I'm going to try re-running this job on the bigger nodes; maybe the real problem is resource exhaustion and it's getting the Docker daemon OOM-killed or something *and* losing the tail end of the test logs.

I'm going to restart all these jobs now that the nodes are bigger.

adamnovak:
Actually, to restart jobs, you might need to be logged in on Gitlab. If you log into Gitlab with Github, I can enable your account there and add you to the vg project.

glennhickey:
Makes sense, thanks. I get an account blocked error when I try to sign in via github on the gitlab page.  Not sure if this happens to anyone else..

adamnovak:
I've unblocked you; getting blocked on first login is how it's *supposed* to work, apparently, so we can restrict access to the Gitlab instance. I might be able to make people accounts by e-mail and preemptively unblock them, but I'm not sure if they will then magically be accessible by Github login.

glennhickey:
That worked, thanks!

PR #2068, opened 2019-01-22
Skip snarls that aren't ultrabubbles with vg call

glennhickey:
This fixes an error like this that I'm getting on a non-ultrabubble snarl:
```
vg: src/traversal_finder.cpp:1176: vg::RepresentativeTraversalFinder::find_traversals(const vg::Snarl&)::<lambda(std::vector<vg::Visit>)>: Assertion `false' failed.
```

Stacks on #2067


ekg:
I think snarls should be ultrabubbles. This might be a mistake.

Issue #2041, opened 2018-12-19
RepresentativeTraversalFinder crashes on cactus graph with failed assertion


eldariont:
Hi,

I was running `toil-vg call` on one of my cactus graph and it crashed:
```
RuntimeError: Docker container for command vg call chunk_S288C.chrXVI_0_aug.vg -t 4 -S SRR4074257.recall.cactus.twoout -z chunk_S288C.chrXVI_0.trans -s chunk_S288C.chrXVI_0.support -b chunk_S288C.chrXVI_0.vg -u -n 0 -r S288C.chrXVI -c S288C.chrXVI -l 954457 -o 0 failed with code 134
```

I can reproduce the error with the following files and commands on vg v1.12.1-83-g9a00b102 "Parolise":
`vg call chunk_S288C.chrI_0_aug.vg -t 4 -S SRR4074257.recall.cactus.twoout -z chunk_S288C.chrI_0.trans -s chunk_S288C.chrI_0.support -b chunk_S288C.chrI_0.vg -u -n 0 -r S288C.chrI -c S288C.chrI -l 219929 -o 0`

Files: https://www.dropbox.com/s/gfpdtmymjfjz90z/Archive.zip?dl=0

Result:
```
...
S288C.chrI	23744	5516691_5514212_5552938_5553675_5540725-5516691_5553675_5540725	A	AGG	4473	PASS	DP=301;SVLEN=2;XSEE=9892,9893,9894,4887785;XREF	GT:DP:XDP:AD:XADL:SB:XAAD	1/0:301:140,237:120,181:-8.250185:60,60,95,86:181
S288C.chrI	23741	5578978_5558693_5533720_5516691-5578978_5558693_4887786_5516691	TTT	GCG,GCA	3402.67	lowxadl	DP=271;SVLEN=0,0;XSEE=9890,9891,9892,3945529,3945530,4887786;XREF	GT:DP:XDP:AD:XADL:SB:XAAD	1/2:271:140,237:0,181,90:-17.880078:0,0,79,102,39,51:271
S288C.chrI	23735	5539370_5561058_5565421	A	G	6762.33	PASS	DP=182;SVLEN=0;XSEE=9886,9887,9888,4777338;XREF	GT:DP:XDP:AD:XADL:SB:XAAD	1/1:182:140,237:1,181:.:0,1,78,103:181
[...]

PR #2070, opened 2019-01-23
Find snarls we are reading out of when finding traversals

adamnovak:
I think this should fix https://github.com/vgteam/vg/issues/2041 again. We weren't catching the case where we read out of a snarl (backward?) in our seed edge.


adamnovak:
@glennhickey Hopefully this helps...

PR #2079, opened 2019-01-29
Fix --gcsa-out option name / help mismatch in vg index

### LINKED ISSUE #2001: unrecognized option --gcsa-out
The long option `--gcsa-out` in v1.12.1 is not recognised:

```
vg: unrecognized option '--gcsa-out'
```

glennhickey:
Resolves #2001

### DIFF
--- a/src/subcommand/index_main.cpp
+++ b/src/subcommand/index_main.cpp
@@ -234,7 +234,7 @@ int main_index(int argc, char** argv) {
             {"exclude", required_argument, 0, 'E'},
 
             // GCSA
-            {"gcsa-name", required_argument, 0, 'g'},
+            {"gcsa-out", required_argument, 0, 'g'},
             {"dbg-in", required_argument, 0, 'i'},
             {"mapping", required_argument, 0, 'f'},
             {"kmer-size", required_argument, 0, 'k'},
--- a/vgci/vgci.py
+++ b/vgci/vgci.py
@@ -663,7 +663,8 @@ class VGCITest(TestCase):
                                      None,
                                      plan.reads_fastq_file_ids,
                                      plan.fasta_file_id, 
-                                     plan.bwa_index_ids, 
+                                     plan.bwa_index_ids,
+                                     None,
                                      plan.bam_file_ids,
                                      plan.pe_bam_file_ids, 
                                      plan.true_read_stats_file_id)
--- a/vgci/vgci.sh
+++ b/vgci/vgci.sh
@@ -30,7 +30,7 @@ KEEP_INTERMEDIATE_FILES=0
 # Should we show stdout and stderr from tests? If so, set to "-s".
 SHOW_OPT=""
 # What toil-vg should we install?
-TOIL_VG_PACKAGE="git+https://github.com/adamnovak/toil-vg.git@9b8b11148c7cd1d559b8fd83521c3f53a645ac36"
+TOIL_VG_PACKAGE="git+https://github.com/vgteam/toil-vg.git@5947af96d806daebd56894b63fc82595a30f545f"
 # What toil should we install?
 # Could be something like "toil[aws,mesos]==3.13.0"
 # or "git+https://github.com/adamnovak/toil.git@2b696bec34fa1381afdcf187456571d2b41f3842#egg=toil[aws,mesos]"


glennhickey:
@adamnovak Can you make sense of this error?
https://ucsc-ci.com/vgteam/vg/-/jobs/1979
I'm trying to get the ci tests running with a recent version of toil-vg and can't get this one to pass.

adamnovak:
@glennhickey Yeah, I've been seeing that kind of failure a bunch. I don't know what is causing the underlying test to fail, but when pytest goes to dump the entire log to the console, I think it is overwhelming whatever in the Gitlab system is capturing its output. The FIFO gets full and throws an error (because I guess it must be in nonblocking mode?).

I'm not quite sure how to solve it. The Right Way would be to work out how pytest gets its output FD into nonblocking mode (maybe it is messed up by Gitlab before pytest is forked off?) and make it not do that, but I don't think we want to mess with the internals of pytest. Dumping the pytest output to disk might help, assuming the local disk can keep up with pytest.

adamnovak:
I don't know if another toil-vg is going to help; it looks like maybe vg genotype is running out of memory and being OOM-killed?

```
ERROR:toil-rt:Docker container for command vg genotype chunk_21_10457102.vg -t 2 -s HG002 -v -G chunk_21_10457102.gam.filter -r 21 -c 21 -l 48129895 -o 10457102 failed with code 137
ERROR:toil-rt:Dumping stderr...
ERROR:toil-rt:0+0 records in
ERROR:toil-rt:0+0 records out
ERROR:toil-rt:0 bytes copied, 111.612 s, 0.0 kB/s
ERROR:toil-rt:/bin/bash: line 1:     7 Killed                  vg genotype chunk_21_10457102.vg -t 2 -s HG002 -v -G chunk_21_10457102.gam.filter -r 21 -c 21 -l 48129895 -o 10457102
ERROR:toil-rt:         8 Done                    | dd of=/control/stdout.fifo
ERROR:root:Genotyping failed. Dumping files.
```

Unfortunately uploading the tests' files is the job of the final report collating job in the CI pipeline, which is no help here for reproducing the problem. I should push that down into the individual test jobs.

glennhickey:
Yeah, it works fine with a really old toil-vg, but gives this crash with
the current one.  The only difference I could spot that would have any
bearing on this is that the new toil-vg doesn't run vg augment before
genotype.  Intead, it uses genotype's built-in augmentation.  This works
fine in general (the test passes on my desktop), but I suspect it somehow
leads to more peak memory usage than the old way of calling vg augment
separately.  This latest commit points to a commit to the latest toil-vg
but patched to run augment separately, just to confirm.

glennhickey:
Using ulimit to help reproduce locally sounds like a great idea.  

It looks like this case was indeed due to the built-in augmentation using too much memory.  I think reverting toil-vg to use separate augmentation is a reasonable short-term fix.  I think it bears looking into some more just to make sure edit isn't leaking memory. 

There are still sporadic can't connect to docker daemon errors (which seem completely unrelated).  Just re-running the tests can get them to pass, but red X in the gitlab status never goes away, making it impossible to merge.  An example is in [this PR](https://ucsc-ci.com/vgteam/vg/pipelines/164/builds), where Job 2215 is listed as failed/retired (and subsequently passed in Job 2276).

adamnovak:
The status not updating on Github might be related to our Gitlab licensing; I think some of the Github integration is a premium feature, and our license ran out overnight. It should be back up and running, but if it persists that's a problem.

I've promoted you to a vgteam owner, so you should have sufficient permission to override the required-ness of the Gitlab tests if it breaks again.

adamnovak:
I'm getting this same failure on test 9 in my PR #2096.

https://ucsc-ci.com/vgteam/vg/-/jobs/2676

```
=================================== FAILURES ===================================
_____________________ VGCITest.test_map_brca1_snp1kg_mpmap _____________________

args = (<vgci.VGCITest testMethod=test_map_brca1_snp1kg_mpmap>,), kwargs = {}
new_seconds = 600, old = 0

    @wraps(function)
    def new_function(*args, **kwargs):
        new_seconds = kwargs.pop('timeout', seconds)
        if new_seconds:
            old = signal.signal(signal.SIGALRM, handler)
            signal.setitimer(signal.ITIMER_REAL, new_seconds)
        try:
>           return function(*args, **kwargs)

.env/local/lib/python2.7/site-packages/timeout_decorator/timeout_decorator.py:81: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
vgci/vgci.py:1363: in test_map_brca1_snp1kg_mpmap
    misc_opts='--filter_opts \"-q 15 -m 1 -D 999 -s 1000\"')
vgci/vgci.py:526: in _test_bakeoff
    self._verify_f1('NA12878', tag)
vgci/vgci.py:499: in _verify_f1
    self.assertGreaterEqual(f1_score, baseline_f1 - threshold)
E   AssertionError: 0.9286 not greater than or equal to 0.9294
```

It looks like updating toil-vg maybe reduced performance in this test.

PR #2081, opened 2019-01-31
VPKG: Encapsulation for indexes and type-tagging for Protobufs

adamnovak:
This is my attempt to revise our IO so that we can keep multiple indexes in one file, and so that we can tell e.g. GAM and multipath GAM files apart when loading them.

Instead of using the various load() and save() methods or SDSL serialization, you make a `src/stream/register_loader_saver_whatever.{cpp,hpp}`, define load/save for your type to streams or messages there, call it from `stream::Registry::register_everything()`, and then use `stream::VPKG::load_one<YourType>()` to get a `unique_ptr<YourType>` from a file or stream, and `stream::VPKG::save()` to save to a file or stream. You can also use `load_all` to get a tuple of several types from a file in one pass.

It still needs:

- [ ] Encapsulation for StreamIndex indexes (easy).
- [ ] Encapsulation for DistanceIndex indexes (hard because @xchang1 made them require a graph to be loaded). I might postpone this.
- [ ] CLI UI for loading everything from one file with fewer options. I will probably postpone this.
- [x] Registration of VG protobuf as a loadable form of HandleGraph (easy).
- [ ] Tests for/usage of magic HandleGraph implementation selection.
- [ ] Some nice developer-facing documentation for the UI.
- [x] Up the stream message size from the tiny 4 kb I made it (easy).
- [ ] Error checking on `load_all()` like `load_one()` has, and a corresponding `try_load_all()` that allows failures.


adamnovak:
I'm going to merge this in now, for the sake of keeping everything synchronized, now that the tests pass.

I still have more things I want to encapsulate, and the magic HandleGraph selection is still basically science fiction, but it doesn't break anything.

PR #2083, opened 2019-02-01
Conservative anchoring and handlification in mpmap

jeizenga:
The main goal in here was to implement an anchoring strategy in mpmap that would be a little more conservative about how much it believes the anchors. To do that, I trim the ends of anchors off if they cause the anchor to make a decision at a branch point. I also modified the graph extraction algorithm so that it won't miss the alternate allele when the MEM hit is on a SNP anymore (a longstanding problem). That helps identify branch points better. 

Along the way I have nearly completed mpmap's transition to HandleGraphs for subgraph extraction. The one remaining place where we are still bound to VG objects is right before the alignment, since the GSSW aligner is still tied to a Protobuf Graph. I also went ahead and switched to a lighter weight graph implementation than VG in many places.

There's a decent chance this will blow up since I touched so much code, but hopefully the regression tests will tell me that.


adamnovak:
@jeizenga This doesn't seem to have blown up. Is there anything left to do before merging?

jeizenga:
I'm experimenting in the cloud to see if it actually helps things, and while there it does blow up. So I'm holding off until I feel a little safer.

PR #2090, opened 2019-02-05
Count reads with no indels in vg stats

adamnovak:
I'm counting reads that don't have any indels in vg stats now, as well as perfect reads. I'm not counting softclips as indels.

I'm also making the graph file optional for vg stats, so you can do `vg stats -a whatever.gam` and get just stats about the reads, without all the stats about how the reads relate to the graph and the alleles in it.


adamnovak:
I don't know if sift needs to have any of this added; I added that Python script for one weird indel analysis we're working on to check the feasibility of Benedict's super-fast gapless mapper idea. But I'll check out sift in case I need to pull out reads with indels.

PR #2091, opened 2019-02-06
Force test IO to be in blocking mode

adamnovak:
I'm trying to fix the problem of pytest dumping logs too fast and crashing due to failed nonblocking IO, on the theory that it's Gitlab's fault that our IO is nonblocking and we can just fix it before running the main part of our test code.

See https://github.com/vgteam/vg/pull/2079#issuecomment-461120947

The actual underlying test failures will still be a problem, but at least we might be able to actually debug them.

Also still a problem is the Docker daemon on the runner worker disappearing as in https://ucsc-ci.com/vgteam/vg/-/jobs/1958. For that see also https://gitlab.com/gitlab-org/gitlab-runner/issues/3612


glennhickey:
Thanks, this actually worked!

adamnovak:
I wouldn't celebrate yet. We weren't seeing it fail *all* the time
without this, and I don't think it can hit what this fixes unless the
tests themselves fail first. But maybe this will help.

PR #2095, opened 2019-02-07
Specify regions in msga

glennhickey:
`-R` option added to accept 4-column BED file mapping fasta sequence names to intervals on reference paths.  When some location information is known a priori, this can be used to speed up alignment as well as force sequences to be aligned about where you expect.


adamnovak:
@glennhickey It looks like t.fa and t.bed never came along with this PR, but the tests didn't notice because we only check for a difference between two pipelines that both use them.

Can you add those files?

PR #2096, opened 2019-02-07
More robust test failure reporting

adamnovak:
This should upload test output when tests fail, and try and run the report job even when tests fail, to report failures.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -3,6 +3,9 @@
 # There's no good way to do that when running Docker containers as siblings of a container containing toil-vg.
 # So we either have to genuinely nest Docker inside another Docker, or we have to run the build on the real host.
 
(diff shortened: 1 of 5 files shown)


adamnovak:
I've pulled in Glenn's PR #2079, because I had to fix (and update to the latest) toil-vg to allow configuring memory requirements for the test call jobs. I'm hoping that if I set it so not too many of them try to run at once, we won't run out of memory.

adamnovak:
OK it looks like the baseline has randomly gotten super good in the `test_map_brca1_snp1kg_mpmap` test, and now we can't merge anything because the 0.9286 F1 that used to be acceptable as of e.g. #2095 is no longer good enough.

@glennhickey We need to rethink how the baseline should work so this stops happening. For now I'm just going to clear the baseline data out from S3.

PR #2099, opened 2019-02-09
Transition to HandleGraphs in alignment code

jeizenga:
This switches (most of) the alignment code away from Protobuf Graphs to HandleGraphs. I haven't touched the X-drop code, mostly because of lack of familiarity. Switching to HandleGraphs allows us to move away from Protobuf-backed implementations (cough cough VG cough) in the alignment-adjacent code, to other potentially better performing implementations.

This factoring also lets us offload some of the graph processing that we had reimplemented multiple places (e.g. sorting) to the HandleGraph algorithms namespace. In turn, this let me easily move the graph processing closer to where it's actually needed in the alignment algorithms. This is a better separation of responsibilities; it makes the alignment methods more self-contained. 

To maintain some of the optimizations in the Mapper, I did have to write one hack into this PR. There's now a thin overlay on an unindexed Protobuf graph that will let you shim it into the new Aligner interfaces. It is not performant, and not really meant to be used for any other purpose than this. However, I think it will help ease the transition to other graph implementations until we do a more substantial refactoring.

### DIFF
--- a/src/algorithms/extract_containing_graph.cpp
+++ b/src/algorithms/extract_containing_graph.cpp
@@ -125,15 +125,15 @@ void extract_containing_graph(const HandleGraph* source,
 void extract_containing_graph(const HandleGraph* source, MutableHandleGraph* into, const vector<pos_t>& positions,
                               size_t max_dist, size_t reversing_walk_length) {
     
-    return extract_containing_graph(source, into, positions, vector<size_t>(positions.size(), max_dist),
-                                    reversing_walk_length);
+    extract_containing_graph(source, into, positions, vector<size_t>(positions.size(), max_dist),
+                             reversing_walk_length);
 }
 
 void extract_containing_graph(const HandleGraph* source, MutableHandleGraph* into, const vector<pos_t>& positions,
                               const vector<size_t>& position_max_dist, size_t reversing_walk_length) {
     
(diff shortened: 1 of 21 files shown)


ekg:
> To maintain some of the optimizations in the Mapper, I did have to write one hack into this PR. There's now a thin overlay on an unindexed Protobuf graph that will let you shim it into the new Aligner interfaces. It is not performant, and not really meant to be used for any other purpose than this. However, I think it will help ease the transition to other graph implementations until we do a more substantial refactoring.

Should it just be made to back into a HandleGraph? I can work on fixing this before you merge in.

jeizenga:
Moving forward, yes, I do think the best move would be to use a generic HandleGraph through as much as the Mapper code as possible so that you can swap the implementation more easily. I think it will take some significant refactoring to achieve that though, and I don't really want to hold up this PR waiting for that to happen. In the meantime, the CI tests don't seem to have shown any dramatic change in run time. I'm taking that to mean that my short-term shims are doing their job, so I'm gonna go ahead and merge. The path to factoring the VG object out of the Mapper is paved though!

jeizenga:
I think the original post may have been unclear on what the shim was. It's basically a HandleGraph overlay on the Protobuf graph that let's you continue to access nodes by index so that you can take advantage of the internal sorting of the Protobuf graph and do basic operations like sequence access, ID access, etc.

PR #2105, opened 2019-02-12
Minimizer index

jltsiren:
This PR adds a prototype minimizer index. The index is implemented as a custom hash table with 64-bit keys and 64-bit values. A separate bitvector tells whether the value in each cell should be interpreted as a single position or as a pointer to `std::vector` of positions. The only dependency to VG is to `types.hpp`.

## Construction

The index can be built from an XG index with the `vg minimizer` subcommand:

```
vg minimizer -i graph.mi graph.xg
```

Options `-k` and `-w` change the minimizer parameters from their default values (21, 11). By default, the index includes all minimizers in the graph. Option `-m N` can be used to exclude minimizers with more than `N` occurrences. Option `-g X` restricts the indexing to haplotype-consistent minimizers using the GBWT index in file `X`.

## Benchmarks

I tested the minimizer index with the 1000GP graph on an AWS i3.8xlarge instance using 32 threads. Indexing haplotype-consistent minimizers took 28 minutes and 98 GB memory, and the final index size was 37 GB. This was before the new `stream::VPKG` serialization interface. Reading/writing tens of gigabytes using a compressed stream may have a significant impact on construction time.

Indexing all minimizers in the graph took 65 hours and 132 GB memory. The final index size was 70 GB. There are too many 32 bp paths in complex regions, but the final number of minimizers is manageable. If we want to make indexing all minimizers a viable option, we need a better algorithm for extracting minimizers from complex regions.

Based on some quick benchmarks, the minimizer index is 3-6 times faster than GCSA2.

## Interface

`MinimizerIndex` is defined in `minimizer.hpp`.

### Finding minimizers

```
std::vector<minimizer_type> minimizers(const std::string& str) const
```

Returns all minimizers in the string as (key, offset) pairs, using the (k, w) values in the index. The offset is the starting position of the minimizer in the string.

This is the only function that deals explicitly with minimizers. The rest of the index sees just integer keys.

### Inserting kmer occurrences

```
void insert(key_type key, pos_t pos)
```

Inserts an occurrence of the kmer encoded in `key` at position `pos` to the index.

### Counting the occurrences

```
size_t count(key_type key) const
```

Returns the number of occurrences of `key`.

### Finding the occurrences

```
std::vector<pos_t> find(key_type key) const
```

Returns the occurrences of `key`.

## Space optimization ideas

The index uses several times more space than a minimizer index for strings. Here are some ideas that could reduce the space usage:

* If we have fast access to the sequences in the graph, we do not need to store keys in the hash table. Hash collisions can be detected by finding a single occurrence and comparing the sequence there to the minimizer.
* We can reduce the number of minimizers in the index by indexing only one minimizer for each window and its reverse complement. This may cause some issues with the mapper, because some occurrences would be reported by their starting positions and others by their ending positions.
* If the graph is small enough, we can use 32-bit positions instead of 64-bit positions.


jltsiren:
This was a bit more interesting test failure than usual. On Mac, parallel `XG::for_each_handle()` visits the nodes in order when it is run with a single thread. On Linux, it first visits the last 4 nodes in order and then continues with the rest of the nodes, at least on the small test graph. Because the insertion order is different, we end up with different index files.

adamnovak:
I've pushed this up to the vgteam/minimizers branch, so to CI updated changes that branch needs to be updated as well.

adamnovak:
You probably want to rebase/merge this with master and try again. It looks like it's failing test 1 on Gitlab by running out of memory, which is a problem we fixed in the master branch.

PR #2109, opened 2019-02-13
Compute read stats without locks for way more speed

### LINKED ISSUE #2092: `vg stats -a` looks like it is spending all its time fighting over per-statistic locks
The code for `vg stats`'s processing of reads tracks a bunch of counters, and for each read it locks and increments the appropriate counters in turn, in parallel over the reads.

I'm pretty sure that this is spectacularly inefficient, especially with lots of threads. Instead, each thread should compute its own stats, and then the stats should be aggregated and reported at the end. We should be able to get away with not doing any locking at all. We should copy the way that `vg filter` collects counts for filtered reads due to each kind of filter, and then aggregates at the end.

Someone should 1) profile `vg stats` to see if I'm right about this and 2) refactor it to use per-thread stat computation with final aggregation.

adamnovak:
This should fix #2092. I've tested it on a set of 1 million reads I have been working with and it is now way faster (at the cost of about double memory usage).

Before:

```
    Command being timed: "vg stats -a trash/HG002_1000G.nodecoy.first1m.gam"
    User time (seconds): 5388.69
    System time (seconds): 54.34
    Percent of CPU this job got: 1822%
    Elapsed (wall clock) time (h:mm:ss or m:ss): 4:58.71
    Average shared text size (kbytes): 0
    Average unshared data size (kbytes): 0
    Average stack size (kbytes): 0
    Average total size (kbytes): 0
    Maximum resident set size (kbytes): 830568
```

After: 

```
    Command being timed: "vg stats -a trash/HG002_1000G.nodecoy.first1m.gam"
    User time (seconds): 61.42
    System time (seconds): 16.02
    Percent of CPU this job got: 215%
    Elapsed (wall clock) time (h:mm:ss or m:ss): 0:35.95
    Average shared text size (kbytes): 0
    Average unshared data size (kbytes): 0
    Average stack size (kbytes): 0
    Average total size (kbytes): 0
    Maximum resident set size (kbytes): 1462188
```

### DIFF
--- a/src/subcommand/stats_main.cpp
+++ b/src/subcommand/stats_main.cpp
@@ -427,21 +427,92 @@ int main_stats(int argc, char** argv) {
             assert(last_underscore != string::npos);
             return path_name.substr(last_underscore + 1);
         };
+        
+        // In order to do stats across multiple threads, we define these add-able bundles of stats.
+        struct ReadStats {
+            // These are the general stats we will compute.
+            size_t total_alignments = 0;
+            size_t total_aligned = 0;
+            size_t total_primary = 0;
+            size_t total_secondary = 0;
+            size_t total_perfect = 0; // Number of reads with no indels or substitutions relative to their paths
+            size_t total_gapless = 0; // Number of reads with no indels relative to their paths
+
+            // These are for tracking which nodes are covered and which are not
+            map<vg::id_t, size_t> node_visit_counts;
+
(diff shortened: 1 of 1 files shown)


adamnovak:
We now might want to go through the parts of `vg stats` that loop over nodes in the graph and use the same lots-of-per-statistic-locks pattern to count up stats about them.

PR #2110, opened 2019-02-13
Upgrade to C++ 14

adamnovak:
I want C++14 because it will simplify some hacks I am doing to present the handle graph interface I want. I think compilers are new enough (C++14 is 5 years old now) that we might be able to just do this with no special changes to our infrastructure.


ekg:
I agree. Compile time seems to increase slightly in my experience, but that
shouldn't stop us. Some newer libraries require C++14

Issue #2104, opened 2019-02-12
Sharing HandleGraph and Algorithms code with dsgvg


adamnovak:
As was discussed at the vgteam call today, we want to share some code between vg and @ekg's https://github.com/vgteam/dsgvg. Specifically, we want to share the `handle.hpp` handle graph interface, and some algorithms from `algorithms`.

Whatever we do, we're going to have to factor out the shared code into its own folder(s) and have it build its own library that does not depend on the greater `libvg`. Maybe it will be `libhandlegraph`? It might be worth it to make the base handle graph interface separate from the algorithms, which would in turn depend on it.

We're also going to have to work out how to actually share the code in the Git repos. We could:

- Just copy-paste files around manually, and track two copies of the code. Simple to start, but potentially terrible if we want to try and merge/synchronize changes.
- Make a new repo for this shared code, and pull into both dependent projects as a Git submodule. We understand how to work with submodules, but this makes PRing changes that touch this shared code very hard, as you have to get your PR against the shared repo merged before any PRs in the dependent repos can even start being tested. This was a huge pain with `xg` and drove the integration of `xg` into the main vg repo.
- Make a repo for the shared code, and pull it in to both projects using [Git subtrees](https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree). This is kind of a git-assisted copy paste. Everything actually exists and can be modified and PR'd as normal in both dependent repos, but Git has some machinery to rewrite changes into/out of commits that just talk about the subtree's code, and by doing that and having a Github repo for just the subtree, we could keep the two dependent repos' copies in sync. The downside is everyone who wants to sync things up has to learn the Git wizardry required to make that happen.
- Try using [git-subrepo](https://github.com/ingydotnet/git-subrepo#description). This is supposed to be "better" than either subtrees or submodules according to its author, but it's an aftermarket add-on to Git that people would have to install to sync up the subrepos.

jeizenga:
I think my vote is for submodule, tbh. I don't think `handle.hpp` is as unstable as the `xg` code was. Although, the same might not be true of `algorithms`...

jeizenga:
I'm considering this resolved now that we have `libhandlegraph`

Issue #2076, opened 2019-01-29
CMake the build


ekg:
This has been discussed previously. We should use cmake and subproject commands to configure and drive the build. We should consider optioally skipping many builds of system installable packages. It is getting somewhat easier to install these via conda and friends, so maybe people can skip long builds of dependencies.

I have some experience with this now, so the best response here would be to send a PR with a working CMakeLists.txt.

Issue #1473, opened 2018-02-28
cmake-ification


ekg:
Can cmake be used to generate more efficient compilation commands and build processed based on the actual dependencies between files? If so, it would be really cool to apply it to vg, for the simple reason that development could be sped up. I've burned a lot of time waiting on our battleaxe of a makefile to do its job after I make a small change in a file that purportedly has a lot of dependencies...

ekg:
In https://github.com/ekg/seqwish I have set up a minimal model for how we
can use cmake to handle what are now submodules in vg. It also let's us
build out of the source tree. We may want to add cmake build codes to a lot
of the repos we control and depend on.

Issue #133, opened 2015-10-27
cmake


ekg:
I'm interested in moving to cmake.

Here is some discussion from `#cmake` on freenode.

```irc
12:48 < erikg> for better or worse, i've been using git submodules for 
               dependency management
12:48 < erikg> strung together with make
12:48 < erikg> i'd like to move to cmake, but i'm having trouble understanding 
               how to do some basic things
12:48 < erikg> say i have protobuf as a submodule
12:49 < erikg> i'd like to build it (using cmake) and use it during the 
               compilation of .proto files (rather than the system-wide 
               protobuf, which is an older version)
12:50 < erikg> i can't find a straightforward way to tell cmake to build it 
               first
12:50 < erikg> what i'd like to do is roughly described here: 
http://stackoverflow.com/questions/11524201/how-to-add-a-cmake-custom-target-for-all-submodules
12:52 < erikg> so it looks like i can only add_custom_target
12:57 <+ngladitz> how are you building your sub-modules? add_subdirectory() or 
                  ExternalProject? what are you using to compile proto files? 
                  the existing cmake module?
17:11 < erikg> ngladitz: my submodules are included via git submodule add... 
               and then i link them in with a handmade makefile
17:11 < erikg> ngladitz: it's not 100% necessary but i'd prefer to use 
               protobuf3, so i have to build it from source
17:11 < erikg> ngladitz: i don't know which i should use, or what pattern--- 
               add_subdirectory() or ExternalProject
17:12 < erikg> one problem i'm having is that i have multiple libs now 
               depending on the same thing. i'd like to avoid building things 
               twice. i'd also like to avoid pushing everything into blocks on 
               biicode or some other kind of dependency mananger.  
17:17 <+ngladitz> well if it were add_subdirectory() then targets of your 
                  parent and the subdirectory would be in the same scope and 
                  could have order dependencies ... e.g. using build targets in 
                  a custom command would automatically set up such a dependency
18:25 < erikg> ngladitz: do you have an example of this handy that i could read?
18:26 < erikg> a CMakeLists.txt or repo
18:30 <+ngladitz> erikg: e.g. http://pastebin.com/UJrjaYbd
18:31 <+ngladitz> the custom command invokes my_tool which itself is an 
                  executable target; cmake implicitly adds a dependency so that 
                  my_tool is build before the custom command is invoked
18:33 < erikg> ngladitz: how do i know when it will do this implicitly and when 
               i have to specify it?
[...]

PR #2114, opened 2019-02-15
Use handle graphs from libhandlegraph

adamnovak:
This PR changes vg to get its HandleGraph interfaces from vgteam/libhandlegraph, where they can be shared with vgteam/dsgvg.

There are some minor changes to the HandleGraph implementations, because I changed the interface to no longer require implementations to `using` in shadowed base definitions. Instead there are front-end templates in the interface for iterating over stuff, and the implementations define `_impl` methods with the most general, early-stopping-supported signature.


adamnovak:
I think that this will fix #2104. We're pulling in the library as a submodule in vg, but as a cmake ExternalProject in dsgvg. If we ever do #133 AKA #1473 AKA #2076 and adopt cmake for vg we can take that approach here too.

PR #2132, opened 2019-02-25
Convert handle

emikoifish:
Convert from HandleGraph to a MutableHandleGraph.
Unittest tests conversion of nodes, edges, paths from xg to vg, PackedGraph, and HashGraph.


jeizenga:
It seems to me that this supersedes `handle_to_vg.[cpp|hpp]`. Think you could delete those files and replace the calls of `handle_to_vg()` with `convert_handle_graph()` as part of this PR?

src/convert_handle.cpp
```diff
@@ -0,0 +1,59 @@
+#include "convert_handle.hpp"
+namespace vg {
+using namespace std;
+    void convert_handle_graph(const HandleGraph* converting, MutableHandleGraph* converted) {
+        assert(converted->node_size() == 0);
+        
+        // If xg is a null pointer, throw a runtime error
+        if (converting == nullptr) {
+            throw runtime_error("There is no graph to convert");
```
jeizenga: You should probably also check whether `converted` is `nullptr` too, right? In fact, I also like to check whether the `converted` is an empty graph using `node_size()` and `get_path_count()`

src/convert_handle.cpp
```diff
@@ -0,0 +1,59 @@
...
+            // Get the sequence of the graph handle
+            string converting_seq = converting->get_sequence(here);
+            // Create a handle in mutable graph using the graph id and sequence
+            converted->create_handle(converting_seq, converting_id);
+        });
+        // add any edges that are not yet present
+        converting->for_each_edge([&](const edge_t& edge) {
+            edge_t converted_edge(converted->get_handle(converting->get_id(edge.first), converting->get_is_reverse(edge.first)),
+                             converted->get_handle(converting->get_id(edge.second), converting->get_is_reverse(edge.second)));
+            if (!converted->has_edge(converted_edge)) {
```
jeizenga: I think this check should be unnecessary. The implementation of `for_each_edge` guarantees that each edge is iterated over only once, so if you have a well-formed graph going in, this should never evaluate to false

src/convert_handle.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include "handle.hpp"
+#include "vg.hpp"
+
+namespace vg {
+    using namespace std;
+    // Takes in a pointer to a HandleGraph and converts it to a MutableHandleGraph graph.
+    void convert_handle_graph(const HandleGraph* converting, MutableHandleGraph* converted);
+    
+    // Change paths to a mutable path.
+    void convert_path_handle_graph(const PathHandleGraph* converting, MutablePathDeletableHandleGraph* converted);
```
jeizenga: I think you should be able to get away with a `MutablePathMutableHandleGraph`, right? You don't actually use the delete methods.

emikoifish: I don't think vg, HashGraph, and PackedGraph have `MutablePathMutableHandleGraph`, I only see `MutablePathDeletableHandleGraph`

jeizenga: `MutablePathDeletableHandleGraph` inherits from `MutablePathMutableHandleGraph`, so that's fine. The point of the distinction is to allow graph implementations that can be mutably "constructed" but don't support efficient delete operations. In fact, this is exactly the sort of use case we had in mind for `MutablePathMutableHandleGraph`.

adamnovak:
@jeizenga Is this ready to merge?

jeizenga:
Yes, this is mergeable. There are a couple of tweaks that could have improved it, but it's correct and merge-worthy as is.

PR #2151, opened 2019-03-07
Option to exclude unaligned reads in output

kevyin:
Option to exclude unaligned reads


adamnovak:
This looks good to me. I've pushed it over to CI to test.

PR #2155, opened 2019-03-11
Add a way to grab the rank of a snarl in its chain

adamnovak:


adamnovak:
I'm going to merge this. The status report-back is broken right now, but the tests pass.

PR #2159, opened 2019-03-13
Continue surject refactor

adamnovak:
I've refactored surject and map to use a unified alignment emitter pattern that can emit single and paired alignments with primaries/secondaries in a variety of formats. We should be able to run all HTSlib output through one place, where all we do is convert from Alignment objects to HTSlib stuff, so if we want to change it we only have to do it in one place.


adamnovak:
This seems to make the tests 2% slower, but I'm going to merge it anyway. There's probably room for speed improvement in the buffering/output logic.

PR #2160, opened 2019-03-13
Add support for reporting the build status directly to the test script

adamnovak:
Until @DailyDreaming can get our Gitlab/Github integration working again by plugging our license into Gitlab, this should serve as a workaround.


adamnovak:
We shouldn't need this; our Gitlab is now licensed again and reporting in.

PR #2162, opened 2019-03-13
Be more careful with dot alts

### LINKED ISSUE #2148: Can anyone please explain what is a meaning of Node with the dot in the variation graph? Please see node 48 in the file (attached)
Please describe:

1. What you were trying to do
2. What you wanted to happen
3. What actually happened
4. What data and command line to use to make the problem recur, if applicable

```
Format code blocks or terminal copy-pastes like this, between triple backticks.
```

glennhickey:
Warn about and don't construct non-DNA alts.  Don't include haploid reference calls in vg call output by default. 
Should fix #2148

### DIFF
--- a/src/constructor.cpp
+++ b/src/constructor.cpp
@@ -177,6 +177,39 @@ namespace vg {
         return std::make_pair(start, end);
     }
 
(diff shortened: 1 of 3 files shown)


glennhickey:
Oops, hadn't noticed #2161

PR #2164, opened 2019-03-14
GBWT-backed HandleGraph

jltsiren:
@adamnovak I modified the makefile to find `jansson` using `pkg-config`. Will this break anything? Apple got rid of `/usr/include` in Mojave, and the Homebrew-installed `jansson.h` can no longer be found in any of the default include directories.

This PR adds `GBWTGraph`, a GBWT-backed runtime-only `HandleGraph` implementation. It uses the graph topology stored in the GBWT index and extracts the sequences from another `HandleGraph`. Sequence access and conversions between handles and node ids are faster than in `XG`, while graph navigation is slower. Minimizer index construction is faster and uses less memory with `GBWTGraph` than with `XG`.

In addition to the standard `HandleGraph` interface, there are a few useful operations:

* `get_sequence_view()` is a version of `get_sequence()` that gives direct access to the stored sequence without copying it.
* `get_state()` and `get_bd_state()` return the `gbwt::SearchState` and the `gbwt::BidirectionalState` corresponding to the handle.
* `follow_paths()` extends the path corresponding to the given search state using an interface similar to `follow_edges()`. If the search state is bidirectional, the path can also be extended backward.


adamnovak:
I don't think using pkg-config to find jansson will cause trouble for anyone. That's weird that Homebrew doesn't drop things in the system include paths on Mojave.

PR #2169, opened 2019-03-18
Faster xg context expansion

glennhickey:
Fixing up some low-haning STL stuff gives a decent speedup on some test data.

old
```
time vg chunk -x component0.xg -p S288C.chrI -c 100 > /dev/null 
real	0m47.752s
user	0m47.260s
sys	0m0.484s
```
new
```
time vg chunk -x component0.xg -p S288C.chrI -c 100 > /dev/null 
real	0m18.226s
user	0m17.788s
sys	0m0.420s
```

### DIFF
--- a/src/xg.cpp
+++ b/src/xg.cpp
@@ -1945,6 +1945,7 @@ vector<Edge> XG::edges_of(int64_t id) const {
     int64_t t = g + G_NODE_HEADER_LENGTH;
     int64_t f = g + G_NODE_HEADER_LENGTH + G_EDGE_LENGTH * edges_to_count;
     vector<Edge> edges;
+    edges.reserve((f - t + G_EDGE_LENGTH * edges_from_count) / 2);
     for (int64_t j = t; j < f; ) {
         int64_t from = g+g_iv[j++];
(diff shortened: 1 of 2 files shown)


ekg:
This might slow down other cases, such as in the mapper. If not then it's
all good.

You can also do much better than unordered map. I would suggest
ska::flat_hash_map as a higher performance alternative in both memory and
runtime.

glennhickey:
@ekg I'll run a mapping check.  Is your concern about ordered sets being
faster for smaller inputs in this case?

glennhickey:
Checking the CI logs, some map times are faster with the unordered maps, but one of the yeast ones looks much slower.  Ironically, it was slow context expansion when sv typing on yeast graphs that lead me to play with this to begin with.
 
Extracting path ranges from these hairball graphs is a losing proposition in general, though, regardless of lookup used. You can't get large SVs without expanding over virtually the entire graph, and the many dangling ends which get created if anything's left out mess up the snarl finder.  Luckily, these graphs are small enough to just run through without any chunking, which is what we're doing now.

PR #2185, opened 2019-03-27
Add node options for vg pack

### LINKED ISSUE #2183: Add a --node-list option for vg pack command
`vg pack` table format itself is too large to handle in MoMI-G. So, I want to generate compressed coverage packs using `--packs-in` option in advance, and then retrieve a subset of coverage packs when requested. 
For retrieving a subset, I will add `--node-list` option for `vg pack` command. Because `vg pack` binary does not have an index for each node, coverages are extracted if and only if a node.id is included in node-list by full search.

6br:
Fix #2183 , as part of the Pangenomics Hackathon at UCSC in March 2019. 
> `vg pack` table format itself is too large to handle in MoMI-G. So, I want to generate compressed coverage packs using `--packs-in` option in advance, and then retrieve a subset of coverage packs when requested. 
> For retrieving a subset, I will add `--node-list` option for `vg pack` command. Because `vg pack` binary does not have an index for each node, coverages are extracted if and only if a node.id is included in node-list by full search.

### DIFF
--- a/src/packer.cpp
+++ b/src/packer.cpp
@@ -1,4 +1,5 @@
 #include "packer.hpp"
+#include "../vg.hpp"
 
 namespace vg {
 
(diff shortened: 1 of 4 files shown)


6br:
It works well. This pull request is ready to merge.

PR #2186, opened 2019-03-27
More GAFFE hackathon work

adamnovak:
I'm fleshing out the pipeline for dealing with reads that don't have a full-length gapless alignment with sufficiently few mismatches.


adamnovak:
I'm going to merge this even though it slows down the new mapper a lot; you can use `-C` to get the old behavior, and we're working on building our funnel of mapping heuristics on top of this.

PR #2190, opened 2019-03-29
Merge nodes and add size info for simple view

jmonlong:
What about we merge nodes as much as possible in the simple DOT mode of vg view and add information about the node size? 

That would be useful for me when looking at SVs. For example a large deletion looked like this:

simple view

With the new version it's easier to see here the 1 kbp deletion in the middle:

simple view 2

If someone prefers to keep the current `-S` behavior, I can add a new argument for it.

### DIFF
--- a/src/packer.cpp
+++ b/src/packer.cpp
@@ -1,4 +1,5 @@
 #include "packer.hpp"
+#include "../vg.hpp"
 
 namespace vg {
 
@@ -383,9 +384,10 @@ vector<Edit> Packer::edits_at_position(size_t i) const {
     return edits;
 }
 
-ostream& Packer::as_table(ostream& out, bool show_edits) {
+ostream& Packer::as_table(ostream& out, bool show_edits, vector<vg::id_t> node_ids) {
 #ifdef debug
-    cerr << "Packer table of " << coverage_civ.size() << " rows:" << endl;
+    cerr << "Packer table of " << coverage_civ.size() << " rows:" << 
+        l;
 #endif
 
     out << "seq.pos" << "\t"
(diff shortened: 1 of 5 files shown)


glennhickey:
I like this idea, but why not leave it in mod: something like `vg mod -u graph.vg | vg view -pdS -` so it can be easily toggled on and off (people using view for debugging often want their ids preserved)?

glennhickey:
Yeah, my vote would be to just add the size info, and leave unchop in mod.  But I don't think it's a big deal either way.  Usually we try to make the command lines as long as possible to necessitate more trips to Portugal.

adamnovak:
@jmonlong I've invited you into vgteam. Can you push this to a branch under vgteam/vg so that Gitlab will test it?

jmonlong:
Thanks for the invitation. I created a branch viz in vgteam/vg. Now I guess I should create a new pull request from that and cancel this one. There is no way to modify this one to point to the new branch, right?

adamnovak:
It's fine to have the PR itself be from your repo. The commit just needs to be somewhere in vgteam/vg to get tested. If you update the PR with new commits, just make sure to update the vgteam/vg branch as well.

adamnovak:
It looks like the commit you made on the viz branch that got tested was a merge commit https://github.com/vgteam/vg/commit/f1b5412048df5659fe153d729ad687b56ffddbc8 that is not the same as the commit this PR points to. You should update your viz branch to point to that commit from the vgteam viz branch, so the PR shows as tested.

ekg:
What's the status of this pr? I'd be happy to merge if it's just adding new functionality.

jmonlong:
My bad I got confused with the branches/CI tests. Thanks @adamnovak, now it marks the GitLab tests as passed.

PR #2207, opened 2019-04-10
Hash graph serialization

jeizenga:
This is a binarized serialization method for the HashGraph. One issue that I know this code has is that it is not endian-independent. It turns out there's no standard way to convert in and out of the architecture-specific endianness for 64-bit integers, which is pretty dumb. If I had a method to do that, this serialization would be pretty easy to make endian-independent.

### DIFF
--- a/src/algorithms/are_equivalent.hpp
+++ b/src/algorithms/are_equivalent.hpp
@@ -15,11 +15,13 @@ namespace vg {
 namespace algorithms {
 using namespace std;
 
-    /// Checks whether two graphs are identical in their IDs, sequences, and edges
+    /// Checks whether two graphs are identical in their IDs, sequences, and edges.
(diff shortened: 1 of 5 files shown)


ekg:
@jeizenga are you familiar with htobe64 and be64toh? https://linux.die.net/man/3/be64toh

jeizenga:
I did run across them when I was looking for a solution to this. For one, it seems they have different names on different platforms, even on Linux. I found a pretty nice macro-based approach to that problem here: https://stackoverflow.com/questions/809902/64-bit-ntohl-in-c

Unfortunately, none of those options compile on my Mac, so I think it might still be a Linux-only solution.

ekg:
It seems we will want a cross-platform compatible thing. I guess we'll cross that bridge when it becomes a problem.

PR #2211, opened 2019-04-11
Store thread names in GBWT

jltsiren:
Store thread names in GBWT instead of XG. The new thread names are structured but compatible with the old `_thread_[sample]_[contig]_[phase]_[count]` names.

Some additional changes:

* Remove a sample from a GBWT index using `vg gbwt`.
* Extract the threads corresponding to a sample using `vg paths`.
* Simplified `vg paths` interface: choose one input format (VG, XG, or GBWT), one output format, and optionally one path/thread selection criterion. (Not all combinations are supported.)
* The VCF parse generated by `vg index -e` now contains sample/contig names, and the external GBWT construction tools use that information for generating thread names.
* `vg index` no longer builds gPBWT indexes or generates thread name databases.
* When a GAM file is indexed as GBWT, alignment names are used as sample names in the GBWT metadata. (GBWT does not support arbitrary strings as thread names.)


jltsiren:
I restored hidden `vg paths` options `--threads` and `--threads-by` for backward compatibility. `--threads` does not do anything apart from displaying a warning, while `--threads-by` displays a warning message and behaves like `--paths-by`.

PR #2215, opened 2019-04-15
Port vg to libvgio

adamnovak:
This PR rips out most of VG's IO logic, in favor of libvgio.

I'm pointing at a vgteam fork of the library, which I've patched to actually work in vg, and to have vg's latest IO system changes.

I've also moved the remaining `stream/` files into `io/` and the `vg::io` namespace, to match libvgio's organization.

I have a new registration function for registering libvg's types with the libvgio `Registry`, for VPKG serialization/deserialization. I'm just calling it from main, rather than trying to get it to magically statically register things.

This PR notably **deletes vg.proto**. The proto file is provided by libvgio now, since you need it to do any decoding of VG and GAM files. According to my logs it hasn't changed since September; hopefully it is evolving slowly enough now that we can bear to have it in another repo. If we need to evolve some parts more quickly, we could break the proto file up.


adamnovak:
OK, the being really slow has been fixed, but now I'm having Protobuf Problems. Since I merged in the Protobuf upgrade from master, on Mac Travis, vg will build fine, but Protobuf in the resulting binary has some kind of version mismatch with itself. It crashes all the time and says things like:

```
[libprotobuf ERROR google/protobuf/descriptor.cc:3561] Invalid proto descriptor for file "google/protobuf/descriptor.proto":

[libprotobuf ERROR google/protobuf/descriptor.cc:3564]   google/protobuf/descriptor.proto: Unrecognized syntax: TACAATCATGGAATTCCAGAAAATG
```

Clearly it's reading at the wrong address here and getting DNA sequence instead of the Protobuf descriptor it is wanting to parse.

adamnovak:
I can reproduce the Mac bug locally; I was managing to link static Protobuf and dynamic libvgio, which depends on Protobuf. That resulted in Trouble.

Linking vgio statically by .a file path solves that problem.

adamnovak:
I'm still not quite happy with the build process here (we build a lot of shared libraries we don't want or use), but this *should* work and be mergeable.

adamnovak:
OK, this passes all the tests and no longer slows down the Gitlab tests noticeably. I'm not sure if the real performance fix was using position-dependent code for the IO or putting `-O3` in the default libvgio build settings, but either way that problem is solved.

I'm going to merge this in.

PR #2217, opened 2019-04-16
optionally trim softclips from paths being embedded in the graph

ekg:
This removes tips that make it really hard to visualize the resulting graphs.

The objective here is to improve visualization for the SV paper.


ekg:
It would have been nice if we had a tool to do this to the alignments, but it seems that nothing exists. `vg filter` appeared to have a trimming option, but it doesn't work for me. Softclips are mentioned in the code but nothing happens when it's applied to alignments with softclips.

PR #2221, opened 2019-04-17
specify the cairo directory when including

ekg:
On some systems, it seems that we need to specify that cairo is within /usr/include/cairo.

Here I do that by writing `#include <cairo/cairo.h>`.

It's not clear to me why this is needed, or if we should make this change. It does fix my build problems though. If testing is fine then we might as well set it up like this to avoid future problems.

### DIFF
--- a/src/viz.hpp
+++ b/src/viz.hpp
@@ -2,8 +2,8 @@
 #define VG_VIZ_HPP_INCLUDED
 
 #include <vector>
-#include <cairo.h>
-#include <cairo-svg.h>
+#include <cairo/cairo.h>
+#include <cairo/cairo-svg.h>
 #include <iostream>
 #include <chrono>
 #include <ctime>


ekg:
The problem here was actually that I didn't have jansson installed in a typical way.

This broke pkg-config, and so the last compile command never got the right set of libraries and paths.

Do we even need jansson anymore?

adamnovak:
We use Jansson *only* for determining where incoming JSON messages end, AFAIK. Protobuf can deserialize incoming JSON, but you have to give it exactly one message, and I don't think it can do a streaming parse and determine where the message ends. But since finding the end of a JSON message requires more or less parsing it (since you have to process string quotes, and escaped quotes, and so on), we rely on the entire Jansson parser just for delimiting the messages.

If we require the JSON to be line-delimited, we break input from jq, which pretty-prints by default.

ekg:
Actually, I had never expected us to be able to process multiline JSON.

jq can output line delimited JSON. Use the compact option: -c

I would prefer to break on non newline delimited JSON than have another
dependency.

adamnovak:
I think it's important to not rely on the input JSON being
line-delimited. Conservative in what you produce, liberal in what you
accept, and all that. And I think it's a slightly different calculus
when we're talking about removing a dependency we already have versus
adding a new one.

But we should try and parse the input JSON another way. We could do
something hacky where we assume that the message ends at a line that
ends in '}'; we'd miss pathological input that has a comma on the next
line, but I don't think we have to tolerate input that runs multiple
objects together on the same line.

We could also write our own semi-parser. It shouldn't be too hard, if
we demand ASCII input; we just need to balance "{" and "}" outside of
strings, and the strings are just balancing quotes.

PR #2222, opened 2019-04-17
updates protobuf

ekg:
This is needed to fix the build on a quirky, ancient system I'm working with.

It seems to be related to this: https://github.com/protocolbuffers/protobuf/issues/5264


adamnovak:
Who wants to run vg on FreeBSD/powerpc64?

PR #2226, opened 2019-04-23
Transition to new interface for path handles

jeizenga:
I switched the HandleGraph code throughout the codebase to the new and improved path interface. This includes removing HandleGraph support for the concept of graph ordering, adding support for circular paths, and retooling iteration along paths. All of the HandleGraph implementations in this repository now implement the new functionality.


ekg:
Maybe we have to add a sortable handle graph interface. Or it can just be
something specific to my implementations.

PR #2239, opened 2019-04-26
Xdrop Documentation and Giraffe Integration

adamnovak:
I'm trying to document how the XdropAligner works internally, so I can modify it so I can implement `align_pinned()` in it, so I can use it for aligning tails in the MinimizerMapper. I would also like to move it from directly using the Protobuf Graph objects to using HandleGraphs, eventually.

I've added some unit tests for the existing behavior, and I'm trying to write out what things actually mean, but I don't yet understand the algorithm for those descriptions to really be correct.

@ocxtal Can you answer some questions for me?

As far as I can tell, I need to feed in a topologically-sorted Graph and a forward-strand Alignment no matter whether I pass `reverse_complement=true` or `reverse_complement=false`. Why would I use one or the other? Can you help me write a unit test to illustrate that the `reverse_complement` parameter is interpreted correctly? What's an example where I would get a wrong answer if I passed the wrong value of `reverse_complement`?

What does `direction` mean? If `direction=true`, does computation proceed over the graph from left to right in topologically-sorted order, or from right to left? In general, why are two passes in opposite directions necessary? The [dozeu example](https://github.com/ocxtal/dozeu#example) shows only one pass.

It seems like maybe we use one pass to find a "head". What is a "head"?

What's the difference between `dz_scan` and `dz_extend`? The dozeu example uses only `dz_extend`, but here we use both of them in different places, and I'm not clear on the semantics of each.

Can you explain the logic behind `index_edges` and `index_edges_head`, and how that relates to the `direction` and `reverse_complement` flags? It looks like depending on the flag, everything about each of them changes: the edges are packed into the int64_t values in a different order, and the values are sorted in the vectors in a different order. Can you sum up what each of them is supposed to represent? Presumably one of them is normal, and one of them is for the head finding pass?

If I feed in no MEMs, it looks like it does some alternative logic instead of one of the passes. I *think* it is looking for a good seed on its own, subject to some constraints, but it's hard for me to tell what the constrants are. Is it something like, "find the best alignment between the first 15 bases of the read and the first node in the graph"? If not, what exactly are the limitations of `scan_seed_position`? What does it find, and when might it not find what it is looking for?

Thanks!


adamnovak:
Also, @ocxtal, is it intended behavior that the full length bonus is only applied once even if both ends of the alignment make it to the end of the read? The normal Aligner applies it once per end.

ocxtal:
Sorry but I haven't read the comment in the codes yet, but here are the specification and what is intended in the X-drop aligner.

Firstly, what are expected for the input are:

* the graph object is topologically sorted.
* the `reverse_complemented` flag represents whether the graph is flipped when extracted from the original graph; false when not flipped.

The X-drop aligner strongly depend on the MEM information to locate the extension beginning point (though the others do not). This is the reason the X-drop aligner requires MEMs along with its direction with respect to the graph. I wonder the direction information could be extracted from the graph object itself (as in ...), but I wasn't sure of the exact behavior of the extraction algorithm. If the direction can be extracted correctly from the graph, it would be better removing the direction argument.

The reason we need two passes of extension is that the dozeu only computes *semi-global (or extension)* alignment, not local nor global. The X-drop aligner first extend toward the downstream of *the original graph* then toward the upstream. The extension direction is normalized to the original graph so that gaps are aligned to the same direction when the resulting alignment path is projected to the original graph. It is also intended to eliminate the influence of suboptimal seeds which might occur around homopolimers or tandem repeats, by starting second extension from the first maximum-scoring position (thus it reports alignment between forward and reverse max positions).

The scan is fallback for empty MEM bin. The difference between `dz_extend` and `dz_scan` is that the scan function do not penalize arbitrary length gap at the head of *reference* sequence. The scan is used to locate the most likely anchor between the graph and the first 15 bases of the query. The first 15 bases is an arbitrary setting; reasonable for Illumina reads which are mostly mapped end-to-end but not for long reads. I believe this is not a problem so far because the scan path (MEM == empty) is only invoked from mate rescuing path but should be fixed in the future.


The index_edges array is an auxiliary data structure to traverse the graph downstream-to-upstream direction on the original graph, regardless of whether the graph of argument is flipped or not. The array is designed to have two properties: incoming edges are obtained as a slice of the array for each node; the slices (each corresponds to each node) are topologically ordered in the direction that the graph is traversed. Since the graph of argument is always topologically sorted, the index_edges array is essentially unnecessary; it is placed to make it easy to write the graph traversing algorithm. The index_edges_head is similar structure for the downward extension from the root; "head" originally meant a node of the upstream max, but the extension direction has flipped for some reason. I should have corrected the name accordingly when the direction was changed, but I would like to give it a more appropriate name if possible.


Lastly, the full-length bonus is given per-end manner.


Regards,

adamnovak:
The output alignment is always in the orientation of the extracted graph, whether `reverse_complement` is true or not, right?

Why does the algorithm care about the orientation of the extracted graph in the original graph, if it has no access to the original graph?

adamnovak:
@ocxtal I'm having trouble getting pinning to work correctly. I'm accomplishing it by sticking an extra base on the front of the graph and the front of the sequence, setting that match as xdrop's head_pos, and doing only the second alignment pass from there and then the traceback.

This doesn't seem to always work. Here's a bit of my debug log showing that I'm getting an alignment that starts with deletions at my head position. Weirdly, the resulting Alignment protobuf object then ends up with an empty Edit in its first Mapping.

```
[src/xdrop_aligner.cpp: calculate_and_save_alignment(513)] ref_length(44), query_length(41), score(38), path(DDDMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(526)] node_id(0), subpath_length(1), subpath(D)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(526)] node_id(1), subpath_length(1), subpath(D)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(526)] node_id(2), subpath_length(1), subpath(D)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(526)] node_id(3), subpath_length(5), subpath(MMMMM)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(526)] node_id(4), subpath_length(1), subpath(M)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(526)] node_id(5), subpath_length(19), subpath(MMMMMMMMMMMMMMMMMMM)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(526)] node_id(6), subpath_length(1), subpath(M)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(526)] node_id(7), subpath_length(15), subpath(MMMMMMMMMMMMMMM)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(587)] rid(7, 8), ref_length(23), ref_offset(0), query_length(41), query_init_length(0)
[src/xdrop_aligner.cpp: calculate_and_save_alignment(592)] i(0), rid(0, 15), ref_length(1), path_offset(0), span->offset(0)
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 0M in {"node_id":"15"}
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 1D in {"node_id":"15"}
[src/xdrop_aligner.cpp: calculate_and_save_alignment(592)] i(1), rid(1, 14), ref_length(1), path_offset(1), span->offset(1)
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 1D in {"node_id":"14"}
[src/xdrop_aligner.cpp: calculate_and_save_alignment(592)] i(2), rid(2, 13), ref_length(1), path_offset(2), span->offset(2)
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 1D in {"node_id":"13"}
[src/xdrop_aligner.cpp: calculate_and_save_alignment(592)] i(3), rid(3, 12), ref_length(5), path_offset(3), span->offset(3)
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 5M in {"node_id":"12"}
[src/xdrop_aligner.cpp: calculate_and_save_alignment(592)] i(4), rid(4, 11), ref_length(1), path_offset(8), span->offset(8)
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 1M in {"node_id":"11"}
[src/xdrop_aligner.cpp: calculate_and_save_alignment(592)] i(5), rid(5, 10), ref_length(19), path_offset(9), span->offset(9)
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 19M in {"node_id":"10"}
[src/xdrop_aligner.cpp: calculate_and_save_alignment(592)] i(6), rid(6, 9), ref_length(1), path_offset(28), span->offset(28)
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 1M in {"node_id":"9"}
[src/xdrop_aligner.cpp: calculate_and_save_alignment(592)] i(7), rid(7, 8), ref_length(23), path_offset(29), span->offset(29)
[src/xdrop_aligner.cpp: push_edit(475)] push_edit: 15M in {"node_id":"8"}
[src/xdrop_aligner.cpp: calculate_and_save_alignment(602)] rv: (8, 0) -> (15, 1), score(38), AAGCTGAGGGAATAGTGCCTGGCATCGAGGAAAGCCTCTGA

Alignment from Dozeu: {"sequence":"AAGCTGAGGGAATAGTGCCTGGCATCGAGGAAAGCCTCTGA","path":{"mapping":[{"position":{"node_id":"15"},"edit":[{},{"from_length":1}],"rank":"1"},{"position":{"node_id":"14"},"edit":[{"from_length":1}],"rank":"2"},{"position":{"node_id":"13"},"edit":[{"from_length":1}],"rank":"3"},{"position":{"node_id":"12"},"edit":[{"from_length":5,"to_length":5}],"rank":"4"},{"position":{"node_id":"11"},"edit":[{"from_length":1,"to_length":1}],"rank":"5"},{"position":{"node_id":"10"},"edit":[{"from_length":19,"to_length":19}],"rank":"6"},{"position":{"node_id":"9"},"edit":[{"from_length":1,"to_length":1}],"rank":"7"},{"position":{"node_id":"8"},"edit":[{"from_length":15,"to_length":15}],"rank":"8"}]},"score":38,"identity":1}
```

Is this supposed to happen? Did I need to do something special to `head_pos`? right now I have it set up so it can be all 0s (base 0 of the query in correspondence to base 0 of node 0 in the topological order). I noticed that some bits of part of `head_pos.query_offset` are being moved into `state` for the traceback; can you explain how that works and why that happens?

ocxtal:
The resulting alignment is supposed to be in the orientation of *the original graph*, not the extracted graph.

This behavior is mainly for variang calling, so that SNV / short indel callers won't be confused with shifted gaps around homopolimers or tandem repeats. (I haven't confirmed if gap shifting really affect the result of variant calling or not, though, I believe this behavior is preferred.)

ocxtal:
The `state` is an intermediate variable to hold a tuple of (counter, op) in the traceback loop. The pair is initialized differently depending on whether there is clip at the head (upstream of the graph) of the alignment. If there is, the initial values are (clip length, 'I'), representing clip at the head, and are dumped as `{ from_length: 0, to_length: <clip length>, sequence: <query subsequence for the clip> }` immediately it enters the loop. If there is not, the initial values are (0, 'M') so that the counter to accumulate the length of the first contiguous matches.

This conversion routine assumes that the alignment begins with clip or match, thus adding mismatching block at the beginning of the alignment by modifying `head_pos` might result in wired behaviour. A simple solution I think is adding dedicated `push_edit`s at the prologue and the epilogue of the traceback loop (though I'm not sure about the detailed pinned-alignment specs).

Thanks,

adamnovak:
I've set up Giraffe so you can use xdrop for tail alignments with `--xdrop`. I want to talk to @jltsiren about #2250 before doing some real speed tests.

adamnovak:
It seems like merging in master has reduced our performance on calling based on vg map. Also, in my preliminary testing, swapping Giraffe over to xdrop seems to make its performance worse.

I'm going to see if I can do something to address that. Maybe by the time I want to merge that, I will be sure that any lower baseline from master has had time to hit S3.

adamnovak:
I'm still having trouble where using `--xdrop` to do xdrop alignment of tails in Giraffe actually ends up being slower than normal alignment. But I want to merge this into master to keep things in sync. The next step on this project is working out why that would be the case; I did some profiling and it looked like only a little time was spent on the actual xdrop matrix fill, and a lot of time was spent e.g. allocating and deallocating XdropAligners.

ocxtal:
It has been the case since the X-drop aligner was first implemented. The execution time was mostly consumed by unfold-dagify operations in the code 6 months ago. In my understanding they are actually hash table retrieval and memcpy so I hope the performance will improve with the new `HandleGraph` interface (though it is still vague how much it will).

As for the decrease in the performance of calling, I think it will improved if the behavior is matched more closely with the original implementation. It was arbitrary decision that the direction of the gap placement was aligned to the original graph, and in fact it might be affecting the calling to some extent that it can not be ignored. If the behavior of the GSSW aligner is described in detail (or preferred behavior of an aligner is defined), I'm ready to modify the dozeu and corresponding XdropAligner codes to follow that.

adamnovak:
We shouldn't have to be doing any unfolding or dagification in the
pinned alignment codepath I am trying to apply X-drop in right now. I
will profile it to see what the holdup is.

PR #2243, opened 2019-04-30
Mapper tweaks

ekg:
This is meant to deal with an annoying problem that I've run into when working with vg for long read mapping.

The default stream parallel chunk size is 256. However, when working on long reads, it's often the case that we have only a few thousand reads. This leads to us running three or four cores when we should be using the entire set.

@adamnovak do you think this is something we could push upstream into the vgio lib? Is that where these things live now?


adamnovak:
Yes, this probably needs to go in libvgio now; `for_each_parallel_impl` moved over there with the rest of the stream functions. It definitely makes sense to bring out the batch size and make it configurable for cases like this.

PR #2250, opened 2019-05-03
Intercept and drop extension sets with negative offsets

adamnovak:
@jltsiren I'm seeing some gapless extensions producing paths with (invalid) negative offsets in them.

With this branch that detects and complains about them, you can do this:

```
aws s3 cp s3://vg-data/pop-map/graphs/21/v4/snp1kg-CHR21_filter.vg .
aws s3 cp s3://vg-data/pop-map/graphs/21/v4/snp1kg-CHR21_filter.xg .
aws s3 cp s3://vg-data/pop-map/graphs/21/v4/snp1kg-CHR21_filter.gbwt .
aws s3 cp s3://vg-data/pop-map/graphs/21/v4/sim-NA12878-90-10000000-32/sim.gam .
vg snarls -t snp1kg-CHR21_filter.vg >snp1kg-CHR21_filter.snarls
vg index -s snp1kg-CHR21_filter.snarls -w 1000 -j snp1kg-CHR21_filter.dist snp1kg-CHR21_filter.vg
vg minimizer -i snp1kg-CHR21_filter.min snp1kg-CHR21_filter.xg

vg gaffe -x snp1kg-CHR21_filter.xg -m snp1kg-CHR21_filter.min -d snp1kg-CHR21_filter.dist -s snp1kg-CHR21_filter.snarls -H snp1kg-CHR21_filter.gbwt -G sim.gam >mapped.gam
```

With this branch you get some warnings:

```
Warning: Skipping extension on read seed_100_fragment_21713_1 due to negative offset in gapless extension: {"mapping":[{"position":{"node_id":"4004152","offset":"-30","is_reverse":true},"edit":[{"from_length":1,"to_length":1}],"rank":"1"}]}
Warning: Skipping extension on read seed_100_fragment_10693_1 due to negative offset in gapless extension: {"mapping":[{"position":{"node_id":"1906814","offset":"-2","is_reverse":true},"edit":[{"from_length":1,"to_length":1}],"rank":"1"}]}
Warning: Skipping extension on read seed_101_fragment_4315_1 due to negative offset in gapless extension: {"mapping":[{"position":{"node_id":"4004152","offset":"-30","is_reverse":true},"edit":[{"from_length":1,"to_length":1}],"rank":"1"}]}
Warning: Skipping extension on read seed_101_fragment_24119_2 due to negative offset in gapless extension: {"mapping":[{"position":{"node_id":"1472368","offset":"-2","is_reverse":true},"edit":[{"from_length":1,"to_length":1}],"rank":"1"}]}
Warning: Skipping extension on read seed_101_fragment_32473_2 due to negative offset in gapless extension: {"mapping":[{"position":{"node_id":"3334744","offset":"-3"},"edit":[{"from_length":2,"to_length":2}],"rank":"1"}]}
Warning: Skipping extension on read seed_101_fragment_39763_1 due to negative offset in gapless extension: {"mapping":[{"position":{"node_id":"1906814","offset":"-2","is_reverse":true},"edit":[{"from_length":1,"to_length":1}],"rank":"1"}]}
Warning: Skipping extension on read seed_101_fragment_47400_1 due to negative offset in gapless extension: {"mapping":[{"position":{"node_id":"4098983","offset":"-2","is_reverse":true},"edit":[{"from_length":1,"to_length":1}],"rank":"1"}]}
Warning: Skipping extension on read seed_101_fragment_45367_1 due to negative offset in gapless extension: {"mapping":[{"position":{"node_id":"4004152","offset":"-30","is_reverse":true},"edit":[{"from_length":1,"to_length":1}],"rank":"1"}]}
```

Can you figure out why we're getting these negative offsets? When I do the same test with master I get crashes on some later asserts, when I just let these invalid paths into the chaining code.


jltsiren:
I can't access the files. It apparently requires some sort of credentials, but I have no idea how S3 credentials work.

adamnovak:
I also have those files on Courtyard in `/public/home/anovak/build/vg/trash/clustering`, if you want to look there.

jltsiren:
This only happens when you build the minimizer index without GBWT. Then you may get seeds in nodes that are not in the GBWTGraph, because the nodes are not used in any haplotypes. For example, 4004152 is one such node.

If we have a minimizer from the reverse strand of the read, we need to reverse the hits before using them. For this, we call `get_length()`. If a node does not exist in the GBWT index but its identifier is between real nodes, `get_length()` returns 0 and the offsets in the reversed hits are negative.

adamnovak:
Hm. That explains why I wasn't seeing this before. I must have built my minimizer index with the GBWT before.

Should the GBWT graph just throw an error when an attempt is made to use an unvisited node, instead of returning a 0 length? Then it would be clearer what the problem is.

jltsiren:
I don't think it's appropriate in such low-level code. Most HandleGraph implementations just assume that the handle is valid and return whatever they can find in their data structures.

adamnovak:
So it looks like the way the GBWTGraph stores its nodes is to have a
succession of indexes in a giant string at which the nodes start, one
per node ID that might be in the graph.

If the node ID idn't in the graph, we have it as taking up no space,
so the next node's offset is the same as the nonexistent node's
offset.

We could add a check in get_handle to verify that the node exists, but
that would slow things down there. We could also add a check in
get_length or get_sequence, where we already have to compute the 0
length or empty sequence of non-existent nodes. Checking the return
value and throwing an error if it indicates that the node never
actually existed would be more efficient there.

We could also just disable generating the minimizer index without the GBWT...

Jouni, how would you suggest we fix things so nobody makes the same
mistake I did again?

jltsiren:
The high-level issue is that we can't detect when we have an index of the wrong graph. The graph induced by the haplotypes is not necessarily the same as the original graph.

We could store a hash of the graph in the minimizer index and check it when we build the GBWTGraph, but this is not a general solution. In many cases, we are deliberately indexing pruned or otherwise simplified graphs and using the indexes with the original graphs.

Low-level checks for node existence would be expensive. GBWT is internally also a byte array with an index mapping node identifiers to intervals in the array. But while GBWTGraph uses an integer array as the index, GBWT uses `sdsl::sd_vector`.

PR #2269, opened 2019-05-15
Plumbing for SV evaluation of Giraffe

adamnovak:
To run and evaluate Giraffe on SV graphs, I needed to get `vg annotate` to search much further for path hits.

To do this, I had to change the functions it was using internally; it looks like `XG::next_path_position` never really did what it was supposed to. It now uses a double-ended Dijkstra-ish search like a lot of Jordan's XG functions, instead of just checking immediately adjacent nodes (and accumulating the lengths of the ones that don't have paths). There is also now a test for it.

I also refactored single-ended Dijkstra out into its own HandleGraph algorithm, even though I turned out not to need it.

### DIFF
--- a/scripts/giraffe-facts.py
+++ b/scripts/giraffe-facts.py
@@ -662,6 +662,9 @@ def main(args):
     # Close the TSV
     tsv.close()
     
+    # Print the table now in case histogram plotting fails
+    print_table(read_count, stats_total)
+    
     # Plot the histogram
     svg_path = os.path.join(options.outdir, 'times.svg')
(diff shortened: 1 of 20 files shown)


adamnovak:
OK, I now have Giraffe generating little reports based on overall runtime, instead of annotating reads with per-read or per-stage runtimes, since @jltsiren found that those were very variable and influenced the code being timed.

The reports look like this (on a toy example):

```
Mapped 163840 reads across 8 threads in 11.2024 seconds.
Mapping speed: 1828.18 reads per second per thread
```

Provenance tracking should still work.

PR #2273, opened 2019-05-20
Use 0-based indexing in BED parsing

jeizenga:
I think we had a mistake in the BED parsing, where we were assuming that the indexes were 1-based. However, the format standard is pretty explicit about the indexing being 0-based:
https://genome.ucsc.edu/FAQ/FAQformat.html#format1

Some chance this has been biasing the RNA mapping experiments. @jonassibbesen, thoughts?

### DIFF
--- a/src/transcriptome.cpp
+++ b/src/transcriptome.cpp
@@ -95,11 +95,11 @@ void Transcriptome::add_transcripts(istream & transcript_stream, const gbwt::GBW
             continue;
         }
 
-        // Parse start and end exon position and convert to 0-base. 
+        // Parse start and end exon position.
         getline(transcript_stream, pos, '\t');
-        int32_t spos = stoi(pos) - 1;
+        int32_t spos = stoi(pos);
         getline(transcript_stream, pos, '\t');
-        int32_t epos = stoi(pos) - 1;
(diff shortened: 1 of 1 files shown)


src/transcriptome.cpp
```diff
@@ -95,11 +95,11 @@ void Transcriptome::add_transcripts(istream & transcript_stream, const gbwt::GBW
             continue;
         }
 
-        // Parse start and end exon position and convert to 0-base. 
+        // Parse start and end exon position.
         getline(transcript_stream, pos, '\t');
-        int32_t spos = stoi(pos) - 1;
+        int32_t spos = stoi(pos);
         getline(transcript_stream, pos, '\t');
-        int32_t epos = stoi(pos) - 1;
```
glennhickey: BED endpoints are exclusive, so you probably wanna leave the -1 for epos

jonassibbesen:
I don't think that there is a mistake. It is not a BED file that is parsed in *transcriptome.cpp*, but instead a GTF/GFF file. Coordinates are 1-based and inclusive in that format: 
https://genome.ucsc.edu/FAQ/FAQformat.html#format3

jeizenga:
Ah, my bad, sorry!

PR #2277, opened 2019-05-22
Revise AlignmentEmitter to be less contentious

### LINKED ISSUE #2271: vg cannot use 32 threads for mapping
I would like to use all the 32 vCPUs on my system to map reads.

However, this is what happens:

Screenshot from 2019-05-17 13-25-38

I can only use about 30% of the available resources.

It used to be that `vg map` would use all 32 threads in this case until it completed its job. What has changed?

adamnovak:
This should fix #2271, I think. @ekg ought to test it before it is merged.

I'm replacing the current `OMPThreadBufferedAlignmentEmitter` with an `OMPSingleWriterBufferedAlignmentEmitter`. It uses @benedictpaten's suggested approach of having one writer thread that goes around emptyiung everone's buffers.

This might be reducing contention: instead of all the worker threads fighting over the lock for the output file when they align their 1000th reads and fill their buffers, they all just keep aligning. They only have to contend for their buffers with the writer thread, and the writer thread doesn't have to contend with anybody for the output file.

I have some logic in there to try and stop the memory usage from growing unboundedly if you have more threads than your disk IO lets the writer service in a timely fashion. But I don't think we're likely to hit it, at least given the mapping speed of vg map.

Possible improvements include real double-buffering or some kind of ring buffer, so the writer thread can quickly unblock the worker thread whose buffer it needs to empty, without making it wait for the serialization to happen.


ekg:
I'm getting a segfault when trying to emit SAM. It seems that the tests are failing due to something similar.

To test I switched to JSON output.

It is definitely true that we can saturate all the CPUs on a multicore system (32 vCPUs in my case). But, this doesn't seem to be stable. Memory usage climbs, and eventually most of the threads are spending much of their time in contention, possibly memory related, see the red bars for CPU usage in the attached screenshot.

Screenshot from 2019-05-22 11-12-09

At the point I took this, we'd added 8GB to RAM over the index cost.

The writing thread doesn't seem to be working all the time. Rather, it occasionally writes at 40k reads/s, then stops. In this particular context, I'd expect vg map to run at at least 1k reads/CPU/s, so we'd need to sustain 40k read outputs a second to keep up.

In bwa mem I see a kind of staggered process, where buffers are read in by each aligning thread, outputs are written out, and then collected and written very quickly. There does seem to be some overlay of writing and aligning. It's not perfectly using the available CPUs, but it seems to be striking a decent balance between IO bounding and buffer filling, and manages to keep up with a process that's running around 2k reads/CPU/s.

ekg:
A few minutes later, and we're back to a similar situation as before.

Screenshot from 2019-05-22 11-17-24

ekg:
We should make sure that all the compute cost of transforming the alignments into the output data type is done in each thread. Only the actual write should be managed by the writer thread. It should just be dumping raw data into the output stream we've chosen and cleaning up the buffer as it goes.

adamnovak:
Yeah, it looks like my HTSlib code doesn't work at all.

It seems like the key thing to do is to minimize the work the writer thread is doing, because on your system at least it can't keep up when doing JSON output. As it is now the AlignmentEmitter design puts the thread caching before actual serialization. I need to flip that around to do per-thread caching of interleave-able blocks of serialized bytes.

I think the right way to do this is to put it after BGZF compression as well. the HTSlib BAM output code we want to use is [tied to writing to a BGZF object](https://github.com/samtools/htslib/blob/5cfc3ede097f20ccea9504b757a402ea5062bb3d/sam.c#L532), and I already have the hard part of implementing an hFILE done for gluing that to C++ streams.

So my new plan is to implement a ThreadSafeStream that goes over a C++ stream and provides a C++ stream interface, plus a method to mark when you have reached a coherent point in your writing and what you have written so far can be shipped off to the backing stream. Internally it can use the box-collecting design, with some limit on how many bytes are allowed to be buffered before writes block.

Then we can push the real serialization work into the worker threads, and let the writer thread just ship bytes like you suggest. This gets us multithreaded BGZF compression for free, too, which is something I've wanted to have. All the caching at the AlignmentEmitter level goes away, with each AlignmentEmitter just being responsible for constructing the ThreadSafeStream and flushing periodically to provide interleaving points.

adamnovak:
OK, HTSlib output still isn't debugged, but I have a multiplexer in libvgio now. You get a stream per thread, and it can ask you to come to a synchronization point, at which point you flush your protobuf/bgzf/etc. and tell it to take what you have written. Then there's a writer thread that writes it all to a single output stream, with some queues inbetween. Locks only happen when enqueueing and dequeueing, which are just moves, so its' super non-contentious.

I can get up to ~5500% CPU with this on Courtyard (although I haven't tested the writer yield I just added).

Still TODO: write htslib output in a way where it can use the multiplexer and still do lazy header initialization. I'm going to need a lot of open HTS streams, one per thread, I think, and then I will just not write the headers to them but write them to the backing stream instead.

ekg:
I'm getting this build failure

```
deps/libvgio/src/stream_multiplexer.cpp: In member function ‘void vg::io::StreamMultiplexer::writer_thread_function()’:
deps/libvgio/src/stream_multiplexer.cpp:156:18: error: ‘std::current_thread’ has not been declared
             std::current_thread::yield();
                  ^~~~~~~~~~~~~~
```

ekg:
Looks like the same problem for travis. Perhaps you didn't update your submodule pointer?

ekg:
Another issue from libvgio:

```
deps/libvgio/src/message_iterator.cpp:67:76: warning: overflow in conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘int’ changes value from ‘10000000000’ to ‘1410065408’ [-Woverflow]
         coded_in.SetTotalBytesLimit(MAX_MESSAGE_SIZE * 2, MAX_MESSAGE_SIZE * 10);
```

adamnovak:
Hm. I must be missing an `#include <thread>` somewhere. And I thought I fixed that overflow warning already, but maybe we have two copies.

adamnovak:
OK, I haven't hit the overflow warning yet, and I may have had a regression in speed between 
4997720  and 8796998  (maybe the writer is yielding too aggressively?) but I do have working SAM output now.

BAM output doesn't work, and I'm not sure we'll ever get CRAM output to work without a bunch of CRAM-specific code. The HTSlib output functions seem to think that you're going to write a header to each output file, which we aren't doing if each thread gets its own multiplexed stream.

ekg:
What's the current state of this?

Would it be unreasonable to just write full BAM files into the buffers and have the merging thread remove 
the headers? We'd have to make the BGZIP block stop at the end of the header in order to avoid recompressing.

ekg:
Perhaps you forgot to push to libvgio?

```
In file included from src/indexed_vg.hpp:15,
                 from src/indexed_vg.cpp:6:
src/stream_index.hpp: In instantiation of ‘void vg::StreamIndex<Message>::index(vg::StreamIndex<Message>::cursor_t&) [with Message = vg::Graph; vg::StreamIndex<Message>::cursor_t = vg::io::ProtobufIterator<vg::Graph>]’:
src/indexed_vg.cpp:44:31:   required from here
src/stream_index.hpp:883:19: error: ‘using cursor_t = class vg::io::ProtobufIterator<vg::Graph>’ {aka ‘class vg::io::ProtobufIterator<vg::Graph>’} has no member named ‘has_next’; did you mean ‘has_current’?
     while (cursor.has_next()) {
            ~~~~~~~^~~~~~~~
            has_current
```

adamnovak:
@ekg The current status of this is sadly that I pulled in changes from master and immediately went on vacation without actually getting the merge commit to build. But I'm back now and that's my job today.

As for merging, I think I'm going to pursue something similar to what you propose. I looked at the htslib code, and it looks like the output is flushed when headers are written, so they should already be getting their own BGZF blocks. I think I am going to have every thread write a header, but then just drop the header data instead of actually sending it off to be multiplexed, since I will always have it alone in its own interleave-able chunk in the multiplexer. All I have to do is *not* put it in the queue to be picked up by the writer thread; the writer doesn't have to do any detection of anything.

ekg:
Oh! Sorry to pester! I didn't realize you were talking time off.

Unfortunately, I've found that hacks like this are basically the "correct" way to do things in htslib land. That said, they don't seem to cost much. They're just funky.

adamnovak:
OK, here's a working (i.e. bash-tap map and surject tests pass) implementation of htslib output, based on writing and discarding the headers. We use multithreaded compression, and interleave runs of pre-compressed blocks in the output file.

However, I think that tcmalloc is to blame for my performance "regression" [I mentioned](https://github.com/vgteam/vg/pull/2277#issuecomment-496700345), and possibly a lot of the the speed improvement I thought I produced by multithreading the serialization. The mutrace profiler only works when you are using the normal allocator, so I did a bunch of my testing with tcmalloc off. With tcmalloc off, on `plaza` with its 64 CPUs, I can use >4000% CPU at peak, with mutrace showing very little contention on the IO mutexes. With it on, I am limited to maybe 2400% CPU at peak.

Here's time results for the same short mapping run (`time vg map -x trash/clustering/snp1kg-CHR21_filter.xg -g trash/clustering/snp1kg-CHR21_filter.gcsa -G trash/clustering/reads/sim.gam > trash/clustering/time.gam`), using this PR's code, without tcmalloc:

```
real	0m26.372s
user	10m49.976s
sys	0m45.359s
```

And with tcmalloc:

```
real	0m40.746s
user	11m26.671s
sys	2m4.472s
```

So @ekg, try building with and without tcmalloc and seeing if that gets your performance up. Maybe there are tcmalloc features we can tune to make it scale better, or maybe we actually shouldn't be using it.

adamnovak:
It could just be that tcmalloc doesn't like the memory patterns I am using win the multithreaded serialization: allocating big blocks of memory in the worker threads, and then freeing them in the writer thread. That's not the pattern it is designed for, given that it is "thread caching" malloc. I could try adding a return queue to send buffers back to the original threads to be deallocated, which might reduce tcmalloc contention.

See this similar issue where people talk about not shipping big buffers between threads in order to make tcmalloc happier: https://issues.apache.org/jira/browse/IMPALA-4923?focusedCommentId=16024131&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16024131

ekg:
As for allowing multithreaded mapping, this definitely works for me.

But you're right that getting the htslib BAM file merging to work is tricky. What I see is that only the first chunk that's written is able to be parsed by samtools. The writing continues but it seems to violate the BAM format somehow, and so later bits can't be read.

adamnovak:
OK, I'm able to reproduce the bad htslib output. I suspect that the problem is that writing the headers flushes the BGZF, but flushing the BGZF doesn't flush the backing hfile, so I need to add code to do that manually or things will get out of order and headers won't actually be discarded.

adamnovak:
OK, this should fix htslib output.

I've also added some code to reconfigure tcmalloc's `tcmalloc.max_total_thread_cache_bytes` property to be higher. I found that setting the corresponding environment variable `TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES` higher. The default is 16 MB spread across all the caches for all the threads, and the docs suggest upping it if you have a lot of threads and are experiencing bad performance and tcmalloc-induced contention.

I've changed it to 16 MB per default-OMP-thread-count thread (i.e. per core, in general). Setting it to 16 MB per actual `-t` thread would be difficult, because there's no single function we go through. I'm doing the configuration during static initialization, in a .o that we won't link in on Mac, where we don't link against tcmalloc.

This lest me use nearly all the cores on Courtyard with `vg map`.

ekg:
@adamnovak curious why this is happening:

Screenshot from 2019-06-04 12-45-24

adamnovak:
OK, it looks like the problem here is a missing hFILE flush in the MessageEmitter. I'm fixing it now.

adamnovak:
@ekg If you want to see the failed tests on Gitlab, you have to scroll down in that annoyingly tiny popup window. You can only see the first 2 of the 19 tests that run.

adamnovak:
OK, I'm pretty sure this is the right answer, so I am going to merge this. @ekg, let me know if this doesn't resolve the issue for you.

PR #2284, opened 2019-05-28
augment refactoring (wip)

glennhickey:
Some refactoring work to better use vg's edit() functionality in variant calling
* Pull most edit logic out of the vg class and into augment.cpp, and use handle graph interface where possible (paths and translations left in protobuf for now)
* Fully deprecate `vg mod -i` (replace with `vg augment -i`)
* Add interface to augment from a GAM stream.  This requires an extra pass but saves memory.  A 30x GAM for chr21 takes now takes 9G instead of 36G memory, for example.
* The old VG::edit interface for lists of paths left unchanged as it's used in a lot of places. 

Would like to get it a bit smaller/faster, but I think it's close to being usable at the chromosome level which is part of what we'll need to pull `vg chunk` and `pileup` out of the calling pipeline.


glennhickey:
It looks like the memory usage is dominated (>90%) by storing the graph itself (1.5G for the original grows to about 8.5G for the augmented).  So trying out a different backing graph does seem like the next step as far as reducing memory goes.

ekg:
I think it's time we bite the bullet and drop in one or several of the HandleGraph implementations for testing.

btw, I've got a new version of xg that's independent of protobuf and builds from GFA: https://github.com/vgteam/xg

glennhickey:
For sure!

glennhickey:
Good idea.  I think the sparse_hash_maps used (via hash_map in hash_map.hpp) are the biggest issue, and will try to globally replace them (along with unordered_maps) with ska_flat_hash in another pr.  

unordered_map may have its problems, but using sparse_hash_map in its place for those handful of augmenting structures more than doubled overall memory usage in my test.

PR #2288, opened 2019-05-30
vg find a path DAG

ekg:
This lets us get all the nodes in a path range including those in bubbles when the path range that's given is for a partially ordered graph.


ekg:
@adamnovak is there a problem with the CI? The report suggests that everything is fine, but the CI failed for a rather vague reason: https://ucsc-ci.com/vgteam/vg/-/jobs/10763

adamnovak:
I think we can have that failure whenever the Gitlab server is too
busy to take uploads from the nodes "fast enough". The monitoring page
at https://ucsc-ci.com/admin/system_info shows that our disk is about
half empty. I'm not sure if artifacts expired since this failed, but I
think the out-of-space error is an internal server error instead.

We could reduce the size of the files our pipeline wants to upload, or
we could try and increase Gitlab's timeout values:
https://forum.gitlab.com/t/timeout-when-uploading-artifacts/14160

PR #2289, opened 2019-05-30
Handle graph conversion CLI

glennhickey:
I'm sure I'm duplicating more code here, but I really want to start playing with other graph implementations and couldn't find any other way to make packed and hash graphs. 

I think this should really live somewhere apart (sglib?), but think it's harmless here in the interim.

### DIFF
--- a/src/subcommand/convert_main.cpp
+++ b/src/subcommand/convert_main.cpp
@@ -1,0 +1,148 @@
+#include "subcommand.hpp"
+#include "../vg.hpp"
+#include "../utility.hpp"
+#include "../xg.hpp"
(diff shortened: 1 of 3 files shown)


ekg:
I think we should add a from_gfa method to the various implementations so that they can be built directly from a text format. But there is no harm to test it out here.

glennhickey:
OK, I'll merge this in for now.  

For gfa, it looks like handelifying the relevant functions in [gfa.cpp](https://github.com/vgteam/vg/blob/master/src/gfa.cpp) would give us a general gfa interface for all the implementations.

PR #2290, opened 2019-05-31
More augment tweaks

glennhickey:
Change up the `edit` function to only make translations when wanted, as they take a lot of memory.   (If translations are needed, streaming them out instead of buffering them all, along with moving away from protobuf, can also fix this.) 

This finally gets the memory in check.  Augmenting a 1.5M node chr21 graph into a 12M node graph using 10M illumina reads goes from over 150G to 4.5G RAM:

* Embed each read in the graph (old `mod -i` logic): `>150G` / `hours`
* and augment graph, but put paths in augmented GAM:  `63G`
* and  don't buffer entire input GAM in memory: `30G` / `17m`
* and don't make translations: `13G` / `12m`
* and use HashGraph instead of VG: `8G` / `9m`
* or use PackedGraph instead of VG: `4.5G` / `11m`

### DIFF
--- a/src/augment.cpp
+++ b/src/augment.cpp
@@ -11,9 +11,13 @@ namespace vg {
 using namespace std;
 
 // The correct way to edit the graph
-vector<Translation> augment(MutablePathMutableHandleGraph* graph, istream& gam_stream,
-                            ostream* gam_out_stream, function<void(Path&)> save_path_fn,
-                            bool break_at_ends, bool remove_softclips) {
+void augment(MutablePathMutableHandleGraph* graph,
+             istream& gam_stream,
+             vector<Translation>* out_translations,
+             ostream* gam_out_stream,
+             function<void(Path&)> save_path_fn,
+             bool break_at_ends,
+             bool remove_softclips) {
     // Collect the breakpoints
     unordered_map<id_t, set<pos_t>> breakpoints;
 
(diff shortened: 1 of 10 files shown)


jeizenga:
Nifty

ekg:
If I match odgi to the new HandleGraph interface would you test it too?

ekg:
And, most importantly: way cool!

glennhickey:
Yeah, I can definitely run the same test on odgi if it implements `MutablePathMutableHandleGraph` from the current libhandlegraph.  

Maybe Monday we can discuss the roadmap for working with all these implementations.  Are we going to hide them behind Adam's magic VPKG framework?  Can we somehow get them all as submodules in vg either directly or recursively in a graph-library repo? 

I'm going to start comparing ways to query the augmented GAM now.  Will try GBWT like you suggested way back,  the current GAM index and a succinct `vg pack`-like support structure.

ekg:
I've updated odgi to match the appropriate handlegraph model. I should probably experiment with bringing it into vg.

glennhickey:
Yeah, I think getting odgi into some kind of library form is going ot be necessary for me to run this experiment on it.

ekg:
I think all that has to happen is to build a static lib out of it. It's
already there.

You're just using the vg resident packedgraph and hashgraph?

glennhickey:
This is probably me not knowing cmake too well.  How do I build the static library for odgi?  I'm only getting a binary executable.

ekg:
Oh, I need to add in the option to do so. Let me do that and I'll ping you.

PR #2293, opened 2019-06-04
Optimize after augment

glennhickey:
Move that path-reindexing and sorting from the end of VG::edit to VG::optimize, and call the latter from augment.  This should let augment() be a better drop-in replacement when dealing with a VG back end.

### DIFF
--- a/src/augment.cpp
+++ b/src/augment.cpp
@@ -147,6 +147,9 @@ void augment(MutablePathMutableHandleGraph* graph,
     if (out_translations != nullptr) {
         *out_translations = make_translation(graph, node_translation, added_nodes, orig_node_sizes);
     }
+
+    // optimize the graph.  VG will use this to rebuild its in-memory path index which is important. 
+    graph->optimize();
 }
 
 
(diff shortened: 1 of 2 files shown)


src/augment.cpp
```diff
@@ -147,6 +147,9 @@ void augment(MutablePathMutableHandleGraph* graph,
     if (out_translations != nullptr) {
         *out_translations = make_translation(graph, node_translation, added_nodes, orig_node_sizes);
     }
+
+    // optimize the graph.  VG will use this to rebuild its in-memory path index which is important.
```
adamnovak: Is this important for correctness? Or just for speed? I'm worried the optimize method might be picking up some new semantics.

glennhickey:
Was trying to shoehorn logic from the old edit function into the new
augment function, without leaving any vg-specific code in the latter.  I
did indeed feel a bit dirty doing it.  I'll take a closer look at how
necessary this is and where else it can go before merging.

adamnovak:
@glennhickey Do you still want to merge this?

glennhickey:
I will do this differently in another PR.  But wouldn't mind keeping this
one open for a bit just as a reminder....

PR #2296, opened 2019-06-05
Option to weight support by quality in vg pack

glennhickey:
If a position in the graph is covered by two reads in the GAM, one with MAPQ=60 and the other with MAPQ=30, it will get a coverage of 2 in the pack.   This PR adds the `-q` option for quality-scaling which would instead give it 90.  The idea is to use this as a very simple heuristic to downweight coverage from low quality alignments.  When a base quality is available, it gets combined with the MAPQ.  For edges, the average base quality of its two incident Mappings is used.  

A variant caller may want to have access to both the scaled and unscaled support at the same time (in addition to some strand information), but we can worry about that later.

### DIFF
--- a/src/packer.cpp
+++ b/src/packer.cpp
@@ -253,7 +253,6 @@ void Packer::add(const Alignment& aln, bool record_edits, bool qual_adjust) {
     for (size_t mi = 0; mi < aln.path().mapping_size(); ++mi) {
         auto& mapping = aln.path().mapping(mi);
         int mapping_quality = aln.mapping_quality();
-        bool has_base_quality = !aln.quality().empty();
         if (!mapping.has_position()) {
 #ifdef debug
             cerr << "Mapping has no position" << endl;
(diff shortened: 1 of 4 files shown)


ekg:
Right now, pack is using only one byte to store values <256 during construction. If you sum qualities, then the memory usage will go up significantly, because it looks into a table for the numbers above 256.

One option is to use a wider integer in this structure, such as 16 bits. That might keep things compact. Or to use a dyn::hacked_vector from DYNAMIC, which will automatically keep the bitwidth at the lowest possible level, and will do so in a semi-local sense. This will mean that regions of higher coverage take more bits, and lower coverage take lower amounts.

I can put this into a future PR. For the moment, this particular PR won't break. It's just that the quality summing mode will be expensive memorywise.

glennhickey:
Good point, thanks!   This didn't pop out in my (augmented) chr21 tests, I think, because half the chromosome is Ns and has no coverage:
RAM / Wall-time / Size on Disk:
No quality: `3.6G` / `2m30` / `79MB`
With quality: `6.5G` / `3m26` / `108MB`
Bumping the width to 16:
With quality: `5.7G` / `3m00` / `108MB`

I'm going to start testing on larger inputs soon. Can try hacked_vector if memory becomes an issue.

Issue #2280, opened 2019-05-24
Compilation error on ubuntu 


RenzoTale88:
Good morning,
I'm trying to install vg on a ubuntu 16.04 machine. I've performed all the steps described in the guide. However, I get a compilation error involving swap in xdrop_aligner. Below I pasted the error I get. Is it because I miss some dependency? Thank you in advance for your help,

Andrea

Error displayed:

```
src/xdrop_aligner.cpp: In lambda function:
src/xdrop_aligner.cpp:184:55: error: ‘swap’ was not declared in this scope
             swap(handle_pair.first, handle_pair.second);
                                                       ^
src/xdrop_aligner.cpp:184:55: note: suggested alternatives:
In file included from src/hash_map.hpp:18:0,
                 from src/handle.hpp:17,
                 from src/xdrop_aligner.hpp:18,
                 from src/xdrop_aligner.cpp:10:
/home/atalenti/Documents/Software/vg/include/sparsepp/spp.h:4329:13: note:   ‘swap’
 inline void swap(spp_::sparse_hash_set<Val, HashFcn, EqualKey, Alloc>& hs1,
             ^
In file included from /home/atalenti/Documents/Software/vg/include/sdsl/structure_tree.hpp:12:0,
                 from /home/atalenti/Documents/Software/vg/include/sdsl/int_vector.hpp:25,
                 from /home/atalenti/Documents/Software/vg/include/sdsl/bit_vectors.hpp:8,
                 from /home/atalenti/Documents/Software/vg/include/sdsl/wt_pc.hpp:25,
                 from /home/atalenti/Documents/Software/vg/include/sdsl/wavelet_trees.hpp:32,
                 from /home/atalenti/Documents/Software/vg/include/gcsa/utils.h:36,
                 from /home/atalenti/Documents/Software/vg/include/gcsa/support.h:29,
                 from /home/atalenti/Documents/Software/vg/include/gcsa/files.h:29,
                 from /home/atalenti/Documents/Software/vg/include/gcsa/gcsa.h:28,
                 from src/mem.hpp:4,
                 from src/xdrop_aligner.cpp:9:
/usr/include/c++/5/sstream:783:5: note:   ‘std::__cxx11::swap’
     swap(basic_stringbuf<_CharT, _Traits, _Allocator>& __x,
     ^
In file included from /usr/include/c++/5/regex:61:0,
                 from src/utility.hpp:16,
                 from src/position.hpp:6,
                 from src/mem.hpp:6,
[...]

PR #2299, opened 2019-06-07
Update elfutils to fix a warning-error on GCC 9.1.0

adamnovak:
This should fix the second iteration of #2280.


adamnovak:
The elfutils build fails on Gitlab with this:

```
make[1]: *** No rule to make target '../lib/libeu.a', needed by 'i386_gendis'.  Stop.
```

I'm not sure the elfutils build process is really robust enough to run it outside maintainer mode.

We should probably just be pulling this in as a system-level dependency instead. But for now I think I will just fix the immediate new warning and leave the build process like it was, because it worked before.

PR #2300, opened 2019-06-07
Turned off tcmalloc check

xchang1:
Turned off assert so that valgrind can be used
@adamnovak


adamnovak:
Sorry, I meant more like #2301 instead. We still want to make the call into tcmalloc; we just don't want to stop the program if tcmalloc doesn't listen.

PR #2307, opened 2019-06-12
Move xg into the vg namespace

jeizenga:
We need to compile libvg and xg0 into the same binary for the handle graph evaluations, which means the XG implementations are clashing. The easiest, guilt-free-est stopgap solution I could think of was to move the VG implementation of XG into the vg namespace, leaving the xg namespace to xg0. I don't think this will create any substantial difficulty for us later.


ekg:
Sounds good to me. We can always move things back around later.

PR #2311, opened 2019-06-17
move to xg0 index

ekg:
This is my work from today in integrating the new xg index into vg.

The new xg index (https://github.com/vgteam/xg), which I've been calling xg0, removes a lot of functionality that was in the old xg, including functions that allow for manipulation of protobuf objects. It also removes the gPBWT, which is not used now that we have a fully developed GBWT. This has resulted in a lot of problems during this merge.

It matters to pull xg0 in because it simplifies path storage and can scale its construction to whole genome splicing graphs with hundreds of thousands of embedded paths. It also caches node to path mappings in a simple array, which should help us improve the performance of this previously slow operation that the mapper clusterers depend on or benefit from. It is also implemented independently of vg, and that will allow other tools to directly produce and use the index without pain.

I'm currently adding functions that are needed in many places, in particular those related to subgraph extraction, to `xg_position.cpp/hpp`. @jeizenga has mentioned that we should probably move these functions into algorithms. I'd be all in favor of that, but was hoping to move quickly on this.

There is a ton of work still to be done on this. If you make progress, please push quickly and mention if you're working on it in slack so we don't step on each others' toes. CC @glennhickey


glennhickey:
I added that `forward_only` flag back to the expand_subgraph functions, as I'm pretty sure tube maps needs it.

jeizenga:
This has been superseded, no?

ekg:
Yes, we should kill this branch.

PR #2315, opened 2019-06-21
Reconstruct deconstruct

glennhickey:
The need for a graph-to-vcf converter has come up as we're looking at SV graphs built from whole genome alignments.  `vg deconstruct` fit the bill but didn't work on many inputs.  This PR is an attempt at fixing it up a bit.  It's now running on my various tests, including graphs from cactus and seqwish, but still needs more testing especially on inversions and nested sites, as well as maybe a way of lumping different path names together in the same genome.  

In addition to trying to fix the existing logic, I added an option `-e` to consider only traversals corresponding to embedded paths.  This gets around issues of not being able to enumerate all traversals in complex sites.  And in this mode, each non-reference path in the graph is listed as a sample in the VCF, and the genotype field is used to link vcf alleles back to their path names in the graph.


ekg:
Awesome

PR #2320, opened 2019-06-25
Remove path to component index from XG

jeizenga:
@ekg I moved the path to component index into its own external object as part of the XG refactor


ekg:
Awesome thanks

ekg:
The path oriented distance stuff is also backed into the old XG index.
Specifically: XG::closest_shared_path_unstranded_distance is not something
that I've pulled into the new XG. Perhaps I should though.

jeizenga:
Now that we have a handle implementation for path offsets, I could move it into an algorithm

PR #2326, opened 2019-07-02
Added missing libcairo-dev dependency

code-s-witch:
VG did not install on my machine using Linux Mint 19.1 Tessa, required this dependency so I updated the files with requirement.


ekg:
Thank you!

PR #2329, opened 2019-07-03
Transition to handle interface for oriented distance code to improve code segmentation with XG

jeizenga:
As requested by @ekg 

Pretty big one here. I removed a lot of the pre-handle dependencies on the internals of XG that were relying on its path position queries. Most of the changes are in the MultipathMapper clustering code and in the Surjector. All of the memoizing code has also now been pushed into an overlay over a PathPositionHandleGraph. 

I've gotten rid of the following functions from XG (and thus they can also be removed from the new XG):
- memoized_paths_of_node
- memoized_oriented_occurrences_on_path
- memoized_oriented_paths_of_node
- memoized_get_handle
- closest_shared_path_unstranded_distance
- closest_shared_path_oriented_distance
- jump_along_closest_path
- validate_strand_consistency

To support this work, I also had to flesh out the libhandlegraph PathPositionHandleGraph interface a bit, but I think in pretty sensible ways. However, to keep this PR self-contained, I added these new methods to the old XG rather than glomming this into the new XG PR. As long as the new XG can keep the same semantics on these new handle methods, it should be fine (they're not very complicated).


ekg:
Great! Thank you!

I guess that protocol here is to merge this into master and then pull that into the xg0 PR branch. That's what you were intending?

jeizenga:
Yep, exactly. Good chance of a few merge conflicts, I'm afraid, but all of the new code I wrote is handle-y. Hopefully that makes it relatively easy to massage with the new Xg.

ekg:
I think it will work out. This was the major blocker for me.

jeizenga:
There's still some similar code emanating from `XG::nearest_offsets_in_paths` that I think @adamnovak wrote? I could take a stab at refactoring that in a similar way, but I have to work on some other things before the holiday. Maybe could get to it Friday or over the weekend.

jeizenga:
This lagging Travis job has actually completed, I don't know why it's not reporting it. I'm going to go ahead and merge.

PR #2330, opened 2019-07-05
Use tree-shaped tail graphs for tail alignment in Giraffe

adamnovak:
This needs to be brought up to speed with @xchang1's new Giraffe heuristics and with @jltsiren's gapless extension refactor.

But, this PR improves mapping speed by about 20%, by using tree-topology graphs for aligning GBWT tails, instead of using one linear graph per possible tail sequence.

Here's some tests from an m5d.metal, with one thread per full core:

```
vg gaffe -p -x graphs/whole_genome_graph/whole_genome.xg -m graphs/whole_genome_graph/whole_genome.min -H indexes/all_full.gbwt -d graphs/whole_genome_graph/whole_genome.dist -f ~/graphs/whole_genome_graph/real_148_200000.fastq -t48 --linear-tails > test/linear-tails/linear.gam

...
Mapped 200000 reads across 48 threads in 14.3036 seconds.
Mapping speed: 291.302 reads per second per thread
            
vg gaffe -p -x graphs/whole_genome_graph/whole_genome.xg -m graphs/whole_genome_graph/whole_genome.min -H indexes/all_full.gbwt -d graphs/whole_genome_graph/whole_genome.dist -f ~/graphs/whole_genome_graph/real_148_200000.fastq -t48 > test/linear-tails/tree.gam

...
Mapped 200000 reads across 48 threads in 11.633 seconds.
Mapping speed: 358.176 reads per second per thread
```


adamnovak:
I also still need to assess the accuracy impact of this change.

adamnovak:
OK, I checked the accuracy of this on my 200k chr21 simulated reads. Tree-topology alignments got 190201 correct, with average identity 0.997086, and linear-topology alignments got 190200 reads correct, with average identity 0.997084. So there's no large difference in read placement accuracy.

adamnovak:
I've just checked this latest commit for accuracy on 200k chr21 simulated reads. All combinations of (xdrop, gssw) and (linear tails, tree tails) get 190189 reads out of 200k reads correct.

adamnovak:
Here are some mapping speeds on whole-genome real reads on an m5d.metal:

| | Xdrop | GSSW |
|---|---|---|
| Linear | 88.1124 | 229.943 |
| Tree | 377.142 | 271.037|

Moving from linear to tree tails increases speed by 18%. Additionally, turning on Xdrop makes us a further 39% faster when using tree tails, although it makes us very slow for some reason with linear tails.

I don't think it's worth trying to debug the speed of linear-tail xdrop when it's clearly not the right algorithm anyway.

Right now, tree-tail GSSW is the default, and xdrop can be enabled with `--xdrop`. Since xdrop now makes us faster, we might consider making it the default.

PR #2331, opened 2019-07-05
Move xg methods

adamnovak:
This is some of the refactoring @jeizenga wanted in #2329. I've only done a couple of the dependency methods of `nearest_offsets_in_paths`. I still need to finish moving/reimplementing `offsets_in_paths` before I can do `nearest_offsets_in_paths` itself.


ekg:
Thanks Adam. I had dropped these in the new XG...

PR #2332, opened 2019-07-07
Finish moving XG methods

jeizenga:
This PR includes and finishes https://github.com/vgteam/vg/pull/2331 (I would have pushed it up to that branch, but I'm having trouble with git)

I finished pulling out `nearest_offsets_in_paths` from XG into the algorithms namespace. I also fixed the problem where it was actually not a Dijkstra traversal while I was at it.

Finishing up Adam's work, I rewrote the offset finding code in `offsets_in_paths` in `nearest_offsets_in_paths` (it's much simplified by the handle graph interface). I also deleted `offsets_in_paths` in XG. It's just a special case of `nearest_offsets_in_paths`, so I replaced it with a wrapper of `nearest_offsets_in_paths` in the algorithms namespace.


jeizenga:
I got annoyed with my bugs and did away with `next_path_position` and `find closest with paths`, which were only ever used inside `nearest_offsets_in_paths`. I replaced them with some extra code in `nearest_offsets_in_paths`. Personally, I think this cleans things up a bit. Also, this now actually works.

PR #2365, opened 2019-07-24
Use uncompressed VPKG for everything

adamnovak:
This makes all VPKG files written by vg uncompressed, as @jltsiren and @ekg suggested.

We retain support for reading the compressed files, and we probably want to actually compress GAM, VG, and other Protobuf-based formats, but I want to make sure everything at least works with compression off.

If you want to compress a VPKG file, you can do it with `bgzip`. The compressed version will be loaded anywhere the uncompressed one is accepted.

As part of this change, I had to disable the non-VPKG "bare" loaders when reading from non-seekable streams, because they don't let me preview enough of the stream data to tell if the VPKG framing is there or not when the GZIP magic number isn't. In practice I don't think we're ever loading things like the minimizer index from pipes, so this shouldn't be a problem, and with the ability to do things uncompressed there's less of a reason not to use the VPKG framing on everything.

@jltsiren Is there a particular speed benchmark that you would suggest to run with this, to test compressed vs. uncompressed vs. completely raw indexes?


adamnovak:
I think that this isn't building on Mac Travis because the system htslib headers (from the htslib from Macports that Samtools needs) are taking precedence over the ones we ship during the vg build.

adamnovak:
OK, the problem we are encountering here with the mac builds is that vg's `include` directory is added to the `CPLUS_INCLUDE_PATH` environment variable by the `source_me.sh` script. The directories from that variable are searched *after* all the `-I` options on the command line, including those added by pkg-config that tell us to look in the Homebrew/Macports include directories. Moreover, if a directory appears both as a `-I` option *and* an entry in `CPLUS_INCLUDE_PATH`, the `-I` option is ignored and the directory is only searched when the `CPLUS_INCLUDE_PATH` directories are searched.

So even though we have an `-I` option to make the vg include directory be searched first, it's actually being searched more or less last.

If you add `-Wp,-v` to the `CXXFLAGS` you can see this in action.

I think I am going to fix this by removing the vg include directory from the directories that `source_me.sh` sets up. We can leave the Makefile itself in charge of providing it.

PR #2366, opened 2019-07-24
Obliterate sglib

adamnovak:
Some bits of sglib survived #2361, presumably because the removed headers and submodules stayed cached on our CI somehow? Master didn't build when I updated to it. This should excise all references to "sglib" from cpp and hpp files and make master build for other people.


jeizenga:
Weird. My bad.

PR #2338, opened 2019-07-10
Move all xg-related methods into handle based algorithms

ekg:
As discussed on this week's call, I'll be pulling in all the changes I made in xg0 here, testing at each step that things are OK. When we don't have any more non-handle xg interfacing, it will be trivial to drop in the new xg.

PR #2368, opened 2019-07-25
Handlify some things

jeizenga:
Starting with `target_alignment`, which I moved to `alignment.[ch]pp`


jeizenga:
We seem to have an intermittent failure in the `vg sim` integration tests due to some changes in https://github.com/vgteam/vg/pull/2338 of the `Mapper` distance function, which is also used to score alignments. I think I can work around it.

PR #2370, opened 2019-07-26
Add an option to skip joining extensions

adamnovak:
Not joining any extensions seems to make Giraffe ~8% faster.


adamnovak:
It looks like there was an intermittent bug in the read scoring in `vg sim` that has been fixed in master, so I merged master into this branch to get the fix. Hopefully it will pass all the tests now.

PR #2374, opened 2019-07-30
Handlify haplotype extracter and expand context

jeizenga:
I handlified the `HaplotypeExtracter`. It still has some Protobuf bits that feel dated, but I think that it's the primary interface for SequenceTubeMaps and I don't want to break that. 

Of note, as part of this I wrote a handle-based replacement for `expand_context` that I think could replace the equivalent method in both `XG` and `VG`. However, the semantics are ever-so-slightly different, but in a way that I think makes it more well-defined. Accordingly, I'm not going through to replace calls to the old versions yet.

### DIFF
--- a/src/algorithms/are_equivalent.cpp
+++ b/src/algorithms/are_equivalent.cpp
@@ -87,7 +87,7 @@ namespace algorithms {
         if (!are_equivalent(graph_1, graph_2, verbose)) {
             return false;
         }
-        
+                
         if (graph_1->get_path_count() != graph_2->get_path_count()) {
             if (verbose) {
                 cerr << "graphs have different numbers of paths: " << graph_1->get_path_count() << " and " << graph_2->get_path_count() << endl;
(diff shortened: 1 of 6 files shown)


jeizenga:
@adamnovak I have some failing builds using GCC, which seem to have something to do with htslib. I see you made some changes involving htslib in https://github.com/vgteam/vg/pull/2365. Does this look like it might have anything to do with those?

adamnovak:
I think the problem might be that your build is running off of master's cache, but without either master's updated htslib or the code I added to the Makefile to clean out the old htslib headers before building the new htslib. So the cached headers are getting into your htslib build and breaking it. You should merge master into your branch, and it will solve the problem.

ekg:
Jordan, check out subgraph.cpp in algorithms.

jeizenga:
Hmm. Yeah, I think those functions could probably be composed in the right way to get something pretty close to the semantics I was trying to copy from XG's `expand_context`. That probably could have saved me some time...

PR #2376, opened 2019-07-31
Handlify sublinear Li & Stevens code and RegionExpander

jeizenga:
In addition, update a lot of the documentation to reflect the fact that it is no longer XG specific, make fewer things fall in the build cascade of XG, and switch to a libhandlegraph that doesn't have a memory leak we'd missed before.

### DIFF
--- a/src/chunker.cpp
+++ b/src/chunker.cpp
@@ -19,24 +19,22 @@ PathChunker::~PathChunker() {
 }
 
 void PathChunker::extract_subgraph(const Region& region, int context, int length,
-                                   bool forward_only, VG& subgraph, Region& out_region) {
+                                   bool forward_only, MutablePathMutableHandleGraph& subgraph, Region& out_region) {
 
-    // extract our path range into the graph
-    Graph g;
-    xg->for_path_range(region.seq, region.start, region.end, [&](int64_t id, bool) {
-            *g.add_node() = xg->node(id);
-        });
-    
-    // expand the context and get path information
-    // if forward_only true, then we only go forward.
-    xg->expand_context(g, context, true, true, true, !forward_only);
-    if (length) {
-        xg->expand_context(g, context, true, false, true, !forward_only);
-    }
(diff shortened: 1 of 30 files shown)


jeizenga:
I think this might need @adamnovak 's attention when he gets back from vacation. Both the Linux and Mac failures seem to be caused by some compilation issue with `libbdsg`. In the Linux tests, it fails at the linking stage with some vtable related error. For Mac, it makes it all the way to run time and then crashes on a simple call to a HashGraph method.

ekg:
Is this crash possibly related to a dynamic cast?

jeizenga:
I don't think it's likely. For one, I think a dynamic cast would cause problems at run time rather than at linking, so it wouldn't explain the Linux failures. Also, on Mac I was able to reproduce the crash with this block:

```
auto cluster_graph = new bdsg::HashGraph();
volatile bool test = cluster_graph->has_node(1);
```

adamnovak:
OK, I think the problem here is that libbdsg isn't actually getting rebuilt to account for the updated libhandlegraph headers. The vg build process sees that it needs a rebuild, but when it goes in and invokes libbdsg's build process, libbdsg's Makefile doesn't see the changes in libhandlegraph's headers, and decides libbdsg is up to date and doesn't need a rebuild. Linking objects built against two different versions of the same headers results in Bad Things Happening.

We can't list the paths to the libhandlegraph headers in the libbdsg Makefile by hand; I think we need to set up automatic dependency rule generation for the libbdsg build, so that when the compiler builds things for the first time it writes out bits of Makefile that make the resulting .o depend on all the headers, even installed headers from other packages, that went into the build. I think you can do that by following https://stackoverflow.com/a/52036564 except using `-MD` (which includes what the compiler sees as system headers) instead of `-MMD` (which only makes rules for user headers).

The other solution is to drop a `make clean` into the vg Makefile rule that invokes the libbdsg build, so every time vg decides it needs to build libbdsg, it does a full rebuild. However, that will be slower to rebuild.

PR #2378, opened 2019-08-02
Mcmc

code-s-witch:
MCMC_Genotyper now passes all unit tests.


adamnovak:
@code-s-witch Are you sure you actually added and committed all the files? Your commit claiming to implement a copy constructor doesn't actually have any code for a copy constructor, and the PR tests fail because there's no PhasedGenome copy constructor.

adamnovak:
We already merged your current work, so I am going to close this old PR.

PR #2379, opened 2019-08-02
[WIP] Turn extensions from a set into alignments best-first

adamnovak:
This PR makes each extension in an extension set get turned into an Alignment, if it would be the best, rather than throwing them all into a MultipathAlignment and letting the multipath logic work out a winner.

I've refactored a bunch of the best-first traversals of things to use a common template function.

What I don't have yet is any early stopping, so this might turn out to be slower than what we have now. I'm running some speed measurements to see if my refactoring is needlessly slow, and how much of a hole the early stopping would have to dig us out of if we took these changes.


adamnovak:
My removal of the multipath alignments from the algorithm seems to have made it much slower. I don't know if the cost is actually sorting the extensions as opposed to filtering them, or if the problem is that I'm splicing together a bunch of complete Alignments that I turn out not to need, or what. I'm going to investigate; I've broken off a faster intermediate version of the code into a new PR.

PR #2386, opened 2019-08-12
Handlify XG out of vg chunk

glennhickey:
Unfortunately, since we don't have path ranks in our interface, chunks can no longer be stitched back together (see https://github.com/vgteam/vg/commit/325531c3205aba0060155b506326e147ce01dfc3)


ekg:
We need to discuss the semantics that are needed. Maybe there is a way to
do this that is compatible with rGFA. Are ranks the best way?

glennhickey:
I think I'd vote for a single positional offset (like S0 in rGFA) for a given path.

PR #2387, opened 2019-08-12
Handle-ify SLLS

adamnovak:
This builds on #2376 and fixes the build system to fully rebuild libbdsg when the libhandlegraph headers change. This ought to pass CI.

PR #2390, opened 2019-08-12
Tolerate -T in vg chunk without a haplotype file to trace by tracing no haplotypes

adamnovak:


adamnovak:
This fixes the tube map crash I complained at @jeizenga about in https://github.com/vgteam/vg/pull/2387#issuecomment-520568883

PR #2398, opened 2019-08-14
Clarify graph building requirements for new users.

djhogan:
I didn't understand that a VCF index was required to build a graph.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ The simplest thing to do with `vg` is to build a graph and align to it. At prese
 vg construct -r small/x.fa -v small/x.vcf.gz >x.vg
 ```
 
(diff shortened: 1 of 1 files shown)


adamnovak:
Hello @djhogan! Thanks for pointing this out!

Unfortunately, I'm not sure we actually do provide a `tabidx` command; it certainly doesn't land in vg's `bin` directory. Can you point out where it is?

When I want to make an index of a VCF, I usually do `tabix -p vcf whatever.vcf.gz`, using the `tabix` command from SAMtools.

djhogan:
My mistake. Thanks for pointing that out.  I fixed the typo (tabidx -> tabix) and referenced SAMtools.

PR #2402, opened 2019-08-16
drop in xg0 (take two)

glennhickey:
This is a fresh stab at #2311, made much easier by all the handlefication efforts in the meantime.

As of this commit, it's compiling but not linking due to a couple trailing XGPath usages..


glennhickey:
Having the libhandlegraph version specified in two places 
* https://github.com/vgteam/libhandlegraph/tree/47791b556bb963d7f1922548cf512749e59dab86
* https://github.com/vgteam/xg/blob/64165b94e1b114d6e42341d1b2fd65a05418f816/CMakeLists.txt#L55

is a wee bit dangerous (a little mismatch gave me weird segfaults)

glennhickey:
From the [first failed bash tap test](https://github.com/vgteam/vg/blob/master/test/t/05_vg_find.t#L93-L96), it looks like creating an xg index is dropping a bunch of paths
```
vg construct -m 1000 -r small/xy.fa -v small/xy2.vcf.gz -R x -C -a 2> /dev/null | vg view -j - | sed s/_alt/alt/g | vg view -Jv - >w.vg
vg paths -v w.vg -L | wc -l
39
vg index w.vg -x w.xg
vg paths -x w.xg -L | wc -l
30
```

ekg:
It might be that the specific code path that is used by vg wasn't tested. I
had been building from GFA during testing.

glennhickey:
Yeah, this error seems to be coming from empty paths getting dropped by `XG::from_path_handle_graph()`.  Pretty minor, I think (provided the format supports empty paths).

glennhickey:
`make test` is finally passing.

glennhickey:
I think this is just about there.  Only problem left I'm seeing is that VGSet::to_xg(), which I lifted from the other PR, is too inefficient in the presence of alt paths, and causes the CI tests to run out of memory. 

The good news is that these graphs can be xg-indexed directly via vg convert in very little memory (though more slowly than the old xg), so it's nothing to do with the index itself -- just the rank merging code in vgset.

glennhickey:
The tests all seem to be passing locally.  

@adamnovak Any idea how to get Travis to rebuild deps/xg, along with its correct cmake submodules?  Also, which node type is gitlab using?  Thanks.

adamnovak:
If Travis is hopelessly broken, you can go into the "More Options" menu on the upper right of the main Travis page for the PR, and go to "Caches". Form there you can delete the cache for the branch, and the cache for the PR, which correspond to the two different Travis checks. Then the next Travis build will use the cache from the current master branch instead.

If your PR can't build with the cache from master, it probably can't build from a non-clean state on people's machines either. You *can* delete the master cache too, but then the next master Travis build will need to get rerun a few times to rebuild all the cached stuff, because we can't fit in a full build with no cache under Travis's build time limit.

The Gitlab CI system runs all the tests on c5.4xlarge machines, which it tags as `vg-ci-runner-machine`.

glennhickey:
Great thanks.  I'd tried cleaning the PR/branch caches.  Will try master
now too.

ekg:
This bug is a potential (but probably minor) issue:
https://github.com/vgteam/xg/issues/4

It seems to be that gfakluge's parsing code isn't handling the last path
step in the file correctly.

glennhickey:
Most of these changes are now in master via #2409.   All the test failures in this PR are therefore due to either
* bugs in new xg
* subtle build problems (My main suspect, as everything runs fine when I build locally)
* bugs in VGSet::to_xg() (only vg code change here that's not in master)

A future PR adding the new xg should only need to adjust the Makefile and submodules as well as the aforementioned VGSet function.  Hopefully the third time is the charm.  Anyone else wanna give it a shot? @adamnovak @ekg ?

adamnovak:
I'm going to set up a PR that makes your other changes on top of this. We will see how that goes.

PR #2406, opened 2019-08-21
CI experiments for new xg

glennhickey:
This is a PR where I fight with FKA Jenkins to see why tests are failing.  Starting with `test_map_mhc_primary_genotype` which times out now.  Seems to run fine locally.  From the logs, it seems like maybe creation of an xg index on a chunk graph is hanging?


glennhickey:
The test seems to pass when run in isolation.  I think there's something going on where we sometimes can't make two xg indexes at once.  Perhaps a temp file collision or running out of disk or memory?

PR #2416, opened 2019-08-27
Add a Dockerfile that specifies a multi-stage build

adamnovak:
With this Dockerfile and Quay.io automated builds, I think we can replace most of the vg_docker repo's functionality.

I have it set up with automated builds for testing at https://quay.io/repository/adamnovak/vgtest?tab=builds so we should check in there to make sure it built properly before merging this.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
 Dockerfile
-.git
 .dockerignore
 Brewfile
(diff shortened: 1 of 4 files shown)


adamnovak:
OK, I managed to get this to not try and ship the source code in the run container (at the cost of some extra tiny layers/commands in the Dockerfile). I'm still not sure *why* that works, but we can probably proceed with this Dockerfile.

adamnovak:
The Travis Mac jobs have been timing out, presumably since we cleared out some caches. I've kicked them again to see if they progress, but the Dockerfile changes shouldn't affect Travis, so I'm not sure we really need to wait for them.

PR #2418, opened 2019-08-28
[WIP] Handlify (and revamp) caller

glennhickey:
This is a  cleanup of `vg call` and all the code behind it.  Major changes are
* Deprecate pileup and go all in on pack.
* Operate directly on xg (required for pack) via handle interface
* Reorganize calling code to fit 2 interfaces
     * `SnarlCaller`: Compute genotype from a set of traversals.  `PackedSupportSnarlCaller` implements this interface to genotype using packed support.  It uses a pared down and cleaned up version of the old `vg call` genotyping code.
    * `GraphCaller`: Compute variants on a graph (using a SnarlCaller).  Two of these are implemented, `VCFGenotyper` which genotypes a given VCF and `LegacyCaller` which is a refactor of the old call logic that heuristically finds traversals top-down on nested snarls.

One of the main goals of this is to be less dependent on `vg chunk`.  Provided the xg can be loaded with alt paths, I think genotyping should be doable directly on whole-genome graphs.  *de novo* calling will use the same pipeline but with an augmentation step that may require working on chromosome-graphs.  

The new code is currently working well on some SV test sets.  Still a fair bit of integration to do (especially in gitlab tests) before merging.


glennhickey:
Will merge this in now.  Next step is to benchmark (and document) whole-genome SV genotyping. 

I've kept all the old logic (and even preserved the old subcommand as vg call-old) just because it made getting through tests so much easier.  I will remove it in a fresh PR within the next week or so.  

Some of the SNP calling benchmarks went down slightly.  In almost all cases, it seemed to be due to going from pileup augmentation to edit-based augmentation as the latter can make more complicated snarls (so using packed support with the old caller gave similar results).

PR #2411, opened 2019-08-23
Change xg

adamnovak:
This makes the changeover to the new XG.

@glennhickey @ekg It's failing some of the unit tests with xg assert failures in my hands. I'm heading out for the weekend and won't be able to fix any of them soon; if either of you happen to have time, since I think you know the new XG better than I do, you might want to take a look. I suspect you might be able to see some of them if you built the xg tests by themselves through cmake in non-release mode.

PR #2430, opened 2019-09-05
Change xg again

glennhickey:
#2411 was pretty close to passing CI tests, but I think merging in master threw a wrench in things.  I'm just trying to roll that back to see if the CI tests pass.  Will make it easier to focus on problems arising from merging from master.


glennhickey:
Oops, the toil-vg this is pointing to with the fixed docker also requires the new caller which is only in master.  sheesh.

PR #2431, opened 2019-09-05
[WIP] Activate overlays in caller

glennhickey:
Goal is to accept any graph implementation instead of just xg, and use path-position and vectorizable overlays as needed.


glennhickey:
The thing about a Travis cache, it's got lifeless eyes, black eyes, like a doll's eyes. When it comes at you it doesn't seem to be livin'... until he bites you, and those black eyes roll over white.

Issue #2453, opened 2019-09-13
vg call breaks on cyclic graphs


ekg:
More feedback from CPANG19.

A lot of student groups here built graphs with methods that result in complex (sometimes cyclic) structures. Usually, the graph is mostly in linear components, with occasional diversions into complex regions. Many tried to run `vg call`, but this usually broke because of cycles.

Is it possible to just skip regions that are cyclic, or make a reasonable decision about what version of the local reference path to use?

PR #2468, opened 2019-09-18
enable vg call on cyclic graphs

ekg:
This doesn't change anything about `vg call`'s `LegacyCaller` logic, but it does change the way that we determine reference positions to report. Previously, as noted in #2453, call would break if it saw a cyclic path. This change will always report the snarl at the first reference position. The snarl start is determined as the first, and the snarl end is found by path iteration in the graph. This could result in mistakes, but at least it won't blow up. I've included a test that verifies this works in a simple graph with a cyclic path.

This should probably only merge after @glennhickey's ok.


ekg:
I think we need tests to make that work correctly. I thought that we would
get the node handle / step handle in the right orientation. This is in the
handle graph documentation at least.

glennhickey:
I see.  I don't think `get_next_step` goes backward when the given handle is reversed, though.  From xg's implementation:
```
as_integers(next_step)[0] = as_integers(step_handle)[0];as_integers(next_step)[1] = as_integers(step_handle)[1] + 1;
``` 
If I'm understanding that, we'd need to manually toggle get_previous if the start step handle is reversed wrt the ref path.  Anyway +1 for a test.

ekg:
Seems I was confused. Maybe we should clarify the semantics then. If we
take a step on a handle, should get next follow the directions implied by
the relative orientation between the handle and the path?

jeizenga:
IMO it would be a pretty big headache to change those semantics now that we've handlified so much of the path-based code, but it would be relatively simple to add a generic wrapper function over `get_next_step`/`get_previous_step` that does what you're describing.

ekg:
Ok. Then in the iteration I need to check that the path is oriented the
same as the handle or not and iterate along the path appropriately.

PR #2483, opened 2019-09-25
Update some overlays

glennhickey:
Switch to `PackedPositionOverlay` in caller.  Thanks @jeizenga

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ vg call xa.xg -k aln.pack -v small/x.vcf.gz > genotypes.vcf
 Pre-filtering the GAM before computing support can improve precision of SNP calling
 ```sh
 # filter secondary and ambiguous read mappings out of the gam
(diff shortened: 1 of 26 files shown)


glennhickey:
Any handle-graph implementation can now be used in place of xg for all the subcommands.  So something like `vg find -x graph.vg -n 1` would work now, for instance.  

Not using an `xg` comes at a performance cost in most cases though, and it will be very noticeable on larger graphs.  For instance, mapping seems to run about 15% slower on packed or hash graphs compared to xg.  There's also an added ~10 minute loading overhead when using the packed or hash graphs for mapping vs xg. 

But sometimes it's nice to avoid the xg.  When doing something like `augment->pack->call`, it helps to not have to make a new xg index between `augment` and `pack`, and there doesn't seem to be much of a performance hit when using packed graph throughout (`vg pack` has even run slightly faster in some cases, I think due to naming synergy).

PR #2496, opened 2019-10-05
Simulate RNA reads from haplotype transcripts

jeizenga:
Adds a bit of machinery to use the output of `vg rna` to simulate reads from haplotype transcript embedded paths, according to an RSEM expression profile.


jeizenga:
The Mac Travis is having a problem with Protobuf. Seems the vg.pb.h that it's using wasn't generated by the same version as the libprotobuf that's getting linked in.

jeizenga:
@adamnovak I'm still seeing the protobuf version mismatch error on the Mac Travis. Do you have any insight? Maybe this is something we could fix by clearing the Travis cache?

adamnovak:
Yeah, I've been seeing it on my branch's Mac Travis too. It *probably*
could be fixed by clearing the master cache and the caches for the PR
and branch you are testing. But that still leaves open the question of
how it got started (maybe through an upgrade of Protobuf in
Homebrew?).

PR #2499, opened 2019-10-08
Reorganize path dropping from graphs

jeizenga:
I moved the methods to add and drop paths from a graph from `vg mod` into `vg paths`, which has a much better interface for selecting paths. I also expanded the interface to include the catch-all "list of names in a file" option and a convenience filter for the embedded paths added by `vg construct -a`.

@glennhickey Do you think toil-vg will need any updating?

### DIFF
--- a/src/algorithms/copy_graph.cpp
+++ b/src/algorithms/copy_graph.cpp
@@ -1,0 +1,45 @@
+#include "copy_graph.hpp"
+
+namespace vg {
+namespace algorithms {
+
+    void copy_handle_graph(const HandleGraph* from, MutableHandleGraph* into) {
+        
+        // copy nodes
+        from->for_each_handle([&](const handle_t& handle) {
+            into->create_handle(from->get_sequence(handle), from->get_id(handle));
+        });
+        
+        // copy edges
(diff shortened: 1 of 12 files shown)


glennhickey:
The only thing I can think of is this:
https://github.com/vgteam/toil-vg/blob/master/src/toil_vg/vg_construct.py#L1447

It gets used by the gitlab tests so would need updating.  If you haven't
done it by then, I'll fix this tomorrow.

glennhickey:
There's also a mod -r in that function.

glennhickey:
@jeizenga The gitlab tests are fixed in @2500.  Clearing the travis cache also gets it to run through.  There's still an error in the last `vg stats` test.  As far as I can tell, it's because `vg paths -d` canonicalizes edges where `mod -D` didn't.  And somehow this affects the output of normalization `vg mod -U`.  I read that as a bug in normalization and not an issue for this PR.  

I see why you did it, but I'm not a huge fan of creating a new vg file and would prefer just dynamic casting to the mutable interface and going through that (and returning an  error if the cast failed on an xg).

PR #2510, opened 2019-10-16
Pack faster

glennhickey:
Using more threads in `vg pack` makes more `Packer` objects, which are used to process the GAM in parallel.  The problem is that creating the Packers at the beginning, and merging them together at the end is sequential.  So in practice after about 5 threads, adding more slows the wall time down.

This PR parallelizes packer construction and merging.  On my (whole-genome SV graph with 30X GAM) test, this lets it use up to about 16 threads before slowing down, and reduces the best wall time from about 80 minutes to under 30.  

Memory usage still scales up linearly with the number of threads, though.

### DIFF
--- a/src/packer.cpp
+++ b/src/packer.cpp
@@ -108,7 +108,7 @@ void Packer::merge_from_files(const vector<string>& file_names) {
             assert(n_bins == c.get_n_bins());
         }
         c.write_edits(tmpfstreams);
-        collect_coverage(c);
+        collect_coverage({&c});
     }
 }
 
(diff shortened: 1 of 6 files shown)


ekg:
We could implement a coarse grained lock system. Say for every 100bp of
sequence in the graph we have a lock. Then we don't waste memory keeping a
lock per base, but the access patterns "should" be sparse enough that we
don't have a lot of collisions (provided randomly sorted input though).

glennhickey:
@ekg Yeah.  I think this PR squeezes about as much speed out of the current setup as possible, and locking's the only way forward.  Especially for keeping memory under control.

I think parameterizing `batch_size` in `for_each_parallel` (hardcoded at 256 now), and having `pack` setting it really would make it fairly robust to the input being sorted or not.

I'll give this a try when I have a moment.  I want to have `augment` optionally use pack information (which means running pack more often), and was hoping this PR would get packing down to 10-15min for a whole genome, but it's not quite there yet.

PR #2516, opened 2019-10-24
WIP: Packer improvements

glennhickey:
* Allocate counter bins as-needed to better support sparse use cases. 
* Make Packer objects multithreaded by locking counter bins to avoid costly merges, and to be more easily re-usable in other code.

### DIFF
--- a/src/packer.cpp
+++ b/src/packer.cpp
@@ -8,10 +8,10 @@ namespace vg {
 const int Packer::maximum_quality = 60;
 const int Packer::lru_cache_size = 50;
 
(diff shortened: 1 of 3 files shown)


ekg:
Nice. Is each base bin getting a mutex lock? What's the scheme.

Sorry that I'm asking without looking at the code.

glennhickey:
I'm trying bins with one lock per bin.  I'm getting better results than the
old way on most tests.  Hoping the difference widens once I try more
cores.

Issue #2511, opened 2019-10-16
Tool/API to efficiently estimate read depth


glennhickey:
I'd like to estimate the sequencing depth from a given GAM file.  A notion of the expected coverage is sorely lacking from `vg call`, and it will be important for `vg augment` soon too.  

Average coverage is easily obtained from BAMs, and can be broken down by contig (`samtools idxstats, depth etc`).   But on a graph, it's a harder problem.  In fact, I don't think we have any way to get contig depth right now, beyond chunking up the graph and gam by chromosome.  And even computing the depth then isn't so straightforward.  One way would be to sum up the base coverages from `pack -d`, then divide by the total length of reference paths as extracted from `paths -E`.

So I think we need something like `vg depth <pack file> <path names>` (or add to `stats`?) that uses context expansion to sample a few subgraphs from each path and compute depth across those (as described above).  It could also accept an indexed GAM as input.  Would be nice to run on non-indexed GAM but I'm not sure that's doable (maybe it just makes a pack on the fly in this case).

ekg:
Let the vg monolith/ecosystem grow. This is a great idea!

Let's add `vg depth`, not add to stats.

PR #2519, opened 2019-10-28
WIP: Faster augment via coverage threshold

glennhickey:
`augment` works in three phases:

1. Find all the breakpoints implied by the GAM
2. Chop of the graph at each breakpoint
3. Add new stuff to the graph

Breakpoints due to read errors saturate the graph (and memory. and running time) for larger inputs.  This PR explores adding a coverage threshold for breakpoints: only those breakpoints with at least `M` support (from command line) will get passed to step 2.  

Read errors can still get into the graph if they occur on the same breakpoints as legitimate stuff.  Getting around this is tough.  Perhaps a re-imagining of Packer's edit index would be needed.  For now, I'm hoping that these are (relatively) rare enough to not blow up memory and/or cause trouble with downstream analysis.

Determining the coverage threshold from the data should be possible, but that may go into a later PR.

### DIFF
--- a/src/augment.cpp
+++ b/src/augment.cpp
@@ -3,6 +3,7 @@
 
 #include "augment.hpp"
 #include "alignment.hpp"
+#include "packer.hpp"
 
 //#define debug
 
@@ -18,15 +19,21 @@ void augment(MutablePathMutableHandleGraph* graph,
              bool embed_paths,
              bool break_at_ends,
              bool remove_softclips,
-             bool filter_out_of_graph_alignments) {
+             bool filter_out_of_graph_alignments,
+             Packer* packer,
+             size_t min_bp_coverage) {
 
-    function<void(function<void(Alignment&)>, bool)> iterate_gam =
-        [&gam_stream] (function<void(Alignment&)> aln_callback, bool reset_stream) {
+    function<void(function<void(Alignment&)>, bool, bool)> iterate_gam =
+        [&gam_stream] (function<void(Alignment&)> aln_callback, bool reset_stream, bool parallel) {
         if (reset_stream) {
             gam_stream.clear();
             gam_stream.seekg(0, ios_base::beg);
         }
-        vg::io::for_each(gam_stream, aln_callback);
+        if (parallel) {
+            vg::io::for_each_parallel(gam_stream, aln_callback, Packer::estimate_batch_size(get_thread_count()));
+        } else {
+            vg::io::for_each(gam_stream, aln_callback);
+        }
     };
 
     augment_impl(graph,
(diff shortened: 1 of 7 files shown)


glennhickey:
Whole genome augmentation now possible.  On a machine with 16 cores and 128G RAM, I tried a 30X GAM on a whole-genome SV graph.  The graph has no small indels or SNPs, so there's lots to augment. 
```
-rw-rw-r-- 1 ubuntu ubuntu 9.7G Oct 24 21:58 HGSVC.xg
-rw-rw-r-- 1 ubuntu ubuntu  75G Oct 24 21:15 hgsvc-map.gam
```
First need to convert to packed graph
```
vg convert HGSVC.xg -p > HGSVC.pg

-rw-rw-r-- 1 ubuntu ubuntu 3.2G Oct 29 19:11 HGSVC.pg
```

The default augmenter ran out of memory after 2.5 hours
```
vg augment HGSVC.pg hgsvc-map.gam -m 0
# Command terminated by signal 9
# Maximum resident set size (kbytes): 124251084
```

But thresholding the coverage let it run through (albeit slowly)
```
# Minimum coverage = 10 : 10h7m / 37G
vg augment HGSVC.pg hgsvc-map.gam -m 10 -A HGSVC_aug10.gam

# Minimum coverage = 5 : 10h40m / 55G
vg augment HGSVC.pg hgsvc-map.gam -m 5 -A HGSVC_aug5.gam

# Minimum coverage = 2 : 12h28m / 90G
vg augment HGSVC.pg hgsvc-map.gam -m 2 -A HGSVC_aug2.gam
```

The resulting sizes are
```
-rw-rw-r-- 1 ubuntu ubuntu  72G Oct 30 08:19 HGSVC_aug10.gam
-rw-rw-r-- 1 ubuntu ubuntu  72G Oct 30 18:57 HGSVC_aug5.gam
-rw-rw-r-- 1 ubuntu ubuntu  84G Oct 31 07:24 HGSVC_aug2.gam

-rw-rw-r-- 1 ubuntu ubuntu 3.3G Oct 30 08:21 HGSVC_aug10.pg
-rw-rw-r-- 1 ubuntu ubuntu 3.5G Oct 30 18:59 HGSVC_aug5.pg
-rw-rw-r-- 1 ubuntu ubuntu 8.3G Oct 31 07:28 HGSVC_aug2.pg
```

The min-coverage of 2 is probably allowing a lot of errors to slip through judging by this.  I've also implemented `-q and -Q` to filter by base and mapping quality, respectively.  They can be used in conjunction with the coverage threshold to further filter read errors, though I haven't benchmarked their effect whole-genome yet. 

Conclusion:  Augmentation is now do-able whole genome with a coverage threshold and 12 hours to spare.  The parameters are a bit fiddly, and I will take a stab at automatically estimating them next, as I work on #2511.  Better wall-time is probably only going to come via using separate processes for different chromosomes.  Will update toil-vg and the wiki with scripts for doing this shortly (need to split by chrom to do snarls anyway).  

@jonassibbesen I think this is ready to try out on your RNA seq data.  If most of your non-zero coverage will be much above 100 (I think you said it was in the 1000s?), make sure to set `-k` accordingly so enough bits get used (not doing so will still produce correct output but may use much more memory).  I'm hoping to get this set automatically in the future as well.

PR #2535, opened 2019-11-07
The VG ontology was missing position.

JervenBolleman:
The `vg:position` predicate has been in use for a while. 
At the same time a `rdfs:range` statement is fixed as well. ?step -> `vg:node` -> ?node .

### DIFF
--- a/ontology/vg.html
+++ b/ontology/vg.html
@@ -144,13 +144,13 @@
 </tr>
 <tr><td><a href="#Properties">
                                     Properties (
-                                    7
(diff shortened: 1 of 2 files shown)


adamnovak:
OK, I've pulled this in for full CI testing. If it passes we can merge.

JervenBolleman:
@adamnovak as it passes. Is it ok if I hit the rebase and merge button?

adamnovak:
Sure!

PR #2536, opened 2019-11-08
WIP: Probablistic calling and genotyping with vg call

glennhickey:
For the longest time `vg call` has used a thresholding approach to come up with genotypes.  This PR finally adds logic to use likelihood instead (keeping around the old way via option for now).

This resembles the Paragraph approach in that it uses a Poisson distribution to model the read coverage.   Some key differences are:
* `vg call` considers the support of the allele's entire path, rather than (independently) looking at breakpoitns
* `vg call` uses only the likelihood, with (for now) no option to specify prior
* `vg call` treats homozygous calls with support N as two alleles with support N/2 to be more comparable to het computation (and because it already needs  to partition support to handle overlapping traversals) 
* `vg call` includes the standard error of the read depth distribution when determining a coverage error rate (which is otherwise hardcoded).  Still work to be done tuning this.

Still a fair bit of testing to do before merging, but on smaller tests the accuracy is already better than the old way, with the difference more pronounced at lower coverages.  One exception being inversion genotypes which need to be looked into.

### DIFF
--- a/src/algorithms/coverage_depth.cpp
+++ b/src/algorithms/coverage_depth.cpp
@@ -90,7 +90,7 @@ pair<double, double> packed_depth_of_bin(const Packer& packer,
             }
         }
     }
-    return wellford_mean_var(bin_length, mean, M2, true);
+    return wellford_mean_var(bin_length, mean, M2);
 }
 
 vector<tuple<size_t, size_t, double, double>> binned_packed_depth(const Packer& packer, const string& path_name, size_t bin_size,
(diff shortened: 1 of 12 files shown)


ekg:
If you have likelihoods, we can pull in the genotyping model from
freebayes. This will let us specify a prior and would work with collections
of samples.

In any event, specifying the prior is not complicated. And you may want to
do it to get good performance.

I'll take a look at this.

glennhickey:
Yeah, thanks.  Any suggestions appreciated for the prior.  Am just running one sample at a time at the moment thouhg.

glennhickey:
The old version can be enabled with `vg call -B`, but the Poisson model is better in many cases (though often only slightly).  I still need to add the `GQ`s and some filters, but will do that within another PR.

PR #2544, opened 2019-11-22
Handlify vg chunk more and fold explode into it

glennhickey:
This adds a samtools-like `-O` option for output format to `vg chunk` (currently supported vg,pg,hg but can be and handle graph).  Would like to add this eventually to other graph-creating tools like `vg construct`.  

The goal for now is to quickly produce a set of chromosome packed graphs from an xg, then augment them in parallel, without dealing with the overhead of working with large vgs.  

Some of the chunk internals still rely on the VG class internally, but I'm not sure how necessary they are any more.  May be better to re-write or phase out than handligraphify.

### DIFF
--- a/src/algorithms/subgraph.cpp
+++ b/src/algorithms/subgraph.cpp
@@ -8,7 +8,7 @@ void expand_subgraph_by_steps(const HandleGraph& source, MutableHandleGraph& sub
     subgraph.for_each_handle([&](const handle_t& h) {
             curr_handles.push_back(h);
         });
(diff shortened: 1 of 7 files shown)


glennhickey:
I'm also deprecating `vg explode` here and folding it into `vg chunk`.  The latter can now split into components using `-C`, and these components can be subselected with some of the existing options. 

`vg chunk -M` will generate a separate component for every path in the graph.  It doesn't yet handle paths that share components (will cause duplicate graphs in output).  `-C` will be better in these cases.

Issue #2546, opened 2019-11-25
Issues with vg call on graph generated from cactus


RenzoTale88:
Hello,
I'm writing again to ask support on the vg call stage for some alignments. 
I've generated cactus alingments for 5 different mammalian genomes, and converted that to vg as described in issue #2514. After that, I've proceeded as follow for each sample analised:

```
sample=sample1
# Map the reads to the graph
vg map -R $sample -N $sample -S 0 -u 1 -t 4 -x all.xg -g all.gcsa my_lib1_R1.fastq.gz my_lib1_R2.fastq.gz

# combine different libraries into one gam file
while read p; do cat ${p} && rm ${p}; done < mappedlist.txt | vg gamsort --threads $NSLOTS -p -i ALIGN/${sample}/${sample}.gam.gai - > ALIGN/${sample}/${sample}.gam

# Filter the alignments
vg filter ALIGN/${sample}/${sample}.gam -r 0.90 -fu -m 1 -q 15 -D 999 -x all.xg > FILTER/$sample/$sample.filtered.gam

# Chunk the gam
vg chunk -x /exports/cmvm/eddie/eb/groups/prendergast_roslin/Andrea/GraphGenomes/NewGraph_03112019/GRAPH_CACTUS/all.xg -a FILTER/${sample}/${sample}.sorted.gam -P ./LISTS/PATHS.txt -c 50 -g -s 2500000 -o 100000 -b FILTER/${sample}/${sample}_call_chunk -t ${NSLOTS} -E FILTER/${sample}/${sample}.chunklist -f

# augment every chunk
while read p; do
reads=$( echo $p | awk '{print $4}' )
bname=$(basename -s ".gam" $reads)
vg augment ./FILTER/$sample/${bname}.vg $reads -p -C -t $NSLOTS -A ./FILTER/$sample/${bname}.aug.gam > ./FILTER/$sample/${bname}.aug.vg
vg index ./FILTER/$sample/${bname}.aug.vg -x ./FILTER/$sample/${bname}.aug.xg
vg snarls ./FILTER/$sample/${bname}.aug.xg > ./FILTER/$sample/${bname}.aug.snarls
vg pack -x ./FILTER/$sample/${bname}.aug.xg -g ./FILTER/$sample/${bname}.aug.gam -Q 15 -o ./FILTER/$sample/${bname}.aug.pack
done < FILTER/${sample}/${sample}.chunklist 

# Do variant call
while read p; do
[...]

PR #2548, opened 2019-11-26
Better handling of reverse-strand cycles in vg call

glennhickey:
Should fix the crash in #2546

More comprehensive testing on cactus graphs still needed.


ekg:
I have a collection of graphs in GFA form. It would be great to develop
some kind of simulation based variant calling test to make sure that we
support all the structures in them. I'll share the link by email.

PR #2552, opened 2019-11-27
Enable vg stats to mostly work on any PathHandleGraph

adamnovak:


adamnovak:
This should not break GBWTGraph stuff, but it does. I suspect we're missing code in the vg Makefile to clean out the GBWTGraph build artifacts when libhandlegraph changes, so it is not getting rebuilt for the new vtable.

PR #2555, opened 2019-12-03
Parallelize snarl computation

glennhickey:
Compute snarls of connected components independently and in parallel when more than one thread available.  Should be much faster.  Memory usage (big issue with snarls) will be lower but only if there are fewer threads than components (ie not operating on the whole graph at once)

Only realized now as new test failed sporadically that cactus isn't thread safe.  Can't merge until fixed, but I think it's just a matter of getting rid of the global variables in https://github.com/benedictpaten/pinchesAndCacti/blob/87b93048b4deb0c77bdf0a36c95dff70b557f81e/externalTools/threeEdgeConnected/impl/3_Absorb3edge2x.c

### DIFF
--- a/src/cactus.cpp
+++ b/src/cactus.cpp
@@ -203,7 +203,8 @@ void addArbitraryTelomerePair(vector<stCactusEdgeEnd*> ends, stList *telomeres) 
 // Step 2) Make a Cactus Graph. Returns the graph and a list of paired
 // cactusEdgeEnd telomeres, one after the other. Both members of the return
 // value must be destroyed.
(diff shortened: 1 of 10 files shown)


glennhickey:
I finally switched my old macbook from ports to brew, so I can reproduce these weird Travis bugs.  This one seems to be coming from iterating paths in the same graph as done here https://github.com/vgteam/vg/blob/packed-call/src/subgraph_overlay.cpp#L108-L119 at the same time in different threads.  Moving from vg to a different backing implementaiton seems to fix, as does adding a print statement.

glennhickey:
Goodness, I think it's the name/id maps in `Path.cpp` here: https://github.com/vgteam/vg/blob/master/src/path.cpp#L341-L355.

PR #2563, opened 2019-12-10
update gssw to get the hardening fixes + cmake

mr-c:


mr-c:
Hey @ekg , any advice here?

https://travis-ci.org/vgteam/vg/jobs/623108356#L3601 
> libc++abi.dylib: terminating with uncaught exception of type std::out_of_range: map::at:  key not found

Here is the gssw diff: https://github.com/vgteam/gssw/compare/79e0599d46133dc1c1c7dce451f29808c2f7e3a9...9f8efbca6dbfa3b7a4e6a65d0ef7cf5009e4aca1

glennhickey:
Yeah, this is almost certainly unrelated to your changes.  It's an error that came up in #2555, but should have been fixed within that same PR.  But I've hit it a few times since as well, and clearing the Travis cache has always resolved it.  I've done that here, and we'll see what happens.  

I'm beginning to suspect it's still just randomly failing and may not be cache related.  I'll try to reproduce locally.

mr-c:
@glennhickey Glad to hear it wasn't this PR :-)

glennhickey:
Can you please try rebasing this PR on master?  I think it should fix the Travis error.  Same thing for any other failing PRs.  Thanks!

PR #2565, opened 2019-12-10
vcflib: switch to upstream's repo for hardening and other goodies

mr-c:


mr-c:
Huh, same macOS only single test failure as https://github.com/vgteam/vg/pull/2563

https://travis-ci.org/vgteam/vg/jobs/623133084#L4202

PR #2567, opened 2019-12-10
update pinchesAndCacti submodule for hardening fixes

mr-c:


adamnovak:
@mr-c Current master has a fix for the Mac CI, which stopped working because of version changes of packages in Homebrew or something.

We can either merge all these PRs you made because the "pr" and Gitlab tests pass (i.e. the merged master will work), or you can pull master's fixes into all of them, and I can go back and pull them all into benaches in vgteam/vg again so the Gitlab tests will run again.

I think just merging them despite the failing "push" test is probably the right call. I can try that on Monday.

mr-c:
> I think just merging them despite the failing "push" test is probably the right call. I can try that on Monday.

Great! Find by me

PR #2573, opened 2019-12-12
correctly produce all kmers for each region provided they are in bounds

ekg:
Hope this resolves the issue found my Manuel.


ekg:
@adamnovak is Travis permabroke?

PR #2578, opened 2019-12-19
Switch to latest toil-vg for ci tests

glennhickey:
Now that `vg call` and `vg augment` scale better, I'm [cleaning up toil-vg call](https://github.com/vgteam/toil-vg/pull/781) to use them more efficiently by following the steps in [new wiki](https://github.com/vgteam/vg/wiki/Whole-genome-calling-and-genoyping).

 A part of this is dropping path-range chunking, and moving to either no, connected component or chromosome chunking.  And a consequence of this is dropping `vg genotype` support.  One day it may get folded into `call` or optimized to be run on a chromosome, and we can add it back.  In the meantime, not running it on chr21 should make the CI tests substantially faster and cheaper.


glennhickey:
@adamnovak I'm trying to remove 3 tests from gitlab
```
vgci/vgci.py::VGCITest::test_map_brca1_snp1kg_genotype
vgci/vgci.py::VGCITest::test_map_mhc_primary_genotype
vgci/vgci.py::VGCITest::test_map_mhc_snp1kg_genotype
```
But they seem to be coming up as errors in this PR

https://ucsc-ci.com/vgteam/vg/pipelines/2222/builds

Any idea what I'm missing, after removing them from `vgci.py` and `test_list.txt`?

adamnovak:
You also should lower the number of parallel Gitlab jobs launched in
.gitlab-ci.yml; there should be one per line in vgci/test-list.txt.

PR #2580, opened 2019-12-19
WIP: Fix construct

### LINKED ISSUE #2579: I think construct is dropping variants
On complex graphs, `vg call -v` sometimes complains that it can't find an alt path for a given VCF variant.  I've found a case where a 1-base deletion is dropped.  The issue (or at least my attempts to reproduce with `-R`) may be related to:

https://github.com/vgteam/vg/blob/master/src/constructor.cpp#L1753-L1755
```
while (variant_source.get() && variant_source.get()->sequenceName == vcf_contig &&
                variant_source.get()->zeroBasedPosition() >= leading_offset &&
                variant_source.get()->zeroBasedPosition() + variant_source.get()->ref.size() <= reference_end) {
```
This loop terminates when a previous variant (which overlaps the deletion) extends past reference_end and the so the variant I'm missing never gets visited.  Dropping the big overlapping variant makes it appear in the graph (using same construct command line)

This may be related to #2571 and #2279.

glennhickey:
Trying to fix #2579.  The code flagged in that issue can be changed (see below) to fix the case where the variant gets dropped when constructing with `-R` and a previous variant extends past the given range.  But it doesn't help not drop the variant when doing the whole contig, so there must be something separate going on (with chunking?).

### DIFF
--- a/src/constructor.cpp
+++ b/src/constructor.cpp
@@ -1751,12 +1751,19 @@ namespace vg {
         }
 
         while (variant_source.get() && variant_source.get()->sequenceName == vcf_contig &&
-                variant_source.get()->zeroBasedPosition() >= leading_offset &&
-                variant_source.get()->zeroBasedPosition() + variant_source.get()->ref.size() <= reference_end) {
+               variant_source.get()->zeroBasedPosition() >= leading_offset &&
+               variant_source.get()->zeroBasedPosition() <= reference_end) {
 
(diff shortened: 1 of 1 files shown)


glennhickey:
The missing deletion seems to get processed normally up until around here:
https://github.com/vgteam/vg/blob/master/src/constructor.cpp#L1268-L1277
Where its end point never gets picked up.

PR #2582, opened 2019-12-23
Fix construct to catch deletion end breakpoints that were previously missed

adamnovak:
I thought I had a good reason to use the exclusive bound query on the deletion endpoints, but changing it to the inclusive one seems to fix @glennhickey's missing deletion, which was caused by the reference never getting broken where the deletion was supposed to end in order to make the deletion arc.

I've changed the bound query and it seems to solve the problem. It might also result in extra/unnecessary reference breaks; we don't have a good test set to screen for those, but they should be mostly harmless.


glennhickey:
Thanks @adamnovak !  This was a long-time annoyance for SV graphs.

PR #2593, opened 2020-01-08
Load Protobuf-format files as HashGraph, add vg combine

adamnovak:
This changes the VPKG loading logic to load Protobuf-format files as HashGraphs instead of vg::VG objects.

Where this upsets code that demands a vg::VG object to work, logic has been added to convert the graph in memory.

This often results in serialized HashGraphs being produced as output where VG Protobuf used to be produced. Some attempts have been made to compensate for this; the tests have been tinkered with to not demand VG Protobuf so much, and a `vg combine` command has been added to do what `cat` on the Protobuf graphs used to do, with some pass-through logic to let you do things like combine a HashGraph and a Protobuf VG file containing just a Path.

I've also converted some logic over to algorithms where it was convenient.


adamnovak:
OK, `vg map` apparently only works tolerably on Cactus graphs if the xg index it is working from is actually in ID-sorted order.

I'm going to merge this now. If it causes problems, the easiest thing to roll it back would be to not revert the whole PR but to change the loaders back so that we load vg::VG objects from Protobufs again.

glennhickey:
@adamnovak toil-vg tests are failing with a `vg map` [crash](https://ucsc-ci.com/vgteam/toil-vg/-/jobs/34942) when I update vg's [docker image to the current master](https://github.com/vgteam/toil-vg/pull/785).  Think this PR could be behind it?

adamnovak:
Yeah, toil-vg needs some changes to be compatible with this. I had them on a branch that I pointed vg's CI to use. I've made a PR https://github.com/vgteam/toil-vg/pull/786 now, but I think that that PR needs to be updated to in turn reference the new vg in order to pass.

glennhickey:
OK, thanks.  I updated your PR to use my docker image.  We'll see if that's enough to get it in.

PR #2596, opened 2020-01-09
Don't try and use the PackedGraph tag for ODGI

adamnovak:
We were trying to use the PackedGraph tag for ODGI as well as PackedGraph files.

@glennhickey Might this be why ODGI conversion didn't work and got stubbed out? Or if I re-enable it will there be more problems?


glennhickey:
Oops, looks an old copy pasto on my part.  That was surely breaking
everything, so I see no harm in trying to re-activate.

PR #2602, opened 2020-01-13
Update to and synchronize latest libhandlegraph and related dependencies

adamnovak:
This replaces #2561 by bumping libbdsg to the latest version.

PR #2606, opened 2020-01-14
Make sure xg edge vectorization accepts all extant edges

adamnovak:
This uses and tests a slightly more robust XG, which won't fall over if asked to get an index for an edge_t that is backward.


adamnovak:
I don't think that this will work at all until https://github.com/vgteam/vg/pull/2602 is merged. It looks like the xg I am trying to pull in here wants the new SerializableHandleGraph setup.

ekg:
Yeah, I just updated that. It's good to sync everything.

PR #2617, opened 2020-01-21
Fix small bug in spliced surject

jeizenga:
I thought of an edge case that my old code didn't behave correctly on


jeizenga:
Travis is actually finished but just not reporting, so I'm gonna merge

PR #2651, opened 2020-02-20
A bit more handles

jeizenga:
Finished up some handlification I started yesterday


glennhickey:
https://github.com/vgteam/vg/blob/master/src/subcommand/explode_main.cpp#L73-L76

jeizenga:
Ah well, doesn't hurt, I guess.

jeizenga:
@adamnovak I think Travis is acting up

adamnovak:
It looks like Homebrew decided to remove the `brew bundle` and Brewfiles feature, so now when we try to do it it fails.

adamnovak:
Actually, it was removed like 6 years ago, but then factored out into its own module that brew just knows to install when you want to use it.

But it also isn't working on Travis today: see https://github.com/Homebrew/homebrew-bundle/issues/646

adamnovak:
I've made https://github.com/vgteam/vg/pull/2652 to try and fix this. If that works it can be merged into this branch.

adamnovak:
Upgrading Homebrew makes it upgrade Protobuf, which means we have to clear the caches because our build system can't detect the dependency on the Protobuf version.

adamnovak:
OK @jeizenga, rebase/merge master into this and clear the branch/PR caches and it should work.

PR #2655, opened 2020-02-25
Experimental: change MAPQ default settings in mpmap

jeizenga:
The RNA evaluation suggests that some features that once helped performance might now be harming it. This PR disables a couple of code paths, but I may reject it if the performance isn't good.


jeizenga:
This will probably fail the cactus graph test, but that's a test that I'm okay with failing. mpmap's default algorithm isn't really appropriate for that kind of graph, so I'm not going to worry about optimizing for it until I change the algorithm.

jeizenga:
This is that test I'm okay with failing, so I'm merging over it

PR #2658, opened 2020-02-28
Use latest toil-vg in Gitlab tests

glennhickey:
Bumps the tests from Python2.7 to Python3.6 in the process

### DIFF
--- a/vgci/vgci.py
+++ b/vgci/vgci.py
@@ -149,7 +149,7 @@ class VGCITest(TestCase):
 
             try:
                 connection = urllib.request.urlopen(url)
-                return str(connection.read())
+                return connection.read().decode('utf-8')
             except urllib.error.HTTPError as e:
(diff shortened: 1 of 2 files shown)


adamnovak:
@glennhickey I'm going to try turning on a Kubernetes runner on our Gitlab. It will probably cause the Gitlab tests here to fail unless you pull in the last PR I merged into master, which does some hosuekeeping that ought to let the tests pass on Kubernetes.

glennhickey:
OK, thanks for the heads up.  I've rebased.  This PR is pretty low priority so no worries if it's delayed due to gitlab work...

adamnovak:
This is still broken @glennhickey, but not for Kubernetes-related reasons now, I don't think.

You can go ahead and fix the remaining bugs, or I could give it a shot.

glennhickey:
Thanks @adamnovak I'll take a look today.

PR #2661, opened 2020-03-03
Remove MAPQ locate cap and extra MAPQ arguments

adamnovak:
This can do ~150 bp real reads at 1412.02 reads/second single-ended and 1060.09 reads/second paired.

I'm still collecting ROC plots, but in theory this should not really be able to be appreciably worse than having the locate cap.


adamnovak:
Before:
locate-and-extend

After:
extend-only

I think we can probably merge this; it doesn't seem to affect accuracy, and it removes code.

PR #2665, opened 2020-03-07
Refactor Indexing Part 1

adamnovak:
We want the mappers to be able to build their own indexes, so I'm refactoring the good parts of `vg index`'s `index_main()` into their own functions and files to support that.

So far, I have GBWT index construction pulled out, and a sort-of sketch of a class that will build just the indexes you need based on whatever files you have.


adamnovak:
OK, this now has I think all the changes to the actual indexing (and construction) code that I will need, and the beginnings of some code to actually build indexes on demand.

I'm going to merge this, and keep working on the on-demand indexing coordination logic. The only thing I'm worried about right now is slightly higher memory usage when building the GBWT, since it's not sensible in the new design for the indexing code to be able to delete the source graph when it is done with it anymore.

PR #2667, opened 2020-03-09
Speed up RNA mapping on real data

jeizenga:
A few small algorithmic changes in mpmap to avoid some computational pitfalls that sometimes occur on real data.


jeizenga:
Looks like there's some issue with the test reporting, but this PR has actually passed all of the tests.

PR #2671, opened 2020-03-11
Partial bug fix for surjection soft clips in splice-variation graphs

jeizenga:
The surject algorithm doesn't properly detect splice junctions in splice-variation graphs. It also had a bug that was interacting with this problem, leading to big soft-clips on transcript alignments. This PR fixes the bug, which will (approximately) fix the problem of soft-clipping. However, the algorithm can still have some micro-alignment issues around splice junctions, and the alignment scores will not be accurate.

@jonassibbesen I think this should work as a stop-gap for your pipeline, since the transcript surjections will be mostly accurate (although some splices will be identified as deletions instead of unaligned sequences, i.e. D rather than N in the CIGAR).

### DIFF
--- a/src/multipath_mapper.cpp
+++ b/src/multipath_mapper.cpp
@@ -2018,6 +2018,7 @@ namespace vg {
                         cerr << "found a duplicate" << endl;
 #endif
                         duplicate = true;
+                        // TODO: should i take the max of the multiplicities, or let it stay at 1?
                         break;
                     }
                 }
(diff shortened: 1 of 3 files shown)


jeizenga:
@adamnovak is this just bad traffic on the travis macs? it seems to get stuck downloading the cache

adamnovak:
Yeah I think it's just cloud weather. I restarted the offending job.

PR #2674, opened 2020-03-12
Use auto-indexing in Giraffe

adamnovak:
This lets Giraffe work with just a FASTA and VCF, building the other indexes dynamically and saving them in sensible locations.

If it can't load an index, it will print a nice error instead of crashing and telling the user to file a vg bug.

All the old switches should still work, so most scripts shouldn't break. But if you have stray `.gg` files laying around that you didn't intend to use, they may now get used, instead of the GBWTGraph being built in memory. And if the GBWTGraph isn't on disk, it may now be saved next to the XG.

### DIFF
--- a/doc/asciidoc/man/vg-gaffe.adoc
+++ b/doc/asciidoc/man/vg-gaffe.adoc
@@ -12,29 +12,40 @@ vg-gaffe - map unpaired short reads using minimizers and gapless extension
 
 == Synopsis
 
-*vg gaffe* [_OPTION_] > output.gam
+*vg gaffe* [_OPTION_]... [_FASTA_ [_VCF_]] > output.gam
+
+== Arguments
+
+_FASTA_::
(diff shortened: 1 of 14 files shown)


jeizenga:
How much heavy lifting would it take to get these to work on map or mpmap?

adamnovak:
We'd need to hook up the GCSA index as one of the indexes that this knows how to make, but that shouldn't be hard because multiple tools (`index` and `msga`) already know how to make it.

Then we'd need to go through the mappers and make them use this system instead of whatever they're using to hold their graphs/indexes now. Plus anything where the behavior depends on the presence or absence of an index would need to change to have its own flag.

Maybe about a day for each?

jeizenga:
GCSA indexing could be difficult since in practice we need to use `vg prune` on complex graphs, and because we might want to include the PhaseUnfolder somehow too.

glennhickey:
From a usability perspective, I think we gain a lot more by pruning within
vg index than we do by doing vg index -g/-x inside vg map.

jeizenga:
I think pruning would be pretty easy to accomplish via an edge-masking overlay, in which case it wouldn't be particularly memory-intensive to prune and index from the same graph in memory.

PR #2684, opened 2020-03-24
Tweak README

adamnovak:
Really I just want to see if we can pass the CI tests from a PR with no actual code changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ On other distros, you will need to perform the equivalent of:
                          
 Note that **Ubuntu 16.04** does not ship a sufficiently new Protobuf; vg requires **Protobuf 3** which will have to be manually installed.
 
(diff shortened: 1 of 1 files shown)


jeizenga:
Well, that's infuriating.

adamnovak:
Sorry. But as far as I can tell, you touched something. Are you sure that map doesn't eventually call anything you touched?

jeizenga:
I'm completely at a loss. The only thing in that PR are a handful of new functions in alignment.[ch]pp and multipath_alignment.[ch]pp, and then using those function in mpmap_main.cpp. I have no idea how that could affect the performance of map.

adamnovak:
Can you replicate the issue running one of the affected tests locally? Maybe you can paste in one function at a time until it breaks or something.

PR #2693, opened 2020-03-26
Update libbdsg so that HashGraph doesn't crash after using increment_node_ids

jeizenga:
Update to include https://github.com/vgteam/libbdsg/pull/53 and also to remove the hack that we used to get around the bug.


glennhickey:
Thanks!

PR #2700, opened 2020-03-31
Tighten handlegraph semantics

adamnovak:
This makes sure that deleting nodes with `destroy_handle` doesn't leave behind dangling paths. It also prohibits empty nodes in everything except HashGraph, where empty nodes are actively used in extraction algorithms that mpmap uses.

It also includes a bunch of int64_t->nid_t changes that somehow got lost in a branch I forgot about. They are needed for vg to work right with new libbdsg, because we changed the type of nid_t to make the libbdsg Python bindings able to be cross-platform.


adamnovak:
@jeizenga This is pretty much the opposite of what you wanted to do. What do you think?

adamnovak:
I merged this on my machine instead of through Github; Github thinks there's a conflict when a later commit advances a submodule further than what's already in master, but (my) Git can auto-merge such cases.

PR #2702, opened 2020-03-31
Turn on ODGI save/load and conversion

### LINKED ISSUE #2687: Converting to ODGI format is disabled
```
vg convert --odgi-out graph.vg | xxd -
error [vg convert]: odgi conversion does not work yet
```

We should make `vg convert` able to write ODGI, in case someone wants to use it with libbdsg. I think this is what the ODGI docs mean about vg's ODGI conversion being broken.

adamnovak:
This will fix #2687.


adamnovak:
I think https://github.com/vgteam/libbdsg/pull/54 broke things for ODGI. It makes the ODGI internal IDs start at 0, so handle rank can be -1, which turns into 0xfff.... and is off the end of the arrays.

PR #2705, opened 2020-04-02
Add secret menu of options to mpmap for a grid search

jeizenga:
I'm exposing a handful of options at the command line so we can do a proper grid search to come up with RNA defaults.


jeizenga:
Travis isn't reporting, but this has passed all the tests.

Issue #2683, opened 2020-03-23
VG call calling odds heterozygotes


RenzoTale88:
Hi,
I've recently started to align short reads to a arthropod-sized graph genome (~250 Mb) created through cactus, aligning two different assemblies.
I've aligned the short reads for the males, which are haploids, to the graph genome, and performed, alignment, filtering, chunking (```-c 50 -s 999999999 -o 100000```), augmenting of each chunk and calling.
I do the call using the command:
```
vg call ./CHUNK/$sample/${chunk}.aug.xg -r ./CHUNK/$sample/${chunk}.aug.snarls -s $sample -k ./CHUNK/$sample/${chunk}.aug.pack -t $NSLOTS -p $chroms -l $clength -o $bpi | vcf-sort | bgzip -c > ./VCALL/$sample/${chunk}.vcf.gz
```
Everything runs just fine, with no errors all along. The samples have a low depth of coverage (~10X), and are all haploid. In this situation, I would've expected the variants to be mostly homozygotes for ref or alt, with the heterozygotes to be mostly low depth sites. 
But when I look at the variants I can see that some of them are called heterozygotes despite having the same coverage and support for the different alleles of others called as homozygotes:
```
NW_0205  242     6740388_6740391 A       C       178.07  PASS    DP=12   GT:DP:AD:GL:GQ:GP:XD:MAD        0/1:12:4,8:-19.252327,-1.922362,-9.938795:80:-1.098612:13.482759:4
NW_0205  262     6740396_6740399 T       G       203.279 PASS    DP=11   GT:DP:AD:GL:GQ:GP:XD:MAD        0/1:11:2,9:-22.474669,-2.623996,-6.040050:34:-1.098996:13.482759:2
NW_0205  264     6740399_6740402 C       G       175.825 PASS    DP=11   GT:DP:AD:GL:GQ:GP:XD:MAD        0/1:11:3,8:-19.252327,-2.146874,-8.068569:59:-1.098613:13.482759:3
NW_0205  266     6740402_6740405 A       G       203.279 PASS    DP=11   GT:DP:AD:GL:GQ:GP:XD:MAD        0/1:11:2,9:-22.474669,-2.623996,-6.040050:34:-1.098996:13.482759:2
[...]

PR #2707, opened 2020-04-03
Support cutoff fow flow traversal enumeration

glennhickey:
Data from #2683 was crippling the latest version of the caller.  The problem seems to be big snarls with no support anywhere crippling the traversal enumeration.  Without any good paths for guidance, the running time in practice tends towards the worst case which is bad.  So when enumerating the K best traversals, it now stops when it hits a traversal with 0 support.  Also, uses an average support heuristic for larger snarls to make it more consistent with computations elsewhere.


glennhickey:
This has gotten too hacky.

PR #2710, opened 2020-04-06
Pangenome Ontology

6br:
Issue: https://github.com/graph-genome/graph-genome.github.io/issues/15

To extend https://github.com/vgteam/vg/tree/master/ontology with the new concepts of binning and sorting on odgi and Pantograph, I added ontology for pantograph format.


6br:
@JervenBolleman Is it possible for you to review the ttl description of `odgi.ttl` and `pantograph.ttl`?

6br:
* json_version is not needed. first_bin and last_bin is not needed, just inferred from the entire data.
* Components are a group of bins.
* Occupants/Participants are represented as whether the Entity is exist or not.
* Arrivals and Departures are encoded as Arrow because they are represented as arrow on Schematize. Arrow are a pair of bins for each path.
* Path names are just vg:paths
I have two questions:
* How should we handle the entire fasta? Should we store the entire sequence data?
* The bins of component should be represented as an array of bin. But how in RDF? (or Just start and end bin as First_bin and Last_bin?)

6br:
We are trying to have an example first, and then write down as an ontology.

```ttl
<pg/zoom10> a vg:ZoomLevel ;
   vg:components <pg/zoom10/group1>, <pg/zoom10/group2> ;
   vg:zoomBinWidth 10 .
<pg/zoom1000> a vg:ZoomLevel ;
   vg:components <pg/zoom1000/group1>, <pg/zoom1000/group2> ;
   vg:zoomBinWidth 1000 .
<pg/zoom1000/group1> a vg:Component ;
   vg:binEdge <pg/zoom1000/bin2> ;
   vg:bins <pg/zoom1000/group1/bin1>, <pg/zoom1000/group1/bin2> ;
   vg:steps <pg/zoom1000/group1/step1>, <pg/zoom1000/group1/step2> ;
   vg:departures <TBD>
   vg:arrivals  <TBD>
<pg/zoom1000/group1/path1> a vg:Path ;
   vg:positionPercent 0.04
   vg:inversionPercent 1
   vg:firstNucleotide <faldo:Position?>
   vg:lastNucleotide <faldo:Position?>
<pg/zoom1000/group1/bin1> a vg:Bin ;
   vg:binEdge <pg/zoom1000/bin2> ;
   vg:binRank 1 ;
   vg:meanCov 0.90 ;
   vg:binRegion <path1/region/6-100>,<path2/region/7-101> .
<pg/zoom1000/group2/bin2> a vg:Bin ;
   vg:binEdge <pg/zoom1000/bin3> ;
   vg:binRank 2 ;
   vg:meanCov 0.94 ;
   vg:binRegion <path1/region/101-1000>,<path2/region/103-1004> .
<path1/region/6-100> a faldo:Region ;
   faldo:begin <path1/position/6>  ;

**Comment by 6br** [comment]
Now I updated ontology schema for the latest example.

**Comment by subwaystation** [comment]
Hi @6br ,
this looks really good now!
I think @JervenBolleman already told me that, but 
```.ttl
<pg/zoom1000/group1/bin1> a vg:Bin ;
   vg:binEdge <pg/zoom1000/bin2> ;
   vg:binRank 1 ;
   vg:binRegion <path1/region/6-100>,<path1/region/200-234>,<path2/region/7-101> .
```
would be possible, too? So we will stay compatible when we have a list of ranges for each step in a bin.
For the sake of completeness let me add such an example in JSON here:
```JSON
{"json_version":10,
    "bin_width":100,
    "first_bin":0,
    "last_bin":26,
    "components": [
        {"first_bin":0,
            "last_bin":0,
            "occupants":[true,true], 
            "matrix": [[0.04,1,[[1,4], [5,10]]],[0.04,0,[[249,262], [345,456]]]], // [percent position in genome,percent inversion, [range_1, range2_ range_n]]
            "arrivals": // incoming links
                [],
            "departures": // outgoing links
                [{"upstream":0, // into which bin enters the link upstream?
                    "downstream":1, // into which bin enters the link downstream?
                    "participants":[true,false]}]}], // which genomes have that link?
    "path_names":
        ["1976","768"],
    "total_nr_files":12,
    "pangenome_length":52441,
    "filename":"chunk00_bin100.schematic.json",
    "fasta_filename":"seq_chunk00_bin100.fa"
} 
```
As discussed before, we can ignore `json_version`, `first_bin`, `last_bin`, `total_nr_files`, `file_name`, `fasta_filename`  and I think `path_names` which are indirectly encoded in a `vg:binRegion`. For browsing we would still need a complete list of available paths.

The trick question now is, how we can ship in the nucleotide sequence of the pangenome. I thin we would only need it once. Assuming we would encode it as one large blob, I would hope there is a SPARQL query that allows accessing only a subsequence of it. Comparable to an FASTA index.
Would that make sense?

6br:
> <pg/zoom1000/group1/bin1> a vg:Bin ;
   vg:binEdge <pg/zoom1000/bin2> ;
   vg:binRank 1 ;
   vg:binRegion <path1/region/6-100>,<path1/region/200-234>,<path2/region/7-101> .
would be possible, too? So we will stay compatible when we have a list of ranges for each step in a bin.

Sure, I will use this notation.

6br:
Here the major problem is how to represent nucleotide sequence on RDF. Apparently, it is undesirable to store the entire DNA sequence in higher zoom factor level (just redundant). We can’t imagine that every zoom layer has every nucleotide sequence.
Instead:
1. RDF stores nucleotide sequence of each node. Because component is a set of bins, and bin is a set of nodes, concatenating each sequence of bins can generate the entire sequence. But we anxious about the efficiency and performance of concatenating all nodes in SPARQL query.
2. Upload the entire sequence for public datastore and just point to the URL of public datasource.
3. Is it possible to store 2-bit compression of DNA sequence on RDF data?

6br:
The 1st seems the way forward, at least for Covid19 data. Which means “store each nucleotide sequence on node level, and use group_concat operation if needed.“

6br:
Is there any comments for the definition?

subwaystation:
Just one addition to the 1st way which @JervenBolleman mentioned:
`
When dealing with larger chromosomes, the implementation of the group_concat function in the database and even basic string types become an issue. For example when dealing with chromosomes of serious length (plants e.g. Paris japonica) the Java and Python string types can't represent the longer chromosomes. This can be worked around and fixed easily enough, but like spodgi a separate project in a way. 
`
So we are aware of this. Else, I never build an ontology. But if I could assist you somehow, please let me know.

ontology/README.md
```diff
@@ -52,3 +52,40 @@ me:example:some_gene rdfs:seeAlso ENSEMBL:ESG00000XXXX . #and then pick up the a
...
+   pg:components <pg/zoom1000/group1>, <pg/zoom1000/group2> ;
+   pg:zoomFactor 1000 . // zoomFactor is binWidth here.
+<pg/zoom1000/group1> a pg:Component ;
+   pg:componentRank 1 ;   # The order of component is inferred by rank.
+   pg:componentEdge <pg/zoom1000/group2> ;
+   pg:bins <pg/zoom1000/group1/bin1>, <pg/zoom1000/group1/bin2> ;
+   pg:steps <pg/zoom1000/group1/step1>, <pg/zoom1000/group1/step2> ;
+<pg/zoom1000/group1/step1> a pg:Step ;
+   pg:positionPercent 0.04
+   pg:inversionPercent 1
```
JervenBolleman, line 74: Is inversionPercent 1 -> 100% or 1% ? as positionPercent is 0.04 I assume it was supposed to be 100% but something to clarify.

ontology/vg.ttl
```diff
@@ -113,6 +114,60 @@
...
+  rdfs:range xsd:positiveInteger ;
+.
+:binRegion
+  rdf:type owl:ObjectProperty ;
+  rdfs:comment "A bin correspnds to the specific region of bins."^^xsd:string ;
+  rdfs:domain :Bin ;
+  rdfs:label "is"^^xsd:string ;
+  rdfs:range faldo:Region ;
+.
+:arrival
```
JervenBolleman, line 150: It looks like `:arrival` is the `owl:inverseOf` of `:departure`.

ontology/pg.ttl
```diff
@@ -0,0 +1,102 @@
+# baseURI: http://biohackathon.org/resource/pg
```
JervenBolleman, line 1: I would be happy to have all of this in the vg file. If we have a separate one we need to ask @ktym to set up the file copying to the biohackathon website. perhaps we can change the html generation to credit your work as well?

ontology/pg.ttl
```diff
@@ -0,0 +1,102 @@
...
+  rdfs:range xsd:positiveInteger ;
+.
+:bins
+  rdf:type owl:ObjectProperty ;
+  rdfs:comment "A Component has one or more Bins."^^xsd:string ;
+  rdfs:domain :Component ;
+  rdfs:label "has"^^xsd:string ;
+  rdfs:range vg:Bin ;
+.
+:steps
```
JervenBolleman, line 69: I am not sure that `:steps` and `:Step` are required. The `:componentRank` and `:binRank` seems to have the same function. In any case when reading vg rdf and pg rdf, having `pg:Step` and `vg:Step` makes debugging harder than necessary (can easily confuse).

ontology/README.md
subwaystation: Both inversionPercent and positionPercent are 1 -> 100%.
I just checked the code and example data.

6br: Sorry for setting an example value as 1, which causes ambiguity on whether the number is float or integer. We wish to regard the number as a float.

ontology/pg.ttl
6br: Sure, then I would merge all data format into `vg.ttl`.

ontology/vg.ttl
6br: That's true. I will add the line.

ontology/pg.ttl
6br: The reason why we need `pg:Step` is that whether a path appears on the component depends on path.
If there is a path on the component, positionPercent and inversion Percent have each path in each component. Therefore, we need to specify a path of the component.
Because it is confusing, I will change its name.

6br: > The :componentRank and :binRank seems to have the same function. 
Should I change the name of `:componentRank` or `:binRank`, or annotate they have the same function?

6br:
I updated the definition.

http://htmlpreview.github.io/?https://raw.githubusercontent.com/6br/vg/pantograph_turtle/ontology/vg.html

ontology/pg.ttl
JervenBolleman: I think it is good the way it is.

JervenBolleman:
I think the ontology changes are fine to go in. I can't see anything wrong with it so lets use it as.
However, do add your organizations grants to the bottom of the HTML output!
You deserve the credit.

6br:
Thank you for your review, Jerven! I will add organization information

ontology/README.md
```diff
@@ -52,3 +52,40 @@ me:example:some_gene rdfs:seeAlso ENSEMBL:ESG00000XXXX . #and then pick up the a
 
 At this moment VG RDF wants a fully embedded variation graph. e.g. all positions in vg json have a single edit which covers a whole node. This is to enable easy SPARQL queries where substring operations are rarely used.
 
+
+## Annotations on pantograph
```
subwaystation, line 56: @dimatr In case you started: This would be an example of a `.ttl`. Each entry can be dumped into any chunked file. The sequence of the entries does not matter.

ontology/vg.ttl
```diff
@@ -113,54 +114,184 @@
...
+  rdf:type owl:ObjectProperty ;
+  rdfs:comment "A start bin of nonlinear link. Outgoing edge of the bin."^^xsd:string ;
+  owl:inverseOf :arrival;
+  rdfs:domain :Link ;
+  rdfs:label "departure"^^xsd:string ;
+  rdfs:range :Bin ;
+.
+:paths
+  rdf:type owl:ObjectProperty ;
+  rdfs:comment "This means that the path participates the non-linear links between bins."^^xsd:string ;
```
subwaystation, line 161: This sentence is somehow confusing me. What is it you want to say?

6br: Link is between two bins, shown as "Arrow" in Schematize. Arrow is shown iff the two bins are not adjacent (I denoted as  non-lenear). Link is composed of multiple paths. So, Link has one or more paths.

6br: As you said, this sentence is unclear. Do you have any recommendations?

subwaystation: `A list of paths that follow the same non-linear link between two bins`.

Would that make sense?

6br: That's it! Thanks!

subwaystation:
@adamnovak or @jeizenga Would you be so kind and initiate the `ucsci-ci`?
Also, if you are happy with this PR, could you please merge it in?
Many thanks from @6br @JervenBolleman @subwaystation

adamnovak:
OK, I've pushed this up to a branch in vgteam for CI to run on it.

subwaystation:
Adding the current class diagram.
[ClassDiagram_Pangenome.pdf](https://github.com/vgteam/vg/files/4522635/ClassDiagram_Pangenome.pdf)
According to the diagram, a Bin has cells, one arrival, and one departure. But in the README.md I can see that a Link has an arrival and a departure which is a bin.
That's confusing. Also, I thought a bin from `odgi bin` can have several arrivals and departures? According to the CS output, we have a list of arrivals and departures for each component. See https://github.com/vgteam/vg/pull/2710#issuecomment-610227866.

Also I see these lines in `ontology/README.md`: 
<pg/zoom1000/component1/bin1/cell2> ;
<pg/zoom1000/component1/bin2/cell2> ; 
Where you did not add any content. These should represent empty cells?

6br:
> Also, I thought a bin from odgi bin can have several arrivals and departures? According to the CS output, we have a list of arrivals and departures for each component. See #2710 (comment).

For representing a list of arrivals and departures, I added forward/reverseLinkEdge and linkRank. The departures and arrivals on each bin can be inferred from component.

> Also I see these lines in ontology/README.md:
<pg/zoom1000/component1/bin1/cell2> ;
<pg/zoom1000/component1/bin2/cell2> ;
Where you did not add any content. These should represent empty cells?
 
I just omitted to write cell2 in the example. If you need, I can add also the instance of cell2.

6br:
> Adding the current class diagram.
ClassDiagram_Pangenome.pdf
According to the diagram, a Bin has cells, one arrival, and one departure. But in the ontology/README.md I can see that a Link has an arrival and a departure which is a Bin.

Actually, Bins and Links are multi-versus-multi relationship, but that is difficult to represent in class diagram...

subwaystation:
> > Also, I thought a bin from odgi bin can have several arrivals and departures? According to the CS output, we have a list of arrivals and departures for each component. See #2710 (comment).
> 
> For representing a list of arrivals and departures, I added forward/reverseLinkEdge and linkRank. The departures and arrivals on each bin can be inferred from component.
> 
> > Also I see these lines in ontology/README.md:
> > <pg/zoom1000/component1/bin1/cell2> ;
> > <pg/zoom1000/component1/bin2/cell2> ;
> > Where you did not add any content. These should represent empty cells?
> 
> I just omitted to write cell2 in the example. If you need, I can add also the instance of cell2.

But what if we would have an empty cell? 
Imagine we have a path2, that is not present in bin1, but in bin2. So cell2 would be empty in bin1, but present in bin2?

subwaystation:
> > Adding the current class diagram.
> > ClassDiagram_Pangenome.pdf
> > According to the diagram, a Bin has cells, one arrival, and one departure. But in the ontology/README.md I can see that a Link has an arrival and a departure which is a Bin.
> 
> Actually, Bins and Links are multi-versus-multi relationship, but that is difficult to represent in class diagram...

You are right, we can leave it as it is.

6br:
> Imagine we have a path2, that is not present in bin1, but in bin2. So cell2 would be empty in bin1, but present in bin2?



> > > Also, I thought a bin from odgi bin can have several arrivals and departures? According to the CS output, we have a list of arrivals and departures for each component. See #2710 (comment).
> > 
> > 
> > For representing a list of arrivals and departures, I added forward/reverseLinkEdge and linkRank. The departures and arrivals on each bin can be inferred from component.
> > > Also I see these lines in ontology/README.md:
> > > <pg/zoom1000/component1/bin1/cell2> ;
> > > <pg/zoom1000/component1/bin2/cell2> ;
> > > Where you did not add any content. These should represent empty cells?
> > 
> > 
> > I just omitted to write cell2 in the example. If you need, I can add also the instance of cell2.
> 
> But what if we would have an empty cell?
> Imagine we have a path2, that is not present in bin1, but in bin2. So cell2 would be empty in bin1, but present in bin2?

Yes, it becomes so.

Bin1 -- Bin1/Cell1     if Path1 is present
Bin1 -- ~Bin1/Cell2~     if Path2 is absent
Bin2 -- Bin2/Cell1     if Path1 is present
Bin2 -- Bin2/Cell2     if Path2 is present

6br:
And @JervenBolleman said SPARQL can definitely deal with non-existance of cells, so it would be no problem.

subwaystation:
@adamnovak Could you please start the CI again?
And if it does not complain, just merge it in, please.

Many thanks from @6br @JervenBolleman @subwaystation .

adamnovak:
OK, when this passes it should be ready to merge.

adamnovak:
One test failed because our CI server couldn't handle the artifacts being uploaded (we really need to provbision a bigger CI server or raise the gateway timeout or something to solve that).

But one test failed because some of the test data we used has moved.

I think we need to merge https://github.com/vgteam/vg/pull/2756 into this branch for it to pass.

6br:
Sorry, but still the CI fails. What should I do?

adamnovak:
I think it might be failing now because the server we want to get the data from can only sometimes see the file we want. I've restarted the offending job to check if that is the case. We should maybe retry the download if it fails.

6br:
Thank you!

PR #2712, opened 2020-04-06
Fix glaring bug in FlowCaller

glennhickey:
where it was assuming the snarl is oriented along the reference path, but not actually enforcing it.  How the tests passed is a mystery.


glennhickey:
These changes are unrelated to to the CI failure (mapping timeout), so merging through.

PR #2717, opened 2020-04-07
Boost timeout for simulated chr21 CI mapping test

glennhickey:
`test_sim_chr21_snp1kg_trained` is often taking longer than the 2400 seconds given to it.  I'm giving it another 2000 seconds here in the hopes that comparing the generated report with an older one will help see if there have been any major running time regressions.  `mpmap` is my primary suspect.


jeizenga:
I don't think I've made any substantial changes to default `mpmap` behavior in probably 2 weeks, so if this is recent I don't it's `mpmap`.

adamnovak:
It looks like this is working; the offending test finished in 2402 seconds.

Maybe something I touched in libbdsg, or that combined with something already in master, made it slightly slower than it used to be? Or maybe it really is Kubernetes load, although the load is really mostly at the pod-scheduling level, and most of the scheduled resources appear to still be unused.

PR #2720, opened 2020-04-07
Remove support for reading GFAs with overlaps

adamnovak:
This PR removes vg's ability to read GFAs with any overlaps or sequence-to-sequence alignments.

I'm ripping out the old GFA parser, which resolved overlaps by doing a transitive merge of all bases mapped to each other. At the [Computational Pan-Genomics Meeting](https://www.uni-bielefeld.de/(en)/ZiF/AG/2019/09-30-Stoye.html) in Bielefeld, we all decided that transitively merging bases wasn't the right approach, because it creates a blunt-ended graph that admits more paths and k-mers (often pathologically many more) than the original graph with overlaps.

Now, when you try and load a GFA graph with overlaps, it stops you and refers you to https://github.com/hnikaein/stark which is the only bluntification tool I know of that implements a vaguely correct algorithm.


adamnovak:
I'm going to cancel CI; I have another commit coming.

adamnovak:
@edawson This is what we want, right? We don't want to be creating super high degree nodes on accident.

edawson:
Nice - I'd referenced this here: https://www.biostars.org/p/428586/ .

I have a new GFA parser that I incorporated but the tests were failing. It looks like the Path lines are just getting ripped out of the tests here?

adamnovak:
I replaced some tests for the overlap functionality I removed with similar tests for input validation that don't include paths, which makes it look a bit like I just removed the paths, in the diff. We still test path import from GFA, I believe.

ekg:
You might also want to refer to seqwish/gimbricate. This implements a
correct process for overlaps that aren't exact. It does blow up the kmer
space of course.

adamnovak:
Can gimbricate bluntify? Or can we add some kind of simple bluntification to it? It looks like it corrects overlaps but has to leave them in.

ekg:
Yes, gimbricate followed by seqwish bluntifies any graph with any kind of
overlaps (including non exact marches).

Issue #2721, opened 2020-04-08
Cannot pull docker image for v1.23.0 


nh13:
```
$ docker pull quay.io/vgteam/vg:v1.23.0
Error response from daemon: manifest for quay.io/vgteam/vg:v1.23.0 not found: manifest unknown: manifest unknown
```

```
$ docker pull quay.io/vgteam/vg:v1.22.0
v1.22.0: Pulling from vgteam/vg
423ae2b273f4: Pull complete 
de83a2304fa1: Pull complete 
f9a83bce3af0: Pull complete 
b6b53be908de: Pull complete 
6138760a51a5: Pull complete 
d069e2a00b19: Pull complete 
7f797bdd79b7: Pull complete 
3011d8cbf968: Pull complete 
03d525cf39c9: Pull complete 
1212db4dbce2: Pull complete 
Digest: sha256:befe7d4d0cbb7e87c47614b9beb6a52742005fa506c5af458a538134348c8dbc
Status: Downloaded newer image for quay.io/vgteam/vg:v1.22.0
quay.io/vgteam/vg:v1.22.0
```

ekg:
FWIW, I think there is a Dockerfile in vg that you can run to build a current version. That might be more reliable. I don't know who's updating quay.io.

nh13:
The Release notes list the quay.io image, so someone must be doing it on here :)

Also, clone took forever so I didn’t want to wait, and then I saw the tarballs on the release page, but also the docker image.

ekg:
There are also static vg binaries on the release page. Those should be fine
if you just want to run vg.

adamnovak:
It looks like the v1.23.0 Docker build failed on Quay: https://quay.io/repository/vgteam/vg/build/e07e4293-3ffe-4a9e-bc72-21ab92cc987f

It looks like it ran out of memory on Quay's build server:

```
3/30/2020, 6:11:57 PM as: out of memory allocating 524411 bytes after a total of 0 bytes
3/30/2020, 6:11:58 PM cc1plus: out of memory allocating 249368 bytes after a total of 8151040 bytes
3/30/2020, 6:11:58 PM virtual memory exhausted: Cannot allocate memory
3/30/2020, 6:11:58 PM Makefile:48: recipe for target 'gbwt.o' failed 
```

We don't have any way to get feedback when the release builds fail unless someone goes to look at it. We should probably add "wait an hour and see if the Docker build worked" to the release checklist.

[...]

PR #2722, opened 2020-04-08
Cap the number of threads used in the Docker build

adamnovak:
We default to 8 threads to avoid trying to use all the cores on a Kubernetes host, but we should never use *more* threads than `nproc` reports, or we risk running out of memory on Quay like in #2721


adamnovak:
One oif the tests failed uploading the results to the coordinator.

I'm going to rerun it. Maybe we need to scale up our coordinator or increase a load balancer timeout or something?

adamnovak:
OK, the test artifacts are gone so the report can't rerun, but I reran the test that couldn't upload before and it passed. I'm going to go ahead and merge this.

PR #2730, opened 2020-04-10
Giraffe improvements and refactoring

jltsiren:
A number of improvements to Giraffe. **Distance indexes and minimizer indexes must be rebuilt.**

* Store (component id, chain offset) for seeds in top-level chains in the minimizer index. This information can be used to speed up seed clustering.
* Combine the information about seeds and clusters in structs instead of having a separate vector for each field. It's now easier to pass the information around and add/remove fields.


jltsiren:
PR #2743 includes this.

PR #2738, opened 2020-04-15
Simulate reads from a specific sample

jltsiren:
Simulate reads from a specific sample using `vg sim` with options `--sample-name` and `--gbwt-name`. By doing everything in a single command, we avoid unnecessary conversions, serialization, and loading. For 1000GP samples, this reduces the time overhead from 6 hours to 75 minutes.

There are also helper functions for inserting a GBWT thread into a graph and extracting it as a `Path` object.


ekg:
Thank you so much for integrating this.

It's been the biggest blocker to reproducing vg mapping results. It's of
course really important for testing.

PR #2740, opened 2020-04-15
Add node qualities to pack

glennhickey:
Doing some experiments here to see if I can add some MAPQ stats to `vg pack` and use them to improve `vg call`.  Currently trying to store an average MAPQ per node, which will hopefully compress pretty small and still be able to help the poisson mode downstream.


glennhickey:
I was going to make this an option, but it doesn't seem to affect performance so will leave it on by default:  On a GIAB chr15 augmented SV graph, running time goes up by 3%, memory by 2%, disk by 4%.

PR #2745, opened 2020-04-17
Add a no clustering option to mpmap

jeizenga:
Intended for very short reads (e.g. miRNA), in which clustering will incur a computational cost but is not likely to significantly improve mapping accuracy.

@jonassibbesen I'll add it to a preset with my grand interface redesign PR later, but if you want to mess around with it now it's on the mpmap secret menu as `--no-cluster`.


jonassibbesen:
Perfect, thanks @jeizenga.

jeizenga:
Seems have passed already, just GitHub being slow on the uptake. Merging.

PR #2753, opened 2020-04-24
Minor Giraffe optimizations

jltsiren:
Faster cluster scoring and subgraph transformations in pair rescue. Overall speedup is around 5%.


jeizenga:
Just an FYI, there's also the SplitStrandGraph overlay, which might be faster than split_strands. I've never properly profiled them to find out.

jltsiren:
There was barely any difference between the old approach and the SplitStrandGraph overlay. With the overlay, we avoid the expensive HashGraph construction. gssw_graph becomes more expensive, because graph navigation is slower in GBWTGraph than in HashGraph.

I think the best approach is to keep using the overlay and then reuse the topological order we get from `is_directed_acyclic()` / `dagify()` for gssw_graph construction.

jltsiren:
In the long term, it would be better to drop the SubGraph / SplitStrandGraph overlays, the `is_directed_acyclic()` / `dagify()` implementations, and the existing `align()` code and write GBWTGraph-aware versions of them. Accessing the edges of a node for the first time is expensive, but we can avoid the cost in subsequent accesses by using the CachedGBWT layer.

jltsiren:
Some new changes:

* `CachedGBWTGraph` overlay that uses the cached GBWTGraph interface. Much faster than the plain GBWTGraph interface when the algorithm repeatedly accesses the edges of the same subset of nodes. The cache is not thread-safe and it will grow to contain all nodes accessed through the overlay.
* A `CachedGBWT` bug fix that that improves the performance of the cache.
* Use the `CachedGBWTGraph` overlay for rescue in Giraffe.
* A version of `Aligner::align()` that aligns the read to a subgraph induced by a subset of nodes. In addition to the node set, this version also expects a topological order of the handles in the subgraph.

PR #2754, opened 2020-04-25
Further integration of dozeu into mpmap

jeizenga:
This is a not-totally-finished update for aligning high error reads in mpmap, but I'm going to do a preliminary PR because I realized that I broke some of the option parsing in mpmap with the last PR and I don't want to leave master in that state.


jeizenga:
Still has some bugs, but at a point that it's worth merging to fix the bugs in master

PR #2764, opened 2020-05-04
GAF output

ekg:
This creates an AlignmentEmitter which writes GAF.

It's not complete, or tested. It needs cigar strings.

### DIFF
--- a/src/subcommand/map_main.cpp
+++ b/src/subcommand/map_main.cpp
@@ -648,7 +648,7 @@ int main_map(int argc, char** argv) {
     unique_ptr<gcsa::LCPArray> lcp;
     unique_ptr<gbwt::GBWT> gbwt;
     // Used only for memory management:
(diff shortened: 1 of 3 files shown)


src/subcommand/map_main.cpp
```diff
@@ -83,6 +83,7 @@ void help_map(char** argv) {
          << "    -R, --read-group NAME         for --reads input, add this read group" << endl
          << "output:" << endl
          << "    -j, --output-json             output JSON rather than an alignment stream (helpful for debugging)" << endl
+         << "    -%, --gaf                     output alignments in GAF format" << endl
```
adamnovak, line 86: Are we sure we want a `-%` option? Couldn't we leave just the long `--gaf` option, or replace this and `-j` with something tat takes a string to define the target format? Maybe rename `--surject-to` and make surjection implicit for the formats that need it?

PR #2767, opened 2020-05-05
Use a single overlay in the IntegratedSnarlFinder

adamnovak:
This updates libbdsg and libhandlegraph so I can use just one overlay in the snarl finder, instead of one per MergedAdjacencyGraph.

@jeizenga If there's any way to make our main scratch graph type vectorizable, or at least to make it have ranked nodes, I could cut another pass over the graph off the snarl finder. And if the ranks were in a sensible traversal order for the graph, rather than jumping all around, I could make things a lot more cache efficient.

### DIFF
--- a/src/algorithms/three_edge_connected_components.cpp
+++ b/src/algorithms/three_edge_connected_components.cpp
@@ -13,7 +13,7 @@ extern "C" {
 #include <iostream>
 #include <sstream>
 
-#define debug
+//#define debug
 
 namespace vg {
 namespace algorithms {
--- a/src/algorithms/three_edge_connected_components.cpp
+++ b/src/algorithms/three_edge_connected_components.cpp
@@ -13,7 +13,7 @@ extern "C" {
 #include <iostream>
 #include <sstream>
 
-//#define debug
+#define debug
 
 namespace vg {
 namespace algorithms {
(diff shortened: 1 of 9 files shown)


jeizenga:
What do you mean by "scratch graph"?

adamnovak:
@jeizenga I mean our default graph type (and the one we load Protobuf files into when we see them). Right now that's HashGraph, but it doesn't even try to have a concept of node rank, so we're always making vectorizable overlays over it when one is needed.

jeizenga:
To me, node rank seems pretty antithetic to the whole idea of a hash table. If we assume the graph is static, then we could use iteration order over the hash table, but there's no easy way to access node IDs by rank or determine rank from node ID without auxiliary data structures. Moreover, this should yield a pretty chaotic ordering, assuming the hash function is doing its job. I think that anything that we could add to the graph internally would look a hell of a lot like the overlay.

adamnovak:
This PR now has a fix for a bug in my 3ECC implementation; it turns out we need to hide the self loops from the degree calculations, or we won't be able to tell 2-edge-connectedness from 3-edge-connectedness properly.

With these changes I get identical snarls (modulo orientation) with both algorithms on  https://vg-data.s3.amazonaws.com/bakeoff/snp1kg-MHC.vg.

PR #2776, opened 2020-05-08
Handle failures to seed from dozeu without crashing

jeizenga:
The seeding heuristic in dozeu is really quite rudimentary. It looks for an alignment of the first 15 bases of the read starting anywhere in the graph, or of any subsequence of the first 15 bases at the start of a source node. In some cases this fails to find any positive scoring alignment, which previously would cause a crash. I've modified the alignment algorithm so that it can detect when the seeding fails and fall back on GSSW.

@jltsiren I haven't implemented this code path in a particularly optimized way, but I think it's understandable enough if you want to work on it.


jltsiren:
This seems to be working. I didn't see the crash in the benchmarks this time.

PR #2778, opened 2020-05-08
mcmc works on x_small dataset

code-s-witch:
works on tests using 
FASTA = test/small/x.fa
TBI = test/small/x.vcf.gz.tbi
VCF = test/small/x.vcf.gz


jeizenga:
Looks to me like you accidentally scooped up some Visual Studios files at some point. Can you remove those before we merge?

jeizenga:
Oh, another thing: if you want to include your scripts in the VG repository, they should go in the scripts directory

code-s-witch:
Should be good now. These were inconsequential files but i think the command i could use in the future was 'git update-index --skip-worktree <file>'

adamnovak:
@code-s-witch It looks like your unit tests for MCMC print so much data that they cause Travis to declare our tests failed because the log is too long. Can you maybe turn off all the printing in the MCMC code?

Usually what we do is put `#ifdef debug`/`#endif` around every print, and only put in `#define debug` temporarily at the top of files we're currently working on.

PR #2785, opened 2020-05-12
Make IntegratedSnarlFinder the default again

adamnovak:
This reverts commit 9e54402cadfd31b5fd97f5650dcf80ba320fd169.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -753,7 +753,7 @@ $(UNITTEST_OBJ_DIR)/%.d: ;
 		echo 'syntax = "proto3";' > $${TESTDIR}/empty.proto; \
 		protoc $${TESTDIR}/empty.proto --proto_path=$${TESTDIR} --cpp_out=$${TESTDIR}; \
 		PROTOC_VER=$$(cat $${TESTDIR}/empty.pb.h | grep GOOGLE_PROTOBUF_VERSION | sed 's/[^0-9]*\([0-9]*\)[^0-9]*/\1/' | head -n1); \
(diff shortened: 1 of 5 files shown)


glennhickey:
@adamnovak can you please switch over `deconstruct`'s snarl finder while you're at it?

adamnovak:
This is now failing because of a Mac Travis problem that also affects master.

That problem is in turn due to Protobuf's `3.12.3` version, the current version, [meaning different things to different people](https://github.com/protocolbuffers/protobuf/issues/7632), and Homebrew [may or may not stop shipping it](https://github.com/Homebrew/homebrew-core/pull/58104).

I'm not quite sure what to do about that. Maybe always rebuild the Protobufs if we see 3.12.3?

adamnovak:
The real problem for our CI was this: https://github.com/Homebrew/homebrew-core/pull/58229

Homebrew *changed which flavor* of 3.12.3 they ship, but *kept the version number* at 3.12.3. So now we look, see out headers were generated by what appears to be the right Protobuf version, don't regenerate them, and fail.

adamnovak:
Now because we have to rebuild everything that depends on the Protobuf headers, we're hitting the timeout on the build. This approach may not work for Travis; maybe we just need to clear the caches instead.

adamnovak:
Yeah it turns out we can't clear out Protobuf headers *every time we run make* and still pass the Travis tests in time. I cleared the cache for this PR, but other busted PRs will need their caches cleared because I don't think we can usefully solve the problem via the build system.

PR #2786, opened 2020-05-12
WIP: Flesh out GAF support

glennhickey:
Adds some logic for reading GAF, as well as support for "CS" cigar strings. Goal is to get the GAF interface to the point where it can replace GAM.  
Still needs
* more testing
* multibase snp support in GAM conversion
* activate multithreaded i/o
* compressed input (and output?)
* benchmarking

as well as, perhaps in a future pr
* a review of command line options, support across tools
* possible reorganization into libvgio 
* ditching reliance on protobuf-based Alignment class
* supporting the stable coordinates GAF paths (they can be parsed now, but not read into or out of  Alignments)


jeizenga:
re: Potentially eliminating the Alignment object internally
I doubt it's worth it. I just tried out replacing the MultipathAlignment in mpmap with an equivalent structure implemented in STL data types, and I saw no difference in performance. I don't think it's a major driver.

glennhickey:
sold!

jeizenga:
Yeah, uhhhhh actually it turned out that I was looking at the baseline profiling data, which is why it looked like nothing had changed. I have a profiling run queued to actual profile it, but it's languishing the queue-bernetes.

Issue #2806, opened 2020-05-22
random_graph occasionally tries to break nodes wrongly


adamnovak:
When I added new randomized tests for the IntegratedSnarlFinder, I managed to start triggering a bug in random_graph where it can try to break nodes past their ends. This is causing intermittent test failures as in #2796 where the tests complain about an attempt to divide_node past the node's end and exit.

We need to make the failure reproducible (by gluing random_graph's seeding to the `--rng-seed` option form catch.hpp), and then we need to fix it.

adamnovak:
I just fixed this in #2810.

PR #2807, opened 2020-05-22
Seed random tests

adamnovak:
This is half of #2806 
With this in, I can reliably crash `random_graph()` with:

```
vg test --rng-seed 21 "NetGraph can traverse all the snarls in random graphs"
```


adamnovak:
[Quay has gone down again for the second time in a week](https://status.quay.io/), so the Gitlab report job can't finish (because it can't get `vg_ci_prebake`), but the tests themselves passed. Apparently you get what you pay for in container registries; maybe we need our own?

I'm going to go ahead and merge this. Then our CI will only be broken for one reason, instead of two.

PR #2815, opened 2020-05-26
Replace gaffe with giraffe

xchang1:
rename gaffe -> giraffe

### DIFF
--- a/scripts/giraffe-facts.py
+++ b/scripts/giraffe-facts.py
@@ -3,7 +3,7 @@
 
 
 """
-giraffe-facts.py: process a GAM file from the new minimizer-based mapper (vggiraffe) and report runtime statistics by filter.
+giraffe-facts.py: process a GAM file from the new minimizer-based mapper (vg giraffe) and report runtime statistics by filter.
 """
 
 import argparse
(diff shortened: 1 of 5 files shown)


adamnovak:
I've restarted Mac Travis for this. They both ran out of time during the tests. It could be just cloud weather, or it could be that something here introduces a hang in the Mac tests.

scripts/giraffe-facts.py
```diff
@@ -3,8 +3,7 @@
 
 
 """
-giraffe-facts.py: process a GAM file from the new minimizer-based mapper (vg
-gaffe AKA giraffe) and report runtime statistics by filter.
+giraffe-facts.py: process a GAM file from the new minimizer-based mapper (vggiraffe) and report runtime statistics by filter.
```
adamnovak, line 6: This looks to be missing a space.

glennhickey:
I'm seeing Mac Travis hang in another PR here:
https://github.com/vgteam/vg/pull/2813

adamnovak:
I think it was cloud weather. The rerun Mac test tool ~46 minutes instead of an hour and a half.

adamnovak:
@xchang1 Do you want to make a corresponding toil-vg fix to make it call `vg giraffe`? Or should we include a backward-compatibility alias?

xchang1:
I can change it in toil-vg

PR #2796, opened 2020-05-14
Get MCMC genotyper working on small test case 

code-s-witch:
This PR replaces #2778.

PR #2819, opened 2020-05-27
Merge early MCMC work

adamnovak:
This pulls in the first part of #2796, with the Mac test patches added. Hopefully it will pass all the tests and we can grab at least the old commits, before @code-s-witch's new line of development started.


adamnovak:
Mac Travis for the push timed out. I suspect cloud weather, but maybe we really do have an intermittent Mac hang bug.

adamnovak:
I think we might just be wasting lots of mac time rebuilding libvgio; I've merged in master which doesn't have that problem.

Issue #2733, opened 2020-04-11
bugs in Makefiles prevent making bin/vg


nadyawilliams:
**1. What were you trying to do?**

Compiling vg v 1.23 from source on CentOS 7.7. 
Environment: gcc/8.4.0, protobuf/3.11.4, cmake/3.12.3. All compiled from source
and used for other software packages builds. 

I use modules to load above prerequisite software before  I build vg.
Modules setup  CPPFLAGS, LDFLAGS, LD_LIBRARY_FLAGS, etc. 

There are 2 separate issues when building vg. 

1. When build target is libvcflib.a,  it fails claiming it can't find the -lvcflib library. 
Considering we are building it, yes its not there yet to be found.
Removing -lvcflib from deps/vcflib/Makefile solves this issue.

2. When building bin/vg it fails with numerous messages about  undefined 
references in google::protobuf:: ... [abi:cxx11]. This points to inconsistent  use of gcc ABI.
Since i am using the same compiler i used to build protobuf, i know that my environment is
correct.  The error steams from the main  Makefile where -lprotobuf is defined as a  required 
library. However, no -L flags are defined for this library, so it is apparently expected to be
in some default area.  As a result, during the linking  protobuf 3 libraries are not 
found but instead the protobuf 2 is used which is  also available on my system. 
I need to keep both protobuf 2 and 3. Reliance of Makefile  on protobuf 3 being installed in 
system default directories  is not a correct assumption. A better approach is to use pkg-config 
during the configuration step to collect needed info. 
Per my installation of protobuf: 
```
pkg-config --libs protobuf
-L/opt/apps/protobuf/3.11.4/lib -lprotobuf  
```
Changing -lprotobuf in Makefile to -L/opt/apps/protobuf/3.11.4/lib -lprotobuf  solves this issue.
To make this generic one needs to define in Makefile something like 
 -L($PROTOBUF_HOME)/lib -lprotobuf  and make sure the rest of the build knows  how to find
and use  protobuf related variables.

Regards

adamnovak:
If we use `pkg-config` to find Protobuf 3, how do we tell it that we want Protobuf 3 and not a simultaneously installed Protobuf 2? Does Protobuf 3 install its pkg-config info as `protobuf` while Protobuf 2 installs as `protobuf2` or something?

If we can get a constant and Protobuf-3-specific pkg-config name, I think we can just replace `-lprotobuf` with `$(shell pkg-config --libs protobuf)` in the Makefile.

[...]

PR #2822, opened 2020-05-28
Update to latest vcflib

adamnovak:
This updates to current vcflib, slightly patched to fix some missing return
statements.

This fixes half of #2733 by getting rid of the `-lvcflib` in vcflib, because
the new vcflib has a completely new CMake build system.

It can't be instructed to use vg's HTSlib (see
https://github.com/vcflib/vcflib/issues/286), but apparently it uses a similar
enough HTSlib that I can get it to build and run tests.


adamnovak:
It looks like builds are failing because of crosstalk between vg's HTSlib headers and vcflib's bundled HTSlib build.

I think the only real right thing to do is, if vcflib has to bundle htslib, then vg has to use that htslib instead of its own.

adamnovak:
I'm taking a different approach and trying to use vcflib's htslib as the One True HTSlib for the build. This PR can't pass until https://github.com/vcflib/vcflib/pull/289 is merged, which can't happen until it passes its tests, which in turn will require https://github.com/ekg/tabixpp/pull/22 being merged.

mr-c:
Hello @adamnovak ; can this be merged?

adamnovak:
Yes. I was holding off because I wanted to make a release for the Giraffe paper before I started messing about with removing our htslib, but now that's done so we can merge this.

mr-c:
Thanks!

PR #2838, opened 2020-06-05
Remove incorrect assert in vg sim

jeizenga:
Fixes an outdated assert statement in vg sim. Also pulls in the branch with the preset fix so that we can get an image built with both of the fixes we need for miRNA experiments. I will probably merge that branch separately though.


jeizenga:
I believe all of the useful bits of this PR already got merged as part of other PRs.

PR #2840, opened 2020-06-06
WIP: vg call option to output traversals to gentoype with rpvg

glennhickey:
From what I understand of it, [rpvg](https://github.com/jonassibbesen/rpvg) has a better genotyping model than `vg call`.  The drawback, currently at least, is that rpvg needs to know the haplotypes *a priori* and we don't have big databases to feed in for SVs. 

So the idea here is to try to run the rpvg model on a GBWT created from traversals that `vg call` comes up with from the data.  It would look something like

`gam -> pack -> call -> traversals -> gbwt + gam -> rpvg -> genotypes`

I think the `pack` in the second step could be sped up by downsampling to very low coverage (some read support is needed to find traversals, but not nearly as much as for genotyping).  The same pipeline out to apply to augmented graphs too.

### DIFF
--- a/src/graph_caller.cpp
+++ b/src/graph_caller.cpp
@@ -6,9 +6,8 @@
 namespace vg {
 
 GraphCaller::GraphCaller(SnarlCaller& snarl_caller,
-                         SnarlManager& snarl_manager,
-                         ostream& out_stream) :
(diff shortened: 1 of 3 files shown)


glennhickey:
Will do -- seems like a good idea and is easy enough to implement.  GAF calls would, of course, allow for calling snarls that are not spanned by a reference path.

PR #2856, opened 2020-06-19
New seeding algorithm targeted at miRNA

jeizenga:
This PR adds an algorithm that finds seeds that contain a small number of mismatches as specified sites (chosen based on base quality). It's very expensive on typical NGS reads, but I think it might be practical when applied to very short reads such as miRNA. mpmap now uses this algorithm to find seeds when the `very-short` preset is selected. 

@jonassibbesen I'm expecting this to be slower, but it would nice to know if it's slow to the point of being impractical. I could still turn some knobs to tweak the tradeoff.

### DIFF
--- a/src/mapper.cpp
+++ b/src/mapper.cpp
@@ -260,6 +260,9 @@ vector<MaximalExactMatch> BaseMapper::find_fanout_mems(string::const_iterator se
                                                        string::const_iterator qual_begin,
                                                        char max_fanout_base_quality) {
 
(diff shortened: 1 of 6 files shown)


jeizenga:
@adamnovak Have there been any changes in toil or toil-vg that might slow down the CI jobs? I'm consistently getting a timeout on a job that only uses `vg map`, and this PR contains exclusively modifications to `vg mpmap`.

PR #2862, opened 2020-06-25
this enables haplotype scoring in vg map based on graphs with embedded paths

ekg:
Previously, these were not counted towards the total haplotype count in the index, and so we were unable to correctly calculate the score.

This updates the index construction code in vg to count the embedded paths in the metadata, which resolves this problem.

Review from someone with more experience with the GBWT and haplotype scoring would be helpful.

### DIFF
--- a/src/haplotype_indexer.cpp
+++ b/src/haplotype_indexer.cpp
@@ -357,6 +357,8 @@ tuple<vector<string>, size_t, vector<string>> HaplotypeIndexer::generate_threads
         size_t path_rank = 0;
         graph->for_each_path_handle([&](path_handle_t path_handle) {
                 ++path_rank;
+                haplotype_count++;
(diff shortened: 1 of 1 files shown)


jltsiren:
The existing logic is based on the assumption that we are adding reference paths as threads (either separately or in addition to the threads generated from a VCF file). Path names are used as contig names, and we assume that they all form a single sample ("ref") with a single haplotype.

If you want to interpret each path as a separate sample over the same contig, you should add that as a new option. The GAM/GAF section that adds each alignment as a separate sample over contig 0 is more or less what you want to do, but you should also avoid adding path names as contig names.

And now that I look at the code, I see a large number of bugs and inconsistencies. Perhaps I should fix it soon.

jltsiren:
PR #2884 added option `--paths-as-samples` that does what this PR was supposed to do. I believe this can now be closed.

PR #2865, opened 2020-06-25
switch to python3

mr-c:
Adapted from https://salsa.debian.org/med-team/vg/-/blob/a5a7bde0b1c3c0a02ed6559cb26989e0ad40df01/debian/patches/python3

### DIFF
--- a/scripts/filter-noisy-assembler-warnings.py
+++ b/scripts/filter-noisy-assembler-warnings.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python3
 # 
 # filter-noisy-assembler-warnings.py
 # Author: Stuart Berg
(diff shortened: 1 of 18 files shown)


mr-c:
This passes both Travis CI and GitLab CI, but one of the Travis CI runs is not reporting back to GitHub..

scripts/filter-noisy-assembler-warnings.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python3
```
adamnovak, line 1: Shouldn't these all be `/usr/bin/env python3`?

adamnovak:
I'll merge this up, we can change the interpreter path if someone complains it isn't working right in a virtualenv.

scripts/filter-noisy-assembler-warnings.py
mr-c: Yes, sorry. This patch mixed two objectives: py3 + absolute paths (for Debian). I should have kept the 2nd part out

PR #2867, opened 2020-06-25
[WIP] Enable building on non-x86 systems

### LINKED ISSUE #2639: Improving portability
I'm looking into the possibility of getting vg to run on machines without SSE4.2, and eventually without x86 (POWER, ARM, etc.).  After looking into the code a bit I think it can be done without too much effort, but I wanted to see if you would be receptive to patches before I start working on them since I don't want to maintain a fork.

Just to give a bit of context, I'm the primary author of [SIMDe](https://github.com/nemequ/simde/), and I think vg would make an interesting test case.

My plan would be to first get the various modules to compile without any particular architectural requirement, largely by using SIMDe for modules with hard dependencies on SSE.  That would give you the *option* of compiling for more basic CPUs, but if you keep using `-msse4.2` performance should remain unchanged.

After that the next step would be to build multiple versions of some code and automatically choose the fastest version at runtime.  I'd like to keep that optional so people who don't need portable builds can just do `-march=native` (or whatever) and get the fastest version for their machine without any runtime overhead.

The first step would require cooperation from upstream modules like Dozeu, but I think I can keep the changes pretty minimal so hopefully it won't be a problem.

Is that something you would be interested in merging, assuming the code quality is acceptable?

mr-c:
- [x] vcflib ~~https://github.com/vcflib/vcflib/pull/292~~ ~~https://github.com/vgteam/vg/pull/2822~~ ~~https://github.com/vgteam/vg/pull/3075~~
- [x] gssw ~~https://github.com/vgteam/gssw/pull/27~~ ~~https://github.com/vgteam/vg/pull/2906~~
- [x] ssw ~~https://github.com/vgteam/Complete-Striped-Smith-Waterman-Library/pull/4/~~ https://github.com/vgteam/vg/pull/3023
- [x] structures ~~https://github.com/vgteam/structures/pull/11~~ ~~https://github.com/vgteam/vg/pull/2907~~
- [x] dozeu ~~https://github.com/vgteam/dozeu/pull/5~~ (~~[upstream](https://github.com/ocxtal/dozeu/pull/2)~~)
- [x] libbdsg ~~https://github.com/vgteam/libbdsg/pull/72~~ ~~https://github.com/vgteam/vg/pull/2909~~
- [ ] fermi-lite vg dep: https://github.com/edawson/fermi-lite/pull/1 upstream: https://github.com/lh3/fermi-lite/pull/14

Fixes #2639

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -54,7 +54,7 @@
 	url = https://github.com/benedictpaten/sonLib.git
 [submodule "deps/fermi-lite"]
 	path = deps/fermi-lite
-    url = https://github.com/edawson/fermi-lite.git
+	url = https://github.com/mr-c/fermi-lite.git
 [submodule "deps/gbwt"]
 	path = deps/gbwt
 	url = https://github.com/jltsiren/gbwt.git
@@ -116,6 +116,9 @@
 [submodule "vgteam_bbhash"]
 	path = deps/BBHash
 	url = https://github.com/vgteam/BBHash.git
+[submodule "src/simde"]
+	path = src/simde
+	url = https://github.com/nemequ/simde-no-tests
 [submodule "doc/wiki"]
 	path = doc/wiki
 	url = https://github.com/vgteam/vg.wiki.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@
 
 language: cpp
 compiler: gcc
-sudo: required
 dist: bionic
 # We have some shenanigans to let us cache submodules, and update changed files
 # without messing up mtimes and triggering rebuilds unnecessarily. Travis checks
@@ -112,8 +111,6 @@ os:
   - linux
   - osx
 osx_image: xcode12.2
-compiler:
-  - gcc
 env:
   global:
     - DOCS_KEY_ENCRYPTION_LABEL=125272388526
@@ -124,3 +121,5 @@ matrix:
     # We have a special entry to do the docs build
     - os: linux
       env: BUILD_DOCS_ONLY=1
+    - os: linux
+      arch: arm64
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,10 @@ before_install:
   - if [[ -e deps ]]; then mv deps deps_cached; fi
   - git submodule update --init --recursive
   - rsync -rv --links --checksum deps/ deps_cached/
-  - rm -Rf deps
+  # - rm -Rf deps
   # Keep the cached deps if the right compiler version was used.
   # Otherwise start fresh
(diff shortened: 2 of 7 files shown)


adamnovak:
I don't think we'll be able to merge this as is even when it passes tests; instead of pointing everything to your forks of submodules, we'll want to merge your changes into our forks.

With Apple moving Macs to ARM now, building on non-x86 is more important. But we also rely fairly heavily on libraries like dozeu and GSSW that are written around compiler intrinsics for particular new-ish-x86 vector instructions that probably don't have exact ARM equivalents, or at least need some #define magic to select them.

We'll have to add slower, plain-C versions, or even ARM-vector-instruction versions, or just fall back to completely different core alignment code on ARM that isn't really vectorized but still lets it run.

@jeizenga You've been diving through the dozeu code; how would you rate making its vector operations swappable between SSE and [NEON](https://en.wikipedia.org/wiki/ARM_architecture#Advanced_SIMD_(Neon)) intrinsics, versus a C fallback implementation (or fallback C implementations of the intrinsics)? Does it use a lot of unique intrinsics or just a few?

adamnovak:
We also don't *always* use byte order conversion when serializing/deserializing some of our file formats. It looks like arm64 as used in Linux is generally also little-endian, but before we let any big-endian ports out of the lab we need to do some serious interoperability testing to catch all the conversions we're missing.

mr-c:
Hello @adamnovak .

The point of this PR is to use the 1https://github.com/simd-everywhere/simde header only SIMD library to gain compatibility for using the X86 SIMD intrinsics on non-X86 systems.

As you mentioned, there are only a few big endian architectures (for Debian they are s390x, sparc64, ppc64, and powerpc) and they can be safe ignored for now. If IBM or an IBM customer wants vg on a big iron system they they can pay for or contribute s390x big endian support themselves 😁

I agree that this shouldn't be merged until the git submodules are back to using the regular sources, not my forks. I'll add a checklist and link to the other pending PRs.

mr-c:
Okay, the x86 build works and the tests complete. On arm64 the build completes, but some of the tests fail.

adamnovak:
@mr-c Is there a way we can bring arm64 under CI? It's definitely going to break if we don't.

I think we might just be able to add Linux/GCC/arm64 to our build matrix for Travis: https://docs.travis-ci.com/user/multi-cpu-architectures/#testing-on-multiple-cpu-architectures

mr-c:
Already part of this PR :-) https://github.com/vgteam/vg/pull/2867/files#diff-354f30a63fb0907d4ad57269548329e3R124

PR #2879, opened 2020-07-02
Handlify mod -X

glennhickey:
## Changelog Entry

 * Chopping with `mod -X` changed to run natively on handle graphs
 * `vg stats -F` prints graph format 

## Description

Handlify `mod -X` to chop graphs without converting to VG.

Also add format printer to `stats -F`.  It can't distinguish between `VG` and `HashGraph` though.  @adamnovak is there an easy way to test this?

### DIFF
--- a/src/algorithms/unchop.cpp
+++ b/src/algorithms/unchop.cpp
@@ -473,6 +473,26 @@ void unchop(handlegraph::MutablePathDeletableHandleGraph* graph) {
     }
 }
 
+void chop(handlegraph::MutableHandleGraph* graph, size_t max_node_length) {
+    // borrowed from https://github.com/vgteam/odgi/blob/master/src/subcommand/chop_main.cpp
+    std::vector<handle_t> to_chop;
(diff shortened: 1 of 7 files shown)


adamnovak:
We load vg files as HashGraph; you should never get a vg::VG from VPKG load functions when you ask for a handle graph generically.

To test it I guess we would `vg convert` to each format and make sure it returns the format we know we gave it, right? Or is that not what you meant?

glennhickey:
That was bad phrasing on my part.  Was wondering what would be the best way
to distinguish a proto file from a hashgraph file.   vpkg load seems to be
out.  I guess it'd have to try to read directly from a VG object, then call
it a vg graph if it doens't crash?

adamnovak:
You could look at it with a MessageIterator and see if it has a VG format tag. If so it's vg, if not it must be something else.

glennhickey:
Thanks @adamnovak, MessageIterator seems to have done the trick.

PR #2904, opened 2020-07-11
Cap or

adamnovak:
## Changelog Entry

 * Allow summing different routes for minimizer disruption when capping MAPQ

## Description

This adds a `--sum-mapq-cap` option to Giraffe to use probability OR in the MAPQ cap DP, instead of just finding the highest-probability case.


adamnovak:
I don't think we're going to take this route anymore. We have a much more empirically effective cap version in the faster-cap branch that we're working on.

PR #2908, opened 2020-07-14
Use a simde that builds on Clang 11

adamnovak:
## Changelog Entry

 * SIMD alternatives compatible with Clang 11

## Description

I just merged https://github.com/vgteam/dozeu/pull/8 and https://github.com/vgteam/gssw/pull/28 so this will build on CI. With this we should benefit from the fix for https://github.com/simd-everywhere/simde/issues/470 and we will be compatible with Clang 11 once again.

This should let @jonassibbesen and @jeizenga build vg on Mac again.


adamnovak:
We might want to consider bumping out Mac Travis to Clang 11 (as well?) as described in https://github.com/simd-everywhere/simde/issues/470#issuecomment-654704440

PR #2916, opened 2020-07-21
Better GBWT construction from GAM/GAF

jltsiren:
## Changelog Entry

* GBWT construction from GAM/GAF is faster and handles duplicate read names correctly.

## Description

GBWT construction did not like it when the number of distinct start nodes of paths/threads was too large. This PR updates GBWT to fix the issue. GBWT can now be built in a reasonable time from GAM/GAF files with tens (hundreds?) of millions of reads.

In many cases, a read and its pair have the same name. When building a GBWT from such reads, we used to create duplicate sample names in the GBWT metadata. This PR changes the behavior. All alignments with the same sample name now correspond to the same sample/haplotype in the GBWT metadata. We use the `count` field in GBWT path names to distinguish between them.

### DIFF
--- a/src/haplotype_indexer.cpp
+++ b/src/haplotype_indexer.cpp
@@ -3,6 +3,7 @@
  */
 
 #include <iostream>
+#include <map>
 #include <vector>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


ekg:
Will this also improve support when these paths are embedded in the graph?

jltsiren:
@ekg I only changed some internals of GBWT construction and added handling of duplicate read names to the GAM/GAF to GBWT code.

PR #2918, opened 2020-07-23
Faster cap branch improvements

adamnovak:
This uses Benedict's faster approach to computing his MAPQ cap, and fixes issues where total probabilities could hit 1 and had to be hackily capped.


adamnovak:
Note that this wants to merge into the faster-cap branch.

adamnovak:
This doesn't seem to have the right Protobuf version on Mac Travis, but that's not its fault, so I'm merging into the branch.

PR #2923, opened 2020-07-25
Fix DP

adamnovak:
## Changelog Entry

 * Actual optimal alignments from `vg giraffe`

## Description

This is my attempt to fix the "DP bugs" Benedict thinks we have. He's right; we do have DP bugs.

I've fixed the one where we aren't passing a sufficient gap length limit to Dozeu. But now I've found a particular left-pinned alignment problem that `align_pinned` with xdrop on can't seem to handle. It doesn't even involve a particularly long gap.

Reference: AATGAGAAAGGAAAAAGCTTTGGGAAAGTAGCTAAGCAGAGGATGCTTCTTAAAAAATGTCAAAAAAA (all in one node)
Query: CACATATGTCACTAGGAATGAGAAAGGAAAAAGCTTTGGGAAAGTAGCTAAGCAGAGGATGCTTCTTAAAAAATGTC

The optimal alignment is:

```
----------------AATGAGAAAGGAAAAAGCTTTGGGAAAGTAGCTAAGCAGAGGATGCTTCTTAAAAAATGTCAAAAAAA
CACATATGTCACTAGGAATGAGAAAGGAAAAAGCTTTGGGAAAGTAGCTAAGCAGAGGATGCTTCTTAAAAAATGTC
```

We can afford the leading gap, and we don't exhaust all the available query sequence. But the aligner wants to produce something like:

```
{"identity": 0.35064935064935066, "path": {"mapping": [{"edit": [{"sequence": "CACATATGTCACTAGG", "to_length": 16}, {"from_length": 27, "to_length": 27}, {"sequence": "GTAGCTAAGCAGAGGATGCTTCTTAAAAAATGTC", "to_length": 34}], "position": {"node_id": "1"}, "rank": "1"}]}, "score": 6, "sequence": "CACATATGTCACTAGGAATGAGAAAGGAAAAAGCTTTGGGAAAGTAGCTAAGCAGAGGATGCTTCTTAAAAAATGTC"}
```

This is just garbage and I have no idea why it's doing this. I've made a test `"XdropAligner can align pinned left across an insertion with extra graph at the end"` that illustrates the problem.

@jeizenga You've been digging around in the depths of Dozeu, right? Do you want to take a crack at fixing it?

### DIFF
--- a/src/algorithms/alignment_path_offsets.cpp
+++ b/src/algorithms/alignment_path_offsets.cpp
@@ -1,5 +1,6 @@
 #include "alignment_path_offsets.hpp"
 
+//#define debug_mpaln_offsets
 
 namespace vg {
 namespace algorithms {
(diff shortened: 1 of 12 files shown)


adamnovak:
OK, this now has Jordan's fix for an internal xdrop problem, and my fix to make sure that, when the gapless extender returns partial-length extensions alongside a "full-length" extension with too many mismatches, we actually turn at least one of those into a candidate alignment.

adamnovak:
@xchang1 is going to see how this affects our Giraffe ROC curves before we merge it.

PR #2927, opened 2020-07-27
Use libvgio that only finds Protobuf once

adamnovak:
## Changelog Entry

 * libvgio now only tries to find Protobuf once

## Description

This should fix @jeizenga's problem where libvgio tries to use two different Protobuf versions in its build.


adamnovak:
This doesn't fix Jordan's problem, but I think we want it anyway.

PR #2955, opened 2020-08-16
Add vg view arg -u for more compact dot

jmonlong:
## Changelog Entry

 * "no sequence" mode using `vg view -du` where the node sequence is replaced by its size, e.g. to visualize large graphs with dot

## Description

It is sometimes difficult to visualize large regions using `vg view -d`. The image quickly becomes very wide. For example:

ex view

To address that I had been using the simplified view (`vg view -Sd`) after unchopping nodes with `vg mod`. It's more compact but we lose the bi-directed representation which complicates the interpretation when the reverse complements of nodes are involved. For example, for the same region, it looks like there is a cycle around node 2439865:

ex view S

The new argument to `vg view`, called `-u`, uses the bi-directed representation but shows the size of a node instead of the full sequence.  I had started tweaking the existing `-S` option but felt bad for destroying a simplified view that some people might like. For example, this view helps us understand that the "cycle" around node 2439865 is actually just a path through the reverse complement:

ex view u

### DIFF
--- a/src/subcommand/view_main.cpp
+++ b/src/subcommand/view_main.cpp
@@ -63,6 +63,7 @@ void help_view(char** argv) {
 
          << "    -d, --dot                  output dot format" << endl
          << "    -S, --simple-dot           simplify the dot output; remove node labels, simplify alignments" << endl
+         << "    -u, --noseq-dot            shows size information instead of sequence in the dot output" << endl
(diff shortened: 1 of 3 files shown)


jmonlong:
I just realized I can also use `vg mod -O` to re-orient the nodes and avoid most of the confusion with the small reverse nodes I had in my graphs. Ah well, this bi-directed view might still be useful to look at inversions or in complex regions where the graph contains nodes that can't be re-oriented easily.

ex view uO

PR #2974, opened 2020-09-11
Changed fast mode hard hit cap from 1000 to 500

xchang1:
## Changelog Entry

Fix bugs in giraffe and gamcompare

## Description
Changed giraffe's fast mode hard hit cap to 500 to match the default


xchang1:
@jeizenga I think this should fix the warnings you saw. Is it easy for you to check?

Issue #2981, opened 2020-09-14
VG convert fails on trivial GFA with no header


rlorigro:
**1. What were you trying to do?**
Convert a GFA to a dot file using 
```
vg convert -g test_alignment_graph_0.gfa -p | vg view -d - | dot -Tsvg -o graph.svg
```

**2. What did you want to happen?**
the dot file be made good

**3. What actually happened?**
crash:
```
ERROR: Signal 6 occurred. VG has crashed. Run 'vg bugs --new' to report a bug.
Stack trace path: /tmp/vg_crash_3uAMp9/stacktrace.txt
Please include the stack trace file in your bug report!
error[VPKG::load_one]: Correct input type not found while loading handlegraph::PathHandleGraph
```

**4. If you got a line like `Stack trace path: /somewhere/on/your/computer/stacktrace.txt`, please copy-paste the contents of that file here:**

```
Crash report for vg v1.24.0 "Montieri"
Stack trace (most recent call last):
#17   Object "/home/ryan/software/vg/vg", at 0x4ddbb9, in _start
#16   Object "/home/ryan/software/vg/vg", at 0x1c41438, in __libc_start_main
#15   Object "/home/ryan/software/vg/vg", at 0x40a627, in main
#14   Object "/home/ryan/software/vg/vg", at 0xa33cf7, in vg::subcommand::Subcommand::operator()(int, char**) const
#13   Object "/home/ryan/software/vg/vg", at 0x9ed9be, in main_convert(int, char**)
#12   Object "/home/ryan/software/vg/vg", at 0xf08062, in vg::algorithms::gfa_to_path_handle_graph(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, handlegraph::MutablePathMutableHandleGraph*, bool, bool)
#11   Object "/home/ryan/software/vg/vg", at 0xf07d27, in vg::algorithms::gfa_to_handle_graph_on_disk(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, handlegraph::MutableHandleGraph*, bool, gfak::GFAKluge&)
#10   Object "/home/ryan/software/vg/vg", at 0xf17520, in gfak::GFAKluge::for_each_edge_line_in_file(char const*, std::function<void (gfak::edge_elem const&)>)
[...]

PR #2983, opened 2020-09-14
Update gfakluge to one that doesn't crash with no header

jeizenga:
## Changelog Entry

 * No longer crashes when converting GFAs that lack a header line

## Description

Resolves https://github.com/vgteam/vg/issues/2981


glennhickey:
I'm seeing the same gitlab error in my PRs where `vgci/vgci.py::VGCITest::test_map_mhc_snp1kg` is apparently timing out while augmenting.  Not sure if we have some kind of speed regression in master or if kubernetes is just going a bit slower these days.

PR #2991, opened 2020-09-17
Handle inability to learn paired end distribution in reasonable time

adamnovak:
If our buffer gets too full, we finalize the distribution as is.

If that results in a distribution that's mean 0 and ends up too short for the
reads, we will instead fall back on single-ended mapping and print warnings.

## Changelog Entry

 * Giraffe now falls back to single-ended mapping instead of quitting when it can't learn a fragment length distribution


adamnovak:
I rebased this because I think I need Glenn's test timeout fixes.

PR #3000, opened 2020-09-24
Handle graph gfa

xchang1:
## Changelog Entry

 * Make vg view convert a handle graph directly into gfa

## Description
vg view used to take a handle graph, convert it to vg, then convert vg to gfa. It can now go directly from a handle graph to gfa


jeizenga:
Kinda silly that we hadn't done this already lol

PR #3010, opened 2020-09-25
Account for score_mismatch scoring negative

### LINKED ISSUE #3001: Giraffe paired-end regression
Giraffe performance on paired-end reads regressed between 43e408c (Sept 17) and 1ee0c8f1 (Sept 20). Single end performance is unchanged

On the 1kg graph, full gbwt, novaseq6000 reads on 43e408c, giraffe gets 1968202 reads correct single end and 1983815 correct paired end
On the same graph and reads on 1ee0c8f1, giraffe gets 1968202 reads correct single end and 1961336 correct paired end.

This plot was run with 1ee0c8f1 with a 1kg graph, novaseq6000 reads, and sampled gbwts with different coverage
plot-roc

adamnovak:
## Changelog Entry

 * Mismatches are now recognized as bad byu alignment rescoring

## Description

This should fix #3001, I think.

### DIFF
--- a/src/aligner.cpp
+++ b/src/aligner.cpp
@@ -807,7 +807,7 @@ int32_t GSSWAligner::score_discontiguous_alignment(const Alignment& aln, const f
                 score += score_exact_match(aln, read_offset, edit.to_length());
                 last_was_deletion = false;
             } else if (edit_is_sub(edit)) {
-                score -= score_mismatch(aln.sequence().begin() + read_offset,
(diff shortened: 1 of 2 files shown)


adamnovak:
OK, I've tested this and it solves the regression. I get `1983815 reads correct`.

Still not as good as 6871da4ec8494530db1ec4f8c76d0f0171b83821 which manages to hit `1984114 reads correct` somehow, but it's not regressed relative to our baseline.

PR #3025, opened 2020-10-05
Emit raw HandleGraphs that libbdsg can read.

adamnovak:
## Changelog Entry

 * Now vg outputs raw handle graph files readable by libbdsg without the need for `vg view -x/--extract-tag`.

## Description

This changes vg to output HashGraph, PackedGraph, and ODGI graphs without any VPKG framing, so that libbdsg can read them. It also switches over to a libvgio that can read them from standard input (by buffering standard input enough that we are guaranteed to be able to unget enough characters to sniff the format safely).

### DIFF
--- a/src/io/converted_hash_graph.hpp
+++ b/src/io/converted_hash_graph.hpp
@@ -1,0 +1,23 @@
+#ifndef VG_IO_CONVERTED_HASH_GRAPH_HPP_INCLUDED
+#define VG_IO_CONVERTED_HASH_GRAPH_HPP_INCLUDED
+
+#include <bdsg/hash_graph.hpp>
+
+namespace vg {
+
+namespace io {
+
+/**
+ * Define a type that inherits HashGraph so we can tell the difference between
+ * a real HashGraph and a HashGraph converted at load time. We care about this
+ * so that vg stats can tell you the original input file format.
+ */
+class ConvertedHashGraph : public bdsg::HashGraph {
+    using bdsg::HashGraph::HashGraph;
+};
+
+}
+
+}
+
+#endif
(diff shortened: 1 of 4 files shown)


glennhickey:
Can we use this buffered input to transparently stream GAF into handlegraphs?

adamnovak:
I don't see how we would. It still only buffers about 2k of data, which is enough for all practical file type sniffing purposes, but not really an amount that will fundamentally change the GFA to HandleGraph problem.

glennhickey:
By "GFA to HandleGraph problem" you mean removing the overlaps?  I'm not referring to that.  I'd like any tool that reads a HandleGraph now to be able to read a **blunt** GFA as well, loading it internally into whatever the default implementation is.  It seems this would be a doable with the buffer, as I'd think 2k would be enough to detect if the input file is GFA or not.   

Hooking this into `vg` would then let all vg tools operate directly on GFA, which I think would be a win in terms of usability.

adamnovak:
Oh, if we had a HandleGraph loader registered that could read a GFA
file, we could definitely leverage this to sniff that the file is GFA
when it's coming from a stream.

On 10/12/20, Glenn Hickey <notifications@github.com> wrote:
> By "GFA to HandleGraph problem" you mean removing the overlaps?  I'm not
> referring to that.  I'd like any tool that reads a HandleGraph now to be
> able to read a **blunt** GFA as well, loading it internally into whatever
> the default implementation is.  It seems this would be a doable with the
> buffer, as I'd think 2k would be enough to detect if the input file is GFA
> or not.
>
> Hooking this into `vg` would then let all vg tools operate directly on GFA,
> which I think would be a win in terms of usability.
>
> --
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub:
> https://github.com/vgteam/vg/pull/3025#issuecomment-707397222


-- 
---
Adam Novak (He/Him)
Senior Software Engineer
Computational Genomics Lab
UC Santa Cruz Genomics Institute
"Revealing life’s code."

Personal Feedback: https://forms.gle/UXZhZc123knF65Dw5

ekg:
This is a great idea. We should work directly on GFAs whenever possible.

Some of the loaders like that for xg and odgi require memory mapping of the
GFA though.

PR #3041, opened 2020-10-13
Unslowify surject

jeizenga:
## Changelog Entry

* `vg surject` is now substantially faster

## Description

I redid the way surject performed `path_handle_t` to minimize interaction with the backing BWT/FM-index, which it turns out is responsible for a very large proportion of surject's run time. I also added an algorithm to filter out redundant partial paths from the GAMP surject code path, which seem to have been driving some of its higher run time.

### DIFF
--- a/src/subcommand/map_main.cpp
+++ b/src/subcommand/map_main.cpp
@@ -731,6 +731,11 @@ int main_map(int argc, char** argv) {
         });
     }
     
+    unordered_set<path_handle_t> paths;
+    for (const auto& path_name : path_names) {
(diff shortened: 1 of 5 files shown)


adamnovak:
Did you actually remove the name-taking version? I did almost the same thing in #3042 but I deprecated them. We have to combine these somehow. Maybe you should merge first.

jeizenga:
@adamnovak If you're volunteering to be the one who has to resolve git conflicts I will take you up on that unhesitatingly. Have fun!

PR #3042, opened 2020-10-13
Bring HTS output formats to Giraffe

### LINKED ISSUE #3029: Add --surject-to to Giraffe
Map can surject as it runs. Giraffe should be able to do this too so it can just write BAM.

Piping Giraffe to surject has been shown to be quite slow by @cmarkello. THis would work around that.

adamnovak:
## Changelog Entry

 * `vg giraffe` can now output to SAM/BAM/CRAM directly

## Description

When a non-GBWTGraph graph is available, `vg giraffe` can surject when you tell it e.g. `--output-format BAM`.

This will fix #3029 (although not with the `--surject-to` option specifically).

Giraffe and map surjection are mostly unified now, using a `SurjectingAlignmentEmitter` wrapper and a new alignment emitter factory function that sets up surjection. Some of the groundwork for sequence dict order support has been laid, and the comments lie about it being there already. I want to finish pulling that through at least on the backend before we merge this.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -94,11 +94,22 @@ bam_hdr_t* hts_file_header(string& filename, string& header) {
 }
 
 bam_hdr_t* hts_string_header(string& header,
-                             map<string, int64_t>& path_length,
-                             map<string, string>& rg_sample) {
+                             const map<string, int64_t>& path_length,
+                             const map<string, string>& rg_sample) {
+    
+    // Copy the map into a vecotr in its own order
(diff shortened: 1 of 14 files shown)


adamnovak:
OK, I've plumbed through enough of the ordering support that the doc comments are now actually true. We just need some way to read a sequence dictionary file in map/giraffe/surject and use that set and order of paths to also fix that issue.

adamnovak:
See: #3043

PR #3070, opened 2020-10-29
XG temp dir support

adamnovak:
## Changelog Entry

 * `xg` indexing should now use the temporary directory fro scratch instead of the current directory


ekg:
Does this default to the current working directory? We had a series of
problems with this due to people running out of space in tmp.

jltsiren:
Before this PR, XG used the temp directory for its own temporary files and the current working directory for the SDSL temporary files created during CSA construction. The latter directory could not be changed. The SDSL temp files are small unless the number of paths is very large. However, if we are building an XG with alt paths, the files can take tens of gigabytes.

After this PR, everything goes to the directory specified in `xg::temp_dir`.

ekg:
Great thanks for the clarification!

PR #3075, opened 2020-10-31
update to latest vcflib

mr-c:
## Changelog Entry

 * update to latest vcflib to get non-x86 support, "Configure for default, libcurl-using HTSLib config"


adamnovak (approved):
This looks like it should work too, and I've pulled it in for testing also.

Issue #3080, opened 2020-11-03
Giraffe/surjector can produce BAM with multiple insertions/deletions in a row


adamnovak:
Andrew Carroll says that, when running GATK4.1.2.0 MarkDuplicates, he encountered a "No M or N operator between pair of D operators in CIGAR" error.

While in GAM we can need multiple deletions in a row on different nodes to delete all of them, in BAM they need to be merged.

We think we can still use these BAMs by telling the GATK validator to chill, but they shouldn't be like this.

adamnovak:
This *shouldn't* happen. CIGAR generation for HTSAlignmentEmitter goes through `append_cigar_operation`, which coalesces adjacent operations of the same type:
https://github.com/vgteam/vg/blob/c4ad64e0a16d6095242545f9cafcc486f661350b/src/alignment.cpp#L818

I suppose we could actually be using the spliced alignment emitter somehow, which has its own non-coalescing append logic:
https://github.com/vgteam/vg/blob/c4ad64e0a16d6095242545f9cafcc486f661350b/src/hts_alignment_emitter.cpp#L732

I could add (time-taking) scans for adjacent identical operations when serializing, or code to fix them up, I suppose.

It would be nice if I could reproduce the problem.

jeizenga:
The spliced version actually does coalesce as well:
https://github.com/vgteam/vg/blob/c4ad64e0a16d6095242545f9cafcc486f661350b/src/hts_alignment_emitter.cpp#L726-L733

adamnovak:
OK, looking at Andrew's dump of offending reads, it seems that the problem isn't duplicate `D` operations, but runs of e.g. `7D8I7D`, where two `D` operations are separated by an `I` operation.

I think the right way to approach this might be to settle on a canonical order (e.g. `D` always comes before `I`), and to coalesce runs of D and I operations as a unit. So when we want to add e.g. `1D`, we look not only at the top operation to see if it is a `D`, but also see if it is an `I`, and if so either insert or add to a `D` one operation below it.

jeizenga:
Sure, that sounds like a pretty easy post-processing step for the CIGAR code.

adamnovak:
I already did #3080 in #3089 by changing the way we add in new CIGAR operations. Do we want that *and* a postprocessing step?

PR #3094, opened 2020-11-12
Fix a few recent crashes in surject

jeizenga:
## Description

Some of the changes I made in https://github.com/vgteam/vg/pull/3087 weren't fully baked, leading to some malformed multipath alignments. In addition, recent changes exposed some of the limitations in the previous surject algorithm at detecting splicing events. This PR includes a more general splice detection algorithm based on bicliques instead of stars.

I also tacked on a solution to the interleaved D and I CIGAR problem reported in https://github.com/vgteam/vg/issues/3080 for multipath surject.


adamnovak:
I already did #3080 in #3089 by changing the way we add in new CIGAR operations. Do we want that *and* a postprocessing step?

PR #3104, opened 2020-11-20
Resolves an indexing error in mpmap

jeizenga:
## Description

I was not adjusting an index to account for the fact that the vector had resized, leading to out-of-bounds accesses.


jeizenga:
@jonassibbesen and I are impatient to get this merged into master so we can peg the preprint against it, so I'm going to merge. If it breaks, blame him.

PR #3105, opened 2020-11-20
Bump macOS version used for Travis CI

adamnovak:
## Changelog Entry

 * CI now tests macOS 10.15
 * vg now uses (and needs!) system Boost for Vowpal Wabbit instead of bundling a subset

## Description

@jltsiren mentioned our macOS for Travis was too old for Homebrew; maybe this will help?


adamnovak:
This looks like it isn't building on Mac due to picking up a conflicting Boost program options library from the environment.

Maybe we can stop shipping it ourselves and just require it to be available? It's got to be in Homebrew and apt.

adamnovak:
We seem to have acquired (or already had?) an intermittent failure in the vg augment tests:

```
ok 20 - augmenting 2 snps with -m 1 produces the same nodes as default
2d1
< C
5,7c4
< T
< TCCAACTCTCTG
< TCTATTATA
---
> TCTATTATATTCCAACTCTCTG
not ok 21 - augmenting 2 snps and 2 errors with -m 11 produces the same nodes as with just the snps
#          got: '1'
#     expected: '0'
```

I can't see how this is related to Mac, Boost, or libvw.

jeizenga:
I merged a couple naughty PRs while the Mac build was failing, but they were pretty narrowly confined to mpmap stuff. I don't imagine they would have messed up augment.

adamnovak:
I ran the augment tests over 100 times and couldn't get them to fail on my Mac.

jeizenga:
I'm so glad I didn't volunteer to fix this 😁

adamnovak:
I think I didn't necessarily *have* to rip out our vendored Boost, but if we can do it I think we should.

adamnovak:
Everything has passed except one of the docs build jobs which is still in the queue. I'm going to go ahead and merge this so other stuff stops failing tests.

PR #3127, opened 2020-12-14
Make Giraffe see the correct cores when working with syncmers for minimizers

adamnovak:
## Changelog Entry

 * Giraffe now maps and caps MAPQs properly with syncmers

## Description

This brings bounded syncmer index (`vg minimizer -b`) support to Giraffe.

I've told it to think of the entire k-mer of the syncmer as a single-window agglomeration, so it's all core with no flank. I think that showing that to the existing mapping quality capping code will Just Work; there won't be any flanking bases at which errors could occur, so we'll bypass that whole case.


adamnovak:
Using syncmers seems to be about half as fast as minimizers at the moment, but it does appear to produce decent results. The ROCs are broadly comparable:
syncmer-roc

PR #3135, opened 2020-12-28
[WIP] Use overlays created at runtime to remove node length restrictions from the user's perspective

jeizenga:
## Changelog Entry

 * VG no longer requires a maximum node length of 1024 bp for mapping

## Description

This isn't done, but I'm putting it up here to get feedback. 

One point of confusion we keep running into is that VG requires a maximum node length of 1024 during mapping because our indexes have bit-packed the position information into 10 bits. There's also a "softer" requirement for short nodes in that we frequently extract subgraphs that we expect to be small by copying full nodes. If nodes are large, performance can suffer. This was something we talked about when plotting the [VG Roadmap](https://github.com/vgteam/vg/wiki/Roadmap), although I don't think it ended up being its own bullet point.

My proposed solution is to have a lightweight overlay to do the node chopping on the fly whenever graphs have nodes that are too long. We can automatically add this on top of the graph before indexing, and then again before mapping. That will ensure that the mapping machinery only ever sees nodes that are below the maximum length, but we can translate mapping back into the original graph coordinates before returning them.

So far, this PR includes the actual overlay, but none of the other plumbing. My current implementation adds `O(1)` to all queries except for `get_handle`, which is `O(log N)` (although I suspect with a very small constant). There's another possible implementation that has `O(1)` `get_handle` queries, but `O(log (node length))` for many node-centric queries like `follow_edges` and `get_sequence`. We could also maintain both implementations in the same data structure to get `O(1)` for all queries.

### DIFF
--- a/src/chopped_graph.hpp
+++ b/src/chopped_graph.hpp
@@ -1,0 +1,850 @@
+/** \file
+ * chopped_graph.hpp: defines a handle graph overlay that splits nodes
+ * into multiple nodes to satisfy a maximum length
+ */
+
+#ifndef VG_CHOPPED_GRAPH_HPP_INCLUDED
+#define VG_CHOPPED_GRAPH_HPP_INCLUDED
+
+#include "handle.hpp"
+#include "utility.hpp"
+
+namespace vg {
+
+using namespace std;
+
+/**
+ * A HandleGraph implementation that overlays some other handle graph and
+ * splits its nodes to satisfy a maximum length
+ */
+template<uint64_t MaxNodeLength = 1024>
+class ChoppedGraph : virtual public HandleGraph {
+public:
+    
+    /// Initialize as the reverse version of another graph, optionally also
+    /// complementing
+    ChoppedGraph(const HandleGraph& graph);
+    
+    /// Default constructor -- not actually functional
+    ChoppedGraph() = default;
+    
+    /// Default destructor
+    virtual ~ChoppedGraph() = default;
+    
+    //////////////////////////
+    /// HandleGraph interface
+    //////////////////////////
+    
+    // Method to check if a node exists by ID
+    bool has_node(nid_t node_id) const;
+    
+    /// Returns true if there is an edge that allows traversal from the left
+    /// handle to the right handle. By default O(n) in the number of edges
+    /// on left, but can be overridden with more efficient implementations.
+    bool has_edge(const handle_t& left, const handle_t& right) const;
+    
+    /// Look up the handle for the node with the given ID in the given orientation
+    handle_t get_handle(const nid_t& node_id, bool is_reverse = false) const;
+    
+    /// Get the ID from a handle
+    nid_t get_id(const handle_t& handle) const;
+    
+    /// Get the orientation of a handle
+    bool get_is_reverse(const handle_t& handle) const;
+    
+    /// Invert the orientation of a handle (potentially without getting its ID)
+    handle_t flip(const handle_t& handle) const;
+    
+    /// Get the length of a node
+    size_t get_length(const handle_t& handle) const;
+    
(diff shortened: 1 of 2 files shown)


src/chopped_graph.hpp
```diff
@@ -0,0 +1,850 @@
...
+    // convert it into a prefix sum
+    chopped_id_begin[0] = 1; // make IDs start at 1
+    for (size_t i = 1; i < chopped_id_begin.size(); ++i) {
+        chopped_id_begin[i] += chopped_id_begin[i - 1];
+    }
+    
+    // inefficient way to find most significant bit, but whatever
+    for (uint8_t i = 1; i <= 64; ++i) {
+        if (max_chopped_count & (1 << (i - 1))) {
+            chopped_index_bit_count = i;
```
adamnovak, line 331: Since we have all the backing graph handles around, should we also check to make sure the handles of the backing graph leave enough high bits for the shift to be safe? We're starting to steal more than just one low bit now.

src/chopped_graph.hpp
```diff
@@ -0,0 +1,850 @@
...
+ */
+
+template<uint64_t MaxNodeLength>
+ChoppedGraph<MaxNodeLength>::ChoppedGraph(const HandleGraph& graph) : graph(&graph) {
+    
+    // note: assumes dense IDs starting at 1 for efficiency
+    // make a vector where every ID can also index before and after by 1
+    chopped_id_begin.resize(graph.max_node_id() + 2, 0);
+    size_t max_chopped_count = 0;
+    graph.for_each_handle([&](const handle_t& handle) {
```
adamnovak, line 314: There's no way to avoid this scan of the whole graph on load, is there?

If we added max node length to the stats we expect a graph to track, maybe we could skip it? Then we could also bypass the overlay when the graph isn't in need of chopping, with an OverlayHelper or something.

src/chopped_graph.hpp
```diff
@@ -0,0 +1,850 @@
...
+protected:
+    
+    const PathPositionHandleGraph* path_position_graph;
+    
+    using ChoppedPathGraph<MaxNodeLength>::chopped_step_handle;
+    using ChoppedPathGraph<MaxNodeLength>::chop_index;
+    using ChoppedPathGraph<MaxNodeLength>::get_underlying_step_handle;
+};
+
+
```
adamnovak, line 294: Shouldn't the chopped graph be vectorizable if the underlying graph is?

adamnovak (commented):
This looks pretty good. I'm also concerned about the indirection overhead and doing a bunch of bit shifts all the time, and also about the startup overhead. If we put this over a memory-mapped graph the first thing it is going to do is scan through the whole graph and get the length of every node in it.

src/chopped_graph.hpp
jeizenga: There's no way to avoid it if you want either a) fast ID -> handle queries, or b) fast handle -> sequence/length queries. I have a code path that does something sort of similar over an XG in mpmap though, and it only takes a couple of minutes at start up.

src/chopped_graph.hpp
jeizenga: Is that necessary in of the places we want to use it? I'm only imagining using this for constructing mapping indexes and then actually doing mappings. I don't think mpmap or map needs the vectorizable interface, but I'm less familiar with giraffe.

src/chopped_graph.hpp
jeizenga: Yeah, that's a good idea.

adamnovak:
@jeizenga Are we ever going to merge this, or should we close it?

jeizenga:
I think we can kill it for now. Part of the use case got handled with the named node back translation features.

PR #3136, opened 2020-12-29
Deprecate a bunch of unmaintained or outdated subcommands

jeizenga:
## Changelog Entry

 * The following subcommands are now deprecated: `combine`, `compare`, `explode`, `join`, `kmers`, `locify`, `msga`, `recalibrate`, `sift`, `sort`, `srpe`, `translate`.
 * The following subcommands have been deleted: `bugs`, `crash`.

## Description

I added a category of commands `DEPRECATED`, which continue to execute but don't appear in the `vg help` output. Also, whenever you run them they print a warning that they are not actively maintained and may be removed in future. I also removed two developer tools that I think have outlived their usefulness.

Most of the choices I included in the deprecated list I expected to be pretty uncontroversial. They are either abandoned or specific to data formats/tasks that we no longer use. Perhaps the most controversial is `msga`, so for that one I included an additional warning message that redirects people to `cactus` and `pggb`.


adamnovak:
Oops, I forgot to wait to remove the exhortation to run `vg bugs --new`.

adamnovak:
Also, if we want to resuscitate any of these back to support, we can.

PR #3137, opened 2020-12-29
Remove RocksDB index

jeizenga:
## Changelog Entry

* `rocksdb` is no longer a dependency.
* `vg index` no longer will produce RocksDB indexes of GAMs or graphs.
* `vg find` can no longer be used to query from RocksDB indexes.
* `vg genotype` no longer accepts RocksDB GAM indexes, and instead accepts only GAMs.
* `vg gamsort` can no longer sort using the RocksDB order.
* `vg compare` has been removed.
* `vg locify` has been removed.

## Description

Thie PR completely removes the RocksDB dependency. It was mostly unused at this point, but there were a few straggler subcommands. If we want to retain the same functionality with them, they will need to be updated to use the `.gai` index from `gamsort`:
* `vg genotype` could optionally use the RocksDB read index, but now only accepts GAMs
* `vg locify` is fully dependent on the RocksDB index to produce sensible input, but it seems to be essentially unmaintained since ~2016. I think this command is a good candidate for being deleted, but this PR just leaves it nonfunctional. **Edit: I deleted it.**


adamnovak:
I don't think we want to say locify is "crippled".

jeizenga:
I could just delete it, but as is on this PR it can't produce sensible results.

adamnovak:
No I mean more that it's ableist.

If it doesn't work, we also should rip it out.

jeizenga:
Got it, language is updated and it is deleted.

adamnovak:
@jeizenga I merged the other deprecating PR so now this conflicts on the removed commands.

PR #3142, opened 2021-01-05
Initial implementation for `vg index --giraffe`

adamnovak:
## Changelog Entry

 * You can now `vg index --giraffe` as long as you specify a graph and a VCF.

## Description

This is an initial implementation of `vg index --giraffe`.

I think it might currently make things more confusing rather than less:

* If it doesn't have a graph, it complains there's no FASTA. `vg index` can't accept a FASTA.
* If you provide e.g. `-G` to build the GBWT, it will build the GBWT through the normal process and then load it again.
* None of the options to customize GBWT or distance indexing propagate through to affect the automatic indexing in the IndexManager.
* Giraffe auto-indexing can't do it either, but there's no support for doing a path cover or subsampling, so this won't produce the indexes we think are the right ones to use for 1KG or HGSVC.

I think the solution might be to move all the indexing processes into the IndexManager, and if we have to support customization we plumb those parameters through.


adamnovak:
I'm going to abandon this in favor of Jordan's approach in #3144.

Issue #3144, opened 2021-01-06
Create a planner with Recipes and Indexes to replace IndexManager


adamnovak:
As part of #3126, we want to make `vg index` understand what indexes are needed for Giraffe and the other tools, and how to make them, and have it have a bit of smarts with regard to what indexes can be made and which jobs can be run in parallel. We also want to push down some of the things we learned about "good" pruning and path covering from toil-vg and/or our custom scripts into vg itself.

I made #3142 based on the existing IndexManager, but it turns out that Jordan has been working on a much cooler implementation that's more of a data-driven planner, which will be easier to extend with new rules and the ability to think about resource constraints.

This issue tracks implementing this new planner-based index management system, using it to replace IndexManager, and plugging it into at least some of `vg index`.

adamnovak:
The planning logic @jeizenga has described to me so far reminds me very strongly of the backtracking used by [logic programming](https://en.wikipedia.org/wiki/Logic_programming) systems like [Prolog](https://www.cis.upenn.edu/~matuszek/Concise%20Guides/Concise%20Prolog.html), or the newer [Dialog](https://linusakesson.net/dialog/docs/choicepoints.html). Maybe some of the algorithms/principles used there will be helpful?

jeizenga:
At this point this goal is reasonably complete, although there are still some developments I'd like to make.

PR #3151, opened 2021-01-14
Beginning of an automatic, tool-oriented indexing interface

jeizenga:
## Changelog Entry

 * Indexes for `vg map` can now be constructed automatically using `vg autoindex`.

## Description

This is the first draft of my super-cool auto-indexer, currently `vg autoindex` although I have dreams of replacing `vg index`. So far, it's only implemented for the indexes in `vg map`, which is by far the simplest mapper to index for. I also haven't included any logic based on the limitations of compute environment (e.g. chunking to control memory usage). I think that behavior can be shoehorned into the framework I've started here though.


jeizenga:
Should probably also mention that this is a partial implementation of https://github.com/vgteam/vg/issues/3144.

adamnovak:
OK @jeizenga, I've looked at this.

The temp file handling is definitely wrong; it seems to be using one fixed prefix in `/tmp` for each index type, with no logic to avoid collisions. This won't even work across multiple users on the same machine. I think you probably want the IndexRegistry to make a unique temporary directory for all its files when it's constructed, and delete it and its contents when it is destroyed or the process exist. We might need to extend the utility.hpp `temp_file` machinery with a function to get a unique temporary directory to make that work.

I don't like that there can be multiple IndexRegistries but only one static IndexingParameters. The IndexingParameters struct should probably be a value you pass to the IndexRegistry constructor/factory, and/or a field you can poke at later, and the recipes should look at their own IndexRegistry's IndexingParameters. Otherwise, how would you test in a unit test whether setting the IndexingParameters works? How would you clean up that global state after every test?

I couldn't figure out why `IndexRegistry::registry` is full of `unique_ptr`s to `IndexFile`s instead of the `IndexFile`s themselves. Since an `IndexFile` is already not necessarily filled in, why not pass them all around by reference? Were you having trouble with the `registry` member reallocating or something?

I feel like there has to be a better way to define what's in a node in the recipe graph, so we don't have to have `VG`, `NodeMapping`, `VG + NodeMapping`, `Pruned VG`, and `Haplotype-Pruned VG + NodeMapping` all as separate nodes, with no commonalities apparent to the system, and with dedicated alias recipes that essentially strip off adjectives or take half of a thing.

My first cut at that would be replacing the string identifiers for index types with, basically, machine-readable parses of the little language you are already using. Each node in the recipe graph ("Index") has "Features", which can recursively have more features. The top-level features correspond to files, and the features they have are additional discardable facts about them. So something like:

```
class Feature {
public:
    Feature(const string& name, std::initializer_list<Feature> features);
};
class IndexName {
public:
    IndexName(std::initializer_list<Feature> components);
    bool matches_query(const IndexName& query);
};
```

So you could have a recipe that produces an `IndexName(Feature("VG"))` and one that produces an `IndexName(Feature("VG", Feature("Pruned", Feature("Haplotype"))), Feature("NodeMapping"))`. If you go looking for `IndexName(Feature("VG"))` you can find both of them. Then the alias recipes can all be implicit. You could retain the requirement to pre-register the kinds of index that can exist, and automatically get the un-bundling and more-specific-to-less-specific rules.

Does that make sense? Or do you have a good reason why all the index types/combinations should remain as special unique snowflakes? Or maybe they should stay strings but you should parse the strings (so "Pruned VG" can grab the first file from "Haplotype-Pruned VG + NodeMapping", and "NodeMapping" can grab the second).

jeizenga:
Re: static indexing parameters. 
Fair point. Another option would be to make the `IndexRegistry` static. I think I initially implemented it static similarly to the subcommand registry. I don't actually remember why I switched it, tbh, but I think the reason why I made the `IndexingParameters` static is that I wanted to be able to scope them in the `VGIndexes::get_vg_index_registry` function that sets up the whole registry. If you have other suggestions about how to set up the registry with less ugliness, we might also be able to make the parameters a member variable in the `IndexRegistry`.

Re: temp files. 
Good idea. I like the sound of the unique temp directory.  

Re: unique pointers. 
Yes, I think I originally did that so I could use raw pointers to refer to the `IndexFile`s and not need to worry about reallocations or something. But everything is held in an STL `unordered_map` now, and I believe they don't invalidate iterators. The unique pointers could probably be dropped, but as is we still have need for `vector<IndexFile*>`'s in the recipe lambdas.

Re: node names. 
I'm kind of unsatisfied with the node naming too (you probably saw that I just commented a bunch of the ugly ones out). In my mind, it seems a bit more difficult to iron out than your suggestion though. I don't know if I have the full picture yet, but my gut tells me that your strategy risks introducing cycles into the dependency graph. Do you have a good reason to think that it won't? Also, how would we determine the relative priority of the recipes that are induced by unboxing/aliasing?

Also, another thing that I want to work into the system is control logic over computational parameters, especially chunked workflows. We use chunks 1) to control memory before constructing XGs, 2) to parallelize GCSA construction, and 3) to parallelize GBWT construction. My understanding is that even if we have sufficient memory to make a single VG, the GBWT and GCSA can be prohibitively slow to index without chunks. Even more complicated is that the graph material should be distributed differently among chunks for different index construction algorithms. My loose idea was to do something adding an intermediate `"GCSA-Chunked VG"` index, whose recipes could decide to either chunk or merely alias depending on system resources and input data size. Would it still be possible to do something like that if the aliasing is all handled by a system of types?

The `NodeMapping` is another annoying case (although I think a less performance-critical one). It's much faster to co-create the `NodeMapping` alongside a `VG` or `VGSet` than it is to deserialize the entire graph(s) just to look up its max node ID. However, the only code path that uses the `NodeMapping` is the `GBWT`-based unfolding during the prune step, which isn't going to follow behind all `VG` or `VGSet` constructions. I'm still unsure how to encode this kind of thing in the recipe priority-based system I designed.

adamnovak:
> I think the reason why I made the IndexingParameters static is that I wanted to be able to scope them in the VGIndexes::get_vg_index_registry function that sets up the whole registry.

I would make `VGIndexes::get_vg_index_registry` take an `IndexingParameters` instance, with the default value being a default-constructed one. Then you can configure your IndexRegistry with the parameters you want it to have.

> I don't know if I have the full picture yet, but my gut tells me that your strategy risks introducing cycles into the dependency graph. Do you have a good reason to think that it won't? Also, how would we determine the relative priority of the recipes that are induced by unboxing/aliasing?

The implicit aliasing would only go in one direction: a less-specific request can be satisfied by a more-specific index. I suppose you could get a cycle: making a pruned vg needs a vg, and one way to get a vg is to have a pruned vg. I suppose you don't have that problem now because a "Pruned VG" isn't actually good enough as a "VG" (because the pruning throws away parts you usually need), and so you don't have that alias. So maybe pruned-ness isn't actually a good example of a safely-removable feature tag. In general, the planner would have to handle the case where you have a recipe to add a trait to something by processing it, and also the processed version is still usable where the original would be.

I'm not sure about how the aliasing would slot into the priority system. Right now it's all definition-order-based, right? So the aliases can be defined before or after other recipes for the same thing and will fall there in the priority order. If they're implicit they can't do that, but always taking an alias first if available isn't good, and neither is not ever taking it until last.

Jumping ahead to the `NodeMapping`, the alias ordering and co-generation problems could both be solved if a recipe cost system was used instead of a priority system. If it cost less to generate the `VG`/`VGset` and `NodeMapping` together than it cost to generate them both separately, but more than it cost to generate just the `VG`/`VGset`, then we would pick strategies that generate them both together only when the `NodeMapping` is actually needed. You could set aliasing/unboxing as zero cost, but still not take that path if you would then be aliasing/unboxing something expensive or unnecessary. If you have it, you would immediately use it for free, while if you don't have it, you would only make it if warranted.

> My loose idea was to do something adding an intermediate "GCSA-Chunked VG" index, whose recipes could decide to either chunk or merely alias depending on system resources and input data size. Would it still be possible to do something like that if the aliasing is all handled by a system of types?

You might need to do more than *just* automatic aliasing to pull this off; you would need to be able to write a recipe that might do work or alias, like you said. Chunked-ness also isn't quite a discardable feature; you can't use a chunked VG anywhere you would use a normal one, because you might have to reassemble it. So if we wanted to support that in a type system we'd need a way to say "possibly-chunked vg" other than "vg[chunked]".


Overall, maybe the right strategy is to fix on a formal language for writing the names of the indexes (documenting what "+" and adjectives mean), and then to build the system by hand so we know what kind of qualifying and unboxing and aliasing operations we actually need. Then if we find ourselves writing a whole bunch of aliasing or unboxing rules of the same general form, we could safely automate rule generation because the names would be predictable. And when we get tired of that we could calcify it with real types.

jeizenga:
In practice, that strategy sounds pretty similar to what I've been doing so far except with explicit semantics for "+". As for cost-based planning, I also toyed around with that idea, but it seemed trickier to me to make a system of costs that were correctly balanced across different recipes rather than a rigid priority. I think it's probably possible though. Also, yes, priority is currently based on the order that recipes are defined in.

adamnovak:
Co-generation is hard.

I've considered some A* or Dijkstra in the space of which indexes you have type systems, but those seem like they'll waste their time considering all possible orders of applying rules, when we can topologically sort the rules. Plus the cost knock-on effects are hard to predict.

What I've done so far in [my branch](https://github.com/vgteam/vg/compare/autoindex?expand=1), besides fixing the temp files, is set it up so that indexes are named with sets of strings, instead of strings. I think that should make the extraction alias rules much simpler, and potentially automateable.

My current best idea for co-generation is to make the plan, and then check if it can be simplified by replacing non-co-generating rules with co-generating rules. It should be pretty easy to do the simplification: any plan that involves generating both halves independently would be improved by just doing the co-generation when you would have generated the first half, instead.

PR #3157, opened 2021-01-15
Make roc plotting script work

xchang1:
## Changelog Entry

 * The roc plot script now runs 

## Description
The roc plotting script was failing with an error message:

Error in max(sapply(name.lists, length)) : 
  invalid 'type' (list) of argument
Execution halted

### DIFF
--- a/scripts/plot-roc.R
+++ b/scripts/plot-roc.R
@@ -45,7 +45,7 @@ if (length(commandArgs(TRUE)) > 3) {
 }
 
 # Determine the order of aligners, based on sorting in a dash-separated tag aware manner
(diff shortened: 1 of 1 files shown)


xchang1:
@adamnovak Can you check this for me please? I don't know R so I don't know really understand what I did

adamnovak (approved):
`factor()` looks like it just tells R to treat strings as distinct conditions in a table. I think it's *supposed* to make strings it loads in a table factors automatically. But `factor()` on a factor appears to be a no-op, and `levels()` only works on a factor, so this shouldn't hurt anything.

PR #3158, opened 2021-01-18
Undrepcate, handlify and generalize vg combine (and deprecate vg concat)

glennhickey:
## Changelog Entry

 * `vg combine` no longer loads graphs as, and writes graph to, protobuf.  Also fixed to join id spaces if necessary.

## Description

In the old days, we could merge together graphs by using `cat`.  This changed at some point due to the vpkg wrapper, so `vg combine` was added to do more or less the same thing that `cat` did.  It still worked by appending protobuf though.  So giving it a list of, say, hashgraphs, would convert them to protobuf in memory, then output the catted protobuf to disk.

This PR changes it to just work on handle graphs directly.  If the input is protobuf, it will use more memory than needed, but that's not our primary use case anymore.  

It will also join IDs as needed (equivalent to running `vg ids -j` first).

The `-p` option is added that will connect paths end-to-end if the same path name is found more than once.  This is equivalent to the old `vg concat -p` option (as `vg concat` is now deprecated).  If the `-p` option is not used, and the same path name is found in more than one input, then it will exit with an error.

### DIFF
--- a/src/subcommand/combine_main.cpp
+++ b/src/subcommand/combine_main.cpp
@@ -18,6 +18,7 @@
 
 #include "../handle.hpp"
 #include "../vg.hpp"
+#include "../io/save_handle_graph.hpp"
 
 using namespace std;
 using namespace vg;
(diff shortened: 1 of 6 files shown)


glennhickey:
Note: old `vg combine` logic added back in via the (deprecated) `-c` option.  It is needed, notably, by the `toil-vg sim` pipeline which still uses it to extract thread graphs.

PR #3183, opened 2021-02-02
Move from Travis CI to our Gitlab and to Github Actions

adamnovak:
## Changelog Entry

 * CI now moved to Gitlab and Github Actions

## Description

This moves ~most of~ what we do on Travis over to Gitlab *and Github Actions*. We want to leave Travis so we can avoid migrating to travis-ci.com and get out of the very long .org queues. ~Travis still has the Mac builds and the documentation upload to `gh-pages` at the moment. The doc upload shouldn't be hard to move over, but the Mac stuff will need to move to e.g. Gitlab Actions.~

This includes some fixes to let the CI run on top of Ubuntu 20.04. I upgraded `vg_ci_prebake` to 20.04 because I was having trouble installing `txm`, which needs Node 10+, which I don't think 18.04 had. But without this PR, CI will fail because it is trying to use too new of a base image that doesn't have exactly the packages our CI system wants.


adamnovak:
OK, this now gets rid of Travis completely, and sets up Mac tests on Github Actions. I'm not sure if we can do ARM tests there too when we finally merge ARM, or if we'll want to add some ARM to our Kubernetes, or what.

PR #3191, opened 2021-02-08
GBWTGraph reference paths

jltsiren:
## Changelog Entry

 * Converting GFA with P-lines and W-lines into GBWTGraph stores the P-lines under sample name `_gbwt_ref`.
 * `vg convert` from GBWTGraph to another graph type stores the threads for a specified sample (default `_gbwt_ref`) as paths.

## Description

Update GBWTGraph to a version that handles GFA files with both P-lines and W-lines better. Instead of skipping the P-lines, the conversion now creates a new sample with name `_gbwt_ref` for them and stores the path names as contig names in GBWT metadata.

When converting GBWTGraphs to other graph types, `vg convert` now assumes that there is a reference sample (default `_gbwt_ref`) in the GBWT. The threads for the reference sample will be converted into embedded paths using the contig name as path name.


ekg:
Would it be simple to treat P-lines as haplotype paths?

Given their trivial equivalence with P-lines, we are unlikely to use
W-lines. It's simpler to just convert them to P-lines with annotations than
go the other way, as most contexts don't have the annotations in the fixed
fields of the W-lines.

jltsiren:
If there are only P-lines, `gbwtgraph::gfa_to_gbwt()` uses them as haplotype paths. Generating GBWT metadata for P-lines is difficult, because there are no standard annotations. The only supported option at the moment is parsing the metadata from path names with a regex.

ekg:
Thanks for the clarification. It seems that for compatibility, we just need
to set up these standards for annotation. If this is needed it should be
straightforward for me to do in a PR.

PR #3215, opened 2021-03-01
Error out before writing invalid rGFA tags

glennhickey:
rGFA doesn't support backwards steps.  `vg convert` changed to throw an error if they are detected instead of silently producing invalid output.  Paths can be changed to the forward orientation with [clip-vg](https://github.com/ComparativeGenomicsToolkit/hal2vg/pull/39).  This logic can be copied into `vg paths` if there's demand for it.

### DIFF
--- a/src/gfa.cpp
+++ b/src/gfa.cpp
@@ -2,6 +2,7 @@
 #include <gfakluge.hpp>
 #include "utility.hpp"
 #include "path.hpp"
+#include <sstream>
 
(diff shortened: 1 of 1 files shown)


glennhickey:
@jeizenga Can you please hold off making the release until this is merged in ?  should be done tests pretty soon I hope.

jeizenga:
Sure, no problem

Issue #3214, opened 2021-02-26
.gfa converted to .vg crashes several vg tools


MariaNattestad:
Hi VG team

I'm experiencing some issues using the vg file produced by `vg convert` from a GFA. In short, several tools in vg crash when I use them with the graph.vg produced by:

```
vg convert -g graph.gfa > graph.vg
```

**1. What were you trying to do?**
Use various vg tools with a graph produced by pggb.

**2. What did you want to happen?**
See several examples below. I wanted these to produce outputs instead of crashing.

**3. What actually happened?**
They all crashed.

**5. What data and command can the vg dev team use to make the problem happen?**

I initially found this using .gfa output graphs from pggb, but here I'm showing the issue with a toy example so it's easier to reproduce. I've confirmed that the behaviors and error messages are the same.

## Create a toy GFA and fasta as inputs

graph.gfa:

```
H       VN:Z:1.0
S       2       CTCACGGCCTGTCGAGAGACCCAGGCTAAT  DP:i:1  RC:i:30
L       1       +       3       +       0M
S       3       TTGCAGACCTTTCTTCTCTCTGGAGCTACAGCGGTGATCGAGGATTATCAAGTTGGCAGCGGAATGCGGTCGTGTCGCGGAAATACATTCCGAGTTTATA    DP:i:1  RC:i:100
S       1       GAAAGTGCCCGACTGGAGTACTATGGTGGGTTGGGGCCTCTGTGTAACACTTGTAGACGACGGACACTGCTAATCTCGGTTGGAGTATTACGAAACCGCT    DP:i:1  RC:i:100
L       1       +       2       +       0M
L       2       +       3       +       0M
P       insertion_path  1+,2+,3+        *
P       reference_path  1+,3+   *
```

sequences.fa:

```
>reference_path
GAAAGTGCCCGACTGGAGTACTATGGTGGGTTGGGGCCTCTGTGTAACACTTGTAGACGACGGACACTGCTAATCTCGGTTGGAGTATTACGAAACCGCTTTGCAGACCTTTCTTCTCTCTGGAGCTACAGCGGTGATCGAGGATTATCAAGTTGGCAGCGGAATGCGGTCGTGTCGCGGAAATACATTCCGAGTTTATA
>insertion_path
[...]

PR #3218, opened 2021-03-02
Address issue with better default behavior in vg chunk

jeizenga:
## Changelog Entry

 * `vg chunk -c` will no longer crash if no paths/nodes are provided

## Description

Previously, `vg chunk` would try to load a path chunk whose start and end coordinates were uninitialized memory. This was one of the problems identified in https://github.com/vgteam/vg/issues/3214.


jeizenga:
It looks like I branched off of my dev branch for the index registry. Oops. In any case, those bits won't hurt anything.

Issue #3224, opened 2021-03-05
Snarl traversing bug


xchang1:
I'm testing my new snarl distance index using the IntegratedSnarlFinder and it's failing a unit test in my branch:
https://github.com/xchang1/vg/blob/distance-index-2/src/unittest/snarl_distance_index.cpp#L1961

- `vg test '[bug]'`  runs only this test and passes my tests
- `vg test '[snarl_distance]'` runs all of my tests and fails with the error 
```src/unittest/snarl_distance_index.cpp:2102: FAILED:
  {Unknown expression after the reported line}
with expansion:

due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
```
Under gdb:
- `vg test '[bug]'` passes
- `vg test '[snarl_distance]'` fails at `unittest/snarl_distance_index.cpp` at line 2104 with the message ```vg: src/integrated_snarl_finder.cpp:1564: void vg::IntegratedSnarlFinder::traverse_computed_decomposition(vg::IntegratedSnarlFinder::MergedAdjacencyGraph&, const vg::IntegratedSnarlFinder::MergedAdjacencyGraph&, std::vector<std::pair<long unsigned int, std::vector<handlegraph::handle_t> > >&, std::unordered_map<handlegraph::handle_t, handlegraph::handle_t>&, std::vector<std::pair<long unsigned int, handlegraph::handle_t> >&, std::unordered_map<handlegraph::handle_t, handlegraph::handle_t>&, const std::function<void(handlegraph::handle_t)>&, const std::function<void(handlegraph::handle_t)>&, const std::function<void(handlegraph::handle_t)>&, const std::function<void(handlegraph::handle_t)>&) const: Assertion `frame.bounds.first != outgoing' failed.```

Under gdb with IntegratedSnarlFinder's debug output turned on
- `vg test '[bug]'` passes, but I was running it earlier and it seemed to get stuck in an infinite loop 
```		Walk edge 6+
			Head: 9-
			Next edge back head: 9-
		Walk edge 9
			Head: 3-
			Next edge back head: 3-
		Walk edge 6+ 
			Head: 9-
			Next edge back head: 9-
		Walk edge 9+
			Head: 3-
			Next edge back head: 3-
```
- `vg test '[snarl_distance]'` segfaults in one of my distance index functions, I think because it's trying to visit a node a second time, when it should have already found the node

[...]

PR #3231, opened 2021-03-11
Stop double-counting single node chains

adamnovak:
## Changelog Entry

 * Single nodes at the end of bridge paths in the Cactus graph won't show up twice in the snarl decomposition

## Description

This will fix https://github.com/vgteam/vg/issues/3224#issuecomment-795971733 by no longer double counting single-node chains in certain cases.


adamnovak:
@xchang1 I copy-pasted your unit test code in here, but I had to split out the double-count checking for snarls and for chains, because a snarl and a chain can start at the same place.

PR #3232, opened 2021-03-12
Min cut graph

code-s-witch:
## Changelog Entry

 * Whatsits now frobnicated


adamnovak:
Hey @code-s-witch, it looks like you `#include "snarl_graph.hpp"`, but `snarl_graph.hpp` isn't included in the PR; I guess it never got `git add`ed. Can you add it to the PR?

adamnovak:
OK I investigated, and I can reproduce at least one test failure on my Linux machine:

```
vg test "min-cut-decompomposition works on a 9 node graph"

4 5 6 7 8 
0 1 2 3 
8 6 5 4 
6 5 8 
6 5 
3 2 1 
2 3 

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
vg test is a Catch v1.8.2 host application.
Run with -? for options

-------------------------------------------------------------------------------
min-cut-decompomposition works on a 9 node graph
-------------------------------------------------------------------------------
src/unittest/min_cut_graph.cpp:670
...............................................................................

src/unittest/min_cut_graph.cpp:823: FAILED:
  REQUIRE( set3.count(x)==1 )
with expansion:
  0 == 1
```

I told it to print out the gamma it's actually getting and it looks like this; it's not very different from what you are looking for, which makes me think it really just is a question of which of several equivalent or nearly-equivalent stochastic min cuts it ends up generating depending on the behavior of the randomized algorithm.

You do say to use a particular, specified generator (`minstd_rand0`), but one subtlety is that the distributions (`uniform_distribution`, `discrete_distribution`) can and do differ between compilers in terms of how they turn the number from `random_engine()` into whatever the distribution is supposed to sample from.

Getting exactly consistent results across platforms is hard, so it might make sense to check the decompositions for sensibility (somehow), rather than to actually check for exactly the sets you expect to see. Maybe just check that the first two sets don't share any nodes and between them have all the nodes, and that no single-node sets are in the decomposition, and that there's a reasonable number of sets in it?

PR #3252, opened 2021-03-29
show haplotypes in kmers from vg find

ekg:
## Changelog Entry

 * enable compilation on gcc 10.2.0

## Description

For GRAFIMO, we need to connect haplotypes with kmers in the graph.


src/algorithms/walk.cpp
```diff
@@ -130,6 +130,36 @@ uint64_t walk_haplotype_frequency(const HandleGraph& graph,
...
+        search_state = haplotypes.extend(search_state, next_node);
+        if (search_state.empty()) {
+            break;
+        }
+    }
+    assert(haplotypes.hasMetadata() && haplotypes.metadata.hasSampleNames());
+    for (auto& thread : haplotypes.locate(search_state)) {
+        // not clear why we have to do this, but it's necessary
+        // perhaps due to fwd/reverse threads
+        auto id = (thread-1)/2; // convert thread id returned by locat to gbwt metadata compatible one
```
jltsiren, line 155: This should be `auto id = thread / 2` or `auto id = gbwt::Path::id(thread)`. The GBWT sequence ids returned by `locate()` encode VG thread id (or GBWT path id) and orientation much in the same way as GBWT node ids encode VG node id and orientation. It would not make much sense to store duplicate path names in a bidirectional GBWT, and it would violate the principle that path names should be unique.

ekg: Thanks @jltsiren !

jltsiren (approved):

ekg:
Well, I merged with master, and then the OSX build broke.

Can anyone with an OSX system confirm?

jltsiren:
Apple changed something, and the old htslib we are using no longer builds. The latest version works, at least on an ARM Mac.

PR #3260, opened 2021-03-31
Add the capacity to directly provide or request any index in the IndexRegistry

jeizenga:
## Description

I don't intend this to be used by the public, but I think it might be helpful for the developers internally.


glennhickey:
Cool, I for one would be a fan of something that can take xg/snarls/gbwt (what comes out of the cactus pipeline) and makes whatever's missing for giraffe (minimizer and distance , I think).

jeizenga:
I think this PR could be used as a vehicle for that, although I'm not sure where the haplotype data would come from in those inputs. I don't want to clutter the help documentation with named options to provide all the intermediate files though.

jeizenga:
Oh wait nvm, I just noticed that GBWT is in the list duh

PR #3270, opened 2021-04-09
Do the setup and arch determination necessary to build old htslib on autoconf 2.70+

adamnovak:
## Changelog Entry

 * vg is now able to build bundled htslib on Autoconf 2.70+

## Description

This should fix the Mac CI build; Homebrew started shipping Autoconf 2.70, the first update in about 8 years, this week.

The htslib PR for fixing this in the install instructions is at https://github.com/samtools/htslib/pull/1199 and hopefully that change will eliminate the need to manually use the host architecture guesser.

### DIFF
--- a/.github/workflows/testmac.yml
+++ b/.github/workflows/testmac.yml
@@ -18,6 +18,14 @@ jobs:
     runs-on: macos-10.15
 
     steps:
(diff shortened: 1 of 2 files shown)


adamnovak:
@jltsiren is right that really we will need to upgrade vcflib to the current htslib-less version, and ship tabixpp and maybe htslib separately, but this seems to fix the build for now.

PR #3272, opened 2021-04-10
Add node chopping to graph construction from GFA in autoindex

jeizenga:
## Changelog Entry

 * `vg autoindex` now chops nodes when constructing graphs from GFA

## Description

Before it was possible to make graphs that were unindexable by GCSA2.


ekg:
Oh nice! I just ran into this problem. Autoindex was otherwise perfect.

Would you please pull in the chop algorithm from odgi? It maintains the
node order (so chopped nodes have consecutive ids) but should be otherwise
identical.

PR #3324, opened 2021-06-18
Single node chains

### LINKED ISSUE #3322: Snarl traversal misses single-node components of the root
The IntegratedSnarlFinder's traverse_decomposition() function doesn't find single-node connected components

In this unit test, the graph only contains one node but the snarl traversal never reaches it
https://github.com/xchang1/vg/blob/distance-index-2/src/unittest/snarl_distance_index.cpp#L39

A node that isn't connected to anything else should still be a component of the root, so the snarl traversal should see the node as the start and end of a chain

adamnovak:
## Changelog Entry

 * `IntegratedSnarlFinder` properly generates single-node chains for bridge edges in the Cactus graph.

## Description

This should fix #3322. We weren't generating chain messages when a bridge edge path was just a single node, but we also weren't noticing because the current snarl manager doesn't need them. The new distance index snarl manager is going to.

### DIFF
--- a/src/integrated_snarl_finder.cpp
+++ b/src/integrated_snarl_finder.cpp
@@ -1691,12 +1691,16 @@ void IntegratedSnarlFinder::traverse_computed_decomposition(MergedAdjacencyGraph
                             // It turns out there's only one edge here.
                             // It is going to become a contained self-loop, instead of a real cycle
                             
-                            // TODO: register as part of snarl in index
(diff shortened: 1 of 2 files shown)


adamnovak:
CI failed because `mirror.gcr.io/library/ubuntu:20.04` couldn't be found as a Docker base image. It definitely exists, so I am retrying.

PR #3359, opened 2021-07-28
Fix cached distances for minimizers

adamnovak:
## Changelog Entry

 * Minimizer indexes no longer incorrectly cache distances for looping chains, fixing a duplicate-alignment, low-MAPQ bug in Giraffe

## Description

@xchang1 thinks this will fix the cause of @cmarkello's problem where HPRC Cactus graphs are having reads map with incorrectly-low MAPQs, due to clusters not joining up and duplicate alignments being generated.


adamnovak:
This definitely fixes my test read. @cmarkello I think you should be able to just swap out your minimizer index for `/public/home/anovak/build/vg/trash/dvdebug/GRCh38-freeze1.cactus.f1g-90-mc-jul7.k29.w11.paths.recached.min`, if you don't want to grab a vg Docker and rebuild it yourself.

PR #3382, opened 2021-08-23
Make multipath alignment tails more decomposable

jeizenga:
## Changelog Entry

 * Multipath alignments from `vg mpmap` are more decomposable in tails near soft clips

## Description

Usually, `mpmap` merges non-branching subpaths into a single subpath. This PR prevents that from occurring when it's possible that we might collapse over a branch whose alignment was soft-clipped.


jeizenga:
All tests pass, but we're having issues with the uploading/reporting end. I'm merging over this "fail".

PR #3383, opened 2021-08-23
Improve error handling for vg clip bed input

glennhickey:
ie don't exit with cryptic assertion error if input BED contains path that isn't in a graph


ekg:
Does VG clip produce a BED?

glennhickey:
No, it only takes it as input atm.

ekg:
How can I produce a BED of path regions inside of difficult bubbles?

PR #3387, opened 2021-08-24
Read randomizer for Giraffe

adamnovak:
## Changelog Entry

 * Giraffe now shuffles tied top read mapping candidates like mpmap and map, to avoid arbitrary biases

## Description

This uses the same general shuffling logic as mpmap and map use in Giraffe.

We might be able to save some copies and duplicate hashings, since we're doing the shuffling at all stages and computing the seed repeatedly.

We also might not actually need to do the shuffling at all stages, if we analyze Giraffe's algorithm a bit more closely.


adamnovak:
This is stuck behind #3412.

PR #3392, opened 2021-08-25
The ref allele gets an AT so change its INFO Number from A to R

### LINKED ISSUE #3390: Number entry for INFO/AT field in VCF from `vg deconstruct` should be 'R'
Since the INFO/AT field in VCF from `vg deconstruct` has one value per allele (including the REF allele), the Number entry should be 'R' instead of 'A'.

I guess this is the only line of code needs to be changed:
https://github.com/vgteam/vg/blob/4b0b427c12b45e71f9e1172cf6c269dfcbf8b839/src/deconstructor.cpp#L671

glennhickey:
## Changelog Entry

 * `deconstruct` AT INFO field fixed to have `Number=R` instead of `A`

## Description

Resolves #3390.  Thanks @wwliao


glennhickey:
Will merge this as I'm pretty sure changing an INFO field character in deconstruct VCFs is unrelated to the autoindex crash in the mac tests.

PR #3410, opened 2021-09-07
Add support for GBWT haplotypes as reference paths in vg rna

jonassibbesen:
## Changelog Entry

 * `vg rna` now supports using GBWT haploytypes as reference paths (`--use-hap-ref` option).
 * `vg rna` can now create a spliced pangenome graph and pantranscriptome with chopped nodes (`--max-node-length` option).
 * Spliced pangenome graphs and pantranscriptomes can be sorted and have their ids compacted in `vg rna` (currently only works for inputs in the PackedGraph format).
 * The transcript haplotype info table from `vg rna` no longer contains the "Reference" column (is part of the "Haplotypes" column now).
 * The names for the options `--use-all-paths` and `--add-non-ref-paths` in `vg rna` have been changed (single-letter option the same) and the option `--gbwt-id-interval` has been removed (not relevant anymore).

## Description

Updated the transcriptome code to allow for GBWT haplotypes being used as reference paths. As part of this several minor improvements were made to the code in addition to the things mentioned above. This includes much faster parsing of input transcript annotations, especially if multiple annotations are given as input. Also, augmenting the pangenome graph with splice-junctions scales better now with the number of reference transcripts.


adamnovak:
@jonassibbesen This is failing one of the autoindex tests in a way that I can't blame on CI.

PR #3416, opened 2021-09-09
Use path jaccard to correctly localize variants in segdups / VNTRs

ekg:
## Changelog Entry

 * vg deconstruct: localize each snarl traversal to a single reference coordinate, avoiding the amplification of false variant calls through collapsed reference positions in the graph. 

## Description
Previously, vg deconstruct would emit the same VCF record for each looped reference position in a segdup, leading to false positive amplification in these regions. This approach ignores the fact that the graph provides a full description of the mapping of the sequences to each other. Inside of such regions, we disentangle the relationships by comparing 10kbp of flanking sequence of the reference on each side of the bubble to equivalent flanks from all path traversals across it. This is done using a jaccard metric over node space, which is efficient to compute. A single best reference position is chosen for each path traversal. In experiments on CYP2D6/7, this resolves 97% of reported conflicts, most of which appear to be cases where the sample is homozygous for the reference allele in this locus.

Possible improvements:
- Configure a different flank size. Currently this is hard-coded to something that should provide unambiguous mappings in most of the human genome.
- Optionally filter or ignore any genoype with an unexpected ploidy.


ekg:
@adamnovak what's up with gitlab?

ekg:
I've now enabled nested parallelism. This should help to speed up deconstruct, which seems to get stuck in a few hard top-levels snarls in general, a problem that's exacerbated with the path jaccard metric calculation.

adamnovak:
Gitlab was broken for a bit, because we had to lock down the AWS buckets where we'd been storing the test data (among other things), to make sure nobody could download loads of stuff out of them and sap our AWS money. Also our Kubernetes cluster had its certificates expire.

But it's all working again now, and I've merged the CI fixes into this branch.

So the way this disambiguates alignments is to say, which traversal of the reference path through this area looks the most like the paths that go through the alternate versions of this snarl? And then it puts that snarl onto that reference traversal? And we hope that two variants that are really on different copies in the reference don't overlap and become a single snarl?

adamnovak:
I merged in the Ci fixes, but it looks like this fails a couple of the deconstruct Bash tests:

```
#30 342.2 t/26_deconstruct.t  (Wstat: 0 Tests: 22 Failed: 2)
#30 342.2   Failed tests:  15-16
```

ekg:
This branch currently lets me generate "untangled" VCFs for the pggb graph in relatively short time. The CHM13 VCFs are bigger than the GRCh38 ones in the same pattern I'd expect based on the gaps filled in CHM13. There is no longer an effect where the centromere or repeat complexity causes an increase in the size of the VCF. On focused tests, it's clear that it removes the vast majority of the conflicted sites.

Remaining conflicts in VNTRs occur due to samples that have more or less than the reference copy number. There are ways of handling these, but I'm evaluating if it's worthwhile to pursue them in vg deconstruct. A potential improveent is to use a reference path set that covers all of the unique sequences in these structurally polymorphic regions, and to adjust vg deconstruct to uniquely map each allele observation onto the best-matching one of these.

It might also be nice to optionally filter out conflicted genotypes. @glennhickey that would also affect your HPRC graphs.

I'm testing the phased VCF output. To simplify things, I dropped the LRUCache used for GBWT positions. I may need to bring this back.

In this branch I've also disabled jemalloc. This, when using nested omp threads, would eventually break with very strange errors indicative of overflow and out-of-bounds memory accesses. This could be due to my compilation setup, where I am making a static binary on a different system from where I'm running it.

adamnovak:
Does this pass valgrind or asan testing with jemalloc off? If it breaks jemalloc it must be doing something wrong somewhere, right? Or do nested OMP threads do some kind of swapping between OS threads that jemalloc can't handle, or that depletes its per-thread free lists in a way it somehow can't refill?

glennhickey:
@ekg 
I am concerned that your changes will break the workflow for non-pggb graphs.  Please test or let me test a cactus graph before merging (it's about 5 hours on 64 cores).  To be safest, I'd propose putting the jaccard stuff behind an option, though it looks like that ship may have sailed already.  

I'm pretty sure the lru cache is necessary to use anything from the GBWT as reference.  It's not ideal, but beats the alternatives of loading up a full path-position index for all alts, or scanning the full path for each query.  It's thread-safe just by having one cache per thread.  In addition to letting you call any reference, the lru cache allows the `-a` option to work, which outputs calls within insertion sequences.  

For the conflicts, I'm happy with an option to filter them.  They are, by default, flagged in the VCF though admittedly not in an easy way to script against (perhaps there's a better way to format it?).  

I know this remains a point of contention, but I never understood why not just run `odgi untangle` to resolve the reference cycles, the run the regular `deconstruct` on that?   That would seem more elegant to me as a user, and also make it easier to go back to GFA to visualize the VCF calls.  It'd also give the graph a better chance of working with `vg call` or any other tool that wants to treat separate reference positions separately.

ekg:
> I'm pretty sure the lru cache is necessary to use anything from the GBWT as reference. It's not ideal, but beats the alternatives of loading up a full path-position index for all alts, or scanning the full path for each query. It's thread-safe just by having one cache per thread. In addition to letting you call any reference, the lru cache allows the -a option to work, which outputs calls within insertion sequences.

With the exception of the LRUCache, functionality should be basically the same for the GBWT based workflow. The fundamental issue is that nested parallelism causes problems for this caching technique. That said, I should probably just rebase my changes and re-enable the LRUCache for GBWT. I ultimately worked around this by using paths for phased VCF output rather than the GBWT, and in this mode the positional index is built or provided.

>  I am concerned that your changes will break the workflow for non-pggb graphs. Please test or let me test a cactus graph before merging (it's about 5 hours on 64 cores). To be safest, I'd propose putting the jaccard stuff behind an option, though it looks like that ship may have sailed already.

Let's make sure that the full range of functionality and performance for your workflow is still preserved before I merge.

The jaccard metric compute won't get run unless the reference traverses a snarl more than once, so this shouldn't be an issue for your graphs. The VCF phasing depends on the use of an option `-H` to provide the path name hierarchy delimiter.

> I know this remains a point of contention, but I never understood why not just run odgi untangle to resolve the reference cycles, the run the regular deconstruct on that? That would seem more elegant to me as a user, and also make it easier to go back to GFA to visualize the VCF calls. It'd also give the graph a better chance of working with vg call or any other tool that wants to treat separate reference positions separately.

This requires the development of a completely new graph building / postprocessing step. We'd have to run odgi untangle, then induce a new graph with seqwish. That would have to be tested extensively. It also changes the graph, and is hidden functionality from a user perspective.

It will always be useful to be able to dentangle a graph in the VCF generation step, without further modification to it. Probably, the context length parameter that's now hard-coded should be exposed more directly, but I see little reason why the default should be the reporting pattern we see on master.

adamnovak:
@ekg Can I get ahold of the files you're running on that break the new deconstructor with jemalloc enabled?

adamnovak:
It looks like the Docker-build tests managed to fail the unit tests with a complaint from the non-jemalloc allocator about something screwing up the heap:

https://ucsc-ci.com/vgteam/vg/-/jobs/96240#L3013

(Then the rest of the Gitlab tests tried to run anyway, which is its own CI bug.)

Maybe we can track down whatever is going wrong with enough `vg test` invocations?

adamnovak:
My build has a pretty reliable segfault in `vg test "Weibull fitting function learns parameters that are approximately correct"`, which has nothing to do with deconstruct. Maybe we don't actually work right with GNU malloc?

There's a pretty clear past-the-end write in `y` here that `valgrind` turns up:

https://github.com/vgteam/vg/blob/22044be23a7fb786376245fd0e993acfbc9c19d2/src/statistics.cpp#L363-L368

I think we have let past-the-end writes into our codebase that mostly work because the allocators we usually use mostly don't store anything important right after an allocated block.

I don't know if we'll catch deconstruct doing anything bad until we sort these out.

To dig up this one, I ran: `vg test -t | rev | cut -f1 -d' ' | rev | grep '\[' | xargs -t -n1 vg test`

jeizenga:
Uh oh, that looks like me...

jeizenga:
Easy fix though: just `y[i]` -> `y[i-1]`

ekg:
> @ekg Can I get ahold of the files you're running on that break the new deconstructor with jemalloc enabled?

Yup, this was the fastest to fail:

https://s3-us-west-2.amazonaws.com/human-pangenomics/pangenomes/scratch/2021_07_30_pggb/chroms/chr11.pan.fa.a2fb268.e820cd3.9ea71d8.smooth.gfa.gz

Try `vg deconstruct -t 24 -P chm13 -e -a -H '#' chr11.pan.fa.a2fb268.e820cd3.9ea71d8.smooth.gfa >chr11.vcf`.

You might want to directly run under gdb to see where it's failing, e.g.:

`gdb vg -ex 'r deconstruct -t 24 -P chm13 -e -a -H '#' chr11.pan.fa.a2fb268.e820cd3.9ea71d8.smooth.gfa >chr11.vcf'`

ekg:
As for shaking out memory access bugs, I would suggest that we use ASAN. It's by far the best (and newest) of the C++ memory access checkers. It tracks everything and has been very reliable at showing where the first invalid memory access occurs, rather than where the effects of out-of-bounds writes eventually manifest.

You can enable it by adding this to the C++ flags: `-fsanitize=address`, and then running the executable with this environment:

`ASAN_OPTIONS=detect_leaks=1:symbolize=1 LSAN_OPTIONS=verbosity=0:log_threads=1 vg test`

https://github.com/pangenome/odgi/blob/master/CMakeLists.txt#L13

It takes ages (20x runtime), but maybe if we run it just on the unit tests we'll get hits in reasonable time. The test example above might take several days to trigger an error.

GDB won't work with the ASAN binaries. They also can't be statically built. But it's definitely the best checker. We've prevented many bugs by running tests with ASAN enabled, which seems to have kept things relatively safe. I need to set up a build environment for vg on the same host to verify that it's not just something related to cross-compilation. @adamnovak  You might help me out just by demonstrating it's not an issue for you when you build and run on the same system (fingers crossed).

adamnovak:
It looks like even after I fix the memory errors that were lurking in other parts of vg, a bunch of the `vg deconstruct` command line tests fail. I think some are related to the removal of the `-A` option, which is still sued in the tests.

I also noticed I can make it crash by just trying to do a cyclic-path deconstruct without `-e`:

```
vg construct -v tiny/tiny.vcf.gz -r tiny/tiny.fa | vg view -g - > cyclic_tiny.gfa
printf "L\t12\t+\t9\t+\t0M\n" >> cyclic_tiny.gfa
printf "P\ty\t1+,3+,5+,6+,8+,9+,11+,12+,9+,10+,12+,14+,15+\t8M,1M,1M,3M,1M,19M,1M,4M,19M,1M,4M,1M,11M\n" >> cyclic_tiny.gfa
vg view -Fv cyclic_tiny.gfa > cyclic_tiny.vg
vg index cyclic_tiny.vg -x cyclic_tiny.xg
vg find -x cyclic_tiny.xg  -n 10 -n 11 -n 12 -n 13 -n 14 -n 15 -c 1 > cycle.vg
vg index cycle.vg -x cycle.xg
vg deconstruct cycle.xg -p y -t 1
```

This appears to be because some of the "dummy" `make_pair(step_handle_t(), step_handle_t())`s in `path_travs.second` are actually getting read and passed to the graph. I think we're just always looking for context for each possible traversal even when the traversals aren't actually coming from embedded paths.

adamnovak:
OK, I built 717ad7a6a dynamically on Courtyard, but with jemalloc turned back on, and did this to convert the GFA to HashGraph and then deconstruct the HashGraph, in hopes that it would fail and I could rerun it quickly (also on Courtyard):

```
time vg convert -a chr11.pan.fa.a2fb268.e820cd3.9ea71d8.smooth.gfa >chr11.pan.fa.a2fb268.e820cd3.9ea71d8.smooth.hg
gdb vg -ex 'r deconstruct -t 24 -P chm13 -e -a -H "#" chr11.pan.fa.a2fb268.e820cd3.9ea71d8.smooth.hg >chr11.vcf'
```

I stopped it a bunch to see what it was up to; it wasted a bunch of time building a path position index, which I could save if I made an XG, and then a bunch of time making snarls, which I could safe if I fed it premade snarls.

It also wasted a bunch of time locking on the `add_variant` critical section, which *shouldn't* be needed, because `add_variant` dumps into per-OMP-thread lists. But presumably that's there because threads were forgetting what threads they are.

But it eventually ran to completion without any complaints of running out of memory, or out-of-bounds access.

Next I'll try the full job all in one call, to see if somehow doing the GFA import and the deconstruct in the same process can upset the allocator.

adamnovak:
@ekg I ran the whole `gdb vg -ex 'r deconstruct -t 24 -P chm13 -e -a -H "#" chr11.pan.fa.a2fb268.e820cd3.9ea71d8.smooth.gfa >chr11.vcf'` command to do everything in one go, with `jemalloc` built in, in a dynamic build, and didn't get any crashes.

I don't *think* I fixed any memory errors that should have affected it, but maybe I did. Or maybe there's something wrong with the static build, or your static build in particular. I can try a static build next.

adamnovak:
OK, I ran a static vg with jemalloc in it, which I built on my machine and ran on Courtyard, and I wasn't able to trigger the bug. Here's the report from `time`:

```
        Command being timed: "vg-static deconstruct -t 24 -P chm13 -e -a -H # chr11.pan.fa.a2fb268.e820cd3.9ea71d8.smooth.gfa"
        User time (seconds): 55282.24
        System time (seconds): 647.42
        Percent of CPU this job got: 717%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:09:52
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 12089972
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 118368074
        Voluntary context switches: 293060
        Involuntary context switches: 1844289
        Swaps: 0
        File system inputs: 5666576
        File system outputs: 2776808
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
``1

Here's the binary if you want to try it on your machine @ekg: https://ipfs.io/ipfs/QmTHDCky6Xkf5bLfyzXVPEKaxs9jYeNdWhR8V2z8uHTLC9/vg

I think we should fix up the failing tests, turn jemalloc back on by default, and merge this; it might be something unique to your setup that upsets jemalloc. It used a bit over 12 GB of memory for me, so I don't think it is truly running out of memory for you.

adamnovak:
I wanted to leave a note here: We figured out that the out-of-memory errors on @ekg's machine with jemalloc are due to https://github.com/jemalloc/jemalloc/issues/1328#issuecomment-927683911 and that machine having memory overcommit disabled. We don't want to ship static binaries with jemalloc linked in if it's going to do that, because users will get weird crashes in their institutional environments, won't be able to change the overcommit setting, won't be able to rebuild vg, and won't be able to conclusively implicate this bug as the problem.

However, I also did some mapping speed tests with `vg giraffe`, and taking out `jemalloc` and using the system allocator cut our mapping speed by about 40%. So I still don't really want to do that by default either.

We could look for another fast allocator that doesn't hit the memory mapping limit, somehow configure jemalloc not to hit the memory mapping limit anymore, see if we can ourselves adjust the memory mapping limit, or see if we can do something to pick the allocator at runtime depending on the overcommit setting.

ekg:
An update on my end. I tested turning jemalloc off on other workloads and
it was completely infeasible. The memory fragmentation was too much to
complete the analyses I was running.

I actually reconfigured my systems to allow overcommit. That's working
great.

My current suggestion would be to trigger a warning when running on a
system without overcommit enabled, and this at least will reduce support
costs.

adamnovak:
I have a Secret Plan to reconfigure jemalloc to hopefully be a bit happier on these systems. I can add a warning to that code.

adamnovak:
OK, I have some hacks in here now that print a warning and try to reconfigure jemalloc at runtime, after it has been initially configured and some memory dynamically allocated. I can still pass the vg unit tests, but I'm not sure about the CI tests yet.

I think the only truly safe thing to do is, when it looks like overcommit is disabled and jemalloc isn't using the right settings, set up the `MALLOC_CONF` environment variable with the settings we want and completely re-`exec()` ourselves. Or just bail out and demand that the user set `MALLOC_CONF` for us.

ekg:
Thanks Adam! I still need to fix some issues still to pass CI, and also
update the phased path based output to include the AT tag. Will do that
early in the week.

ekg:
The AT tag is still there :facepalm: so I was just confused. That means the real problem is those failing tests. Checking.

PR #3426, opened 2021-09-17
Use libsnarls

adamnovak:
## Changelog Entry

 * vg now uses `libsnarls`, which brings along many of the vg-ecosystem libraries

## Description

This PR will make vg use the new `libsnarls` for snarl finding, and its CMake build system to build many of vg's dependencies.

Right now the build system stuff if in place, but we still use the vg copies of all the files I copied into libsnarls, and don't really link code from libsnarls.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,9 +22,6 @@
 [submodule "gcsa2"]
 	path = deps/gcsa2
 	url = https://github.com/jltsiren/gcsa2.git
-[submodule "sdsl-lite"]
-	path = deps/sdsl-lite
-    url = https://github.com/vgteam/sdsl-lite.git
 [submodule "deps/libVCFH"]
 	path = deps/libVCFH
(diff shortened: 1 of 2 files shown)


adamnovak:
It looks like CI can't build this because we no longer have a `deps/sdsl-lite/CMakeLists.txt` to `sed` the `march=native` out of.

We control `sdsl-lite` for the build now, so we should just remove `march=native` from it in our fork, and pass the architecture in via CXXFLAGS at the top of the build, which we support now since we set everything up for arm.

jltsiren:
We are also using our fork of `sdsl-lite` in standalone tools, where `march=native` is often the right choice. Changing the `sdsl-lite` build system to take the target architecture from an environmental variable (if present) should be a better idea.

adamnovak:
I think we might actually end up using libbdsg and its snarl distance index to provide this stuff, instead.

PR #3438, opened 2021-09-29
Improve surjection in low-complexity regions

adamnovak:
## Changelog Entry

 * Surjection can now prune back low-complexity or short anchors to give more-optimal alignments against the target paths

## Description

Surjection will take all matches against the target path as true, and preserve them, even if a downstream tool is expecting locally-optimal input alignments against that path. This PR adds some options to allow particularly suspicious-looking matches to be removed and the relevant parts of the read realigned against the target path.


adamnovak:
I tested this and we seem to be happy with the performance, so I'm going to merge it.

PR #3441, opened 2021-10-04
Report not-really-paired alignments to the funnel

adamnovak:
## Changelog Entry

 * Giraffe `--track-provenance` will no longer crash when considering improperly paired mappings

## Description
We were letting pairings into our list of pairings to consider without telling the funnel about them. Then when we tried to use them in the funnel in later stages we would crash. This fixes that.


adamnovak:
This should fix the provenance-tracking issue that @jltsiren discovered; I don't think we ever opened a ticket for it.

PR #3468, opened 2021-10-27
Deduplicate endpoints

adamnovak:
## Changelog Entry

 * Mappings with shared endpoints are now treated as equivalent for MAPQ by `vg giraffe` 

## Description

This fixes @jmonlong's read where several mappings ending at the same point managed to produce a MAPQ 0.

I haven't looked at all into *why* we're having multiple alignments that nearly coincide, and I'm also not removing all the both-reads-coinciding secondaries like `vg mpmap` does, since they seem useful as `-M` output and also they *should* fall out for MAPQ purposes when grouped in just like one-read-coinciding secondaries. @jeizenga does that make statistical sense?

I haven't yet tested what the impact of this is on the DeepVariant pipeline.

### DIFF
--- a/src/minimizer_mapper.cpp
+++ b/src/minimizer_mapper.cpp
@@ -586,6 +586,9 @@ vector<Alignment> MinimizerMapper::map(Alignment& aln) {
     // Grab all the scores in order for MAPQ computation.
     vector<double> scores;
     scores.reserve(alignments.size());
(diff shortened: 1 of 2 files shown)


adamnovak:
We still need to test this on Terra I think; we aren't using it right now in our DV results.

adamnovak:
We didn't like the results of these tests and we're not going to do this, I don't think.

PR #3481, opened 2021-11-09
Fix bug in trimming logic in spliced alignment

jeizenga:
## Changelog Entry

 * Fixes a crash in `vg mpmap` that could happen during spliced alignment of low quality reads

## Description

Small bug-fix.
Edit: fixed a couple more subtle but nasty bugs that came up in the course of testing the earlier bug fix.

### DIFF
--- a/src/algorithms/component_paths.cpp
+++ b/src/algorithms/component_paths.cpp
@@ -227,7 +227,6 @@ vector<unordered_set<path_handle_t>> component_paths_parallel(const PathHandleGr
             }
         });
     }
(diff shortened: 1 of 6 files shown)


adamnovak:
`1206994` added a bug that manifests on my machine (with my RNG seed for the tests?). Some of the values we end up needing to store are actually greater than the GCSA index size by a bit, for some reason.

jeizenga:
In the unit tests, I assume? Is it happening in `index`, `map`, or `mpmap`?

PR #3488, opened 2021-11-23
Don't generate or expect empty allele paths

adamnovak:
## Changelog Entry

 * `vg construct -a` no longer generates, and other vg tools no longer expect to see, empty paths for some variants

## Description

We used to use empty paths (I think for the `_0` reference path?) when generating alt allele paths for some indel variants. We didn't actually need those paths to be there, and now we should be able to use nonexistent paths instead of empty ones.

### DIFF
--- a/src/constructor.cpp
+++ b/src/constructor.cpp
@@ -1548,16 +1548,16 @@ namespace vg {
         max_rank.clear();
         // We have this many nonempty paths at the start of the collection at
         // the start of every loop.
-        size_t nonempty_through = 0;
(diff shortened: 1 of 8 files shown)


adamnovak:
@jltsiren I think with this we won't need to care so much about empty paths in https://github.com/jltsiren/gbwtgraph/pull/15

adamnovak:
Looks like this isn't going to pass; I need to make some of the unit and CLI tests account for different path counts in the output from what they check for right now.

PR #3491, opened 2021-11-24
Fix mem accelerator

adamnovak:
## Changelog Entry

 * `MEMAccelerator` can no longer slice bits off the numbers it is supposed to be storing

## Description

Commit 120699454c92b5fb668420ddc3322b4d29bbe14c seemed to introduce a bug on my machine where the `MEMAccelerator` tests would fail, probably because of the particular seeds/RNG behavior my machine uses to make the random graphs for the test.

The underlying problem is that the size of the GCSA index isn't the maximum number that can appear in a GCSA range. Perhaps @jltsiren can explain what the maximum number actually is guaranteed to be; I've observed the values going up to 256 in a 250-sized index.

This PR adds better logging if the test fails, and also sets up the `MEMAccelerator` build to make room for an arbitrary 100 numbers beyond the end of the GCSA index, and to increase the integer width during the build (with a warning) if it determines that it has a bigger number to store.

The hooks to do this are going to make it a little slower; we should eventually work out what the actual maximum values can be and just allocate bits to store those up front.

### DIFF
--- a/src/mem_accelerator.cpp
+++ b/src/mem_accelerator.cpp
@@ -13,16 +13,34 @@ namespace vg {
 MEMAccelerator::MEMAccelerator(const gcsa::GCSA& gcsa_index, size_t k) : k(k)
 {
     // Compute the minimum width required to express the integers.
-    // We need to be able to represent numbers up to th index size + 1, to
-    // account for empty ranges where the first number is greater than the
-    // second and the second is the index size. 
+    // We need to be able to represent numbers up to a bit over the index size,
+    // for some reason. We're not sure how much over the index size is needed.
+    // Start with a more or less arbitrary guess of the max needed value
+    size_t max_needed_value_guess = gcsa_index.size() + 100;
+    
+    // Compute the bit width needed to represent it
     uint8_t width = 0;
-    size_t log_cntr;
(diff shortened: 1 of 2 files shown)


jltsiren:
@adamnovak Can you clarify the terminology? What do you mean by a GCSA range?

jeizenga:
Maybe the issue is empty ranges? @jltsiren, can either of the integer values of an empty query-result from LF be greater than `gcsa->size()`?

jltsiren:
There are no guarantees on the values in empty ranges. In particular, some branches may return a closed range [0, -1] as a pair of unsigned integers. Other branches may return empty edge ranges, where the numbers are bounded by `edgeCount()`, which is usually a bit larger than `size()`.

jltsiren:
Also, you can use `sdsl::bits::length(max_value)` for computing the width.

jeizenga:
Okay, I imagine that's the problem then (which is why it doesn't show up on larger graphs, only in the unit tests). I might clean this up a bit in a later commit, but this should be fine to merge to fix your crashes.

adamnovak:
Yeah, by "range" I mean `gcsa::range_type` values, and I on;y actually saw their component numbers get too big in semantically-empty ranges.

PR #3494, opened 2021-12-09
Fix capture in Autoindex recipes

adamnovak:
## Changelog Entry

 * `vg autoindex` no longer calls out-of-scope lambdas captured by reference

## Description

`vg autoindex` has been suspiciously broken in some builds since forever; rebuilding with slightly different code would cause it to crash constantly or work perfectly.

I think the problem has been that we've been &-capturing lambdas by reference, out of a setup-function scope that goes away after the indexer is set up. These lambdas don't ultimately hold any real data, but capturing the lambdas by reference out of a local scope that is gone when we call them is undefined behavior, and so we were seeing undefined behavior.

This should remove some of the helper lambdas and make them real functions, and the rest are now captured by value instead of by reference.


jeizenga:
My bad

PR #3495, opened 2021-12-09
GBZ as PathHandleGraph

adamnovak:
## Changelog Entry

 * `.gbz` files can now be used anywhere a graph with embedded paths is expected.
 * `vg autoindex` will embed graph paths into Giraffe GBZ files.

## Description

This sets up a `GBZGraph` wrapper (since `gbwtgraph::GBZ` doesn't itself implement PathHandleGraph) which lets us load GBZ files as graphs and read paths out of them. It also lets us embed named paths into GBZs when building from VCF. I'm not sure yet if they make it through from e.g. GFA.


jltsiren:
Does this handle the case where we create multiple parallel GBWT construction jobs from a single graph and multiple VCF files? Or does `vg autoindex` even support that? If it does, the embedded paths should go to the same jobs as the corresponding haplotypes, because otherwise the fast merging algorithm won't work.

PR #3504, opened 2021-12-20
fix nesting tags bug in vg call

glennhickey:
Fixes bug introduced in #3497 where LV/PS tag code brought into vg call assumed snarls were never flipped (true in deconstruct, not so, apparently, in call)


glennhickey:
@adamnovak Please include this PR in the release (I'll merge it in as soon as tests pass).  Thanks

PR #3524, opened 2022-01-28
GFA space output for `vg snarls`

adamnovak:
## Changelog Entry

 * `vg snarls` now supports `--named-coordinates` for reporting snarls and snarl traversals in input GFA coordinates.

## Description

This adds the `--named-coordinates` option to `vg snarls`, so you can get snarls on a GFA file or GBZ graph and get your output in terms of GFA space. I had to add a new `name` member on the `Visit` to support this.

I'm also dong a little hacky rewriting of traversals, to elide the multiple visits to trivial snarls and different fragments of a segment that you get when a segment is split. The result of this is that if the traversal finder reports cyclic traversals that loop around on a node, the looping will be elided. I probably want to fix that.


adamnovak:
Here's a static binary of this: https://ipfs.io/ipfs/QmQH3C1GUUVekLDMYfXwESLFUL3sJU3JrJXN8gVkHv78jD

PR #3611, opened 2022-03-21
Chunk GAM as unparsed messages instead of Alignment objects

adamnovak:
## Changelog Entry

 * `vg chunk -m` no longer bothers parsing `Alignment`s

## Description

We want GAM chunking to be faster. This should save us some time by making it operate on un-parsed serialized read strings, instead of parsing into Protobuf Alignment objects and then re-serializing them.

### DIFF
--- a/src/subcommand/chunk_main.cpp
+++ b/src/subcommand/chunk_main.cpp
@@ -14,6 +14,7 @@
 
 #include "../vg.hpp"
 #include <vg/io/stream.hpp>
+#include <vg/io/stream_multiplexer.hpp>
 #include <vg/io/vpkg.hpp>
 #include "../utility.hpp"
 #include "../chunker.hpp"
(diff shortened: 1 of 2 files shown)


adamnovak:
This doesn't actually work when I test it manually; the output files are empty. The `vg chunk` CLI tests don't catch that, so I will have to add some more.

adamnovak:
OK, I've fixed this not to drop reads when parsing untagged files, and added a test.

It's only about 20% faster than the existing code; it was a lot better than that when I was discarding all the reads. I guess the compression takes a lot of CPU and happens all in the main thread still.

adamnovak:
I'm now using `StreamMultiplexer` to let me do the compression in multiple threads. It should now be about as fast as `vg stats -a`, although a bit inefficient if you use a lot of threads. You only really need a few.

PR #3646, opened 2022-04-20
Path metadata support

adamnovak:
## Changelog Entry

 * vg now has metadata for paths; haplotype paths can now exist in places other than a GBWT. Try `vg paths -M -x <graph>` on a vg, GFA, or GBZ file to see it.
 * `vg convert` now has a `--ref-sample` option for promoting a haplotype to a reference during conversion.
 * `vg gbwt` and `vg index` no longer have `--paths-as-samples`; prepare input with paths of the appropriate senses instead.
 * vg will interpret W lines in input GFA files as haplotypes
 * vg will interpret W lines with `*` as the sample name as generic named paths.
 * vg will interpret PanSN P line names in input GFA files as haplotypes
 * vg will interpret PanSN P line names with `#<number>` after them as (pieces of) haplotypes
 * vg will interpret string-type `RS` tags in GFA headers as space-separated lists of samples whose paths are references, not haplotypes.
 * vg will generate GFA files using W lines for reference and haplotype paths, and containing `RS` header tags to distinguish them.

## Description

This uses https://github.com/jltsiren/gbwtgraph/pull/28 and https://github.com/vgteam/libbdsg/pull/153 and adds path metadata support.


adamnovak:
The CI tests look to be failing because I'm ending up with an empty xg that vg sim can't deal with. Maybe I actually broke something about xg in removing the GFAKluge dependency? I will have to see if I can replicate the problem locally and outside toil-vg.

PR #3658, opened 2022-05-06
Allow manipulating GBWT tags to efficiently change references in a GBWT or GBZ

adamnovak:
## Changelog Entry

 * `vg gbwt` can now manipulate GBWT tags and change reference samples in a GBWT or GBZ, as [documented on the wiki](https://github.com/vgteam/vg/wiki/Changing-References#large-gbz-and-gbwt-files-with-many-haplotypes).

## Description

This should help @StephenHwang with the long read Giraffe work by making it easy to change which samples in a GBWT or GBZ are references.

If you follow https://github.com/vgteam/vg/wiki/Changing-References#large-gbz-and-gbwt-files-with-many-haplotypes you can set different sample(s) to be reference samples vs. haplotypes without having to completely rebuild the GBWT or dump out its thousands of paths in a non-compressed format. You do still have to rewrite a new copy of the file, though.

### DIFF
--- a/src/subcommand/gbwt_main.cpp
+++ b/src/subcommand/gbwt_main.cpp
@@ -56,7 +56,7 @@ struct GBWTConfig {
     // Other parameters and flags.
     bool show_progress = false;
     bool count_threads = false;
-    bool metadata = false, contigs = false, haplotypes = false, samples = false, list_names = false, thread_names = false;
+    bool metadata = false, contigs = false, haplotypes = false, samples = false, list_names = false, thread_names = false, tags = false;
     bool include_named_paths = false;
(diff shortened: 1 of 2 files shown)


jltsiren (approved):
This looks reasonable.

I've been trying to keep https://github.com/vgteam/vg/wiki/VG-GBWT-Subcommand up to date, so the new options should be documented there (pipeline, step 3, step 7). It would also be good to mention which paths from the original GBWT are included in the subsampled GBWT (step 4) by default and the minimum VG version for that behavior.

adamnovak:
It looks like Jouni did most of the wiki editing already. I added a note about what paths the path cover will try to use right now.

PR #3663, opened 2022-05-14
autoindex will save XG file

StarSkyZheng:
autoindex will save XG file

## Changelog Entry

 * Whatsits now frobnicated

## Description
autoindex will save XG file


jeizenga:
It's admittedly a bit opaque, but this change can actually create mismatched indexes where the XG and GBZ files don't have the same node IDs. What pipeline do you need the XG for? It isn't required by `vg giraffe`. If you do need the XG, you can make a matching one with `vg convert -x`.

PR #3665, opened 2022-05-20
Enable capping number of minimizers and filtering of overlapping minimizers

StephenHwang:
## Changelog Entry

 * Enable capping the number of selected minimizers
 * Enable filtering of overlapping minimizers

## Description
For selecting a subset of minimizers to be used in long read giraffe. Added parameters and code to enable capping the number of selected minimizers. Added parameter to enable the filtering of overlapping minimizers.

### DIFF
--- a/src/minimizer_mapper.cpp
+++ b/src/minimizer_mapper.cpp
@@ -2874,6 +2874,11 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
         }
     }
 
+    // bit vector length of read to check for overlaps
+    size_t num_unique_min = 0;
+    size_t read_len = aln.sequence().size();
+    std::vector<bool> read_bit_vector (read_len, false);
+
     // Select the minimizers we use for seeds.
     size_t rejected_count = 0;
     std::vector<Seed> seeds;
@@ -2903,6 +2908,16 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
         // of the selected minimizers is not high enough.
         const Minimizer& minimizer = minimizers[i];
 
+        // minimizer information
+        size_t min_start_index = minimizer.forward_offset();
+        size_t min_len = minimizer.length;
+        bool overlapping = false;
+        if (this->exclude_overlapping_min) {
+          if (read_bit_vector[min_start_index] == true || read_bit_vector[min_start_index + min_len] == true) {
+            overlapping = true;
+          }
+        }
+
         if (minimizer.hits == 0) {
             // A minimizer with no hits can't go on.
             took_last = false;
@@ -2910,10 +2925,22 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
             if (this->track_provenance) {
                 funnel.fail("any-hits", i);
             }
-        } else if (minimizer.hits <= this->hit_cap ||
-            (run_hits <= this->hard_hit_cap && selected_score + minimizer.score <= target_score) ||
-            (took_last && i > start)) {
-            
+        } else if (  // passes reads
+              ((minimizer.hits <= this->hit_cap) ||
+              (run_hits <= this->hard_hit_cap && selected_score + minimizer.score <= target_score) ||
+              (took_last && i > start)) &&
+              (num_unique_min < this->max_unique_min) &&
+              (overlapping == false)
+            ) {
+
+            // set minimizer overlap as a reads
+            num_unique_min += 1;    // tracking number of minimizers selected
+            if (this->exclude_overlapping_min) {
+              for (size_t i = min_start_index; i < min_start_index + min_len; i++) {
+                read_bit_vector[i] = true;
+              }
+            }
+
             // We should keep this minimizer instance because it is
             // sufficiently rare, or we want it to make target_score, or it is
             // the same sequence as the previous minimizer which we also took.
--- a/src/minimizer_mapper.cpp
+++ b/src/minimizer_mapper.cpp
@@ -2875,7 +2875,7 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
     }
 
     // bit vector length of read to check for overlaps
-    size_t num_unique_min = 0;
+    size_t num_minimizers = 0;
     size_t read_len = aln.sequence().size();
     std::vector<bool> read_bit_vector (read_len, false);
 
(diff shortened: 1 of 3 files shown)


src/subcommand/giraffe_main.cpp
```diff
@@ -350,12 +350,14 @@ void help_giraffe(char** argv) {
     << "  -s, --cluster-score INT       only extend clusters if they are within INT of the best score [50]" << endl
     << "  -S, --pad-cluster-score INT   also extend clusters within INT of above threshold to get a second-best cluster [0]" << endl
     << "  -u, --cluster-coverage FLOAT  only extend clusters if they are within FLOAT of the best read coverage [0.3]" << endl
+    << "  -U, --max-min INT             use at maximum INT unique non-overlapping minimizers [1000]" << endl
```
adamnovak, line 353: When you don't pass `--exclude-overlapping-min`, this is just a limit on the maximum number of minimizers overall, right?

src/minimizer_mapper.hpp
```diff
@@ -94,6 +94,9 @@ class MinimizerMapper : public AlignerClient {
     /// of total score
     double minimizer_score_fraction = 0.9;
 
+    /// Take at maximum number of unique non-overlapping minimizers
```
adamnovak, line 97:
```suggestion
    /// Maximum number of distinct minimizers to take
```
"Take at maximum" doesn't make sense to me, and they aren't non-overlapping unless the flag is set later. And we should make it clear we mean unique as in different form each other, not as in single-hit.

src/subcommand/giraffe_main.cpp
adamnovak:
```suggestion
    << "  -U, --max-min INT             use at most INT minimizers [500]" << endl
```
The default in the code is 500 for the argument value.

src/minimizer_mapper.hpp
```diff
@@ -94,6 +94,9 @@ class MinimizerMapper : public AlignerClient {
...
+    size_t max_unique_min = 1000;
```
adamnovak, line 98:
```suggestion
    size_t max_unique_min = 500;
```
We probably should set this to match the command-line default.

src/subcommand/giraffe_main.cpp
```diff
@@ -399,7 +402,10 @@ int main_giraffe(int argc, char** argv) {
     Range<size_t> distance_limit = 200;
     Range<size_t> hit_cap = 10, hard_hit_cap = 500;
     Range<double> minimizer_score_fraction = 0.9;
+    Range<size_t> max_unique_min = 500;
```
adamnovak, line 405: Here's where the 500 default is set.

src/subcommand/giraffe_main.cpp
```diff
@@ -387,6 +389,7 @@ int main_giraffe(int argc, char** argv) {
     #define OPT_REF_PATHS 1010
     #define OPT_SHOW_WORK 1011
     #define OPT_NAMED_COORDINATES 1012
+    #define OPT_SHOW_EXCLUDE_OVERLAPPING_MIN 1013
```
adamnovak, line 392:
```suggestion
    #define OPT_EXCLUDE_OVERLAPPING_MIN 1013
```
There's no reason to put "SHOW" in here.

src/subcommand/giraffe_main.cpp
```diff
@@ -541,6 +548,8 @@ int main_giraffe(int argc, char** argv) {
             {"cluster-score", required_argument, 0, 's'},
             {"pad-cluster-score", required_argument, 0, 'S'},
             {"cluster-coverage", required_argument, 0, 'u'},
+            {"max-min", required_argument, 0, 'U'},
+            {"exclude-overlapping-min", no_argument, 0, OPT_SHOW_EXCLUDE_OVERLAPPING_MIN},
```
adamnovak, line 552:
```suggestion
            {"exclude-overlapping-min", no_argument, 0, OPT_EXCLUDE_OVERLAPPING_MIN},
```

src/subcommand/giraffe_main.cpp
```diff
@@ -856,6 +865,18 @@ int main_giraffe(int argc, char** argv) {
                     cluster_coverage = score;
                 }
                 break;
+
+            case 'U':
+                {
+                    auto maxmin = parse<Range<size_t>>(optarg);
+                    if (maxmin <= 0) {
```
adamnovak, line 872: I wonder if this will always quite work with the Range type, if someone specifies a range to values to try.

Maybe it's time to rip out the Range feature; we used it for like one grid search once and I don't think we use it anymore.

src/subcommand/giraffe_main.cpp
```diff
@@ -907,6 +928,10 @@ int main_giraffe(int argc, char** argv) {
                 }
                 break;
 
+            case OPT_SHOW_EXCLUDE_OVERLAPPING_MIN:
```
adamnovak, line 931:
```suggestion
            case OPT_EXCLUDE_OVERLAPPING_MIN:
```

adamnovak (approved):
This looks like it should work. I made a bunch of review suggestions to rename stuff, and I will apply those and merge it.

src/minimizer_mapper.cpp
```diff
@@ -2903,17 +2908,39 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
         // of the selected minimizers is not high enough.
         const Minimizer& minimizer = minimizers[i];
 
+        // minimizer information
```
adamnovak, line 2911: What "information", specifically, might want to be specified here.

src/minimizer_mapper.cpp
```diff
@@ -2903,17 +2908,39 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
...
             took_last = false;
             // We do not treat it as located for MAPQ capping purposes.
             if (this->track_provenance) {
                 funnel.fail("any-hits", i);
             }
-        } else if (minimizer.hits <= this->hit_cap ||
-            (run_hits <= this->hard_hit_cap && selected_score + minimizer.score <= target_score) ||
-            (took_last && i > start)) {
-            
+        } else if (  // passes reads
```
adamnovak, line 2928: I don't understand what "passes reads" means.

src/minimizer_mapper.cpp
```diff
@@ -2903,17 +2908,39 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
...
+              ((minimizer.hits <= this->hit_cap) ||
+              (run_hits <= this->hard_hit_cap && selected_score + minimizer.score <= target_score) ||
+              (took_last && i > start)) &&
+              (num_minimizers < this->max_unique_min) &&
+              (overlapping == false)
+            ) {
+
+            // set minimizer overlap as a reads
```
adamnovak, line 2936: I don't really understand what this means either; it's not quite a sentence I can parse.

Do you mean to explain how you are setting bits in the bit vector to track how the minimizers might overlap in the read?

PR #3708, opened 2022-07-25
Update Mac CI image

adamnovak:
## Changelog Entry

 * Mac CI now runs on MacOS 12
 * `uniform_int_distribution` unit tests no longer hang on MacOS 12

## Description

As noted in https://github.com/actions/virtual-environments#available-environments, the MacOS image we are using for Github Actions, `macos-10.15`, is deprecated. Github has started turning it off occasionally to annoy us enough to switch, and it is due to go away entirely sometime in August.

I tried `macos-11`, and ran into a Clang crash when building `aligner.cpp`, which I think we never fixed and just waited for Clang to fix it in the new Clang version that runs on MacOS 12.

On `macos-12`, I have the build successfully building, but running the unit tests appears to hang. I need to see if I can reproduce this hang on an interactive machine somewhere, so I can try to fix it.

### DIFF
--- a/.github/workflows/testmac.yml
+++ b/.github/workflows/testmac.yml
@@ -46,10 +46,10 @@ jobs:
         # We uninstall everything we don't need in order to prevent linking
         # conflicts with existing/outdated packages, which we can't resolve
         # because there's no way to tell Homebrew to force-link when installing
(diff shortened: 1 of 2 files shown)


adamnovak:
I've glommed #3709 into this PR to hopefully make Mac tests actually pass.

PR #3714, opened 2022-08-12
hack ref interval assertion fail into warning

glennhickey:
This changes a (very rare) assertion failure into a (cryptic) message, allowing HG005 to be run on the CHM13-based year-1 HPRC graph.  

This is probably not the idea fix, but is better than a crash.


adamnovak:
@glennhickey Should we close this because #3715 solved the problem? Or do you still want to merge this?

PR #3729, opened 2022-08-31
Make duplicate GFA paths a warning to fix #3726

### LINKED ISSUE #3726: vg chunk on a minigraph-CACTUS graph returns an error because vg can no longer accept a GFA with P lines and rGFA lines that describe the same path
My aim is to split the graph in chromosomes (ideally) so that I can visualize them in `Bandage`

I was expecting the following command
```
vg chunk -x /g100_scratch/userexternal/mungaro0/pangenomes/2.human-pgV2/human-pgV2.gfa -t 48 > /g100_scratch/userexternal/mungaro0/human-pgV2_chunk.vg
```
to use `vg chunk` to divide the graph into major chromosomes to be visualized in a 2D view.
The tool returned an error related to possible file corruption. However, I checked both the integrity and consistency of the original _.gz_ output from `minigraph-CACTUS` as well as the _.gfa_ file I fed to `vg`. They both seem fine but this is the output for the error I get
```
error[register_loader_saver_gfa] GFA file /g100_scratch/userexternal/mungaro0/pangenomes/2.human-pgV2/human-pgV2.gfa is corrupt and cannot be loaded.
GFA format error: On pass 1: On line 53471402: Duplicate path GRCh38.chr1 would be created in graph
```

```
Place stacktrace here.
```

A `minigraph-CACTUS` graph built upon GRCh38 for five human individuals on which to run the `vg chunk` command.

```
vg version v1.42.0 "Obolo"
Compiled with g++ (Ubuntu 11.2.0-19ubuntu1) 11.2.0 on Linux
Linked against libstd++ 20220324
Built by ubuntu@ip-172-31-12-246
```

P.S. I'm not sure about the second line for the error..., it seems something is wrong with the name of the FASTA headers in the reference (which I changed to be more comprehensible); nonetheless, there are no headers with the same name. Hence, I'm not sure about the "Duplicate path" statement

adamnovak:
## Changelog Entry

 * vg main GFA loader can now handle HPRC-style GFAs where the same path exists as rGFA tags and P lines.

## Description

It turns out that we can't just reject GFA files that have the same path twice, because the HPRC release graphs do that (as do new Minigraph-Cactus graphs) and we need to support those.

So this downgrades the duplicate path error to a warning.

### DIFF
--- a/src/algorithms/gfa_to_handle.cpp
+++ b/src/algorithms/gfa_to_handle.cpp
@@ -154,7 +154,7 @@ static void add_path_listeners(GFAParser& parser, MutablePathMutableHandleGraph*
                                                                   subrange);
         if (graph->has_path(implied_path_name)) {
             // This is a duplicate.
-            throw GFAFormatError("Duplicate path " + implied_path_name + " would be created in graph");
+            throw GFADuplicatePathError(implied_path_name);
         }
         
         // Create the path.
@@ -246,7 +246,7 @@ static void add_path_listeners(GFAParser& parser, MutablePathMutableHandleGraph*
                                                                   assigned_subrange);
         if (graph->has_path(implied_path_name)) {
             // This is a duplicate.
-            throw GFAFormatError("Duplicate path " + implied_path_name + " would be created in graph");
+            throw GFADuplicatePathError(implied_path_name);
         }
         
         // Create the path.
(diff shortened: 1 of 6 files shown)


glennhickey:
Yeah, it's not a great situation but was kind of a stop-gap to let the graphs kind of work with gfatools even though they aren't full rgfa, while letting other people use the path lines and ignore the tags. 

But I don't think rGFA's going anywhere.  And we probably want to improve support by
* creating mc pangenomes with full rGFA coverage -- now it's just the reference path
* have fairly transparent surjecting to rGFA coordinates (could be used eg to run DV inside insertions rather than attempt impossible projections to ref)

At the same time, it would be nice if the graphs and haplotypes were readable as regular GFA (everything W-lines, I still plan on dropping the P/W mixing).

So the solution would be either to make two graphs, rGFA and GFA or to stick with 1 graph and let vg do it's best to figure it out...

PR #3734, opened 2022-09-09
Selecting the number of minimzers for Giraffe based on length of aligned read

StephenHwang:
## Changelog Entry

 * Add the flag `--num-bp-per-min` to adjust the number of selected minimizers based on the read length.

## Description
Add command line flag `--num-bp-per-min INT`
for calculating the number of minimizers to select based on `read_length / INT`. The actual number of minimizers selected is the maximum of this calculated value and the value specified by `--max-min INT`.


adamnovak:
I find the help text confusing, maybe because the interaction between this new flag and `--max-min` is not how I would think to do it. I would expect `--max-min` to be the largest possible cap, but really it's smallest possible cap, and then if your read is (by default) 500,000 bp you will start to be able to exceed `--max-min`, and after that there's no limit beyond read length.

But we probably shouldn't change it now that we have models in these terms. Maybe we can come up with better help text later.

PR #3736, opened 2022-09-11
Distance index 2

xchang1:
## Changelog Entry

 * Update distance index 2 to be more efficient and make clustering faster
 * Minimizers now have a payload with two ints
 * The index registry now defaults to building a new distance index for giraffe, but it will still work with the old version and it will still make minimizers for the old version if it's given an old distance index
 * DI2's will need to be rebuilt as well as minimizers that use the old distance index because the payload is bigger

## Description
This updates the distance index and makes the new version the default for giraffe (but not mpmap). 
Giraffe is almost as fast as with the old distance index for hgsvc and 1000gp graphs, and is faster for the hprc graph
```
graph/cache       dist size (GB)	min size (gb)	giraffe memory (GB)	speed single end	speed paired end
1000gp old	    19	                26	                68.3	                 4382.17	                4607.25
1000gp new	    11	                43	                62.7	                 4023.52                   4019.48

hgsvc old	    8.5	                14	                36.3	                 5805.82	                5408.39
hgsvc new	    1.5	                23	                35.2	                 5571.27	                4853.42

hprc old	    4.3	                26	                44.8	                         3494.98	                3442.74
hprc new	    1.9	                43	                59.4	                         4007.94	                3538.62
```


xchang1:
This is actually a bit slower since I merged vg/master, but I don't think the slow down came from anything in the distance index or clustering
ddd07b is before merging master, 413887 is after merging master, eebc21is this pr
```
commit  |  1000gp single |  1000gp paired |  hgsvc single |  hgsvc paired |  hprc single |  hprc paired
ddd07b  |  4142.25       |  4186.96       |  5814.54      |  5057.88      |  4084.02     |  3721.58
413887  |  3989.71       |  4052.68       |  5567.94      |  4861.54      |  4004.69     |  3560.97
eebc21  |  4023.52       |  4019.48       |  5571.27      |  4853.42      |  4007.94     |  3538.62
```

xchang1:
It's also a bit wrong if I run giraffe with the clustering check turned on but not by much I think

xchang1:
@adamnovak @jltsiren Could you look this over for me please? All the clustering code is in src/snarl_seed_clusterer

src/minimizer_mapper.cpp
```diff
@@ -82,11 +84,12 @@ struct seed_traits<NewSnarlSeedClusterer::Seed> {
     /// What minimizer index payload type should we use for decoding minimizer index payloads?
     using MIPayload = vg::MIPayload;
     /// What chain info should we keep around during clustering?
-    using chain_info_t = tuple<size_t, size_t, size_t, size_t, bool>;
+    using chain_info_t = tuple<size_t, size_t, size_t, size_t, bool, bool, bool, bool, size_t, size_t>;
```
jltsiren, line 87: This could be a struct with named fields.

jltsiren (commented):
I left a few comments, but I can't really say anything about the distance index / seed clustering code. There is so much code that it would take me weeks to understand it.

I also have a couple of higher-level comments:

* What happens when the user tries to load an old distance index? Can the loading code identify the old index and give an understandable error message?
* Once this PR is merged, https://github.com/vgteam/vg/wiki/Index-Types and https://github.com/vgteam/vg/wiki/Index-Construction should be updated in the wiki.

xchang1:
Thanks, Jouni!

src/snarl_seed_clusterer.hpp
```diff
@@ -381,59 +453,68 @@ class NewSnarlSeedClusterer {
...
-        //Chain children are tuples<net_handle, (child index, inf) or (seed read num, seed index), chain component, prefix sum>
+        //chain_range_start/end are iterators to the start (inclusive) and end (exclusive) of the
+        //range in chain_to_children representing children of this chain
+        //The range must be ordered
+        //
         //If the children of the chain are only seeds on nodes, then cluster as if it is a node
-        void cluster_one_chain(TreeState& tree_state, size_t chain_clusters_index, vector<tuple<net_handle_t, size_t, size_t, size_t, size_t>>& children_in_chain, bool only_seeds, bool is_top_level_chain) const;
+        void cluster_one_chain(TreeState& tree_state, size_t chain_clusters_index, 
+            const std::vector<ParentToChildMap::ParentChildValues>::iterator& chain_range_start,
+            const std::vector<ParentToChildMap::ParentChildValues>::iterator& chain_range_end,
```
adamnovak, line 496: I think this indicates that the `ParentToChildMap` isn't actually *for* mapping from a parent to its children. It is laid out organized by parent, but really it's holding scratch for the clustering problem, for all the children, which you are going to visit organized by parent.

src/snarl_seed_clusterer.hpp
```diff
@@ -381,59 +453,68 @@ class NewSnarlSeedClusterer {
...
 
         //Helper function for adding the next seed to the chain clusters
         void add_seed_to_chain_clusters(TreeState& tree_state, NodeClusters& chain_clusters,
-                                        std::tuple<net_handle_t, size_t, size_t, size_t, size_t>& last_child, net_handle_t& last_child_handle, 
+                                        ParentToChildMap::ParentChildValues& last_child,
                                         size_t& last_prefix_sum, size_t& last_length, size_t& last_chain_component_end, 
-                                        vector<pair<pair<size_t, size_t>, pair<size_t, size_t>>>& cluster_heads_to_add_again,
-                                        bool& found_first_node, vector<bool>& found_first_node_by_read,
-                                        tuple<net_handle_t, size_t, size_t, size_t, size_t>& current_child_indices, bool is_first_child, bool is_last_child,
+                                        vector<ClusterIndices>& cluster_heads_to_add_again,
```
adamnovak, line 503: This argument sounds very exciting and needs to be explained.

src/snarl_seed_clusterer.hpp
```diff
@@ -381,59 +453,68 @@ class NewSnarlSeedClusterer {
...
+                                        bool& found_first_node, pair<bool, bool>& found_first_node_by_read,
```
adamnovak, line 504: These look like output parameters, and output parameters should probably go last.

src/snarl_seed_clusterer.hpp
```diff
@@ -381,59 +453,68 @@ class NewSnarlSeedClusterer {
...
+                                        const ParentToChildMap::ParentChildValues& current_child, bool is_first_child, bool is_last_child,
```
adamnovak, line 505: I feel like this should really be some kind of identifier (`net_handle_t`? Reference to record for the child but with a better type name?) for the child, and then we should just ask something (the `TreeState`?) if it is first/last in its parent.

Also `last_child` above is really `prev_child`, right? Whereas here "last" is really rightmost?

src/snarl_seed_clusterer.hpp
```diff
@@ -444,15 +525,14 @@ class NewSnarlSeedClusterer {
         //SeedIndex is used to store the seeds
         //left offset is the distance from the left side of the structure
         template <typename SeedIndex>
```
adamnovak, line 527: Why do you need to support multiple types here?

src/snarl_seed_clusterer.hpp
```diff
@@ -444,15 +525,14 @@ class NewSnarlSeedClusterer {
...
 
         //Compare two children of the parent and combine their clusters, to create clusters in the parent
-        //This assumes that the first node hasn't been seen before but the second one has, so all of the
-        //first node's clusters get added to the parent but assume that all of the second ones are already
-        //part of the parent
-        //old_distances contains the distances for cluster heads in the children, 
-        //since the distances in tree_state.read_cluster_heads_to_distances will get updated
+        //child_distances contains the distances for cluster heads in the children, 
+        //since the distances in the seeds will get updated to be the distances in the parent
         //First child is true if this is the first time we see child_clusters1. If first_child is true and this is
```
adamnovak, line 536:
```suggestion
        //First child is true if this is the first time we see child_clusters1. If first_child is true and the parent is 
```

src/snarl_seed_clusterer.hpp
```diff
@@ -444,15 +525,14 @@ class NewSnarlSeedClusterer {
...
-        //This assumes that the first node hasn't been seen before but the second one has, so all of the
-        //first node's clusters get added to the parent but assume that all of the second ones are already
-        //part of the parent
-        //old_distances contains the distances for cluster heads in the children, 
-        //since the distances in tree_state.read_cluster_heads_to_distances will get updated
+        //child_distances contains the distances for cluster heads in the children, 
+        //since the distances in the seeds will get updated to be the distances in the parent
         //First child is true if this is the first time we see child_clusters1. If first_child is true and this is 
         //a snarl, then we need to update the snarl's distances to its parents
         void compare_and_combine_cluster_on_child_structures(TreeState& tree_state, NodeClusters& child_clusters1,
```
adamnovak, line 538: If the node clustering records knew if they were roots or whether they had their distances to parents filled in yet, you could drop the flag arguments.

adamnovak (commented):
I've once again gotten only through the header, but I'm out of time to work on this for now, unfortunately.

I think a lot of the names for the inner classes here (`NodeClusters`, `ParentToChildMap`, `ParentChildValues`, and *maybe* `TreeState`) are lies. Really it looks like most of these represent various kinds of scratch data for the clustering problem (and `TreeState` *is* the clustering problem, though it's also kind of the state of the tree). `ParentChildValues` is scratch state for all the children at the current level, which is sorted out so you can go through it in order within each parent. `ParentToChildMap` is just a container for those, or like scratch/an intermediate result for a particular level of the clustering problem. Then `TreeState` holds the scratch for each level and lets you swap the upper level down so you can go up the tree. And then `NodeClusters` is really per-snarl-tree-node scratch.

It looks like across these types there's a nearly complete forgery of the relevant subtree of the `SnarlDecomposition` that the `SnarlDistanceIndex` implements. It might make sense that you need a copy of just this subtree, because asking it what the children of something are and then subsetting down to just the ones that have any relevance to the clustering problem might be too slow. But the way in which the subtree is stored seems pretty odd; you're defining a completely new space of numbers for all the different snarl tree nodes, and working there instead of with `net_handle_t`s, and doing a lot of tupling things with their parents at the front and sorting them.

Is that really justified? It's complex, and it's not really clear to me that you are gaining a lot of speed from this, and it is clear that you are going to have to be doing a lot of sort copying and doing binary searches to find things when you could have been doing O(1) hash table lookups if you were willing to maintain lists of children.

What I would do would be something like:

`TreeState` -> `ClusteringProblem`, and also it gains most of the methods as instance methods. Maybe it keeps a pointer back to the `SnarlSeedClusterer` for grabbing configuration.
`NodeClusters` -> `SnarlTreeNodeProblem`, and then put them in a big hash table by net_handle_t.
`ParentChildValues` -> Just merge this into `SnarlTreeNodeProblem`, because each snarl tree node is one child of one parent. Might only get used for levels that are chains.
`ParentToChildMap::parent_to_children` -> Give each `SnarlTreeNodeProblem` a vector of some kind of index/pointer to its childrens' problems in order, if you can afford the allocations. If not, maybe make a linked list out of the `SnarlTreeNodeProblem`s, where each has something like a `next_sibling` index/pointer, and a `first_child` index/pointer. Or at least keep a sorted list of indexes/pointers, and not the scratch records themselves; copying a `ParentChildValues` from one vector slot to another involves copying at least 48 bytes, which I think might be worse than having to follow an index indirection?

Then `chain_to_children` and `parent_chain_to_children` could be pointers/indexes for the heads of linked lists, or they could be `vector<net_handle_t>` sorted by parent via lookup in the sort comparison function, or something.

I think then it would be clearer what the parts are doing, and that might make it easier to make the actual code (which I still have to look at) simpler and faster.

PR #3760, opened 2022-10-24
Statically link libvgio if we are statically linking Protobuf

adamnovak:
## Changelog Entry

 * vg should now only link one copy of Protobuf into the non-static build
 * vg should now build against newer libomp from Homebrew which is keg-only

## Description

Since `libvgio` has a `libvgio.so`, which transitively links against `libprotobuf.so`, but since `vg` statically links Protobuf, we were ending up with a static Protobuf in the `vg` binary and then another, dynamic Protobuf being loaded when `libvgio.so` was loaded, in non-static `vg` builds.

This would *almost* work, except somewhere between Protobuf 3.6.1 (on Ubuntu 20.04) and Protobuf 3.12.4 (on Ubuntu 22.04), this started breaking Protobuf reflection, which broke JSON IO in `vg`.

This PR fixes the problem by using the static `libvgio` in the `vg` build, just like we use the static `libprotobuf`.

It also adds a test specifically to make sure that Protobuf reflection is working.

### DIFF
--- a/.github/workflows/testmac.yml
+++ b/.github/workflows/testmac.yml
@@ -48,8 +48,9 @@ jobs:
         # because there's no way to tell Homebrew to force-link when installing
         # from a Brewfile. We also update Protobuf to make sure we have 3.21.3+
         # to avoid problems with ABI changes with/without -DNDEBUG.
+        # And we update libomp to make extra sure it will be picked up by the compiler.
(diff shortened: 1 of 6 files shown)


adamnovak:
Looks like this PR is working but Mac CI is borken on master due to not being able to find omp.h properly: https://github.com/vgteam/vg/actions/runs/3246144050/jobs/5491873608#step:6:1711

PR #3784, opened 2022-11-15
Extract every snarl in region with vg chunk -S

glennhickey:
## Changelog Entry

 * New `vg chunk` option `-S` can be used to extract every snarl that is fully contained in the the given path region (ie specified with `-p`).  This can be used instead of context steps `-c`.  The advantage is that `-S` will return everything inside the region and nothing outside the region (barring the start and end nodes), which helps with the problem of pulling out massive amounts of neighbouring regions when jacking up `-c` for complex subgraphs.  The disadvantage is that if the region specified contains only parts of snarls, the results will be a misleadingly simple graph.

### DIFF
--- a/src/chunker.cpp
+++ b/src/chunker.cpp
@@ -4,6 +4,7 @@
 #include "chunker.hpp"
 #include "algorithms/subgraph.hpp"
 #include "vg.hpp"
+#include "clip.hpp"
 
 //#define debug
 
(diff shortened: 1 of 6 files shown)


jmonlong:
Thanks for implementing that! I've always been annoyed at the huge flanks that context adds when trying to extract large insertions.

glennhickey:
Just don't try it directly on GBZ (it's super slow)! 

Which reminds me, I think that's easily fixed...

PR #3795, opened 2022-11-21
Run edge filter as well in vg clip

glennhickey:
## Changelog Entry

 * `vg clip -d` now removes *edges* not meeting the depth threshold, in addition to nodes
 * `vg paths -C` reports both undirected and directed cyclicity


adamnovak:
@glennhickey If this passes in time, should we include it in the release today? Or would you rather it merge later?

adamnovak:
I'm going to start on the release without this.

glennhickey:
`I'm going to start on the release without this.`

Noooo!

PR #3803, opened 2022-11-30
Make vg chunk and vg trace be able to get GBWT from input graph

adamnovak:
## Changelog Entry

 * `vg chunk` and `vg trace` can now get haplotypes from an input GBZ file

## Description

We need this fro the tube map to work well with a GBZ file.


adamnovak:
Here's a static Linux binary with this change: [vg.zip](https://github.com/vgteam/vg/files/10144715/vg.zip)

PR #3836, opened 2023-01-30
Allow local graph files in read simulation script

adamnovak:
## Changelog Entry

 * `scripts/make_pbsim_reads.sh` now works with local graph files in addition to S3 URLs
 * `scripts/lr_benchmark.sh` now downloads and uses CHM13 graphs


adamnovak:
When I run the benchmark on PacBio reads, I get:

```
OUT_DIR=./lr_benchmark_HG002_hifi WORK_DIR=./trash/work_hifi INPUT_READ_PATH=/public/groups/cgl/graph-genomes/anovak/data/hprc-lrgiraffe/reads/sim/hifi/HG002/HG002-sim-hifi-1000.gam scripts/lr_benchmark.sh
...
Correct reads: 922
Total reads: 1000
Reads with truth positions: 951
Correct fraction: .92200000000000000000 of all reads, .96950578338590956887 of reads with truth positions
```

If I use the default nanopore simulated input reads, I manage to crash vg:
```
error:[gssw] Stuck in main matrix!
vg: src/gssw.c:3572: gssw_alignment_trace_back_word: Assertion `0' failed.
````

It happens inside this left tail alignment:
```
warning[MinimizerMapper::find_chain_alignment]: Falling back to non-GBWT alignment of 47589 bp left tail against 28436835-124 in S477_680
```

I don't think this is related to the benchmark itself; there's a preexisting bug in GSSW somewhere.

adamnovak:
I had to aggressively limit the lookback again to make it not run for 24 hours on the nanopore set, and bail out on a lot of pinned alignments on suspicion of the DP matrices being too big. We might want to relax these constraints later but I think it makes sense to fill in the new distance measurement and chaining stuff first.

Issue #3886, opened 2023-03-15
vg autoindex not work in rpvg module


ld9866:
Dear developer:
When using the module of vg autoindex, we provided the required files as required, but we still kept being reminded that we were missing the required files. May I ask how to solve this problem?
Code:
vg autoindex --threads 10 --workflow mpmap --workflow rpvg --prefix vg_rpvg --ref-fasta Sscrofall.fa --vcf primates-pg.vcf.gz --gfa primates.gfa --tx-gff ref.gtf
Result:
[IndexRegistry]: Checking for phasing in VCF(s).
[IndexRegistry]: Checking for haplotype lines in GFA.
error:[vg autoindex] Input is not sufficient to create indexes
Inputs
        GTF/GFF
        Reference FASTA
        Reference GFA w/ Haplotypes
        VCF
are insufficient to create target index Haplotype-Transcript GBWT

jeizenga:
You've run across a missing feature in `vg autoindex` that I've been meaning to implement for a while now. We have external pipelines that can create spliced pangenome graphs and pantranscriptomes from GFA input, but I haven't imported that logic into `vg autoindex` yet. For the time being, `vg autoindex` only supports FASTA + VCF input, and the VCF inputs need to contain phased haplotypes. The FASTA and VCF that you provided would be sufficient, except that it seems that the VCF isn't phased.

ld9866:
Thank you for your prompt reply!
I still have some questions I'd like to ask you.
Firstly, we used the minigraph-cactus generated vcf file, I don't really understand how you say "phase" should work, I found pggb working well, but minigraph-cactus is not very good.
Secondly, We took code and compared a standard 5 G transcriptome data to the human pan-genome, which was built using pggb, but we found that it was very slow, it took 3 days to complete a transcriptome, and we used an AMD7763 processor, which is very fast in other ways, I wonder if there is any way to speed up this process.
Thank you!
Best Yours
# Map simulated RNA-seq reads using vg mpmap
vg mpmap -t 20 -x vg_rna.spliced.xg -g vg_rna.spliced.gcsa -d vg_rna.spliced.dist -f example_data/x_rna_1.fq -f example_data/x_rna_2.fq > mpmap.gamp

jeizenga:
[...]

PR #3889, opened 2023-03-16
Treat haploid VCFs as phased in vg autoindex

jeizenga:
## Changelog Entry

 * `vg autoindex` recognizes haploid VCFs as phased

## Description

Some people have been trying to use VCFs from minigraph-cactus in `vg autoindex`, which are often output as haploid VCFs (because the input haplotypes have not been grouped into diplotypes). Haploid VCFs are trivially phased, but the phase-checking code wasn't written to recognize this case.

Resolves https://github.com/vgteam/vg/issues/3886. Also mentioned in https://github.com/jonassibbesen/rpvg/issues/43.

### DIFF
--- a/src/index_registry.cpp
+++ b/src/index_registry.cpp
@@ -168,10 +168,10 @@ double approx_num_vars(const string& vcf_filename) {
     // estimate of the number of variants contained in a VCF
     if (is_gzipped(vcf_filename)) {
         // square root got a pretty good fit, for whatever reason
-        return 0.255293 * file_size / sqrt(num_samples);
+        return 0.255293 * file_size / sqrt(std::max(num_samples, (int64_t) 1));
     }
     else {
-        return 0.192505 * file_size / num_samples;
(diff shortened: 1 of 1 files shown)


adamnovak:
This is failing the same Giraffe tests on Mac and on Gitlab's Linux build:

```
#30 456.8 Test Summary Report
#30 456.8 -------------------
#30 456.8 t/50_vg_giraffe.t   (Wstat: 0 Tests: 45 Failed: 3)
#30 456.8   Failed tests:  43-45
```

adamnovak:
Those tests use autoindex with `test/1mb1kgp/z.vcf.gz`, which has no samples. We're looking at it and deciding it must be a phased haploid VCF because it doesn't have any non-haploid genotypes (because it has no genotypes, because it has no samples). And then we don't do something we need to do (probably generating haplotypes) for the tests to work.

jeizenga:
Thanks!

Issue #3890, opened 2023-03-17
vg surject behave unexpectedly when converting a PGGB .gam file


Overcraft90:
**1. What were you trying to do?**
After running `vg giraffe` on a PGGB pangenome graph, I obtained a GAM file which I need to convert to BAM in order for `DeepVariant` to perform variant calling. However, once I launched the _surject_ command as follow
```
vg surject -x pggb.snarl.removal.p99.chronly.xg -F pggb_snarl_removal_p99_paths_list.txt -i -b -t 16 full_output-HG002.gam > full_output-HG002.bam
```
I get prompted with this error

> error: couldn't identify a path corresponding to surjected read HISEQ1:18:H8VC6ADXX:1:2115:2713:22626

What I did to avert any possible other option/scenario was to run `vg giraffe` in BAM output mode with the below command
```
vg giraffe -p -f /media/scratch/7.PGGB_Giraffe-DV_samples/HG002_R1.fastq -f /media/scratch/7.PGGB_Giraffe-DV_samples/HG002_R2.fastq -x pggb.snarl.removal.p99.chronly.xg -H pggb_snarl_removal_p99.gbwt -g pggb_snarl_removal_p99.gg -d pggb_snarl_removal_p99.dist -m pggb_snarl_removal_p99.min -o BAM --ref-paths pggb_snarl_removal_p99_paths_list.txt -t 16 > test.bam
```
But again I got prompted a similar error just listing a different header from the dataset of reads I'm using for the HG002 sample in question:

> error: couldn't identify a path corresponding to surjected read HISEQ1:18:H8VC6ADXX:1:2115:8207:18345

**2. What did you want to happen?**
Ideally, I should get a viable BAM file out of this surjection step (or mapping i ftalking about `vg giraffe`)

**3. What actually happened?**
The process quickly aborted returning an error message related to specific headers in the FASTQ that couldn't be identified.

**4. If you got a line like `Stack trace path: /somewhere/on/your/computer/stacktrace.txt`, please copy-paste the contents of that file here:**

```
N/A
```

**5. What data and command can the vg dev team use to make the problem happen?**
Either the `vg surject`, as shown above, or `vg giraffe` in BAM output mode, as previously outlined.

**6. What does running `vg version` say?**

```
vg version v1.46.0-49-g163339e6a "Altamura"
Compiled with g++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0 on Linux
Linked against libstd++ 20220421
[...]

PR #3892, opened 2023-03-23
Improve Surject Validation and Error Reporting

adamnovak:
## Changelog Entry

 * `vg surject` will now enforce that the reads it is surjection actually were mapped against the graph you are surjecting against. Right now it checks node IDs and lengths. You can turn this off with `-V`/`--no-validate`.
 * `vg gbwt` now accepts a `-I`/`--gg-in` option, which lets you load a `.gg` file and a `.gbwt` file and combine them into a `.gbz` graph.
 * `vg validate` now accepts a `-A`/`--gam-only` option which will validate only the provided alignment's agreement with the graph, and not the graph itself.
 * The `vg surject`/`vg giraffe` `error: couldn't identify a path corresponding to surjected read` error message has been improved to dump more information about the offending read and path.
 * When selecting paths to surject to, a warning will now be printed if the user asks for a path with a `[]`-enclosed subrange at the end. The base path name without the `[]` subrange coordinates should usually be used instead, because that is the space in which the SAM/BAM output will have its coordinates specified.
 * The `vg surject` `Graph does not have a path named` error message should now no longer print pointer values, and is extended to explain a bit more about subpaths.

## Description

This makes a bunch of small improvements to the surjection pipeline to try and make it easier to use and harder to use wrong.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -2081,24 +2081,32 @@ void alignment_set_distance_to_correct(Alignment& aln, const map<string ,vector<
     }
 }
 
-bool alignment_is_valid(Alignment& aln, const HandleGraph* hgraph) {
+AlignmentValidity alignment_is_valid(Alignment& aln, const HandleGraph* hgraph) {
     for (size_t i = 0; i < aln.path().mapping_size(); ++i) {
         const Mapping& mapping = aln.path().mapping(i);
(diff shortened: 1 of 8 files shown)


adamnovak:
This will fix #3890 by making sure that reads are surjected to the right graphs.

PR #3903, opened 2023-04-03
Distance indexing bugfix

### LINKED ISSUE #3884: [VG Crashed] Constructing distance index for Giraffe

`autoindex` to generate `gbz`,`dist`,`min` files

get excepted files

```shell
[IndexRegistry]: Constructing GBWT from VG graph and phased VCF input.
[IndexRegistry]: Merging contig GBWTs.
[IndexRegistry]: Stripping allele paths from VG.
[IndexRegistry]: Constructing XG graph from VG graph.
[IndexRegistry]: Downsampling full GBWT.
[IndexRegistry]: Not enough haplotypes; augmenting the full GBWT instead.
[IndexRegistry] forked child 450645
[IndexRegistry]: Constructing GBZ.
[IndexRegistry]: Constructing distance index for Giraffe.
terminate called after throwing an instance of 'std::runtime_error'
  what():  error: trying to access a snarl tree node of the wrong type
ERROR: Signal 6 occurred. VG has crashed. Visit https://github.com/vgteam/vg/issues/new/choose to report a bug.
```

```
Place stacktrace here.
```

data: the vcf file and reference fasta file
command: `vg autoindex --prefix auto_ref_vcf --workflow giraffe --ref-fasta genome.chr.fa --vcf all.ins.vcf.gz --target-mem 300G --threads 16`

```
vg version v1.46.0 "Altamura"
Compiled with g++ (Ubuntu 10.3.0-1ubuntu1~20.04) 10.3.0 on Linux
Linked against libstd++ 20210408
Built by xian@octo
```

xchang1:
## Changelog Entry

 * Predict the bit width for distance indexing more accurately

## Description

This should fix #3884

### DIFF
--- a/src/snarl_distance_index.cpp
+++ b/src/snarl_distance_index.cpp
@@ -111,15 +111,19 @@ SnarlDistanceIndex::TemporaryDistanceIndex make_temporary_distance_index(
         pair<SnarlDistanceIndex::temp_record_t, size_t> chain_index = stack.back();
         stack.pop_back();
 
+#ifdef debug_distance_indexing
         assert(chain_index.first == SnarlDistanceIndex::TEMP_CHAIN);
+#endif
         SnarlDistanceIndex::TemporaryDistanceIndex::TemporaryChainRecord& temp_chain_record = temp_index.temp_chain_records.at(chain_index.second);
         nid_t node_id = graph->get_id(chain_end_handle);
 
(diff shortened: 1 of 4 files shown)


xchang1:
This should work now. There was a problem with the way that I was adding size_t's that made it overflow when the number exceeded 32 bits

Issue #3917, opened 2023-04-11
VG Giraffe indexing terminating with Warnings (complex insertions not canonicalizeable) [canonicalize]


kvaddad1:
Hi, 
I am trying to build VG index using giraffe pipeline on 1KGP_phase3 VCF files on selected samples. The indexing command is terminating after throwing a bunch of warnings as follows.
`Warning: insertion END and POS do not agree (complex insertions not canonicalizeable) [canonicalize]`

There was no error message as such, the command abruptly ends after showing these warnings and the indexing command was not successful. Can you please guide me on what could be the issue here.  I am using autoindex command as follows 
`vg autoindex --workflow giraffe -r ${ref} "${VCF_ARGS[@]}" -p parent --threads 32
`
and my vg version is `vg: variation graph tool, version v1.46.0 "Altamura"`
Please suggest. 
Thank you.

jeizenga:
Are you able to share the data that you used when you ran into the error?

kvaddad1:
Hi @jeizenga,
Thank you so much for your kind response. I am trying to run autoindex giraffe pipeline with GRCh38.fasta (GCA_no_alt_version) along with th chromosome vcf files repository available at 
`http://ftp.1000genomes.ebi.ac.uk/vol1/ftp/data_collections/1000G_2504_high_coverage/working/20220422_3202_phased_SNV_INDEL_SV/`

Please suggest. Thank you so much.

jeizenga:
And where did you get the HG38 version that you used as input?

kvaddad1:
Hi @jeizenga 
The fasta I tried with the above VCF files is `GCA_000001405.15_GRCh38_no_alt_analysis_set.fna` and 
available at 
`https://ftp.ncbi.nlm.nih.gov/genomes/all/GCA/000/001/405/GCA_000001405.15_GRCh38/seqs_for_alignment_pipelines.ucsc_ids/`

Please suggest, thank you so much for following up.

adamnovak:
I think this might be a duplicate of #3922 and thus also fixed by #3933, according to the log @jeizenga gave me, which ended with:

```
Inversion can start after 53964
Inversion can end by inverting 54062
Buffered final node becomes: 34933728
[...]

PR #3939, opened 2023-04-25
Fix slow construct on giant clumps

adamnovak:
## Changelog Entry

 * `vg construct` should now be faster when variants are extremely long and overlap each other

## Description

On top of #3934, this adds speed improvements that seem important for running the data from #3917 in a timely fashion. I'm now using an interval tree instead of a scan when looking for other variants' ref paths to put nodes in. This should make giant clumps not quadradic to construct as long as only a small constant number of variants actually overlap any given point on the linear reference.


adamnovak:
This seems to work, on Linux, when it builds.

PR #3982, opened 2023-06-07
Bump libvgio and C++ standard version to work with Homebrew's current Protobuf

adamnovak:
## Changelog Entry

 * vg now builds with C++17 on Mac, as required byu the version of Protobuf packaged in Homebrew
 * vg now deduplicates arguments from pkg-config, to limit command line length with Protobuf's 30-odd Abseil dependencies.

## Description

This should fix the Mac CI failures. It might interfere with the static builds actually using a static Protobuf.


adamnovak:
CI is failing because the new version of Protobuf is no longer letting us serialize infinities in the Giraffe annotations, because Protobuf can't round-trip them type-safely.

PR #3985, opened 2023-06-10
Add a warning to giraffe when using minimizers without distance hints

xchang1:
## Changelog Entry

 * Add a warning to giraffe when using minimizers without distance hints

## Description
Running giraffe with minimizers that don't include the distance hints in the payload is very slow. Warn users when this happens


xchang1:
Actually this is too slow and it would be too difficult to try to go through just a handful of minimizers or reads before mapping. I think it would be better to try to change the minimizer index to remember if it has distances or not

adamnovak:
@xchang1 Did you ever actually do the other solution you proposed here? Should we close this?

xchang1:
Whoops no, I don't think so and this looks pretty outdated so I"ll just close it

PR #3992, opened 2023-06-15
Test workflow docker image caching

adamnovak:
## Changelog Entry

 * CI test jobs now cache pulled Docker images

## Description

This tries to set up Gitlab caching for the Docker image storage in the test jobs. This will hopefully stop 4 test jobs from starting on a node and thrashing the disk by all trying to do docker pulls at the same time.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -98,6 +98,8 @@ local-build-test-job:
     - nvm use
     - npm ci
     - CI=true npm run test
+  after_script:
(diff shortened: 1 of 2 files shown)


adamnovak:
I think I need a step to delete all the Docker containers; caching all of /var/lib/docker might bring along the state of what containers exist (or they might just still exist from the test run), so I can't untag the image sometimes because it thinks containers are left behind.

Issue #4022, opened 2023-07-12
vg rna crashes while trying to create spliced graph


JoshLangman:
**1. What were you trying to do?**
I was attempting to use `vg rna` to construct a pantranscriptome using a `gff` file and `pg` file made with `vg construct`

**2. What did you want to happen?**
I wanted to produce a `gbwt` file to use in downstream analyses using `rpvg`

**3. What actually happened?**
The program crashes immediately after running `vg rna`

**4. If you got a line like `Stack trace path: /somewhere/on/your/computer/stacktrace.txt`, please copy-paste the contents of that file here:**

```
Crash report for vg v1.48.0 "Gallipoli"
Stack trace (most recent call last):
#6    Object "/home/jpl298/.conda/envs/pantranscritome/bin/vg", at 0x5f0e3d, in _start
#5    Object "/home/jpl298/.conda/envs/pantranscritome/bin/vg", at 0x1ed71af, in __libc_start_main
#4    Object "/home/jpl298/.conda/envs/pantranscritome/bin/vg", at 0x5c0ade, in main
#3    Object "/home/jpl298/.conda/envs/pantranscritome/bin/vg", at 0xd3143b, in vg::subcommand::Subcommand::operator()(int, char**) const
#2    Object "/home/jpl298/.conda/envs/pantranscritome/bin/vg", at 0xc2ce88, in main_rna(int, char**)
#1    Object "/home/jpl298/.conda/envs/pantranscritome/bin/vg", at 0x1e97298, in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::assign(char const*)
#0    Object "/home/jpl298/.conda/envs/pantranscritome/bin/vg", at 0x1f4caea, in __strlen_sse2
ERROR: Signal 11 occurred. VG has crashed. Visit https://github.com/vgteam/vg/issues/new/choose to report a bug.
Please include this entire error log in your bug report!
```

**5. What data and command can the vg dev team use to make the problem happen?**
gff file `genomic_chr8_truncated.gff`: 
```
##gff-version 3
NC_050103.1	BestRefSeq	gene	263720	266280	.	+	.	ID=gene-LOC100273199;Dbxref=GeneID:100273199;Name=LOC100273199;description=uncharacterized LOC100273199;gbkey=Gene;gene=LOC100273199;gene_biotype=protein_coding;gene_synonym=cl25415_1a,GRMZM6G040576
[...]

Issue #3999, opened 2023-06-21
Trying to make a spliced pangenome for RNAseq alignment


oakeley:
**1. What were you trying to do?**
I made a pangenome using cactus pangenome like this:
---
rm -rf js; source /da/cbt/GX/oakeled1/genomes/cyno/pangenome/cactus-bin-v2.6.0/cactus_env/bin/activate; cactus-pangeno
me ./js cyno.txt --outDir /da/cbt/GX/oakeled1/genomes/cyno/S-210316-00677/pangenome --outName cyno-pg --reference Maca
ca_fascicularis_6.0.109 --vcf --giraffe --gfa --gbz
---
This works fine as a reference for aligning Illumina reads to. But I would like to align RNAseq reads so I tried to make a spliced reference using either the GTF or GFF3 file and either the "cyno-pg.gbz" or "cyno-pg.d2.gbz" generated by cactus-pangenome.
I also tried with and without the --use-hap-ref flag. The most recent command I tried was (but all failed the same way):

---
source /da/cbt/GX/oakeled1/genomes/cyno/pangenome/cactus-bin-v2.6.0/cactus_env/bin/activate; vg rna -p -t 8 --transcripts /da/cbt/GX/oakeled1/genomes/cyno/S-210316-00677/GFF3/Macaca_fascicularis.Macaca_fascicularis_6.0.109.gff3 --gbz-format cyno-pg.gbz > cyno-spliced.pg
---

**2. What did you want to happen?**
I wanted a spliced pangenome graph to be made using the GFF3 or GTF annotations.

**3. What actually happened?**
"vg rna" failed

**4. If you got a line like `Stack trace path: /somewhere/on/your/computer/stacktrace.txt`, please copy-paste the contents of that file here:**
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```
Crash report for vg v1.48.0 "Gallipoli"
Stack trace (most recent call last):
#12   Object "/da/cbt/GX/oakeled1/genomes/cyno/pangenome/cactus-bin-v2.6.0/bin/vg", at 0x5f0e3d, in _start
#11   Object "/da/cbt/GX/oakeled1/genomes/cyno/pangenome/cactus-bin-v2.6.0/bin/vg", at 0x1ed71af, in __libc_start_main
#10   Object "/da/cbt/GX/oakeled1/genomes/cyno/pangenome/cactus-bin-v2.6.0/bin/vg", at 0x5c0ade, in main
[...]

Issue #3997, opened 2023-06-20
vg rna crash with GBZ graph


brettChapman:
Hi

I'm using vg version v1.40.0 "Suardi" packaged along with PGGB version 0.5.3.

I have ran Minigraph-cactus and generated a GBZ graph. The graph is clipped and filtered. I try and generate a spliced pangenome graph and I get this error:

```
vg rna -p --threads 32 --transcripts /data/pangenome_phase2/minigraph-cactus_splice_graph/pangenome.gff --use-hap-ref --gbz-format /data/pangenome_phase2/minigraph-cactus_splice_graph/barley-pg/barley-pg.gbz
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
ERROR: Signal 6 occurred. VG has crashed. Visit https://github.com/vgteam/vg/issues/new/choose to report a bug.
Stack trace path: /tmp/vg_crash_pyR3rA/stacktrace.txt
Please include the stack trace file in your bug report!
srun: error: node-9: task 0: Exited with exit code 134
```

The GFF is basically all genome GFF concatenated together and the chromosome names modified to match the names of paths in the graph.

Thanks for any help you can provide.

brettChapman:
Here is the stack trace. I've also tried by using the GFA and converting to a PG instead of using GBZ, and I get the same error.

```
Crash report for vg v1.40.0 "Suardi"
Stack trace (most recent call last):
#11   Object "/usr/local/bin/vg", at 0x5b95cd, in _start
#10   Object "/usr/local/bin/vg", at 0x1d5dc8f, in __libc_start_main
#9    Object "/usr/local/bin/vg", at 0x58d20e, in main
#8    Object "/usr/local/bin/vg", at 0xc7ee5b, in vg::subcommand::Subcommand::operator()(int, char**) const
#7    Object "/usr/local/bin/vg", at 0xc52e0b, in main_rna(int, char**)
#6    Object "/usr/local/bin/vg", at 0x58bb10, in std::__throw_logic_error(char const*)
#5    Object "/usr/local/bin/vg", at 0x1c99d08, in __cxa_throw
#4    Object "/usr/local/bin/vg", at 0x1c99ba6, in std::terminate()
#3    Object "/usr/local/bin/vg", at 0x1c99b3b, in __cxxabiv1::__terminate(void (*)())
[...]

PR #4023, opened 2023-07-14
Fix long option parsing in vg rna

jeizenga:
## Changelog Entry

 * `vg rna` can parse long options

## Description

I'm sort of amazed that this bug lasted this long, but `vg rna` has apparently never been able to parse long options correctly. Only the single-letter options worked correctly. Perhaps this worked on some systems/compilers?

Resolves https://github.com/vgteam/vg/issues/4022. I believe it also resolves https://github.com/vgteam/vg/issues/3999.


jonassibbesen:
Thank you for fixing this @jeizenga. I have always used the single letter options when running and testing `vg rna`, but I am still really surprised that I never noticed this. It should also fix #3997. I will add a comment about it to the wiki with the guide.

jeizenga:
Yeah, I had the same blind spot, lol.

Issue #4043, opened 2023-08-03
vg 1.50 -  Reference path must have a sample name


colindaven:
**1. What were you trying to do?**

Run  an internal nextflow pipeline which creates vg indices from a  odgi->GFA and maps to them with giraffe.
Odgi files were created in a previous pipeline with minigraph-cactus.

**2. What did you want to happen?**

Successful conversion and mapping, as previously with vg1.48 I believe

**3. What actually happened?**
`
    what():  Reference path must have a sample name`

Presumably the vg mod or more likely vg convert failed. I probably need to add sample path information, but how - by which vg command, and where ?

**4. If you got a line like `Stack trace path: /somewhere/on/your/computer/stacktrace.txt`, please copy-paste the contents of that file here:**

```

Error executing process > 'make_indices (AT_Col0.gfa)'

Caused by:
  Process `make_indices (AT_Col0.gfa)` terminated with an error exit status (134)

Command executed:

  vg mod -t 8 -X 1000 AT_Col0.gfa > AT_Col0.gfa.fixed.gfa
  vg convert -t 8 -x -g AT_Col0.gfa.fixed.gfa > AT_Col0.gfa.xg

  vg gbwt --num-threads 8 --gbz-format -G AT_Col0.gfa.fixed.gfa -g AT_Col0.gfa.gbz
  vg gbwt --num-threads 8 -g AT_Col0.gfa.gg  -G AT_Col0.gfa.fixed.gfa -o AT_Col0.gfa.gbwt   

  vg snarls -t 8 -T AT_Col0.gfa.xg > AT_Col0.gfa.snarls
  vg index -t 8 -j AT_Col0.gfa.dist AT_Col0.gfa.gbz
  vg minimizer -t 8 -d AT_Col0.gfa.dist AT_Col0.gfa.gbz -o AT_Col0.gfa.min

Command exit status:
  134

Command output:
  (empty)

Command error:
  terminate called after throwing an instance of 'std::runtime_error'
    what():  Reference path must have a sample name
  ━━━━━━━━━━━━━━━━━━━━
  Crash report for vg v1.50.0 "Monopoli"

```

**5. What data and command can the vg dev team use to make the problem happen?**

As above.

**6. What does running `vg version` say?**

```
vg version v1.50.0 "Monopoli"
Compiled with g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 on Linux
Linked against libstd++ 20230528
Built by hickey@boreale

```

[...]

PR #4045, opened 2023-08-04
Handle two-part contig names again and put them under test

adamnovak:
## Changelog Entry

 * `GRCh38#chr1` style path names in GFA P lines should now be parseable again

## Description

This will fix https://github.com/vgteam/vg/issues/4043 where path names like `GRCh38#chr1` were breaking GFA import. We were losing the sample name when parsing them, and then crashing with a `Reference path must have a sample name` error when trying to reassemble a string name from what we had parsed.

This also adds some tests so this shouldn't be able to break again.


adamnovak:
This uses https://github.com/vgteam/libhandlegraph/pull/94

PR #4059, opened 2023-08-23
Diploid sampling

jltsiren:
## Changelog Entry

 * Diploid mode for haplotype sampling: first select N haplotypes, then choose the best pair.

## Description

A quick implementation of diploid sampling in `vg haplotypes`. This mode, enabled with option `--diploid-sampling`, first generates the specified number of candidate haplotypes with the existing greedy algorithm. Then it considers all pairs of candidates and chooses the highest-scoring pair according to a simple +1/-1 scoring scheme:
* Present kmer: -1 / 0 / +1 for 0 / 1 / 2 copies in the haplotypes
* Heterozygous kmer: 0 / +1 / 0
* Absent kmer: +1 / 0 / -1

As usual, this process is repeated independently in each subchain.

Haplotype sampling now also tries to use reference contig names in the generated haplotypes. This requires rebuilding the haplotype information.

### DIFF
--- a/src/recombinator.cpp
+++ b/src/recombinator.cpp
@@ -920,6 +920,7 @@ void RecombinatorHaplotype::insert(gbwt::GBWTBuilder& builder) {
         builder.index.metadata.addSamples({ sample_name });
     }
 
+    // FIXME: Get the contig name from the graph if possible.
     std::string contig_name = "chain_" + std::to_string(this->chain);
     gbwt::size_type contig_id = builder.index.metadata.contig(contig_name);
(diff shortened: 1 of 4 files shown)


jltsiren:
I added a check that there is only one top-level chain in each graph component (see #4060). Right now, `vg haplotypes` refuses to generate the haplotype information if the check fails. I'm not sure if it's better to tell the user to simplify the graph (by removing tips?) or remove such components manually, maybe depending on the size of the component, or if we should skip such components.

PR #4065, opened 2023-08-31
Save paths for alts using plain variant ID (when possible)

ldenti:
Hi all,
I needed to store alt paths by variant ID (and not SHA1 hash). To do this without (hopefully) breaking anything, I added a new option to the CLI. I noticed that there is already a function to retrieve the variant ID (if present), so I used it. Any drawbacks? (I'm curious to know why by default it uses the `make_variant_id` and not the `get_or_make_variant_id` function)

If you think this can be useful, let me know how to improve it to make it mergeable.

Best,

### DIFF
--- a/src/constructor.cpp
+++ b/src/constructor.cpp
@@ -696,7 +696,7 @@ namespace vg {
 
                     // Name the variant and place it in the order that we'll
                     // actually construct nodes in (see utility.hpp)
-                    string variant_name = make_variant_id(*variant);
+                    string variant_name = sha1_variant_name ? make_variant_id(*variant) : get_or_make_variant_id(*variant);
                     if (variants_by_name.count(variant_name)) {
                         // Some VCFs may include multiple variants at the same
                         // position with the same ref and alt. We will only take the
--- a/src/constructor.hpp
+++ b/src/constructor.hpp
@@ -80,6 +80,7 @@ public:
     // _alt_6079b4a76d0ddd6b4b44aeb14d738509e266961c_0 and
     // _alt_6079b4a76d0ddd6b4b44aeb14d738509e266961c_1?
     bool alt_paths = false;
+    bool sha1_variant_name = true;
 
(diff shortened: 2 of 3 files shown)


adamnovak:
We can definitely add this. I think we have been always using the hashes on the theory that the variant IDs might not actually be unique across a set of files, or even worse might uniquely identify a variant but not a ref/alt designation or an allele numbering order for the variant. But if you know that's not going to happen, we ought to be able to use the stored IDs.

adamnovak:
I've pushed this over to a branch https://github.com/vgteam/vg/tree/plain-altid so I think CI will test it.

ldenti:
> But if you know that's not going to happen, we ought to be able to use the stored IDs.

Well, I'm not sure of that: I always try to make them unique.. But for this reason I believe that an optional arg is the best option.. The default behavior is still hashing the variants, but if a user needs to use plain IDs, he can but he must be aware that IDs must be unique. This shouldn't break anything in vg and will place the burden on the user that doesn't use default parameters (I can add this warning to the `--help`).. Don't know if this is good or not btw (in terms of vg project)

adamnovak (approved):
This looks good to me.

ldenti:
I've updated the `--help`.

adamnovak:
OK, I've merged this in.

But there are [several other places](https://github.com/search?q=repo%3Avgteam%2Fvg%20make_variant_id&type=code), besides in `vg construct`, where we unconditionally call `make_variant_id()` and expect to find the resulting path in the graph. The most important is probably [here, where we are making the haplotype paths for the GBWT](https://github.com/vgteam/vg/blob/c7d6bbb34b9e3c806f49005e0d0f0a760a89b916/src/haplotype_indexer.cpp#L172).

So unless we also add this option to the other subcommands, or we figure out how to make them look for the matching path by ID if they can't find it by hash, then graphs made with this new option won't really work with those other operations.

PR #4072, opened 2023-09-05
Restrict CI to resources available

adamnovak:
## Changelog Entry

 * vg CI expects only the allocated cores on the GItlab runners
 * vg CI Buildkit docker builds use the local Docker Hub mirror
 * vg production Docker build timeout is now twice as long

## Description

This should help fix problems like those that happened with merge commits e173d0a81c6aa0e5b93a863f19b73ee4200177dc and 647d9356aaab4ac7968cb03d4b75a3b2b2345e2c, where CI tests/the ARM Docker build time out.


adamnovak:
Separating out the cache directories by image does seem to make it not rebuild the `build` stage when building the `run` stage. Or at least it changed behavior when I made that change.

I am rerunning the job that failed when I deleted the Docker Hub cache, now that I have recreated the Docker Hub cache, and hopefully it will pass and we can merge this. (Thought I think the timeout gets ignored)

PR #4085, opened 2023-09-15
Add ability to list reference and haplotype paths

adamnovak:
## Changelog Entry

 * `vg paths` now supports `-H` for selecting haplotype paths and `-R` for selecting reference paths

## Description

When listing or removing paths with `vg paths`, we were able to ask for the generic paths, but not for other kinds of paths. Now we can ask for each of the three path types separately or in any combination. If none of the options are specified, we get all path types (the default).


netwon123:
It is equal to the function 'vg convert -H' in v1.49?

adamnovak:
I think the sense is reversed; `vg convert -H` is `vg convert --drop-haplotypes`, while `vg paths -H` is `vg paths --haplotype-paths`.

I think you can now use `vg paths` to get a graph without the haplotype paths in it, like `vg convert -H`, but you can also now use `vg paths` to e.g. get a list of the haplotype paths.

We might want to remove one of the two ways to do the same thing here?

PR #4121, opened 2023-10-17
Better error messages from vg haplotypes

jltsiren:
## Changelog Entry

 * Better error messages from `vg haplotypes`.

## Description

Better error messages from `vg haplotypes`, particularly when the KFF file or the haplotype information file cannot be opened. Also try to catch more exceptions and report the errors without the garbage surrounding uncaught exceptions.

### DIFF
--- a/src/kff.hpp
+++ b/src/kff.hpp
@@ -84,7 +84,7 @@ public:
     typedef gbwtgraph::Key64::value_type kmer_type;
 
     /// Creates a new reader for the given file. Throws `std::runtime_error` if the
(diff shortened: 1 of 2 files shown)


adamnovak:
It looks like CI failed when testing the current version of the tube map against vg. I think the problem is an intermittent test failure in the tube map, caused possibly by cross-talk between different tests, or by not using the testing library right.

I've slightly improved the tube map tests, and I haven't been able to trigger it locally. I've rerun the job here; hopefully it passes now.

PR #4156, opened 2023-11-11
Ensure that vg rna produces unique HST names

jeizenga:
## Changelog Entry

* `vg rna` no longer projects transcripts twice onto a reference given by `RS` tag in a GFA
* `vg rna` assigns unique names to twice-projected transcripts on cyclic haplotypes

## Description

There were a few overlapping bugs having to do with `vg rna`'s projection algorithm when applied with new haplotype naming conventions. First, we were projecting reference transcripts back onto the reference whenever the reference was given as a W line with an RS tag. Second, when a transcript projected multiple times to the same haplotype (which can happen in cyclic, copy number variant genes) they were given the same identifier, which violated `rpvg`'s assumptions downstream, leading to a crash.

### DIFF
--- a/src/subcommand/rna_main.cpp
+++ b/src/subcommand/rna_main.cpp
@@ -292,7 +292,7 @@ int32_t main_rna(int32_t argc, char** argv) {
 
         // Load GBZ file 
         unique_ptr<gbwtgraph::GBZ> gbz = vg::io::VPKG::load_one<gbwtgraph::GBZ>(graph_filename);
-
+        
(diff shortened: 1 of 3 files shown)


jeizenga:
Also, the development highlighted that some of the default VCF to GBWT behavior can be problematic for `vg rna`. Currently, haplotypes around unphased variants are extended into the region between variants from both sides. This leads to two haplotype blocks from the same haplotype covering the region, which can confuse `vg rna` into projecting onto the haplotype twice. This will only really be a problem in VCFs of quite low variant density, because the region between variants needs to cover a full transcript, so this probably won't show up in most use cases.

PR #4182, opened 2023-12-05
Zipcode tree

xchang1:
## Changelog Entry

 * Add non-dag snarl handling to zip trees


src/zip_code_tree.hpp
```diff
@@ -68,140 +74,96 @@ class ZipCodeTree {
...
-      These chains would be considered to be reversed in the zip code tree, so the order of the children
-      of the chain may be backwards relative to their order in the variation graph.
-      If a snarl is the child of a chain that is traversed backwards in the zip tree, then that snarl
-      and all its children are also traversed backwards.
+
+      For snarls that aren't dags (called cyclic snarls, even though they could have an inversion 
+      and no cycles), the zip tree should represent all possible paths that the read could take 
+      through the snarl. All seeds on the snarl are split up into "runs" of seeds on the same chain
+      that are "close" to each other. The runs are sorted and orientated by their read coordinate 
+      and each run is made into a separate child chain like normal. A run occur twice, once in
```
adamnovak, line 99:
```suggestion
      and each run is made into a separate child chain like normal. A run occurs twice, once in 
```

Are we doubling the run even if the non-dag snarl is a cyclic directed graph and not actually bidirected?

src/zip_code_tree.hpp
```diff
@@ -405,114 +426,231 @@ class ZipCodeForest {
...
+        // We build a forest of trees. A new tree is formed either when a new top-level chain is 
+        // found (or a slice of a top-level chain if it is far enough away from the previous thing 
+        // in the chain), or when part of a chain in a snarl is too far from everything else in the
+        // snarl. In the second case, the entire subtree is found before determining that it should 
+        // be a subtree, and then it is copied into a new zip_tree_t in the forest.
+        // So only one tree is actively being added to at a time.
+        // This keeps track of which is the active tree, as an index into trees
+        // Note that this can't be an actual pointer to the forest because the address may move if 
+        // the vectors get shifted around in memory.
+        size_t active_zip_tree_i;
```
adamnovak, line 526: Is the `_i` doing any useful work here?

src/zip_code_tree.hpp
```diff
@@ -405,114 +426,231 @@ class ZipCodeForest {
...
+        //For children of cyclic snarls, an entire chain may be duplicated in the opposite 
+        // orientation immediately after the first copy. In this case, when the second copy is 
+        // processed, the entire interval is already in the correct order, just reversed. 
+        //If this is_reverse_ordered true, then the interval is sorted in the reverse order, so it
+        // needs to be flipped before processing
+        bool is_reverse_ordered;
+
+        //After flipping a reverse-ordered interval, all of the child intervals will be sorted
+        //So remember if the interval doesn't need sorting
+        bool is_ordered;
```
adamnovak, line 623: If you really want to bit pack this maybe the flags should be together?

Also, I'm a little worried this type is getting away from being what its name says it is (a pair of an interval and an orientation?) and I kind of don't understand what it really is. It looks like maybe it is some kind of scratch space for doing bookkeeping about what range is covered by part of the in-progress zip tree and also how the things are laid out in that space and how they are ordered. Maybe it's really an `interval_state` or `interval_metadata` something.

src/zip_code_tree.hpp
```diff
@@ -544,96 +682,139 @@ class ZipCodeForest {
...
+    //////////////////   Functions for sorting and finding intervals of seeds along the snarl tree
+    /////////////////////////////////////////////////////////////////////////////////////////////
 
-        vector<size_t> seed_sort_order;
 
+    /// Sorts the given interval (which must contain seeds on the same snarl/chain/node at the given
+    /// depth)  Sorting is roughly linear along the top-level chains, in a topological-ish order in
+    /// snarls. Uses radix_sort_zipcodes and default_sort_zipcodes
+    void sort_one_interval(forest_growing_state_t& forest_state, 
+                           const interval_and_orientation_t& interval, size_t interval_depth) const;
```
adamnovak, line 694: Doesn't the interval already know its own depth?

src/zip_code_tree.hpp
```diff
@@ -691,4 +872,758 @@ struct iterator_traits<vg::ZipCodeTree::reverse_iterator>{
...
+    print_self(&seeds);
+    validate_zip_forest(distance_index, &seeds, distance_limit);
+    assert(forest_state.open_chains.empty());
+    assert(forest_state.open_intervals.empty());
+#endif
+
+}
+
+template<typename Minimizer>
+vector<ZipCodeForest::interval_and_orientation_t> ZipCodeForest::get_cyclic_snarl_intervals(
```
adamnovak, line 1247: This is another one of those big algorithm functions we always write. Would it be possible to break any of the `/******* *******/` pieces out into their own functions to split this up?

src/zip_code_tree.hpp
```diff
@@ -691,4 +872,758 @@ struct iterator_traits<vg::ZipCodeTree::reverse_iterator>{
...
+
+
+
+
+
+
+
+
+/// Implementations for the templated functions using Minimizers since the definition is in the minimizer_mapper
+//TODO: This really shouldn't be in the hpp file
```
adamnovak, line 897: Can this move to the CPP?

src/zip_code_tree.hpp
```diff
@@ -405,114 +426,231 @@ class ZipCodeForest {
     /// If a distance limit is given, then also partition the tree into subtrees that are
     /// farther than the distance_limit from each other
     /// Otherwise, the forest will just be connected components
-    /// If a distance limit is given, then distances larger than the distance limit are not
-    /// guaranteed to be accurate
-    void fill_in_forest(const vector<Seed>& all_seeds, const SnarlDistanceIndex& distance_index,
+    /// The gap_distance_limit is the limit for making runs of seeds in a cyclic snarl- it 
+    /// should be roughly the distance that the dynamic programming is willing to jump to 
+    /// connect two consecutive minimizers 
+    //TODO: I think the distance_limit should just be the same as the gap_distance_limit
```
adamnovak, line 432: I think having this line breaks up the doc comment.

src/zip_code_tree.hpp
```diff
@@ -691,4 +872,758 @@ struct iterator_traits<vg::ZipCodeTree::reverse_iterator>{
...
+    After sorting an interval at one depth, sub-intervals representing the children can be found.
+    
+    Intervals are stored in a stack. The algorithm starts with an interval for each child of the
+    root snarl. An interval is popped from the stack. Any incomplete snarls or chains that the
+    interval is not a child of must be completed. Then, the snarl or chain that the interval
+    represents is added to the zip tree, along with any relevant distances. Intervals representing 
+    the children of the snarl or chain are found and added to the stack. This repeats until the
+    stack is empty.
+
+    */
```
adamnovak, line 939: This is an important high-level overview of the whole zip code tree build process, but it's buried in this function. Maybe it wants to be in the header somewhere, before the definitions of all the functions that represent pieces of this algorithm?

src/zip_code_tree.hpp
```diff
@@ -691,4 +872,758 @@ struct iterator_traits<vg::ZipCodeTree::reverse_iterator>{
...
+    }
+
+    /*
+    Make a ZipCodeForest
+    Takes a vector of seeds and their minimizers and fills in the forest
+
+    The zip forest is made by sorting the seeds along chains/snarls, then adding each seed, 
+    snarl/chain boundary, and distance to zip_code_tree.
+
+    Sorting and tree-making is done at the same time, in a depth-first traversal of the snarl tree.
```
adamnovak, line 924:
```suggestion
    Sorting and tree-making are done at the same time, in a depth-first traversal of the snarl tree.
```

src/zip_code_tree.hpp
```diff
@@ -405,114 +426,231 @@ class ZipCodeForest {
...
+      components are saved as "intervals" that remember the range in the sort order that the seeds 
+      occur on. Each interval of seeds represents a root-level snarl or chain. Each interval is then
+      sorted to order the seeds along the snarl or chain, and new intervals are found representing 
+      ranges of seeds on the children.
+
+      Each snarl and chain is comprised of the start and end bounds, the children, and distances 
+      between children/bounds. So as each child is added, we will need to know what came before it
+      in the parent snarl/chain so that we can add the distances. We also need to remember the 
+      ancestors of each snarl and chain as we are building them, so that we can close each structure
+      properly. All of this information is stored in a forest_growing_state_t as the zip trees are
```
adamnovak, line 467: I wonder if instead of passing around a state to everything, it might be better to make some kind of builder class and have it own the state, to reduce the number of arguments to all the functions? Or does it need to be like this for testability?

A builder class might also help you break the build code out into its own file which might be clearer?

src/zip_code_tree.hpp
xchang1: It's just so I know that it's an index into the vector of trees, not the tree itself. What I really wanted was a `ZipCodeTree& active_zip_tree`, but since the trees are stored in a vector, the pointers can move around

src/zip_code_tree.hpp
xchang1: Yeah good point

src/zip_code_tree.hpp
xchang1: Maybe it would be better to move the `Minimizer` class (and maybe `Seed`, `Cluster`, etc) to a separate file?

src/zip_code_tree.hpp
adamnovak: I might say `_index`? Maybe just `active_index`? I guess `_i` means the same thing and is shorter but I don't know if we name anything else like that in the codebase.

src/zip_code_tree.hpp
adamnovak: I think you want to put the template declaration in the header, and the template definition in the CPP, but also explicitly instantiate it for the right minimizer type with a line in the CPP:

```
template void void ZipCodeForest::fill_in_forest<MinimizerMapper::Minimizer>(const vector<Seed>&, const VectorView<MinimizerMapper::Minimizer>&, const SnarlDistanceIndex&, size_t, size_t);
```

I think you can also leave the variable names in?

Then it knows to actually make code for that instantiation of the template in the `.o` file that comes from that CPP file, and then the linker is happy.

xchang1: Thanks! I could not figure out how to get it to work in the CPP file

src/zip_code_tree.hpp
```diff
@@ -691,4 +872,758 @@ struct iterator_traits<vg::ZipCodeTree::reverse_iterator>{
...
+    /*******************
+
+      For each child of the snarl, walk through the seeds and build runs of seeds that are close
+      For each seed, compare it to all other seeds found so far to see if they can be merged
+
+      *****************/
+
+
+    forward_list<run_t> all_runs;
+    vector<std::tuple<size_t, size_t, bool>> read_and_chain_values (snarl_interval.interval_end-snarl_interval.interval_start);
```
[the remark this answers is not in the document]
xchang1: They're actually the offsets in the read and the chain for each seed. I renamed it and explained better.

I think this is the most efficient way to do it, even though I'm copying the values later. I need to associate each value with a run, but I don't know which seed belongs to which run, and if I waited until I had the runs, I'd have to re-compute the values, which I think will be slower than just copying them.

PR #4220, opened 2024-01-26
Add notion of eligible reads to GAM comparison

adamnovak:
## Changelog Entry

 * `vg gamcompare` now keeps track of whether reads could have possibly been gotten correct, based on there being a truth position for them in the truth set, and excludes reads without a truth from plots and accuracy statistics.

## Description
This should help with cases where we want to simulate reads from a sample that doesn't 100% align along the reference we are using for coordinates for accuracy measurement. Reads from diverged parts of the sample that can't be given correct linear coordinates won't be used when assessing accuracy.

### DIFF
--- a/scripts/plot-pr.R
+++ b/scripts/plot-pr.R
@@ -11,7 +11,7 @@ require("ggrepel")
 
 # Read in the combined toil-vg stats.tsv, listing:
 # correct, mapq, aligner (really graph name), read name, count, eligible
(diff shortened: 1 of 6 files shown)


adamnovak:
I think e09fcc700156c0326929d3fcbd428dbed5d20583 will fix CI; it seems like it broke because the R we were using could no longer install the R packages we use for the plots.

PR #4229, opened 2024-02-23
Mostly consistently left align in surject

jeizenga:
## Changelog Entry

 * `vg surject` will left align relative to the forward strand most of the time.

## Description

`vg surject` will now left-align on the forward strand and right-align on the reverse strand of a node. The net effect is to produce more identical edits from reads mapped to opposite strands. There are still some weird cases when indels span node boundaries, but this should handle the vast majority of cases.


jeizenga:
Something is being held up on the reporting, but tests have passed. I'm merging without the report so I can make a release.

PR #4236, opened 2024-03-01
Annotate subpath

parsaeskandar:
## Description
Added the subpath support to the vg annotate.


adamnovak:
I have pulled this into the `annotate_subpath` branch in this repo for testing. @parsaeskandar I have also invited you to the vgteam organization, so you should be able to update that branch and keep it in sync with the branch this PR is from, in case you change the code.

PR #4334, opened 2024-07-08
Update version.cpp

rlorigro:
## Changelog Entry
- Update version.cpp

## Description
Make sure `Bologna` is in the release name list


src/version.cpp
```diff
@@ -101,7 +101,8 @@ const unordered_map<string, string> Version::codenames = {
     {"v1.55.0", "Bernolda"},
     {"v1.56.0", "Collalto"},
     {"v1.57.0", "Franchini"},
-    {"v1.58.0", "Cartari"}
+    {"v1.58.0", "Cartari"},
+    {"v1.69.0", "Bologna"}    // Reserved
```
jeizenga, line 105: Can confirm. Does say Bologna.

PR #4338, opened 2024-07-09
Make the default surjection max_anchors unlimited

### LINKED ISSUE #4336: Turn off surjection anchor limit by default
@jeizenga thinks that the surjection anchor count limit from https://github.com/vgteam/vg/pull/3920 should be off by default to get better default behavior on long reads.

adamnovak:
## Changelog Entry

 * `vg surject` now uses an unlimited `--max-anchors` by default

## Description
This should fix #4336 and make the defaults make more sense for long reads.

### DIFF
--- a/src/subcommand/surject_main.cpp
+++ b/src/subcommand/surject_main.cpp
@@ -28,7 +28,7 @@
 
 using namespace std;
 using namespace vg;
-using namespace vg::subcommand;
+using namespace vg::subcommand; 
 
 void help_surject(char** argv) {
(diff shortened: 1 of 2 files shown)


adamnovak:
@jeizenga When you were looking at lifting the anchor limit and changed the default in the header, were you testing with Giraffe-integrated surjection and not `vg surject`? Since it looks like this PR still changes the default in` vg surject` itself.

jeizenga:
Honestly, I didn't really plan to test these changes, but I think it makes sense to raise the limit in both.

Issue #4257, opened 2024-03-27
MacOS compile requires 


adadiehl:
Compilaltion on MacOS Sonoma 14.2 fails with the following error:
 
```error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: changing install names or rpaths can't be redone for: lib/libdeflate.dylib (for architecture arm64) because larger updated load commands do not fit (the program must be relinked, and you may need to use -headerpad or -headerpad_max_install_names)
make: *** [lib/libdeflate.dylib] Error 1
```
Adding the ```-headerpad``` and ```-headerpad_max_install_names``` to LD_LIB_FLAGS fails to resolve the issue, though it does make it much further through compilation, to eventually fail with the error:

```clang: error: unknown argument: '-headerpad'```

**Comment by stacy-l** [comment]
Managed to resolve this:

https://github.com/vgteam/vg/blob/963d8d9003884c6c25480215497701871dd4da89/Makefile#L658-L659

Drop in change, doesn't need to be reverted to finish building.
```
$(LIB_DIR)/libhandlegraph.a: $(LIBHANDLEGRAPH_DIR)/src/include/handlegraph/*.hpp $(LIBHANDLEGRAPH_DIR)/src/*.cpp
	+. ./source_me.sh && cd $(LIBHANDLEGRAPH_DIR) && rm -Rf build CMakeCache.txt CMakeFiles && mkdir build && cd build && CXXFLAGS="$(CXXFLAGS) $(CPPFLAGS)" cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_PREFIX=$(CWD) -DCMAKE_INSTALL_LIBDIR=lib .. && $(MAKE) $(FILTER) && $(MAKE) install
```

PR #4347, opened 2024-07-15
Remove extra `-headerpad` linker argument

adamnovak:
## Changelog Entry

 * vg mac builds no longer pass `-headerpad` without an argument to the linker.

## Description

According to the Clang 15 man page, `-headerpad` needs a size argument:

>     -headerpad size
>             Specifies the minimum space for future expansion of the load commands.  Only useful
>             if intend to run install_name_tool to alter the load commands later. Size is a
>             hexadecimal number.
>
>     -headerpad_max_install_names
>             Automatically adds space for future expansion of load commands such that all paths
>             could expand to MAXPATHLEN.  Only useful if intend to run install_name_tool to alter
>             the load commands later.
>

I don't think it does anything we need that `-headerpad_max_install_names` doesn't do. And it is causing build failures for some people (see <https://www.biostars.org/p/9598655/>) because it doesn't have an argument.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -216,8 +216,9 @@ ifeq ($(shell uname -s),Darwin)
     START_STATIC =
     END_STATIC =
 
-    # We need to use special flags to let us rename libraries
-    LD_RENAMEABLE_FLAGS = -Wl,-headerpad -Wl,-headerpad_max_install_names
+    # We need to use special flags to let us rename libraries.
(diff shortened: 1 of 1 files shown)


adamnovak:
This might be needed to *actually* fix https://github.com/vgteam/vg/issues/4257; we never worked out how to reproduce it, so it's hard to tell.

PR #4350, opened 2024-07-16
Turn surject's hard filter for long tails into a soft filter

jeizenga:
## Changelog Entry

 * `vg surject` is faster on long reads

## Description

Long tails will now be aligned up to some maximum length and then softclipped beyond that. The default value is 300 bp.


jeizenga:
Merging for now as a better alternative to what is currently in master, but still plan to do some more analysis to decide if this is the best approach.

PR #4363, opened 2024-07-31
Stop dumping GAM to stdout when reading >1k BED or GFF records

adamnovak:
## Changelog Entry

 * When reading more than 1000 BED or GFF records, vg will no longer dump the first records to standard output and forget about them

## Description
In https://github.com/vgteam/vg/commit/ae44fca61eeb17200c8768301c6c5c84ae3c9bd9 @jmonlong improved memory usage when converting large BED and GFF files to GAM, by adding points at which we flush the output buffer. But when I merged https://github.com/vgteam/vg/pull/4248 I didn't notice that that buffer isn't *always* an output buffer and is sometimes an internal collection of annotations to work with later.

So whenever we tried to load *and keep* a lot of annotations, we would dump them out in GAM chunks of 1000 records each and throw them away, whenever we hit 1000 loaded records.

This fixes that.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -1798,7 +1798,7 @@ void normalize_alignment(Alignment& alignment) {
     if (doing_normalization) {
         // we found things we needed to normalize away, so we must have built the normalized
         // path, now replace the original with it
-        *alignment.mutable_path() = move(normalized);
+        *alignment.mutable_path() = std::move(normalized);
     }
 }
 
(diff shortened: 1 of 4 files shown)


jmonlong:
Sorry about the bug I introduced @adamnovak. I didn't notice that region-parsing function was used somewhere else...

I noticed the issue this week when working to make `vg annotate` output GAF. In  #4364 I think we get both the fix and GAF output feature. If it looks good to you, could we merge this one (or add a GAF output mode to yours)?

PR #4364, opened 2024-08-01
Annotate writes GAF too and fix to unwanted GAM dumping to stdout

jmonlong:
## Changelog Entry

 * `vg annotate` can output alignments in GAF.
 * When reading more than 1000 BED or GFF records, vg will no longer dump the first records to standard output and forget about them.

## Description

In  #4363, @adamnovak is fixing the bug I introduced when flushing the output buffer when `vg annotate` output alignments. I happened to have also noticed and corrected this when I added a *GAF output* option to `vg annotate`. 

Basically, I added an optional AlignmentEmitter which, when provided, will be used to write the alignments. That emitter can be set to output either GAM or GAF. In `vg annotate`, we prepare an emitter and pass it to the function that parses the regions, and it outputs the alignments (flushing every 1000 alignments). In the rest of the code, where an emitter is not passed, it will return the vector of alignments and **not** do any GAM dumping and buffer flushing.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -1798,7 +1798,7 @@ void normalize_alignment(Alignment& alignment) {
     if (doing_normalization) {
         // we found things we needed to normalize away, so we must have built the normalized
         // path, now replace the original with it
-        *alignment.mutable_path() = move(normalized);
+        *alignment.mutable_path() = std::move(normalized);
     }
 }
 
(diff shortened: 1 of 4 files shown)


adamnovak:
I think we probably want to glom both of these together; we want to use the AlignmentEmitter, but I think it also makes the most sense for the reader to use a callback to keep the input task and the output task separate, instead of having output logic make its way in there in case the caller was hoping to do output.

jmonlong:
Sounds good, I think I understand your code so I can give it a try if you want.

adamnovak:
I already merged them and put it into #4363.

jmonlong:
Thanks @adamnovak !

jmonlong:
Closing this one then.

PR #4394, opened 2024-09-10
Fix build on Ubuntu 24.04.1 / GCC 13.2.0

glennhickey:
My video drivers bugged out and I ended up just re-installing Ubuntu from scratch instead of wasting more time trying to fix it.  This put me on Ubuntu 24.04.1 with GCC 13.2.0,  seeming to give rise to a situation I always try to avoid: me being the first developer to build on a new compiler release....

This PR fixes the usual issues like GCC  getting stricter about which standard headers you can expect to include others, but also one on `string command = argv[1]` error in `main.cpp` that I still don't understand (but smells like a gcc bug).  

Ideally I'll hold off merging until @yoheirosen merges [this pr](https://github.com/yoheirosen/sublinear-Li-Stephens/pull/11), to avoid switching to repo to my fork as done currently...

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -57,7 +57,7 @@
 	url = https://github.com/jltsiren/gbwt.git
 [submodule "deps/sublinear-Li-Stephens"]
 	path = deps/sublinear-Li-Stephens
-	url = https://github.com/glennhickey/sublinear-Li-Stephens.git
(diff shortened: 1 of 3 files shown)


yoheirosen:
Merged that pr!

Issue #4381, opened 2024-08-21
Throw an error "error: trying to access a snarl tree node of the wrong type" in `haplotypes`


wjwei-handsome:
**1. What were you trying to do?**

Haplotype sampling

First step is preprocessing the graph

**2. What did you want to happen?**

Successfully generate `sample.hapl` file

**3. What actually happened?**

```
Loading GBZ from test.gbz
Generating haplotype information
Guessing that distance index is test.dist
Loading distance index from test.dist
Building minimizer index
Built the minimizer index in 918.913 seconds
Guessing that r-index is test.ri
Loading r-index from test.ri
Determining construction jobs
Using contig name chr12 for chain 0
Partitioned 1 components into 1 jobs in 2.37036 seconds
Running 16 jobs in parallel
error: [job 0]: error: trying to access a snarl tree node of the wrong type

```

**4. If you got a line like `Stack trace path: /somewhere/on/your/computer/stacktrace.txt`, please copy-paste the contents of that file here:**

```
Place stacktrace here.
```

**5. What data and command can the vg dev team use to make the problem happen?**

`vg haplotypes -v3 -t16 -H test.hapl test.gbz`

**6. What does running `vg version` say?**

```
vg version v1.58.0 "Cartari"
Compiled with g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 on Linux
Linked against libstd++ 20230528
Built by anovak@courtyard.gi.ucsc.edu
```

Interestingly, I encountered this problem when building dist index before: https://github.com/vgteam/vg/issues/3884

So I guess, is there some problem in the distance index when doing haplotypes sampling?

Looking forward to your reply :)

jltsiren:
This sounds like an issue with the graph or the distance index. What kind of data do you have, how did you obtain/build the graph, and how did you obtain/build the distance index?

wjwei-handsome:
The graph is `clip.gfa` from the Minigraph-Cactus pipeline, and the distance index was built using the command `vg index -t 32 -j test.clip.dist test.clip.gbz`. 

Additionally, I ran vg haplotype on the graph for each chromosome and found that **only chr12** didn't work.

jltsiren:
Can you share the graph? I don't think we can figure this out without it.

wjwei-handsome:
Of course, thank you very much for your assistance, which will greatly help us！

I sent a Google drive link to your work mailbox (uscs.edu), please let me know if you have other data needs.

[...]

PR #4395, opened 2024-09-11
Looping distance bug

xchang1:
## Changelog Entry

 * Update bdsg to fix distance index bugs with looping chains and distance-less indexes
 * Add a topological sort of child ranks for snarls in the distance index

## Description
Also silence warnings about oversized snarls by default


xchang1:
This should fix #4381

PR #4430, opened 2024-10-28
Add an option to output a haplotype GBWT from vg rna with node IDs that match the spliced graph 

jeizenga:
## Changelog Entry

 * `vg rna` can now output a haplotypes GBWT with node IDs matching the spliced graph

## Description

This representation was already used internally, and now it is available as an output option from the command line.


jeizenga:
@adamnovak Does this need to pull in your fix to the Mac CI?

PR #4433, opened 2024-11-04
Faster GBWT construction from GAM/GAF

jltsiren:
## Changelog Entry

 * GBWT construction from a GAM/GAF file now uses parallel construction jobs.

## Description

If instructed to use m construction jobs, the new algorithm will partition the graph into approximately m jobs, but often a bit more. A separate `GBWTBuilder` will be created for each job. The algorithm reads the GAM/GAF file using m threads, and each read is sent to the appropriate builder. This obviously works better with shuffled reads than sorted reads.

Some additional changes:

* I got rid of the old path/thread terminology in `vg gbwt` and in some other places.
* GBWT construction options have been removed from `vg index`. They had likely diverged from the supported ones in `vg gbwt` anyway.
* SDSL, GCSA2, GBWT, and GBWTGraph have been updated.


adamnovak:
It looks like we need to make a corresponding change in `toil-vg` for the removal of `vg index --gbwt-name`, or else rip `toil-vg` out of CI and do something else for testing the pipelines.

Issue #4434, opened 2024-11-04
VG STATS: 2000 Properly Paired Inquiry


AbdelF6:
**1. What were you trying to do?**
VG stats on 3 files after aligning them using vg giraffe and the T2T-CHM13 reference.

**2. What did you want to happen?**
vg giraffe to use paired-end mapping and properly paired values to be accurate 

**3. What actually happened?**
Got an error for vg giraffe: 
warning[vg::giraffe]: Cannot cluster reads with a fragment distance smaller than read distance
warning[vg::giraffe]: Falling back on single-end mapping

For VG stats, had a properly paired value of 2000 for all 3 files
After removing duplicates form the files using samtools markdup, no errors about cluster reads came up. I was wondering what the 2000 value means (is this a default output for when single-end mapping occurs? Also, is there a command in vg that can allow duplicates to be removed/notify user of duplicates in file?

**4. If you got a line like `Stack trace path: /somewhere/on/your/computer/stacktrace.txt`, please copy-paste the contents of that file here:**

```
Place stacktrace here.
```

**5. What data and command can the vg dev team use to make the problem happen?**

**6. What does running `vg version` say?**
vg version v1.60.0 "Annicco"
```
Place vg version output here
```

adamnovak:
I think the feature request for a vg tool for marking/removing duplicates might be already tracked in https://github.com/vgteam/vg/issues/2662? The request is also mentioned in https://github.com/vgteam/vg/issues/3283. So people definitely want it.

Giraffe insists that your paired end fragment distribution (which you can set manually with `--fragment-mean` and `--fragment-stdev`) produce a maximum fragment size (of mean + 2 standard deviations) that is longer than the first read length + 50 bp, and also longer than the clustering `--distance-limit` (default 200 bp). If you have oddly long reads, or oddly short fragments (so that e.g. your reads are always sequencing the whole fragment), this might not be the case.

It sounds like the fragment length distribution, as learned from the non-deduplicated input reads, didn't satisfy this constraint, which is why Giraffe produced that `Cannot cluster reads with a fragment distance smaller than read distance` message. I'm not sure why that problem would go away when you removed duplicate read pairs from the input; maybe that also changed the order of reads in the file and made the statistics come out in a way Giraffe can handle? (Are you sure the correct pairs of reads are being fed in as pairs?)

[...]

PR #4438, opened 2024-11-06
Giraffe paired end fragment length

xchang1:
## Changelog Entry

 * Change giraffe's measurement of distance between pairs of reads to measure between the inner edges plus the lengths of the reads

## Description

This might address #4434


xchang1:
I haven't tested this yet

adamnovak:
@xchang1 Have you tested/will you test this?

xchang1:
Sorry, I forgot about this. I will test it

xchang1:
Ok I tested it and it's a bit worse. Weird. Maybe we just optimized for the underestimate.
But I don't really want to go hunting this down now and it looks like the issue that started this is quiet so I'm going to close this

Issue #4420, opened 2024-10-21
distance index slow to compute snarls


glennhickey:
I think I've found a graph where using the distance index without distances is much slower than the old snarl manager.  

This came up for me because I've started running path normalization `vg paths -n` as a part of the mc pipeline on the unclipped graphs.  And on this graph it's been running for more than a day.  I've had it in gdb for a few hours, and it's still on `vg::fill_in_distance_index()`.  

It also reproduces on the command line: `vg snarls` finishes in about 10 minutes but the `vg index` version's been going for hours.  

```
cd  /private/home/hickey/dev/work/path-norm/

# slow
vg index chr2.vg.gfaffixed.clip -j chr2.vg.gfaffixed.clip.dist --snarl-limit 0

# fast
vg snarls chr2.vg.gfaffixed.clip > chr2.vg.gfaffixed.clip.snarls
```

xchang1:
@glennhickey Do you still have this graph somewhere? I think I have a fix but I want to test it

glennhickey:
yeah it looks like I deleted that folder, but I'm 95% sure you can reproduce with this graph

```
/private/home/hickey/dev/work/hprc-v2-prerelease/hprc-v2.prereease-mc-chm13/hprc-v2.prereease-mc-chm13.chroms/chr2.vg
```

xchang1:
Thanks!

faithokamoto:
Seems this was handled at the time

PR #4446, opened 2024-11-14
Speed up building the distance index without distances

xchang1:
## Changelog Entry

 * speed up building the distance index without distances


xchang1:
This should help  #4420

xchang1:
I tested this and it should be fine to merge once the CI tests finish

adamnovak:
I think CI was never started since this came from an external repo. I've pulled 87df7fc95bfbbd2f18f56a429534b6f17bed7ead into `distanceless-distance-index` in the main vgteam repo so the tests should actually run now.

PR #4455, opened 2024-11-25
Autoindex zipcodes

xchang1:
Make the autoindexer and giraffe subcommand make minimizers and zipcodes


src/index_registry.cpp
```diff
@@ -4106,17 +4113,78 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                                                         IndexingParameters::minimizer_s :
                                                         IndexingParameters::minimizer_w,
                                                     IndexingParameters::use_bounded_syncmers);
+
+
+        // Find frequent kmers.
+        std::vector<gbwtgraph::Key64> frequent_kmers;
+        //TODO: maybe we want to add this too? I left it as the default
+        bool space_efficient_counting=false;
+        size_t threshold = 500;
```
adamnovak, line 4122: These probably should in fact be in IndexingParameters and not just locals.

src/index_registry.cpp
```diff
@@ -3846,7 +3848,9 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
         
         ifstream infile_gbz;
         init_in(infile_gbz, gbz_filename);
-        unique_ptr<gbwtgraph::GBZ> gbz = vg::io::VPKG::load_one<gbwtgraph::GBZ>(infile_gbz);
+        cerr << "LOad gbz from " << gbz_filename << endl;
+        unique_ptr<gbwtgraph::GBZ> gbz = vg::io::VPKG::load_one<gbwtgraph::GBZ>(gbz_filename);
+        cerr << "Finished loading gbz" << endl;
```
adamnovak, line 3853:
```suggestion
        if (IndexingParameters::verbosity != IndexingParameters::None) {
            cerr << "[IndexRegistry]: Loading gbz from " << gbz_filename << endl;
        }
        unique_ptr<gbwtgraph::GBZ> gbz = vg::io::VPKG::load_one<gbwtgraph::GBZ>(gbz_filename);
        if (IndexingParameters::verbosity != IndexingParameters::None) {
            cerr << "[IndexRegistry]: Finished loading gbz" << endl;
        }
```

adamnovak (changes requested):
If we anticipate people mapping both long-read and short-read data against the same stored reference I think we do have to keep the indexes straight from each other with different names, or else have some kind of error to stop you accidentally using a preexisting short-read index when you meant to be using long-read mode or visa-versa. But since we want to pick up k and w from the index itself, setting up the error would be hard.

Right now if you map with the Illumina preset against a GBZ and then map with the HiFi preset against the same GBZ, you will get different mapping results than if you mapped the HiFi data first, because you will generate an Illumina-parameters minimizer index and then pick it up for HiFi.

But the IndexRegistry system doesn't really understand numerically-parametrized index types.

Since we really only have two minimizer indexing modes (the short read one and the long read one), maybe we should split `"Minimizers"` into `"Short Read Minimizers"` and `"Long Read Minimizers"`. Then we can give them different extensions (`.withzip.min` for short and `.long.withzip.min` for long, or something), and have one of those lambda-function-based recipes so the indexing implementations share code.

src/subcommand/giraffe_main.cpp
```diff
@@ -1373,6 +1375,9 @@ int main_giraffe(int argc, char** argv) {
                         found->second.apply(*parser);
                     }
                 }
+                if (param_preset == std::string("hifi") || param_preset == std::string("r10")) {
```
adamnovak, line 1378: I don't think you really need to call the string constructor here.

adamnovak (approved):

PR #4461, opened 2024-11-29
Add man page

xchang1:
## Changelog Entry

 * Add a man page

## Description
@jmonlong and I made a man page for vg.
It can be automatically generated to stay up to date with changes to the CLI


doc/README.md
```diff
@@ -0,0 +1,25 @@
+# Automated markdown manpage
+
+Make a markdown document with the usage messages of (selected) `vg` subcommands. 
+Calls the `vg` command, so it will match the version available in the command line.
+
+```sh
+python3 vgmanmd.py > man.md
+```
+
+Then copy the markdown content to a wiki page, for example, or move the markdown page to somewhere in the vg repo?
```
adamnovak, line 10: There's a checkout of the wiki as a submodule, so this maybe should involve writing to `wiki/vg-manpage.md` and committing and pushing.

This could also become a Makefile target e.g. `make manpage`.

adamnovak (commented):
This looks super useful!

Should we as part of this cut the previous attempt at "man pages" like https://github.com/vgteam/vg/blob/36c322a4e0e6ad2939aed4f01c97bf6788905da3/doc/asciidoc/man/vg-test.adoc where we had individual hand-maintained AsciiDoc format files that we never finished?

Is this meant to ever get you a "real" man page (i.e. something you can read with `man` in its weird troff format)? It might be good to have a step that calls something like:
* https://github.com/sunaku/md2man
* https://rtomayko.github.io/ronn/ronn.1.html
* https://github.com/remarkjs/remark-man?tab=readme-ov-file#remark-man

Otherwise it's a little confusing to say that this is a "man page".

adamnovak:
Here's something on writing Markdown that Pandoc knows how to translate for `man`: https://gabmus.org/posts/man_pages_with_markdown_and_pandoc/

PR #4323, opened 2024-07-01
Long Read Giraffe

adamnovak:
## Changelog Entry
To be copied to the [draft changelog](https://github.com/vgteam/vg/wiki/Draft-Changelog) by merger:

 * `vg giraffe` now has `--parameter-preset hifi` and `--parameter_preset r10` for using a new chaining-based algorithm to map long reads. `--parameter-preset sr` uses the new algorithm for single-ended short reads; the old `--parameter-preset default` and `--parameter-preset fast` remain available with the old non-chaining algorithm.
 * `giraffe-facts.py` script now knows how to read GAM files internally and no longer needs JSON preprocessing.
 * Short read chaining preset renamed to `chaining-sr` and locked out for paired end inputs

## Description
This adds the much-anticipated Long Read Giraffe to mainline vg.

PR #4488, opened 2025-01-06
Ensafen short read chaining preset

adamnovak:
## Changelog Entry

 * Short read chaining preset renamed to `chaining-sr` and locked out for paired end inputs

## Description
This renames the "sr" preset so it is more clearly using chaining and not the recommended short read mode. It also stops you from accidentally using chaining presets or the chaining flag with paired-end inputs, which can't actually run on the chaining codepath.


adamnovak:
I'm making this against `lr-giraffe`, so we'd want to merge it before https://github.com/vgteam/vg/pull/4323

PR #4489, opened 2025-01-06
Graceful error on gff3.gz files

### LINKED ISSUE #4459: vg rna can't parse bgzipped GFF3

Create a pantranscriptome from a GBZ graph and bgzipped GFF3 annotation set, following the [transcriptomic analysis wiki](https://github.com/vgteam/vg/wiki/Transcriptomic-analyses).

```
vg rna --progress --transcripts annotations.gff3.gz --transcript-tag Parent --use-hap-ref --gbz-format input_graph.gbz > pantranscriptome.pg
```

(My exact command on the cluster was this:)

```
srun -c 16 --mem 40G --time 20:00 vg rna --progress --transcripts /private/groups/patenlab/fokamoto/lr-pan-rna/data/chm13v2.0_RefSeq_Liftoff_v5.2.chm13_prefix.gff3.gz --transcript-tag Parent --use-hap-ref --gbz-format /private/home/hickey/dev/work/hprc-v1.1-jul4/hprc-v1.1-mc-chm13/hprc-v1.1-mc-chm13.gbz > test.pg
```

Automatic detection & handling of the bgzipped annotation file, or at least a graceful error message along the lines of "`.gff3` file cannot be compressed".

When `vg rna` tried to parse the annotation file to add transcript splice junctions, it errored on the compressed data. Note that it fails on the first line of the file. That's the header `##gff-version 3`.

```
[vg rna] Parsing graph file ...
[vg rna] Converting graph format ...
[vg rna] Graph and GBWT index parsed in 328.684 seconds, 14.7488 GB
[vg rna] Adding transcript splice-junctions and exon boundaries to graph ...
        ERROR: Chromosome path "�BCw�Z]o�8}��" not found in graph or haplotypes index (line 1).
```

No stacktrace, but the output of `--progress` is in #3.

* [Annotation file](https://s3-us-west-2.amazonaws.com/human-pangenomics/T2T/CHM13/assemblies/annotation/chm13v2.0_RefSeq_Liftoff_v5.2.gff3.gz) - note that this uses the `chr1` naming system, and the graph uses `CHM13#0#chr1` system, so you have to edit the file before using it successfully. However, that's not necessary to reproduce the error, which occurs before `vg` has a chance to look at any chromosome names.
    On the cluster, the chromosome-edited file is `/private/groups/patenlab/fokamoto/lr-pan-rna/data/chm13v2.0_RefSeq_Liftoff_v5.2.chm13_prefix.gff3.gz`. The same command works with the non-gzipped version of the file, which is the same but without the `.gz` prefix.
* I used a GBZ from the HPRC pangenom v1.1, on the cluster at `/private/home/hickey/dev/work/hprc-v1.1-jul4/hprc-v1.1-mc-chm13/hprc-v1.1-mc-chm13.gbz`. I suspect this will occur for any valid `.gbz` graph.

I built `vg` from source with the following code:

```
git clone --recursive https://github.com/vgteam/vg.git
cd vg 
git checkout lr-giraffe
git checkout autoindex-zipcodes
git submodule update --init --recursive
srun -c 16 --mem=80G --time=00:30:00 make -j16
```

Running `vg version` gives:

```
vg version v1.61.0-1291-g1c97950d1 "Plodio"
Compiled with g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 on Linux
Linked against libstd++ 20230528
Using HTSlib headers 101990, library 1.19.1-29-g3cfe8769
Built by fokamoto@phoenix-17.prism
```

If you want to run my exact installation, y'all can find it in `/private/groups/patenlab/fokamoto/lr-pan-rna/vg` :)

faithokamoto:
## Changelog Entry

 * Since `vg rna` can't parse `.gz` files, have it give a useful error for those inputs (resolve #4459)

## Description

Currently, `.gz` files will be accepted for `vg rna`'s `--transcripts` or `--introns` flags, but when we actually try to parse them it fails. This change auto-detects such files via the `.gz` extension and gives a useful error message. This will have false positives and false negatives, since it relies on the filenames following convention, but actually reading the file and detecting compression is much more difficult.

### DIFF
--- a/src/subcommand/rna_main.cpp
+++ b/src/subcommand/rna_main.cpp
@@ -251,8 +251,8 @@ int32_t main_rna(int32_t argc, char** argv) {
         return 1;       
     }
 
-    // taken from https://stackoverflow.com/a/64794991/
-    for (const auto & f : {std::cref(transcript_filenames), std::cref(intron_filenames)}) {
+    for (const auto& filenames : {std::cref(transcript_filenames), std::cref(intron_filenames)}) {
+        // For each collection of filenames (see https://stackoverflow.com/a/64794991/)
 
         for (auto filename : f.get()) {
 
--- a/src/subcommand/rna_main.cpp
+++ b/src/subcommand/rna_main.cpp
@@ -21,6 +21,7 @@ using namespace std;
 using namespace vg;
 using namespace vg::subcommand;
 
+string GZ_SUFFIX = ".gz";
 
 void help_rna(char** argv) {
     cerr << "\nusage: " << argv[0] << " rna [options] graph.[vg|pg|hg|gbz] > splicing_graph.[vg|pg|hg]" << endl
@@ -92,7 +93,6 @@ int32_t main_rna(int32_t argc, char** argv) {
     string hap_gbwt_out_filename = "";
     int32_t num_threads = 1;
     bool show_progress = false;
-    string gz_suffix = ".gz";
 
     int32_t c;
     optind = 2;
@@ -254,11 +254,11 @@ int32_t main_rna(int32_t argc, char** argv) {
     for (const auto& filenames : {std::cref(transcript_filenames), std::cref(intron_filenames)}) {
         // For each collection of filenames (see https://stackoverflow.com/a/64794991/)
 
-        for (auto filename : f.get()) {
+        for (auto filename : filenames.get()) {
 
             // taken from https://stackoverflow.com/a/20446239/
-            if (filename.size() >= gz_suffix.size() &&
-                filename.compare(filename.size() - gz_suffix.size(), gz_suffix.size(), gz_suffix) == 0) {
+            if (filename.size() >= GZ_SUFFIX.size() &&
+                filename.compare(filename.size() - GZ_SUFFIX.size(), GZ_SUFFIX.size(), GZ_SUFFIX) == 0) {
 
                 cerr << "[vg rna] ERROR: Annotation file " << filename << " appears to be gzipped. Decompress it before use." << endl;    
                 return 1;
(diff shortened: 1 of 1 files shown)


faithokamoto:
This is meant to resolve my issue #4459

src/subcommand/rna_main.cpp
```diff
@@ -250,6 +251,21 @@ int32_t main_rna(int32_t argc, char** argv) {
         return 1;       
     }
 
+    // taken from https://stackoverflow.com/a/64794991/
+    for (const auto & f : {std::cref(transcript_filenames), std::cref(intron_filenames)}) {
```
adamnovak, line 255: This might be too structural to *need* to attribute, and "taken from" makes it sound like you've grabbed a whole snippet or block and not a technique. But it's also something I've never seen before, and the link is less useful for explaining what you mean than a comment and a good variable name.

I feel like the way I would write this wold be:
```suggestion
    for (const auto& filenames : {std::cref(transcript_filenames), std::cref(intron_filenames)}) {
        // For each collection of filenames (see https://stackoverflow.com/a/64794991/)
```

adamnovak (approved):

src/subcommand/rna_main.cpp
```diff
@@ -92,6 +92,7 @@ int32_t main_rna(int32_t argc, char** argv) {
     string hap_gbwt_out_filename = "";
     int32_t num_threads = 1;
     bool show_progress = false;
+    string gz_suffix = ".gz";
```
adamnovak, line 95: This might want to be `GZ_SUFFIX` and placed not right next to all the CLI parameters. Otherwise I expect there to be a CLI option to change it.

faithokamoto: I've done this

PR #4541, opened 2025-03-12
Fix subgraph extraction in Long Read Giraffe

adamnovak:
## Changelog Entry

 * Giraffe will no longer cut nodes wrong when anchoring an extracted subgraph's end on the reverse strand of the same node as its start
 * Giraffe will make fewer tips, by enforcing path length strictly for connecting subgraph extraction. Dagification will still produce tips that will still be trimmed.

## Description
This changes the way Giraffe extracts connecting subgraphs to fix some bugs, add some tests, and remove ways in which we could lose track of which nodes in the processed extracted subgraph are actually the anchoring ones.

I now think I know where tips in the graph are coming from, but I'm not currently ready to write a dagification algorithm that avoids generating them by construction. I've turned on strict path length checking in the connecting graph extraction, which cuts out a lot of tips that we didn't need to have.

I'm also replacing *more* `ChainAlignmentFailedError`s with actual fatal errors, for cases I really think should not happen. If we detect any of these, I don't understand what's going on and we shouldn't be shipping a paper until someone does.

### DIFF
--- a/src/algorithms/extract_connecting_graph.cpp
+++ b/src/algorithms/extract_connecting_graph.cpp
@@ -6,8 +6,9 @@
  
 #include "extract_connecting_graph.hpp"
 #include <structures/updateable_priority_queue.hpp>
+#include "../crash.hpp"
 
-//#define debug_vg_algorithms
+#define debug_vg_algorithms
 
 namespace vg {
 namespace algorithms {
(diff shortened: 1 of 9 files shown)


adamnovak:
I checked against commit 71428e on some simulated HiFi reads, and this didn't dramatically increase wrong reads and may have dramatically increased speed.

PR #4547, opened 2025-03-17
Multi-reference surjection flags

adamnovak:
## Changelog Entry

 * `vg giraffe` and similar tools now have a `--ref-name` or similar option to pick one reference assembly by name (`CHM13`) when a graph contains several.

## Description

This makes it less of a distributed map reduce function to pick a reference from a multi-reference-assembly graph.

You still might want a `.dict` if the references is only partial paths and you want the right lengths in your BAM header.


jltsiren:
Giraffe already has option `--set-reference`, which is used for selecting reference samples for haplotype sampling. It should probably be a synonym for `--ref-name` to avoid confusion.

adamnovak:
I looked at that a bit, and I'm not sure we can use just one option. What if you want to haplotype sample and keep all the references in the sampled graph, but then surject to one particular reference when writing BAM? They control separate steps.

I guess we could lock out non-single-reference haplotype sampling and make you always provide it, and make it a generic "the linear reference to use" option.

adamnovak:
@jltsiren I'm going to merge this like it is, but if you think it's safe to lock in to single-linear-reference haplotype sampling, we can do that and combine these options and make them synonyms.

jltsiren:
I think my perspective is that we don't need to support special use cases in `vg giraffe` itself. There is always the option to do manual sampling with `vg haplotypes` for more flexibility.

I'll look into this once we have determined the best sampling parameters for short/long reads.

PR #4549, opened 2025-03-19
Change `vgmanmd.py` to Handle Subcommand `--help` Failures

SauersML:
## Description

Modified the `vgmanmd.py` script to better handle cases where `vg` subcommands fail to process the `--help` option correctly (e.g., `vg annotate` reporting `'--help' requires an argument` or `vg construct` reporting `unrecognized option '--help'`).

Previously, the try/except block did not trigger fallback behavior because subprocess.run without `check=True` doesn't raise `CalledProcessError` on non-zero exit codes.

This led to error messages appearing in the generated [manpage](https://github.com/vgteam/vg/wiki/vg-manpage). For example, ```vg: unrecognized option `--help'``` appears many times currently.

Adding `check=True` to the initial `subprocess.run` call, means that failures trigger the `except` block, which falls back to running the subcommand without `--help`.

### DIFF
--- a/doc/vgmanmd.py
+++ b/doc/vgmanmd.py
@@ -12,7 +12,11 @@ cmds.sort()
 
 
 # parse short descriptions
-desc_inf = open('./doc/vgmanmd.desc.md', 'rt')
+try:
+    desc_inf = open('./doc/vgmanmd.desc.md', 'rt')
+except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


adamnovak:
I've pushed this up to https://github.com/vgteam/vg/tree/fix-manpage-help so it should get tested.

PR #4550, opened 2025-03-19
Added .fq as a known Fasta format

parsaeskandar:
## Changelog Entry

 * Fixed the bug that giraffe failed with .fq/.fq.gz format.

### DIFF
--- a/src/subcommand/giraffe_main.cpp
+++ b/src/subcommand/giraffe_main.cpp
@@ -1510,7 +1510,7 @@ int main_giraffe(int argc, char** argv) {
         if (fasta_parts.second == "gz") {
             fasta_parts = split_ext(fasta_parts.first);
         }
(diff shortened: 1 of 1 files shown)


adamnovak:
I don't think this is right. This is changing the positional argument parsing to take a FASTQ where we would expect a FASTA *reference*. Do you really want to use a FASTQ as a reference and then map *to* a graph made from the FASTQ?

The right way to take a FASTQ is as a collection of reads to be mapped, and we already have a `-f`/`--fastq-in` option that takes an argument, for doing that:
https://github.com/vgteam/vg/blob/82ae7eda4206c3662eee3028b53f23c08e893614/src/subcommand/giraffe_main.cpp#L1157

So if you want to feed in a FASTQ of reads to map, you say `-f file.fq`.

PR #4569, opened 2025-04-19
Fix CRAM surjection

adamnovak:
## Changelog Entry

 * vg can now surject to CRAM. To use an external reference, you will need to pre-populate the reference cache (see `REF_PATH` and `REF_CACHE` documentation at https://www.htslib.org/workflow/cram.html).
 ~* vg will attempt to download CRAM references to the CRAM cache from the official CRAM reference archive server. This involves sending MD5 hashes of CRAM reference sequences to the server. Don't output CRAM against a secret reference without pre-populating the cache (see `REF_PATH` and `REF_CACHE` documentation at https://www.htslib.org/workflow/cram.html).~

## Description
This fixes CRAM output by not stopping if we have to go to no-reference mode, and also by tracking MD5 values through from dict files. It also checks the MD5 values against the actual graph paths and bails if they don't match.

To get CRAM checksums right, I had to turn off vg's custom multithreaded output multiplexing for CRAM, since that only applies to BGZF files, and CRAM isn't BGZF-compressed. Instead, we're now using htslib's multithreaded compression for CRAM files, with a static number of output threads.

vg is still not clever enough to dump your reference contigs to FASTA to use as the CRAM reference. I wanted to avoid creating a CRAM file against a reference that *only* exists in your graph and your local cache, since that file won't be interpretable on any other machine. If the reference isn't a standard one on the CRAM reference server, you will need to either make do with non-reference-based fallback CRAM files, or put it into a FASTA and then load that into your local CRAM reference cache with samtools's `seq_cache_populate.pl`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ RUN apt-get -qq -y update && \
 RUN apt-get -qq -y update && apt-get -qq -y upgrade && apt-get -qq -y install \
     make git build-essential protobuf-compiler libprotoc-dev libjansson-dev libbz2-dev \
     libncurses5-dev automake gettext autopoint libtool jq bsdmainutils bc rs parallel npm \
-    samtools curl unzip redland-utils librdf-dev cmake pkg-config wget gtk-doc-tools \
+    samtools curl libcurl-dev unzip redland-utils librdf-dev cmake pkg-config wget gtk-doc-tools \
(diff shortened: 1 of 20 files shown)


adamnovak:
I think I can't really do the reference download, because it requires bringing in libcurl, which has dependencies on Ubuntu that don't ship static libraries. So we'd have to [build our own libcurl without the dependencies enabled]( https://stackoverflow.com/a/32117592), which seems like a pain.

adamnovak:
I'm just subsuming this into #4585.

PR #4578, opened 2025-04-23
Allow cutting reads in `vg chunk`

adamnovak:
## Changelog Entry

 * Added `-u`/`--cut-alignment` option to `vg chunk` to allow cutting long alignments down to the portion(s) actually in the chunk.
 * Renamed GAM-related `vg chunk` options to be about generic alignments (with fallback aliases of the old names).

## Description
To debug surject, I wanted to be able to cut out the interesting parts of alignments I don't think are surjecting right, to make small example surjection tasks that exhibit an issue. This adds that feature to `vg chunk`.


adamnovak:
All the tests are passing except the report-generating job, which now can't run because the input artifacts expired. So I'm going to merge this.

PR #4579, opened 2025-04-24
add is_perfect option

faithokamoto:
## Changelog Entry

 * Add `is_perfect` option to `vg filter --tsv-out` to calculate whether a given alignment is perfect

## Description

`vg stats -a` has a statistic for whether an alignment is "perfect", defined as [having no non-match edits](https://github.com/vgteam/vg/blob/1d109aa7bec5034cc92cd8fdbec8644b352e1b35/src/subcommand/stats_main.cpp#L851). That is, it is aligned with no insertions (including soft clips), no deletions, and no substitutions. This PR simply copies the logic into a new function which takes an `Alignment` and spits out whether it is perfect. This allows easy granular investigation of specific reads which are or are not perfect. Perfection is calculated on-the-fly by `vg filter` whenever `is_perfect` is part of the fields requested by `--tsv-out`.

### DIFF
--- a/src/alignment.cpp
+++ b/src/alignment.cpp
@@ -1892,6 +1892,24 @@ int softclip_trim(Alignment& alignment) {
     return cut_start + cut_end;
 }
 
+bool is_perfect(const Alignment& alignment) {
+    // Non-aligned paths can't be perfect (code from subcommand/stats_main.cpp)
+    bool has_alignment = alignment.score() > 0 || alignment.path().mapping_size() > 0;
+    if (!has_alignment) return false;
+
+    // Check that the path is perfect
+    for (size_t i = 0; i < alignment.path().mapping_size(); ++i) {
+        auto& mapping = alignment.path().mapping(i);
+        for (size_t j = 0; j < mapping.edit_size(); ++j) {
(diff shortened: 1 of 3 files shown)


src/alignment.cpp
```diff
@@ -1892,6 +1892,24 @@ int softclip_trim(Alignment& alignment) {
...
+
+    // Check that the path is perfect
+    for (size_t i = 0; i < alignment.path().mapping_size(); ++i) {
+        auto& mapping = alignment.path().mapping(i);
+        for (size_t j = 0; j < mapping.edit_size(); ++j) {
+            auto& edit = mapping.edit(j);
+            if (!edit_is_match(edit)) {
+                return false;
+            }
+        }
```
adamnovak, line 1908: I think we have a function for this part already:
https://github.com/vgteam/vg/blob/0b8b7c2b415cd99b5abd6766265a28857b7c3099/src/path.hpp#L346-L347

```suggestion
        if (!mapping_is_match(mapping)) {
            return false;
        }
```

adamnovak (changes requested):
I think we can re-use some existing code more here.

adamnovak (approved):

Issue #480, opened 2016-09-20
overview of components of vg


ekg:
@vnsk asks:

> Is there any presentation or tutorial kind of thing on how the modules
work in vg?

@edawson has hugely cleaned up the wiki: https://github.com/vgteam/vg/wiki

Have you looked through that recently?

edawson:
Here: https://github.com/vgteam/vg/wiki/File-Formats
 and here : https://github.com/vgteam/vg/wiki/Basic-Operations

are especially helpful sections. I'm still working on improving the Quickstart to include more functionality and haven't begun to explain the data model in wiki format, but those two improvements will come in time.

@vnsk if you have any requests please post them and I will do my best to address them specifically!

vnsk:
Thanks !..will go through them.

PR #4584, opened 2025-04-29
Have autoindex add paths to spliced graph before saving

faithokamoto:
## Changelog Entry

 * Add transcript paths to spliced transcriptome graph before saving in autoindex

## Description

From debugging [SequenceTubeMap #480](https://github.com/vgteam/sequenceTubeMap/issues/480): autoindex was successfully adding transcript paths to the spliced graph, but only to the Transcriptome's `._transcript_paths` private field. We were neglecting to embed those paths into the Transcriptome's graph. Thus when the graph was saved, it didn't have the paths in it, since the Transcriptome kept that knowledge separate from its graph.

### DIFF
--- a/src/index_registry.cpp
+++ b/src/index_registry.cpp
@@ -3011,6 +3011,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
             }
             
             // save the graph with the transcript paths added
+            transcriptome.embed_transcript_paths(true, making_hsts);
             transcriptome.write_graph(&tx_graph_outfile);
             
             tx_graph_names[i] = tx_graph_name;


adamnovak (approved):
The fact that this mistake is possible suggests we might be able to improve the design. If there's 2 places we could be saving something, we're breaking the no forgery principle of how it's better to store one true copy.

faithokamoto:
I think `Transcriptome` separates "find paths" from "add paths" because those are separate options within `vg rna` . It's possible to, for example, ask for transcripts to be projected onto haplotypes (`-l haplotypes.gbwt -e`) but then _not_ ask for those transcripts to be embedded into the graph (fail to use `-a`). There are separate options for whether you want to embed reference transcripts `-r` and whether you want to embed haplotype transcripts `-a`. The only reason I see for such a separation is if you want to project transcripts and then only save the projected paths, instead of the reference paths. In most use cases it's just a user error to run `vg rna` without using one of the embedding options.

We could **add a warning to `vg rna`** if the user asks to parse transcripts and/or project transcripts but _not_ to embed those paths into the graph. Since I think `vg autoindex` is now more or less fine, that should solve the current user interface. Another option is to **change the transcript-parsing so that it embeds paths** after finding them, perhaps depending on a new parameter.

[Can discuss further in Slack if needed]

adamnovak:
We should probably make it an error to parse/project the transcripts but not actually send them anywhere (to a file or into the graph as embedded paths). Then the user would have to know where they are going and won't be able to say to make them but forget to put them anywhere.

PR #4589, opened 2025-05-05
Tsv cigar

faithokamoto:
## Changelog Entry

 * Add a CIGAR option to `--tsv-out`

## Description

I wanted to be able to output CIGARs to compare mappings against minimap2. The code here is patterned off of the [alignment logging](https://github.com/vgteam/vg/blob/61cfb60bb4963b9f378ebcf2babe240b1e0a6a33/src/minimizer_mapper.cpp#L100) in `minimizer_mapper.cpp`. Also I did minor tidying up to the `emit_tsv()` function in general (the later commits).

### DIFF
--- a/src/readfilter.hpp
+++ b/src/readfilter.hpp
@@ -1628,7 +1628,7 @@ inline void ReadFilter<Alignment>::emit_tsv(Alignment& read, std::ostream& out) 
                 }
             }
         } else {
-            cerr << endl << "Available fields: <https://github.com/vgteam/vg/wiki/Getting-alignment-statistics-with-%E2%80%90%E2%80%90tsv%E2%80%90out>" << endl;
+            cerr << endl << "Available fields: <https://github.com/vgteam/vg/wiki/Getting-alignment-statistics-with-vg-filter>" << endl;
             throw runtime_error("error: Writing non-existent field to tsv: " + field);
         }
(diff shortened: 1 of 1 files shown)


src/readfilter.hpp
```diff
@@ -1622,7 +1628,7 @@ inline void ReadFilter<Alignment>::emit_tsv(Alignment& read, std::ostream& out)
                 }
             }
         } else {
-            cerr << "I didn't implement all fields for tsv's so if I missed something let me know and I'll add it -Xian" << endl;
+            cerr << endl << "Available fields: <https://github.com/vgteam/vg/wiki/Getting-alignment-statistics-with-%E2%80%90%E2%80%90tsv%E2%80%90out>" << endl;
```
adamnovak, line 1631: We might be able to get away with the non-percent-encoded versions of these dash characters here. But then we have special dashes that look identical to normal dashes but aren't.

adamnovak: That might make more sense; right now the page is named after the option but the option only applies to one particular subcommand, which I thought was a little weird.

faithokamoto:
Hey Adam, can you re-run the checks on this? I really don't think changing the link (the commit in between passing and failing) would have broken anything. I think the checks timed out?

PR #4592, opened 2025-05-09
Pomote generic-sense paths to haplotypes

xchang1:
## Changelog Entry

 * Add option to `vg convert` to promote generic-sense paths to haplotype-sense paths

## Description
This adds option `vg convert --hap-sample` to convert generic-sense paths to haplotype-sense paths.
Since generic-sense paths only have a locus, the new haplotype-sense path will have that locus as its sample name and everything else defaults to 0.

I'm not sure if this makes sense to do, or if the way I formulated the new path is correct. I added it because I have a graph with only generic-sense paths and I want one of them to be the reference. I couldn't find another way to edit a path's metadata

### DIFF
--- a/src/subcommand/convert_main.cpp
+++ b/src/subcommand/convert_main.cpp
@@ -39,12 +39,14 @@ void help_convert(char** argv);
 void no_multiple_inputs(input_type input);
 // Generate an XG with nodes, edges, and paths from input.
 // Promote haplotype-sense paths for the samples in ref_samples to reference sense.
+// Promote generic-sense paths for the samples in hap_samples to haplotype sense.
 // Copy across other haplotype-sense paths if unless drop_haplotypes is true.
-void graph_to_xg_adjusting_paths(const PathHandleGraph* input, xg::XG* output, const std::unordered_set<std::string>& ref_samples, bool drop_haplotypes);
+void graph_to_xg_adjusting_paths(const PathHandleGraph* input, xg::XG* output, const std::unordered_set<std::string>& ref_samples, const std::unordered_set<std::string>& hap_samples, bool drop_haplotypes);
 // Copy paths from input to output.
 // Promote haplotype-sense paths for the samples in ref_samples to reference sense.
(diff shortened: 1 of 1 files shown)


src/subcommand/convert_main.cpp
```diff
@@ -469,6 +485,7 @@ void help_convert(char** argv) {
          << "    -r, --in-rgfa-rank N   import rgfa tags with rank <= N as paths [default=0]" << endl
          << "    -b, --gbwt-in FILE     input graph is a GBWTGraph using the GBWT in FILE" << endl
          << "        --ref-sample STR   change haplotypes for this sample to reference paths (may repeat)" << endl
+         << "        --hap-sample STR   change generic paths for this sample to haplotype paths (may repeat)" << endl
```
adamnovak, line 488: I don't think this (or the docs for the modified functions) captures what's really happening. What we're doing is taking generic paths and changing them to haplotype paths, using the generic path's name (in "locus" currently) as the new sample name, and making up a new locus name.

The wording for the option makes it sound like the user is specifying a sample name, but there's no such thing as "generic paths for this sample" because generic paths don't belong to samples. Really the user is specifying a locus/full generic path name, which gets transformed into a sample name in the output.

The option here is sort of structured as if it would do the opposite of `--ref-sample`, but it operates on generic paths instead, so it probably should be structured and described more distinctly.

adamnovak (changes requested):
Do we need to make each generic path into *its own* sample?

I think it would make more sense to have the operation be: take these generic paths, and change them to haplotype paths belonging to this newly-defined sample. The generic locus stays the locus, and a user-specified sample is added on. It might need 2 options and a map-from-generic-locus-to-assigned-sample-name as the input to the conversion functions, but I think it would be easier to explain and more likely to do what people will want.

xchang1:
Ok, I changed it so that you have to pass `--hap-locus` to specify the locus name of the generic path you want to change, and it requires `--new-sample` also, which is the sample name to give the new haplotypes.

PR #4605, opened 2025-05-15
Allow tracking either last correct or last possible alignment in giraffe-facts.py

faithokamoto:
## Changelog Entry

 * Add new option `--track-last` to track either last correct alignment (current/default) or last existing alignment in `scripts/giraffe-facts.py`

## Description

I wanted to track where I was losing mappings without having to set correct positions for each read. Using `--track-last any` finds the first stage with `"passed": {"count_total": 0}` but `"failed": {"count_total": X}` (X > 0) for each read. This should be the stage at which that read lost all of its remaining alignment possibilities, if I'm understanding the JSON correctly. I had to make some annoying hacks to pass the necessary parameter into `make_stats()`, but it works.

Also, I rewrote the argument parsing to have argparse find the arguments itself, instead of pre-parsing `sys.argv` for it.

### DIFF
--- a/scripts/giraffe-facts.py
+++ b/scripts/giraffe-facts.py
@@ -117,6 +117,9 @@ def parse_args():
                         help="show stage associated with each filter")
     parser.add_argument("--filter-help", "-f", action="store_true", default=False,
                         help="show documentation for each filter")
+    parser.add_argument("--track-last", choices=["correct", "any"], 
+                        default="correct", 
+                        help="count read as lost if it loses these alignments.")
     parser.add_argument("outdir",
                         help="directory to place output in")
     
@@ -295,7 +298,7 @@ def get_flat_key(struct, key, default=None):
 # Stats under NO_FILTER are not associated with a filter
 NO_FILTER = (-1, "__none__", "")
 
-def make_stats(read):
+def make_stats(read, track_last):
     """
     Given a read dict parsed from JSON, compute a stats OrderedDict for the read.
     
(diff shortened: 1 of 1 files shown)


faithokamoto:
Also note that this only affects the "Lost reads" column, not the "Lost" column. I didn't know what the latter was really doing so I left it alone. Could change it too if desired.

scripts/giraffe-facts.py
```diff
@@ -1403,12 +1425,13 @@ def main(args):
     read_lines = read_read_lines(options.vg, options.input, threads=max(1, options.threads // 2))
    
     # Map it to stats and reduce it to total stats
-    stats_total = map_reduce(read_lines, read_line_to_stats, add_in_stats, collections.OrderedDict, threads=max(1, options.threads // 2))
+    stats_total = map_reduce(read_lines, read_line_to_stats, add_in_stats, collections.OrderedDict, 
+                             options.track_last, threads=max(1, options.threads // 2))
```
adamnovak, line 1429: Instead of adding a `map_function_arg` to `map_reduce`, we should use [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial) to close over the configuration flag we want to pass to `read_line_to_stats` and make a single-argument function.

```suggestion
    stats_total = map_reduce(read_lines, functools.partial(read_line_to_stats, track_last=options.track_last), add_in_stats, collections.OrderedDict, threads=max(1, options.threads // 2))
```

To get out of having an extra parameter on `read_line_to_stats` to pass through, we could get rid of it and compose `read_line_to_read_dict` and `functools.partial(make_stats, track_last=options.track_last)` right here. And maybe store that in a variable here. But Python doesn't have a handy function composition widget.

adamnovak (changes requested):
This looks like it should work, but I want to keep the `map_reduce` function from getting complicated with the extra parameter, and I think we can do it with partials.

adamnovak:
It's been too long since the main jobs ran to run the report job, since the artifacts have expired. So I'm going to declare this to have passed CI and merge it.

Issue #4566, opened 2025-04-13
vg autoindex fails in surprising ways


ohell:
**1. What were you trying to do?**
Create a pantranscriptome from pansified FASTA and GTF

**2. What did you want to happen?**
Pantranscriptome to be created, or get helpful error messages that indicate why it can't be

**3. What actually happened?**
program stops with obtuse messages

**4. If you got a line like `Stack trace path: /somewhere/on/your/computer/stacktrace.txt`, please copy-paste the contents of that file here:**

```
Place stacktrace here.
```

**5. What data and command can the vg dev team use to make the problem happen?**
```
vg autoindex --workflow rpvg --gfa /data/ecoli-small.pggb/few-ecoli-genomes.smooth.final.gfa --ref-fasta /data/few-ecoli-genomes.fasta --tx-gff /data/final.gtf --hap-tx-gff /data/final.gtf --prefix /data/ecoli-small --threads 4
[IndexRegistry]: Checking for haplotype lines in GFA.
[vg autoindex] Guessing that /data/ecoli-small.spliced.xg is Spliced XG
error:[vg autoindex] Input is not sufficient to create indexes
Inputs
	GTF/GFF
	Haplotype GTF/GFF
	Reference FASTA
	Reference GFA
	Spliced XG
```
I created the gfa from 10 ecoli assemblies with contig names pansnified (only 1 contig per sample). I combined the accompanying GTFs using a script that only retains the contigs in the FASTA, pansnifies the contig names, replaaces blank transcript names with gene name and ensures transcript record fields are tab separated (last two were needed to make `vg autoindex --workflow mpmap` work).

**6. What does running `vg version` say?**

```
1.64.0
```

adamnovak:
It looks like you're missing the last line of the error message, where it reports what kind of index it couldn't figure out how to make:

https://github.com/vgteam/vg/blob/1d109aa7bec5034cc92cd8fdbec8644b352e1b35/src/index_registry.cpp#L5710-L5711

I think your problem is probably that the `rpvg` workflow needs haplotypes. Those should come from the GFA file, given how you say you built it from 10 assemblies. But autoindex is seeing it as a `"Reference GFA"` and not a `"Reference GFA w/ Haplotypes"`, which is what it would need to see in order to use haplotypes from it for pantranscriptime generation.

I think this is because it looked for haplotypes in it and didn't find any:

[...]

PR #4622, opened 2025-05-30
Stop using PathMetadata::parse_sense() to try and detect haplotype vs. reference from name alone

adamnovak:
## Changelog Entry

* Fix GFA haplotype sniffing for GFAs with P-lines
* Use graph metadata and not path name to determine reference/haplotype status for paths in `vg call` and `vg deconstruct`.
* Loading transcript files will now produce a human-readable error message when there are duplicate transcripts with the same ID on different paths.

## Description

This will hopefully help with #4566. I've stopped trying to use `PathMetadata::parse_sense()` to determine if a path is reference or haplotype from just its name. This fixes GFA haplotype sniffing in `vg autoindex` for GFAs with P-lines. Probably we should deprecate that function, or just make it tell generic from non-generic.

This doesn't fully solve #4566 though. Now when I run:

```
vg autoindex --workflow rpvg --gfa few-ecoli-genomes.fasta.gz.bf3285f.11fba48.42e55e5.smooth.final.gfa --tx-gff final.gtf --hap-tx-gff final.gtf --prefix ecoli-small --threads 4 --verbosity 2
```

I get an assertion failure:

```
[IndexRegistry]: Constructing haplotype-transcript GBWT and spliced graph from GBZ-format graph.
Assertion failed: (transcript->is_reverse == is_reverse), function parse_transcripts, file transcriptome.cpp, line 680.
```

### DIFF
--- a/src/index_registry.cpp
+++ b/src/index_registry.cpp
@@ -2098,7 +2098,6 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                                                 transcriptome.graph().max_node_id());
                 
                 // save the file
-                transcriptome.embed_transcript_paths(true, false);
                 transcriptome.write_graph(&outfile);
             }
             else {


adamnovak:
I've changed the problem I get with that input data to an error message: `ERROR: There are multiple distinct transcripts with ID unassigned_transcript_198. We have already seen a transcript unassigned_transcript_198 on strand 1 of contig GCF_000005845#1#NC_000913.3 length 4294967295 but on line 18815 there is a transcript unassigned_transcript_198 on strand 0 of contig GCF_000008865#1#NC_002695.2 length 4294967295. Fix the input transcript set to not re-use transcript IDs.`

@faithokamoto do you have insight into what they did in their GTF file? I guess we expect them to uniquely name transcripts across all input haplotypes?

faithokamoto:
Of course that prefixing could also be done by the user outside of vg, which is probably better?

faithokamoto:
Got rid of one of the changes, hopefully that was the issue? The graph is still made as desired.

PR #4629, opened 2025-06-06
Add options to annotate HTSlib with graph-space alignment

jeizenga:
## Changelog Entry

 * `vg giraffe` and `vg surject` can annotate HTSlib output with graph-space alignment

## Description

Hi! This PR adds options to `vg giraffe` and `vg surject` that can add tags to HTSlib output with the alignment of the read in graph-space (minus the node IDs but allowing the graph sequence of the alignment to be reconstructed). The format is the same as the minimap2-style cs tag that we currently use for GAF output.


jeizenga:
Not hearing any objections, I'm just gonna go ahead and merge this

PR #4640, opened 2025-06-24
Fix Giraffe correctness-tracking crash from off-path refpos

adamnovak:
## Changelog Entry

 * `vg giraffe --track-correctness` should no longer crash when read truth positions are on paths that exist in the graph, but are too short to reach where the read is.

## Description

This might fix @dcmonti's `--track-correctness` crash.


adamnovak:
Apparently this doesn't solve the underlying problem yet; Davide still reports crashes. But I have the files to reproduce them now.

adamnovak:
Davide reports that this fixes his issue now.

PR #4654, opened 2025-07-09
Test command line options

faithokamoto:
## Changelog Entry

 * Register command line options correctly & put them under test (`scripts/check_options.py`). This involved a _lot_ of minor bugfixes and helptext modifications, collected in a [Google Doc](https://docs.google.com/document/d/1V4wj4c7OPFfcZ4nsd5hnRWgR_lfSzfMXvVHjEhodAZo/edit?usp=sharing).
 * Manually wrap option helptext lines after 80 characters
 
## Description

This is what I've been working on during the week since I left on vacation. I had to deal with minor option errors one too many times (see #4648, #4649, etc.) and decided I would fix them once and for all. In addition, I've been minorly miffed at when too-long helptext exceeded the width of my (perfectly normal) terminal, and ended up wrapping around to under the list of options instead of staying in the description column. Thus, my quality-of-life PR to end all quality-of-life PRs.

The script itself explains what it does, so I'll just quote that here:

https://github.com/vgteam/vg/blob/a94eea53c2b1be4ae5142babf0052bd3c26429d4/scripts/check_options.py#L1-L61

To bring the subcommand files into compliance with `scripts/check_options.py`, I ended up editing—I wish this were a joke—literally every single file in `src/subcommand/*_main.cpp` except for the two I manually excepted (`test_main.cpp` & `help_main.cpp`) on account of them being exceptions. I don't _think_ I did anything with functional consequences except bugfixes. The majority of the edits were purely whitespace/formatting. However, as my six-page (abridged) changelog Google Doc attests, I did change an awful lot. For what it's worth the current commit passes all of the `vg test` suite.

Most of the commits here are just messing with the option-checking script itself. Only a few touch the subcommand files:
- Commit https://github.com/vgteam/vg/commit/9986d94b232c7c0164390ddff04a11af9914bb0c, which fixes bugs with command-line options (i.e. the original impetus for this project)
- Commit https://github.com/vgteam/vg/commit/5fcd40dbc962e4d52c7c236a73b690fede8ddb1a, which removes a duplicate options and deletes useless comments (this commit wasn't really supposed to be its own thing; it's an accident)
- Commit https://github.com/vgteam/vg/commit/1cbee96ec2f637979479d4dbc3d04fcc9f2c952b, which does whitespace/formatting elements to make the helptext line up in neat 80-char columns. Also breaks up some non-helptext lines (no functional changes, just extra newlines) to keep them from being too obscenely long.

I have taken pains to make sure that the script is well-documented. Ideally it should be accessible to other people. That includes both that they can understand what the script is complaining about, and that they can understand how the script works in case they need to update it themselves. Yet, no matter how long my docstrings are, this is indeed a 750-line script developed by a single hyperfocused person over a week. I've definitely missed something. Y'all are free to ping me about it.

@adamnovak I'd appreciate help figuring out how to add my option-checking script to the automated tests that run on PRs. Currently, it expects to be run from within the vg directory as `python scripts/check_options.py`. If it finds a problem it prints a summary; on a success it prints nothing. I'm open to changing any of that functionality because it's all just a wrapper for the all-important `check_file()` function.

### DIFF
--- a/scripts/check_options.py
+++ b/scripts/check_options.py
@@ -13,11 +13,15 @@ from typing import Dict, Optional
 from dataclasses import dataclass
 
 SUBCOMMAND_DIR = 'src/subcommand'
-# Some subcommand files have a different format
+"""Where to search for subcommand files."""
 SKIP_FILES = {'test_main.cpp', 'help_main.cpp'}
+"""Files to skip in the consistency check."""
 # Some giraffe arguments are processed outside of the switch block
 # (at least I think so? I'm not entirely sure)
 GIRAFFE_EXCEPTIONS = {'max-multimaps', 'batch-size'}
+"""Options for giraffe_main.cpp that are not in the switch block."""
+CRASHES = {'exit', 'return', 'deprecated', 'abort', 'throw'}
+"""Keywords that indicate a crash/deprecation in the switch block."""
 
 @dataclass
 class OptionInfo:
@@ -274,7 +278,7 @@ def extract_switch_optarg(text: str) -> Dict[str, Optional[bool]]:
     (i.e. nested curly braces are handled correctly)
 
     Ignores lines with comments. Handles fallthroughs,
-    breaking cases on either a return or a `break;`.
+    breaking cases on either a crash or a `break;`.
 
     Looks for whether each case uses `optarg` or not.
 
@@ -327,7 +331,7 @@ def extract_switch_optarg(text: str) -> Dict[str, Optional[bool]]:
 
         # Detect new case
         case_match = re.match(r'case\s+(.+)\s*:', stripped)
-        if case_match:
+        if case_match or stripped == 'default:':
             # Flush old cases to avoid optarg
             # being present after fallthrough
             if has_optarg:
@@ -335,7 +339,7 @@ def extract_switch_optarg(text: str) -> Dict[str, Optional[bool]]:
             
             # Remove comments
             stripped = stripped.split('//')[0].strip() 
-            case_value = case_match.group(1)
+            case_value = case_match.group(1) if case_match else 'default'
 
             if case_value.startswith("'") and case_value.endswith("'"):
                 # Single-character variable
@@ -350,8 +354,8 @@ def extract_switch_optarg(text: str) -> Dict[str, Optional[bool]]:
         has_optarg = has_optarg or 'optarg' in stripped
 
         # On crash/deprecation, flush the current case group
-        if (('return' in stripped or 'exit(' in stripped 
-             or 'deprecated' in stripped) and curly_brace_nesting == 0):
+        if (any(keyword in stripped for keyword in CRASHES)
+            and curly_brace_nesting == 0):
             # None is a marker for a crash
             set_optarg_usage(current_cases, None)
 
@@ -389,7 +393,7 @@ def check_file(filepath: str):
 
     all_longform = set(help_opts) | set(long_opts)
 
-    # Confirm that the help options are present
+    # Overall check 1: are help options present?
     for help_alias in ['?', 'h']:
         if help_alias not in getopt_opts:
             print(f"{filepath}: help alias -{help_alias} is missing from "
@@ -408,6 +412,16 @@ def check_file(filepath: str):
         # Change help alias to be non-argument instead of crash
         # in order to match long_options[]
         switch_opts[help_alias] = False
+    
+    # Overall check 2: does the default case crash?
+    if 'default' not in switch_opts:
+        print(f"{filepath}: switch block is missing a default case")
+    else:
+        if switch_opts['default'] is not None:
+            print(f"{filepath}: switch block's default case should crash")
+        
+        # Get rid of default once it's checked; will appear nowhere else
+        switch_opts.pop('default')
 
     # Check longform options between the four sources
     for longform in all_longform:
@@ -476,19 +490,19 @@ def check_file(filepath: str):
                     f"{should_str} use optarg")
             continue
     
-    # getopt is allowed to have ? as a help alias
+    # Overall check 3: getopt is allowed to have ?
     # without that being in long_options[]
     if '?' in getopt_opts:
         getopt_opts.pop('?')
     if getopt_opts:
-        print(f"{filepath}: getopt string has options not in long_options[]: "
+        print(f"{filepath}: getopt string has option(s) not in long_options[]: "
               f"{', '.join(getopt_opts)}")
     
-    # Similarly for switch block
+    # Overall check 4: similarly for switch block
     if '?' in switch_opts:
         switch_opts.pop('?')
     if switch_opts:
-        print(f"{filepath}: switch block has options not in long_options[]: "
+        print(f"{filepath}: switch block has option(s) not in long_options[]: "
               f"{', '.join(switch_opts)}")
 
 if __name__ == "__main__":
--- a/scripts/check_options.py
+++ b/scripts/check_options.py
@@ -333,6 +333,17 @@ def check_file(filepath: str) -> list:
 
     all_longform = set(help_opts) | set(long_opts)
 
+    # Confirm that the help options are present
+    for help_alias in ['?', 'h']:
+        if help_alias not in getopt_opts:
+            print(f"{filepath}: help alias -{help_alias} is missing from "
+                  "getopt string")
+            return
+        if help_alias in long_opts and long_opts[help_alias].takes_argument:
+            print(f"{filepath}: help alias -{help_alias} should not take an "
+                  "argument")
+            return
+
     # Check longform options between the four sources
     for longform in all_longform:
         cur_help = help_opts.get(longform, OptionInfo())
--- a/scripts/check_options.py
+++ b/scripts/check_options.py
@@ -4,41 +4,94 @@ Reads the options within the helptext (help_<command>() function),
 the long_options[] array, the getopt_long() string, and the switch(c)
 block, and checks that they are consistent with each other.
 
-Mostly written by ChatGPT with small tweaks by @faithokamoto
+The base of this script was written by ChatGPT.
+It has since been developed by @faithokamoto.
 """
 import os
 import re
 
 def extract_help_options(text: str) -> dict:
-    """
-    Extract options from the help_<command>() function.
-    Returns a dict mapping long option names to tuples: (short_opt or None, takes_argument: bool)
-    """
-    help_opts = {}
-    help_pattern = re.compile(r'<< "\s+(?:^|[^\S\r\n])(-\w),?\s+(--[a-zA-Z0-9\-]+)(?:\s[A-Z_]+)?')
-    long_only_pattern = re.compile(r'<< "\s+(?:^|[^\S\r\n])(?!-)(--[a-zA-Z0-9\-]+)(?:\s[A-Z_]+)?')
+    """Extract options from help_<command>()
     
+    Looks for `<< "    -<short>, --<long> <arg>  <desc>`
+    (well actually it ignores the description)
+    The shortform option is optional, as is the argument.
+
+    If the shortform option is present, it must be followed by
+    a comma and a single space before the longform option.
+    
+    If the argument is present, it is expected to be in all-caps,
+    have nothing surrounding it (e.g. no <>), and be exactly
+    one space after the long option.
+
+    Parameters
+    ----------
+    text : str
+        The text of the file to search for help options.
+    
+    Returns
+    -------
+    dict
+        A dictionary mapping long option names to tuples:
+        (short_opt or None, takes_argument: bool)
+    """
+
+    help_opts = {}
+    # Match the line's prefix, which is `<< "    `
+    prefix = r'<< "\s+'
+    # Match a shortform option: `-<short`
+    shortform_patten = r'-.'
+    # Match a longform option: `--<long>`
+    longform_patten = r'--[a-zA-Z0-9\-]+'
+    # Match an optional argument in all-caps
+    arg_pattern = r'[A-Z_]+'
+
+    help_pattern = re.compile(
+        rf'{prefix}({shortform_patten}),\s({longform_patten})\s({arg_pattern})?'
+        )
+    # Same pattern but without the shortform option
+    long_only_pattern = re.compile(
+        rf'{prefix}({longform_patten})\s({arg_pattern})?'
+        )
+
     inside_help = False
+    curly_brace_nesting = 0
+
     for line in text.splitlines():
-        if re.search(r'void\s+help_\w+\s*\(', line):
+        stripped = line.strip()
+        # Are we inside the helptext printing function?
+        if re.search(r'void\shelp_\w+\s*\(', line):
             inside_help = True
-        elif inside_help and '}' in line:
+            curly_brace_nesting = 0
+            continue
+        # End at end of outermost curly braces
+        elif inside_help and stripped == '}' and curly_brace_nesting == 0:
             inside_help = False
-        if not inside_help or line.strip().startswith('//'):
+
+        # Ignore comments and lines outside the helptext
+        if not inside_help or stripped.startswith('//'):
             continue
 
-        match = help_pattern.search(line)
+        # Keep track of curly brace nesting
+        if '{' in stripped:
+            curly_brace_nesting += 1
+        if '}' in stripped:
+            curly_brace_nesting -= 1
+
+        # Parse out sections of full pattern
+        match = help_pattern.search(stripped)
         if match:
             short_opt = match.group(1)[1]
             long_opt = match.group(2)[2:]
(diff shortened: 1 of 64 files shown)


src/subcommand/minimizer_main.cpp
```diff
@@ -111,12 +122,12 @@ int main_minimizer(int argc, char** argv) {
...
-    constexpr int OPT_FAST_COUNTING = 1003;
-    constexpr int OPT_SAVE_MEMORY = 1004;
-    constexpr int OPT_HASH_TABLE = 1005;
-    constexpr int OPT_NO_DIST = 1100;
+    #define OPT_THRESHOLD 1001
+    #define OPT_ITERATIONS 1002
+    #define OPT_FAST_COUNTING 1003
+    #define OPT_SAVE_MEMORY 1004
+    #define OPT_HASH_TABLE 1005
+    #define OPT_NO_DIST 1100
```
adamnovak: I would favor switching these all to be `constexpr int`s instead of switching them all to be `#define`s, since `constexpr` is there partly to solve fun problems you can have with defines, like them not having a notion of scope or type.

But since these are all source files the macros won't get very far, so it's not a huge problem to use them.

src/subcommand/sort_main.cpp
```diff
@@ -55,12 +56,13 @@ int main_sort(int argc, char *argv[]) {
...
                 {"without-grooming", no_argument, 0, 'w'},
-                {"index-to", no_argument, 0, 'I'},
+                {"index-to", required_argument, 0, 'I'},
+                {"help", no_argument, 0, 'h'},
                 {0, 0, 0, 0}
             };
 
         int option_index = 0;
-        c = getopt_long (argc, argv, "a:gr:wI:",
+        c = getopt_long (argc, argv, "a:gr:wI:h?",
```
adamnovak, line 65: Sometimes the help options are at the front of this string, and sometimes they are at the back.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+Reads the options within the helptext (help_<command>() function),
+the long_options[] array, the getopt_long() string, and the switch(c)
+block, and checks that they are consistent with each other.
+
+There are a lot of rules; this script has *strong* opinions about how
+options should appear, though I've been flexible about valid
+variations. If something looks off, a detailed message will be
+printed to stdout. If that message doesn't make sense, something
+might've been ignored due to a wrong format, so check "Format".
+If it still doesn't make sense, ping me.
```
adamnovak: "Me" only makes sense if one puts on one's `git blame` goggles, and even then only really works until you finish your thesis.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+
+If you get one of these errors, then, you might have to
+fix multiple things in the file before the script can run well.
+
+Checks performed on the *file as a whole*.
+
+1. The help options (-h, -?) must be present in the helptext,
+   getopt string, and the switch block, and must not take an
+   argument. In the switch block, they must crash.
+2. The default case in the switch block must crash.
```
adamnovak: I don't think that "crash" is the right word here. We crash when something *internal* goes wrong, and when we crash we try to do things like print a stack trace. If we can't continue because we know the user told us to do something impossible or we didn't understand them, then I'd say we *fail*, or generically exit with a nonzero exit code.

It might also be good to put the magic words that the script greps for here; I'm likely to be reading this text because I wrote a call to my `chastise_the_user_gently()` function instead of an `exit(1)` or `<command>_help()` or whatever the script actually promises to detect.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+and not in long_options[] or the helptext.
+
+### Each option separately
+
+Longform options are pulled from the helptext and long_options[]
+
+1. All longform options must be in long_options[]
+2. The long_options[] entry must be either `no_argument`
+   or `required_argument`.
+3. The long_options[] entry must not have a numeric shortform.
```
adamnovak: This means `1` (the number) is disallowed, but `'1'`, the `char` with value 49, is allowed, right?

But I only know that because I saw we were creatively using bare numbers instead of defined constants and you removed them. So maybe you want to say here that it must be either a quoted `char` or an all-caps variable name.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+you may have to do multiple runs/fixes to see all the problems.
+
+## Format
+
+- helptext: within the `help_<command>()` function,
+  options must be printed with `"  -<short>, --<long> <arg>  <desc>"`
+  (the description is ignored, and the shortform and argument are optional).
+  The longform option must be composed of alphanumeric characters
+  and hyphens, and the argument must be in all-caps.
+  In addition, there must be a "usage:" line somewhere.
```
adamnovak, line 79: One thing that tripped me up trying to test this was I tried to set a short option name in the help text to a special character, to create a mismatch between the help and the parsing code.

That wasn't detected, because the checker script won't parse a special character as a short option name. We don't actually use any such options, but some programs will use e.g. `-@` as the option for number of threads.

If we want to keep the rule that only some characters are allowed short option names, we should mention that here.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+1. The help options (-h, -?) must be present in the helptext,
+   getopt string, and the switch block, and must not take an
+   argument. In the switch block, they must crash.
+2. The default case in the switch block must crash.
+3. All options in the getopt string must be in long_options[].
+4. All options in the switch block must be in long_options[].
+
+While long_options[], the getopt string, and the switch block
+must have the same option set, the helptext is allowed to be
+missing some options (e.g. deprecated/developer-only options).
```
adamnovak, line 44: We don't seem to enforce that, if an option has character shortform, and the option appears in the help text, the shortform *also* appears in the help text. Combined with only allowing certain characters for the shortform in the help text parser, this lets us write a help text entry where the shortform in the help text is not the same as the one actually parsed, as long as in the help text it is one of the characters we won't parse as a shortform.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+"""Files to skip in the consistency check."""
+HELPTEXT_FUNCTION = r'void\shelp_\w+\s*\(char\*\* argv.*\) \{'
+"""Regex to match the start of a helptext function."""
+HELP_DESC = 'print this help message to stderr and exit'
+"""Expected description for the --help option."""
+ANNOTATE_EXCEPTIONS = {'xg-name', 'bed-name'}
+"""annotate_main.cpp lets these appear twice in helptext."""
+# Some giraffe arguments are processed outside of the switch block
+# (at least I think so? I'm not entirely sure)
+GIRAFFE_EXCEPTIONS = {'max-multimaps', 'batch-size'}
```
adamnovak: I think these are really errors in Giraffe. We have logic in Giraffe to define command line options programmatically and add them to e.g. the `long_options`, but we also seem to have duplicate static `long_options` entries that should be removed.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+"""Keywords that indicate a crash/deprecation in the switch block."""
+
+@dataclass
+class OptionInfo:
+    """Information about a command line option."""
+
+    takes_argument: bool = None
+    """Whether the option takes an argument."""
+    shortform: Optional[str | int] = None
+    """Short option name, or None if not present."""
```
adamnovak: So than what does this mean when it's an `int`?

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+    def is_unset(self) -> bool:
+        """Check if this option is unset."""
+        return self.takes_argument is None and self.shortform is None
+    
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, OptionInfo):
+            return False
+        # If either shortform is unset, only compare takes_argument
+        if self.shortform is None or other.shortform is None:
+            return self.takes_argument == other.takes_argument
```
adamnovak: This is a creative notion of equality that makes equality not actually an equivalence class anymore; it's now easy to have `a == b` and `b == c` but `a != c`.

Maybe this should be some kind of `can_unify_with()` or `is_compatible()` method instead?

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+"""annotate_main.cpp lets these appear twice in helptext."""
+# Some giraffe arguments are processed outside of the switch block
+# (at least I think so? I'm not entirely sure)
+GIRAFFE_EXCEPTIONS = {'max-multimaps', 'batch-size'}
+"""Options for giraffe_main.cpp that are not in the switch block."""
+CRASHES = {'exit', 'return', 'deprecated', 'abort', 'throw'}
+"""Keywords that indicate a crash/deprecation in the switch block."""
+
+@dataclass
+class OptionInfo:
```
adamnovak, line 145: This doesn't include `longform` because it's meant to always live in a dict keyed by the longform, right?

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+    """Information about a command line option."""
+
+    takes_argument: bool = None
+    """Whether the option takes an argument."""
+    shortform: Optional[str | int] = None
+    """Short option name, or None if not present."""
+    errors: List[str] = None
+    """Some special errors to print about this option."""
```
adamnovak: An option having errors in its formatting isn't a special case; what makes these errors special?

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+            errors.append(f"Help option line '{stripped}' has less than "
+                            "two spaces between option and description")
+        if (shortform == '-h' and description.strip() != HELP_DESC
+            and not is_giraffe):
+            errors.append(f"Help option -h does not have the expected "
+                            f"description '{HELP_DESC}'")
+        
+        # Count of blank spaces between option and description
+        text_offset = re.match(r'\s+(.+)', description).start(1) - match.start()
+        # Column within printed text which starts thr description
```
adamnovak:
```suggestion
        # Column within printed text which starts the description
```

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+
+        prefix = match.group(1)
+        shortform = match.group(2)[1:] if has_shortform else None
+        longform = match.group(3 if has_shortform else 2)[2:]
+        takes_arg = match.group(4 if has_shortform else 3) is not None
+        description_num = 5 if has_shortform else 4
+        description = match.group(description_num)
+
+        if len(prefix) != (6 if shortform is None else 2):
+                errors.append(f"Help option line '{stripped}' "
```
adamnovak, line 296: If we could track the line number through, instead of just re-printing the line, that might be more useful for finding the error in the code. We don't want to see `file.cpp: Help option line '' does not start with two spaces`.

If you print the file path and line number the right way I think fancy editors will even let you jump to it. I know iTerm2 will parse "whatever.cpp line 123", for example.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+        if re.search(HELPTEXT_FUNCTION, stripped):
+            inside_help = True
+            curly_brace_nesting = -1
+        # End at end of outermost curly braces
+        elif inside_help and stripped == '}' and curly_brace_nesting == 0:
+            break
+
+        # Ignore comments and lines outside the helptext
+        if not inside_help or stripped.startswith('//'):
+            continue
```
adamnovak: So single-line comments are supported but multi-line comments are not.

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+
+        # Keep track of curly brace nesting
+        if '{' in stripped:
+            curly_brace_nesting += 1
+        if '}' in stripped:
+            curly_brace_nesting -= 1
```
adamnovak, line 345: Good code will have only one block-defining brace per line, but the language allows more. So maybe we need to note somewhere that the script only parses nice neat code?

We also might have braces in initializer lists in e.g. default values, somehow. Though then we'd expect both an open and a close brace on the line, and no net change in the nesting tracker, so it *should* be fine?

scripts/check_options.py
```diff
@@ -0,0 +1,776 @@
...
+            has_usage = True
+            # Don't check usage lines for length
+            continue
+
+        printed_line = re.search(quoted_pattern, stripped)
+        if printed_line:
+            printed_len = printed_line.end(1) - printed_line.start(1)
+            if printed_len > 80:
+                raise ValueError(f"Helptext line `{stripped}` is {printed_len}"
+                                 ">80 characters long")
```
adamnovak: This doesn't measure the actual *printed* length, though, just what's between the outermost quotes, right? So it would not like:

```
std::cerr << "  -a, --an-option VAL         an option to do the thing (Default: " << namespace::SomeClass::Subclass::get_default_value_of_the_option(SomeOtherClass::MODE_NORMAL_MODE, std::vector<AnotherLongTypeName>()) << ")" << std::endl;
```
Even if the net result of that long expression was a short value.

I guess that's maybe OK since we also want to wrap the code to a plausible length? But the code should be aware that it's not *really* looking at the final printed line length, and that the problem it is reporting could be too long help text *or* too long code on a single line to generate it.

adamnovak (approved):

src/subcommand/minimizer_main.cpp
faithokamoto: I can change them to `constexpr int`s if you like; was just trying to make everything match the most common form

src/subcommand/sort_main.cpp
faithokamoto: And sometimes they're actually in the middle ... I considered checking the order of options in getopt_long but decided it probably didn't matter as long as all of them are there. Would you prefer forcing everything to have a consistent order?

scripts/check_options.py
faithokamoto: I'm changing the regex to "any non space character" since some options do actually have non-letter shortforms

scripts/check_options.py
faithokamoto: changing this to "not be a bare non-char int."

scripts/check_options.py
faithokamoto: made it so it's never an int now

scripts/check_options.py
faithokamoto: Changed to "Extra things to complain about at the end." since that's what I was using it for: it's things that I catch within the helper functions, tied to a specific option, that I want to handle later (since I was trying to avoid ValueErrors and the resulting inability to continue that helper function)

scripts/check_options.py
faithokamoto: yeah, put a more explicit callout

scripts/check_options.py
faithokamoto: fair, made is_compatible()

scripts/check_options.py
faithokamoto: changed to "If it still doesn't make sense, ping @faithokamoto (if I'm still here) and/or open an issue on GitHub.

scripts/check_options.py
faithokamoto: took away giraffe's exceptions and in doing so realized that --batch-size was duplicated (in helptext both automatically and manually) so I took away its manual appearance

scripts/check_options.py
faithokamoto: made explicit

scripts/check_options.py
faithokamoto: changed to phrasing about "intentionally not working" and pointed to the global variable which stores the list of keywords

scripts/check_options.py
faithokamoto: now it tracks line numbers

scripts/check_options.py
faithokamoto: made more explicit - and yeah, that kind of nesting passes without an issue.

scripts/check_options.py
faithokamoto: I documented this as an official workaround (since it's what I did too)

scripts/check_options.py
faithokamoto: I added this check; it actually caught some stuff

src/subcommand/minimizer_main.cpp
faithokamoto: decided to add a check to force everything to be constexpr int

faithokamoto:
Pushed spot-fixes for all but one of the things brought up in review. Ended up adding two new checks
- force options with shortforms to match between `long_options[]` and the helptext
- force all-caps variables to be set as `constexpr int`s
Edited a few helptexts etc. to force them to comply with the new rules. Updated the Google Doc to reflect my changes.

Not sure what you want to do about order within the getopt string, so I didn't modify anything for that. In my opinion the order there doesn't matter.

src/subcommand/sort_main.cpp
adamnovak: That might be too annoying to deal with. But if we ever made a tool that could sort them, that would be useful.

PR #4660, opened 2025-07-16
Add -w option to vg index to allow upweighting nodes in snarl finding

adamnovak:
## Changelog Entry

 * `vg index` now accepts a `-w` option to up-weight nodes to push the top-level chain through them when finding snarls

## Description

This gives you some control over snarl finding by letting you up-weight nodes during distance indexing. You can pass `-w` to `vg index -j` with a node ID and it will imagine it to be 10 billion bases longer than it actually is, which *should* force the top-level chain through it. You can repeat the option if you have multiple nodes you think should be on the top-level chain path (such as two known tip nodes you want at either end of the chain). If it somehow isn't working (maybe you have a *really* long genome), you can use it on the same node multiple times.

In the future we can make the weight amount configurable. Genuinely scripting the order of the snarl decomposition rooting process, rather than just up-weighting nodes, would require a lot more refactoring, so I didn't do it.

### DIFF
--- a/src/integrated_snarl_finder.cpp
+++ b/src/integrated_snarl_finder.cpp
@@ -34,6 +34,14 @@ protected:
     /// massaging and aren't const.
     /// TODO: this makes read operations not thread safe!
     mutable structures::UnionFind union_find;
+
+    /// Mapping full of extra weight to assign to nodes, in addition to their
+    /// actual length in bases, to control snarl tree rooting.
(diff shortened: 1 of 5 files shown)


faithokamoto:
hehe. It's *so easy* to accidentally introduce a command-line option bug, which is why we had so many. Good that we can auto-catch them now.

PR #4666, opened 2025-07-23
[WIP] Deconstruct -f writes off-reference fasta

glennhickey:
## Changelog Entry

 * `deconstruct -f` option to write fasta file of off-reference sequence, as well as a tsv table describing its locations

## Description

This is an old local branch I found that needs testing and merging -- Doing the PR now so I don't forget it.


faithokamoto:
Looks like this managed to run afoul of my new `scripts/check_options.py` - specifically, its rule that helptext strings must wrap after 80 characters. Just pick some spot to manually wrap part of the option description onto the next line.

PR #4679, opened 2025-08-14
Giraffe debug

adamnovak:
## Changelog Entry

 * `vg giraffe` now has a `--softclip-penalty` flag to reduce alignment scores per-base for softclips
 * `vg filter` now has a `-W/--overwite-score` flag to save the scores from `--rescore`.
 * `vg filter` now checks to make sure you aren't using `--rescore` or related options when they would do nothing.
 * Internal changes in `vg giraffe` to allow multiple presets to potentially share settings.
 * Bug fixes for chain transition distance measurement with the zip code tree in `vg giraffe`

## Description
This has several small changes I've needed when trying to finalize Long Read Giraffe for the paper. It originally had a `--min-chain-score-per-explored-minimizer` filter in Giraffe, but I didn't find a use for it and it makde the Giraffe Facts filter table weirdly wide to have it in there, so I backed it out again.


faithokamoto:
merging this since I want to use the backported bug fixes in my tests

PR #4680, opened 2025-08-15
add --invert-keep-paths option to vg mod

faithokamoto:
## Changelog Entry

 * `vg mod` now has an `--invert-keep-paths` option to save the complement of path names passed to `--keep-paths`

## Description

Another "I wanted it so I tweaked a thing" feature. I have a graph with over a hundred paths and I wanted to remove a single one. Technically I could've generated the >100 `--keep-path` options and tacked them all on, but this way is easier.

### DIFF
--- a/src/subcommand/mod_main.cpp
+++ b/src/subcommand/mod_main.cpp
@@ -59,6 +59,7 @@ void help_mod(char** argv) {
          << "  -N, --remove-non-path    keep only nodes and edges which are part of paths" << endl
          << "  -A, --remove-path        keep only nodes and edges which aren't part of a path" << endl
          << "  -k, --keep-path NAME     keep only nodes and edges in the path (may repeat)" << endl
+         << "  -V, --invert-keep-path   keep only nodes and edges in paths not passed to -k" << endl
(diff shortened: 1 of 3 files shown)


faithokamoto:
As far as I can tell the error on Mac isn't something I touched. The same Mac error came up when I merged the `giraffe-debug` branch, despite that branch passing CI tests pre-merge. Something got borked today.
```
/Users/runner/work/vg/vg/deps/libvgio/include/vg/io/registry.hpp:594:23: error: no viable conversion from 'absl::string_view' (aka 'basic_string_view<char>') to 'const string' (aka 'const basic_string<char>')
  594 |         const string& tag = Message::descriptor()->full_name();
```

adamnovak:
I've tinkered with libvgio where it touches Protobuf to copy things back to `std::string` in a way that works on old and new Protobufs.

I'm not sure when Linux distros will start shipping Protobufs that are too new for C++14 support. We can raise the C++ standard version for Linux whenever we want; we have the min supported *compiler* version under CI now.

faithokamoto:
Yay that fixed it!

PR #4689, opened 2025-08-27
Utility functions for option parsing, validity checks, and pretty printing errors/warnings

faithokamoto:
## Changelog Entry

 * Create utility functions for basic parsing/validity checks, and use them in subcommands + `src/index_registry.cpp`
     * `parse_thread_count()`, `parse_split_string()`, and `assign_fastq_files()` are meant to accept `optarg` and then do common nontrivial parsing
     * `error_if_file_does_not_exist()`, `error_if_file_cannot_be_written()`, and `error_if_file_is_gzipped()` are meant to accept `optarg` and perform standardized checks for file validity
* At least attempt to enforce the use of some new standardized parsing/validity functions
* Create utility functions for pretty error/warning printing, and use them in subcommands + `src/index_registry.cpp`
* Fix two broken tests in `test/t/03_vg_view.t`

## Description

Welcome back to Faith's Nitpick Corner, where today we are serving a sizzling side project of centralized utility functions.

### Motivation

Many `vg` subcommand options have similar basic parsing logic. In the best-case scenario, this led to copious copy-pasting, which is why so many subcommands have e.g. [the](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/subcommand/paths_main.cpp#L306) [same](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/subcommand/zipcode_main.cpp#L127) [code](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/subcommand/call_main.cpp#L303) for parsing `--threads`. In the worst-case scenario, people didn't bother adding validity checks. In the confusing-case scenario, people copy-pasted badly, which is how you get [`vg augment` claiming to be `vg call`](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/subcommand/augment_main.cpp#L233).

Initially this was only supposed to be `parse_thread_count()` (errors if given nonpositive input) and `error_if_file_does_not_exist()` (errors if file can't be read). Then it ballooned into adding functions that check for if a file is gzipped (`error_if_file_is_gzipped()`), use the same option to accept up to two FASTQ files (`assign_fastq_files()`), accept an argument of the form `X<delimiter>Y` and extract `X` and `Y` (`parse_split_string()`), or check if a file can be _output_ to (`error_if_file_cannot_be_written()`).

Since all of these new utility functions needed to print out errors, I also made `error_and_exit()` and `emit_warning()` to print an error/warning in a standardized format via `emit_with_indent()`. `error_and_exit()` additionally exits with an error code. These pretty printing functions I used all over the subcommands as well as `src/index_registry.cpp`. I don't _have_ to use them; I just like them. I could revert to the previous error/exit stuff if we want. Though if we do that, we should make sure that all errors actually *error* ([example](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/subcommand/gbwt_main.cpp#L682)).

### Fiddly bits

* Currently `error_if_file_cannot_be_written()` checks if a file can be written to by attempting to open a file for writing. This means it will overwrite whatever was there before. I had to [change a test](https://github.com/vgteam/vg/commit/7f4171cd8f99e05eaaf325ce0a1f2e64db2bb3d1) to move an error test to later, to prevent it from overwriting a file used by a different test.
* I belatedly realized that `split_delims()` is a thing that [already exists](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/utility.cpp#L191-L209), but I think `parse_split_string()` is still useful for its simpler interface in an argument parsing context.
* In places where I'd already validated a filename validity during argument parsing, I removed later duplicate validity checks ([example](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/subcommand/add_main.cpp#L159)). I think it's better to have the validity checks centralized and, crucially, _early_: this way we'll error before doing any expensive algorithms.
* While I _generally_ kept the error/warning messages the same as before, some of the `vg mpmap` error messages used the wrong shortforms for options ([example](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/subcommand/mpmap_main.cpp#L1249)), which I fixed.
* The pretty printing functions are [thread-safe](https://github.com/vgteam/vg/blob/27a9042e8c53d36c0f72a3f7784bedae58c656a4/src/utility.cpp#L28), which feels slightly overkill but at least one of the subcommand files had a thread-safe error print so I ported that functionality.
* I had to slightly rewrite some tests to make them expect my new pretty-printed errors. Their intent/purpose was not changed.
* [Two tests](https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/test/t/03_vg_view.t#L65-L73) in `test/t/03_vg_view.t` were accessing bad file names, which led to empty output. Since these empty outputs were compared to each other, the test passed. I swapped to the correct file names.
* I used `error_if_file_does_not_exist()` in `IndexRegistry`. To then get it past unit tests which intentionally use dummy files, I added a flag to the object which tells it to skip checking the files passed to `provide()`.
* `scripts/check_options.py` attempts to enforce the usage of `error_if_file_does_not_exist()`, `error_if_file_cannot_be_written()`, and `parse_thread_count()`. Unfortunately it has to use heuristics for all of them: filenames variables are auto-detected as `<variable name ending> = optarg` where the known endings are in `FILENAME_VAR_ENDS`. This will fail to catch other name formats such as `xg_name = optarg` as `_name` would lead to false positives. It also won't catch things like `vcf_filenames.append(optarg)` where there's a list of filenames in use. The thread count heuristic requires `omp_set_num_threads()` and `parse_thread_count()` to be on the same line.

### DIFF
--- a/src/index_registry.cpp
+++ b/src/index_registry.cpp
@@ -87,6 +87,7 @@ static string to_string(const vg::IndexGroup& name) {
 
 namespace vg {
 
+const string context = "[IndexRegistry]";
 
 IndexingParameters::MutableGraphImplementation IndexingParameters::mut_graph_impl = PackedGraph;
 int IndexingParameters::max_node_size = 32;
@@ -5630,6 +5631,7 @@ void IndexFile::provide(const vector<string>& filenames) {
     // append all filenames
     // TODO: would it be better to sometimes error check that the file isn't a duplicate?
     for (const string& filename : filenames) {
+        error_if_file_does_not_exist(context, filename);
         this->filenames.emplace_back(filename);
     }
     provided_directly = true;
--- a/src/index_registry.cpp
+++ b/src/index_registry.cpp
@@ -125,16 +125,10 @@ double IndexingParameters::thread_chunk_inflation_factor = 2.0;
 IndexingParameters::Verbosity IndexingParameters::verbosity = IndexingParameters::Basic;
 
 void copy_file(const string& from_fp, const string& to_fp) {
+    error_if_file_does_not_exist(context, from_fp);
+    error_if_file_cannot_be_written(context, to_fp);
     ifstream from_file(from_fp, std::ios::binary);
     ofstream to_file(to_fp, std::ios::binary);
-    if (!from_file) {
-        cerr << "error:[IndexRegistry] Couldn't open input file " << from_fp << endl;
-        exit(1);
-    }
-    if (!to_file) {
-        cerr << "error:[IndexRegistry] Couldn't open output file " << to_fp << endl;
-        exit(1);
-    }
     to_file << from_file.rdbuf();
 }
 
@@ -156,8 +150,7 @@ bool is_gzipped(const string& filename) {
 int64_t get_num_samples(const string& vcf_filename) {
     htsFile* vcf_file = hts_open(vcf_filename.c_str(),"rb");
     if (!vcf_file) {
-        cerr << "error:[IndexRegistry]: Failed to open VCF file: " << vcf_filename << endl;
-        exit(1);
+        error_and_exit(context, "Failed to open VCF file: " + vcf_filename);
     }
     bcf_hdr_t* header = bcf_hdr_read(vcf_file);
     int64_t num_samples = bcf_hdr_nsamples(header);
@@ -196,9 +189,8 @@ double format_multiplier() {
         case IndexingParameters::VG:
             return 2.91;
         default:
-            cerr << "error:[IndexRegistry] unrecognized mutable graph implementation format" << endl;
-            exit(1);
-            return 0.0;
+            error_and_exit(context, "unrecognized mutable graph implementation format");
+            return 0.0; // never reached
     }
 }
 
@@ -320,7 +312,7 @@ vector<string> vcf_contigs(const string& filename) {
     
     htsFile* vcf = hts_open(filename.c_str(),"rb");
     if (vcf == nullptr) {
-        cerr << "error:[IndexRegistry] Could not open VCF" << filename << endl;
+        error_and_exit(context, "Could not open VCF: " + filename);
     }
     
     bcf_hdr_t* header = bcf_hdr_read(vcf);
@@ -387,26 +379,19 @@ size_t xg_index_size(const xg::XG& index) {
 // would go away when the setup function returns.
 
 static void init_in(ifstream& in, const string& name) {
+    error_if_file_does_not_exist(context, name);
     in.open(name);
-    if (!in) {
-        cerr << "error:[IndexRegistry] could not open input file '" << name << "'" << endl;
-        exit(1);
-    }
 }
 
 static void init_out(ofstream& out, const string& name) {
+    error_if_file_cannot_be_written(context, name);
     out.open(name);
-    if (!out) {
-        cerr << "error:[IndexRegistry] could not write output to '" << name << "'" << endl;
-        exit(1);
-    }
 }
 
 static void init_in_out(fstream& strm, const string& name) {
     strm.open(name);
     if (!strm) {
-        cerr << "error:[IndexRegistry] could not open '" << name << "'" << endl;
-        exit(1);
+        error_and_exit(context, "could not open '" + name + "'");
     }
 }
 
@@ -423,8 +408,7 @@ static auto init_mutable_graph() -> unique_ptr<MutablePathDeletableHandleGraph> 
             graph = make_unique<VG>();
             break;
         default:
-            cerr << "error:[IndexRegistry] unrecognized mutable graph implementation format" << endl;
-            exit(1);
+            error_and_exit(context, "unrecognized mutable graph implementation format");
             break;
     }
     return graph;
@@ -457,8 +441,7 @@ int execute_in_fork(const function<void(void)>& exec) {
     pid_t pid = fork();
     
     if (pid == -1) {
-        cerr << "error:[IndexRegistry] failed to fork process" << endl;
-        exit(1);
+        error_and_exit(context, "failed to fork process");
     }
     else if (pid == 0) {
         // this is the child process that will actually make the indexes
@@ -473,7 +456,7 @@ int execute_in_fork(const function<void(void)>& exec) {
     } else {
         // This is the parent
         if (IndexingParameters::verbosity >= IndexingParameters::Debug) {
-            cerr << "[IndexRegistry]: Forked into child process with PID " << pid << "." << endl;
+            cerr << context << ": Forked into child process with PID " << pid << "." << endl;
         }
     }
     
@@ -483,7 +466,8 @@ int execute_in_fork(const function<void(void)>& exec) {
     
     // pass through signal-based exits
     if (WIFSIGNALED(child_stat)) {
-        cerr << "error:[IndexRegistry] Child process " << pid << " signaled with status " << child_stat << " representing signal " << WTERMSIG(child_stat) << endl;
+        cerr << "error" << context << ": Child process " << pid << " signaled with status " 
+             << child_stat << " representing signal " << WTERMSIG(child_stat) << endl;
         if (raise(WTERMSIG(child_stat)) == 0) {
             // TODO: on Mac, raise isn't guaranteed to not return before the handler if it succeeds.
             // Also the signal might not be one that necessarily kills us.
@@ -499,8 +483,8 @@ int execute_in_fork(const function<void(void)>& exec) {
     return WEXITSTATUS(child_stat);
 #else
     // We can't stop OpenMP, so we can't actually fork, so we can't actually do our smart retry.
-    cerr << "warning:[IndexRegistry] vg was built with an OpenMP which is too old to safely support forking."
-         << "We will not be able to automatically retry with a simpler graph if a resource limit is hit." << endl;
+    emit_warning(context, "vg was built with an OpenMP which is too old to safely support forking. "
+                          "We will not be able to automatically retry with a simpler graph if a resource limit is hit.");
 
     // Just run the work in-process
     exec();
@@ -649,7 +633,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                             bool phased_vcf) {
         
         if (IndexingParameters::verbosity != IndexingParameters::None) {
-            cerr << "[IndexRegistry]: Chunking inputs for parallelism." << endl;
+            cerr << context << ": Chunking inputs for parallelism." << endl;
         }
                         
         // boilerplate
@@ -710,14 +694,14 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                     }
                     stat(vcf_filenames[i].c_str(), &stat_vcf);
                     if (stat_vcf.st_mtime > stat_tbi.st_mtime) {
-                        cerr << "warning:[IndexRegistry] Tabix index " + tabix_name + " is older than VCF " + vcf_filenames[i] + " and will not be used. Consider recreating this tabix index to speed up index creation.\n";
+                        emit_warning(context, "Tabix index " + tabix_name + " is older than VCF " + vcf_filenames[i] 
+                                              + " and will not be used. Consider recreating this tabix index to speed up index creation.");
                         continue;
                     }
                     
                     tabix_index = tbx_index_load(tabix_name.c_str());
                     if (tabix_index == nullptr) {
-                        cerr << "error:[IndexRegistry] failed to load tabix index " << tabix_index << endl;
-                        exit(1);
+                        error_and_exit(context, "failed to load tabix index " + tabix_name);
                     }
                 }
                 
@@ -739,8 +723,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                     bcf_hdr_destroy(header);
                     int close_err_code = hts_close(vcf);
                     if (close_err_code != 0) {
-                        cerr << "error:[IndexRegistry] encountered error closing VCF " << vcf_filenames[i] << endl;
-                        exit(1);
+                        error_and_exit(context, "encountered error closing VCF " + vcf_filenames[i]);
                     }
                     continue;
                 }
@@ -758,8 +741,9 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                             vcf_contigs_with_variants[i].push_back(chrom);
                         }
                         else if (curr_contig > chrom) {
-                            cerr << "error:[IndexRegistry] Contigs in VCF must be in ASCII-lexicographic order. Encountered contig '" << chrom << "' after contig '" << curr_contig << "' in VCF file" << vcf_filenames[i] << "." << endl;
-                            exit(1);
+                            error_and_exit(context, "Contigs in VCF must be in ASCII-lexicographic order. Encountered contig '" 
+                                              + string(chrom) + "' after contig '" + curr_contig + "' in VCF file " 
+                                              + vcf_filenames[i] + ".");
                         }
                     }
                     else {
@@ -770,8 +754,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                     err_code = bcf_read(vcf, header, vcf_rec);
                 }
                 if (err_code != -1) {
-                    cerr << "error:[IndexRegistry] failed to read from VCF " << vcf_filenames[i] << endl;
-                    exit(1);
+                    error_and_exit(context, "error reading VCF file " + vcf_filenames[i]);
                 }
                 // we'll be moving on to a different file, so we won't demand that these
                 // be in order anymore
@@ -779,8 +762,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                 bcf_hdr_destroy(header);
                 err_code = hts_close(vcf);
                 if (err_code != 0) {
-                    cerr << "error:[IndexRegistry] encountered error closing VCF " << vcf_filenames[i] << endl;
-                    exit(1);
+                    error_and_exit(context, "encountered error closing VCF " + vcf_filenames[i]);
                 }
             }
         }
@@ -832,8 +814,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
         for (auto it = sample_set_contigs.begin(); it != sample_set_contigs.end(); ++it) {
             for (const auto& contig : it->second) {
                 if (contig_to_group.count(contig)) {
-                    cerr << "error:[IndexRegistry] Contig " << contig << " is found in multiple VCFs with different samples" << endl;
-                    exit(1);
+                    error_and_exit(context, "Contig " + contig + " is found in multiple VCFs with different samples.");
                 }
                 contig_to_group[contig] = contig_groups.size();
             }
@@ -1016,7 +997,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
         }
         
         if (IndexingParameters::verbosity != IndexingParameters::None) {
-            cerr << "[IndexRegistry]: Chunking FASTA(s)." << endl;
+            cerr << context << ": Chunking FASTA(s)." << endl;
         }
         
         output_fasta_names.resize(buckets.size());
@@ -1062,7 +1043,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
         }
         
         if (IndexingParameters::verbosity != IndexingParameters::None) {
-            cerr << "[IndexRegistry]: Chunking VCF(s)." << endl;
+            cerr << context << ": Chunking VCF(s)." << endl;
         }
         
         // open all of the input VCF files
@@ -1071,8 +1052,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
         for (int64_t i = 0; i < input_vcf_files.size(); ++i) {
             htsFile* vcf = bcf_open(vcf_filenames[i].c_str(), "r");
             if (!vcf) {
-                cerr << "error:[IndexRegistry] failed to open VCF " << vcf_filenames[i] << endl;
-                exit(1);
+                error_and_exit(context, "failed to open VCF " + vcf_filenames[i]);
             }
             bcf_hdr_t* header = bcf_hdr_read(vcf);
             bcf1_t* vcf_rec = bcf_init();
@@ -1082,8 +1062,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                 input_checked_out_or_finished[i].store(true);
             }
             else if (err_code < 0) {
-                cerr << "error:[IndexRegistry] failed to read VCF " << vcf_filenames[i] << endl;
-                exit(1);
+                error_and_exit(context, "failed to read VCF " + vcf_filenames[i]);
             }
             input_vcf_files[i] = make_tuple(vcf, header, vcf_rec);
         }
@@ -1154,20 +1133,17 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                         if (phased_vcf) {
                             int sample_add_code = bcf_hdr_add_sample(header, "dummy");
                             if (sample_add_code != 0) {
-                                cerr << "error:[IndexRegistry] error initializing VCF header" << endl;
-                                exit(1);
+                                error_and_exit(context, "error initializing VCF header");
                             }
                         }
                         int hdr_write_err_code = bcf_hdr_write(vcf, header);
                         if (hdr_write_err_code != 0) {
-                            cerr << "error:[IndexRegistry] error writing VCF header to " << output_vcf_name << endl;
-                            exit(1);
+                            error_and_exit(context, "error writing VCF header to " + output_vcf_name);
                         }
                         bcf_hdr_destroy(header);
                         int close_err_code = hts_close(vcf);
                         if (close_err_code != 0) {
-                            cerr << "error:[IndexRegistry] encountered error closing VCF " << output_vcf_name << endl;
-                            exit(1);
+                            error_and_exit(context, "encountered error closing VCF " + output_vcf_name);
                         }
                         output_vcf_names[i] = output_vcf_name;
                     }
@@ -1236,8 +1212,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                         bcf_hdr_t* header_in = get<1>(input_vcf_file);
                         header_out = bcf_hdr_merge(header_out, header_in);
                         if (header_out == nullptr) {
-                            cerr << "error:[IndexRegistry] error merging VCF header" << endl;
-                            exit(1);
+                            error_and_exit(context, "error merging VCF header");
                         }
                         
                         // add the samples from every header
@@ -1251,8 +1226,7 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                                 int sample_err_code = bcf_hdr_add_sample(header_out, header_in->samples[j]);
                                 // returns a -1 if the sample is already included, which we expect
                                 if (sample_err_code != 0) {
-                                    cerr << "error:[IndexRegistry] error adding samples to VCF header" << endl;
-                                    exit(1);
+                                    error_and_exit(context, "error adding samples to VCF header");
                                 }
                             }
                         }
@@ -1261,35 +1235,34 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
                     // documentation in htslib/vcf.h says that this has to be called after adding samples
                     int sync_err_code = bcf_hdr_sync(header_out);
                     if (sync_err_code != 0) {
-                        cerr << "error:[IndexRegistry] error syncing VCF header" << endl;
-                        exit(1);
+                        error_and_exit(context, "error syncing VCF header");
                     }
                     if (phased_vcf && bcf_hdr_nsamples(header_out) == 0) {
                         if (!vcf_indexes.empty()) {
-                            cerr << "warning:[IndexRegistry] VCF inputs from file(s)";
+                            stringstream warning_msg;
+                            warning_msg << "VCF inputs from file(s)";
                             for (auto vcf_idx : vcf_indexes) {
-                                cerr << " " << vcf_filenames[vcf_idx];
+                                warning_msg << " " << vcf_filenames[vcf_idx];
                             }
-                            cerr << " have been identified as phased but contain no samples. Are these valid inputs?" << endl;
+                            warning_msg << " have been identified as phased but contain no samples. "
+                                        << "Are these valid inputs?" << endl;
+                            emit_warning(context, warning_msg.str());
                         }
                         
                         // let's add a dummy so that HaplotypeIndexer doesn't get mad later
                         int sample_add_code = bcf_hdr_add_sample(header_out, "dummy");
                         if (sample_add_code != 0) {
-                            cerr << "error:[IndexRegistry] error initializing VCF header" << endl;
-                            exit(1);
+                            error_and_exit(context, "error initializing VCF header");
                         }
                         // and re-sync, not sure if necessary, but it will be cheap regardless
                         sync_err_code = bcf_hdr_sync(header_out);
                         if (sync_err_code != 0) {
-                            cerr << "error:[IndexRegistry] error syncing VCF header" << endl;
-                            exit(1);
+                            error_and_exit(context, "error syncing VCF header");
                         }
                     }
                     int hdr_write_err_code = bcf_hdr_write(vcf_out, header_out);
                     if (hdr_write_err_code != 0) {
-                        cerr << "error:[IndexRegistry] error writing VCF header to " << output_vcf_name << endl;
-                        exit(1);
+                        error_and_exit(context, "error writing VCF header to " + output_vcf_name);
                     }
                     
                     // remember these so that we can check them out later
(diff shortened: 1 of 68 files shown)


src/subcommand/convert_main.cpp
```diff
@@ -202,83 +204,66 @@ int main_convert(int argc, char** argv) {
...
-                num_threads = parse<int>(optarg);
-                if (num_threads <= 0) {
-                    cerr << "error:[vg convert] Thread count (-t) set to " << num_threads
-                         << ", must set to a positive integer." << endl;
-                    exit(1);
-                }
-                omp_set_num_threads(num_threads);
-            }
+            omp_set_num_threads(parse_thread_count(context, optarg));
+            num_threads = omp_get_num_threads();
```
jltsiren: `omp_get_num_threads()` returns the number of threads in the current parallel region, or 1 if called in a sequential region. To get the maximum number of threads set with `omp_set_num_threads()`, you should call `omp_get_max_threads()` instead.

adamnovak: Should we always be using `omp_get_max_threads()`, instead of the `get_thread_count()` utility function we have right now, for things like setting up per-thread arrays?
https://github.com/vgteam/vg/blob/e78e9d57dbd9777c9430ae0eb21ca6f82b880b4f/src/utility.hpp#L37-L39

src/index_registry.cpp
```diff
@@ -472,7 +456,7 @@ int execute_in_fork(const function<void(void)>& exec) {
     } else {
         // This is the parent
         if (IndexingParameters::verbosity >= IndexingParameters::Debug) {
-            cerr << "[IndexRegistry]: Forked into child process with PID " << pid << "." << endl;
+            cerr << context << ": Forked into child process with PID " << pid << "." << endl;
```
adamnovak: I feel like at this point we might be building a logging framework by degrees. We should look to make sure we wouldn't be better off adopting someone else's logging framework.

src/utility.hpp
```diff
@@ -25,6 +25,21 @@ namespace vg {
 
 using namespace std;
 
+/// Error (and crash) with a standard format
+/// If you're erroring because a file doesn't exist,
+/// consider using error_if_file_does_not_exist() instead.
+void error_and_exit(const string& context, const string& message);
+/// Warn the user with a standard format
+void emit_warning(const string& context, const string& message);
```
adamnovak: Adopting these instead of `cerr << "[leading string]..."` style messages means changing over from building the message into a stream to copying everything into a real `std::string`. This means that we then want to reach for repeated string concatenation to build the message, which isn't very efficient. When we need to talk about something that only has an `oprtator<<`, or some kind of dump-to-stream function, and doesn't have an `std::to_string` overload, then we end up needing to manually set up an `std::stringstream`.

One alternative approach might be to bump up to C++20 and wrap around its `std::format` and its Python-style format strings; apparently it auto-magically uses `operator<<` to stringify things. (Of course, C++ is getting Python-style format strings just as Python and all the cool languages are moving to interpolated strings with the expressions inside them instead.) But that might be worse since then your message is all at the beginning and your values are all at the end and you can get the ordering/count wrong.

Another approach might be to do some magic to make these things themselves work like/be streams:
```
warning(context) << "something is bad" << std::endl;
fatal_error(context) << "something is bad and we'll somehow work out when to kill the program because of it" << std::endl;
```
This would need some cleverness with temporary stream objects and actually doing exits from their destructors, or something; it might be better to see if there's a library that we could use instead of building that ourselves.

src/utility.hpp
```diff
@@ -25,6 +25,21 @@ namespace vg {
...
+
+/// Write a message to an output stream with indentation
+/// Inserts "indent" spaces on the beginning of each line
+/// Honors existing newlines, but otherwise adds newlines between words
+/// such that everything gets printed with width no more than 80
+/// Assumes that the first line is already indented
+/// (because you put something there and you want others to line up)
+void emit_with_indent(ostream& outstream, const string& message, size_t indent);
```
adamnovak: This doesn't promise that things like dumped JSON, FASTA sequence headers, malformed GFA lines, or weird but revealing garbage where the program expected important data to be, will survive the reflow process intact or copy-paste-able. If we feed in something like:
```
warning: we couldn't handle quality string because it is malformed: character ' ' is not allowed in the string.
```
I think that would wrap to the somewhat confusing:
```
warning: we couldn't handle quality string because it is malformed: character '
         ' is not allowed in the string.
```
Which is a lot harder to grasp.

And JSON with spaces in quoted strings could wrap to JSON that can't actually be loaded, so `error_and_exit(context, "There's a very big problem with this alignment: " + pb2json(alignment))` would potentially get into trouble if somehow there were strings with spaces in there.

*Mostly* things would work, though, and maybe we shouldn't be dumping anything we need to be machine-readable through the system we're using for human-readable errors?

We also don't know that the user's terminal is 80 characters wide, but we do have access to APIs that will tell us the current width of the terminal, or whether our output stream is connected to the terminal at all. If the user has a narrower terminal, we probably don't want to be wrapping to 80 characters, because that will be harder to read than just letting the terminal reflow the text.

We can't really do anything with long nucleotide sequences that might appear and upset the formatting. (But maybe we could if we were *really* fancy and sniffed for them?) And if we output an actual tab character, the extra width (4 or 8 characters?) isn't accounted for in the wrapping.

src/utility.hpp
```diff
@@ -821,6 +835,40 @@ string file_base_name(const string& filename);
...
 
+/// Determine if a file can be written to.
+/// Only works for files writable by the current user.
+/// Meant to be used to validate output file names.
+/// Will actually open the file, so it will overwrite the current contents.
+bool file_can_be_written(const string& filename);
+
+/// Check if a file exists and return its name (if so) or error.
+/// Uses file_exists() and is intended to be called when parsing arguments.
+string error_if_file_does_not_exist(const string& context, const string& filename);
```
adamnovak: This function and the others that have it here should probably explain their `context` argument. I guess it's the context to attach to any errors?

It might also be good to specify that this will "print an error message and stop the program", rather than `throw`ing an error, because just saying it will "return ... or error" makes it sound like it might throw or like it might even return an error message.

src/utility.hpp
```diff
@@ -821,6 +835,40 @@ string file_base_name(const string& filename);
...
+
+/// Error if a file looks like it's gzipped.
+void error_if_file_is_gzipped(const string& context, const string& filename);
```
adamnovak: Why doesn't this one return the file's name like the others do?

src/utility.hpp
```diff
@@ -821,6 +835,40 @@ string file_base_name(const string& filename);
...
+/// If max_threads is non-zero, it will also decrease to that maximum.
+/// If max_threads is zero, max is omp_get_max_threads()
+int parse_thread_count(const string& context, const string& arg, int max_threads = 0);
+
+/// A special parser for possibly-paired FASTQ files
+/// If the first file is unset, then it gets the input filename
+/// Else if the second file is unset, then it gets the input filename
+/// If both are already set, then this errors
+/// Also calls error_if_file_does_not_exist() on the filenames
+void assign_fastq_files(const string& context, const string& input_filename, string& fastq1, string& fastq2);
```
adamnovak: OK, so a proper usage of this function involves calling it twice? Maybe it should have a name that reflects that it only does a single file assignment somehow?

I might write this:
```suggestion
void push_fastq(const string& context, const string& input_filename, string& slot1, string& slot2);
```
Or
```suggestion
void add_fastq(const string& context, const string& input_filename, const std::pair<std::string, std::string>& fastq_pair);
```

src/utility.hpp
```diff
@@ -821,6 +835,40 @@ string file_base_name(const string& filename);
...
+
+/// Parse an argument that should be a pair of strings separated by a delimiter.
+/// Errors if the string does not contain exactly one delimiter.
+pair<string, string> parse_split_string(const string& context, const string& arg,
+                                        const char& delimiter, const string& option_name);
```
adamnovak: If this is a pair-only function and not a general split, maybe it should be `parse_pair`?

Since it gets [used to parse things that aren't *really* strings and immediately get `parse<>`'d themselves](https://github.com/vgteam/vg/pull/4689/files#diff-732a7706426698185d84aac919c3b444e8748a107ee89a367645e937d24f47ddR177-R180), this should maybe integrate with that system and become more like:

```suggestion
template<typename T1 = std::string, typename T2 = std::string>
std::pair<T1, T2> parse_pair(const string& context, const string& arg, const char& delimiter, const string& option_name);
```

Then it can call `parse<>` to build the pair items internally (maybe with a `parse<string>` specialization getting defined that does nothing), and you can `parse_pair<string, size_t>(...)` when that's what you need, and it will Just Work.

src/utility.hpp
```diff
@@ -821,6 +835,40 @@ string file_base_name(const string& filename);
 /// Only works for files readable by the current user.
 bool file_exists(const string& filename);
 
+/// Determine if a file can be written to.
+/// Only works for files writable by the current user.
+/// Meant to be used to validate output file names.
+/// Will actually open the file, so it will overwrite the current contents.
+bool file_can_be_written(const string& filename);
```
adamnovak, line 828: I'm worried about having this function blow away the current contents of the file as a side effect of the test.

It's hard for me to come up with a good example of how it would go wrong, though; the best I have is if you're going to run a command that would regenerate several files, and one of them isn't writable, so it aborts as soon as it notices, but at that point it has already touched and clobbered the other files and you have neither your old nor your new index. Or maybe in a similar case where your command line is very wrong and you've put an input file in an output slot, and the command figures out that you've asked for obvious nonsense, but not before blowing away the input file.

It makes it hard to use this in something like `vg add-to-database whatever.db input.dat`, but we don't actually use that pattern at all.

It's not *worse* than `>`'s behavior, but I feel like it could be better. *Actually* opening and writing to the file, and then closing it again and returning `true`, doesn't help at all with time-of-check/time-of-use confusion, because we're just going to re-open the file again by name later.

src/utility.hpp
```diff
@@ -821,6 +835,40 @@ string file_base_name(const string& filename);
...
+void error_if_file_is_gzipped(const string& context, const string& filename);
+
+/// Check if a file can be written to and return its name (if so) or error.
+/// Uses file_can_be_written() and is intended to be called when parsing arguments.
+string error_if_file_cannot_be_written(const string& context, const string& filename);
+
+/// A special parser for thread count which errors if non-positive
+/// If max_threads is non-zero, it will also decrease to that maximum.
+/// If max_threads is zero, max is omp_get_max_threads()
+int parse_thread_count(const string& context, const string& arg, int max_threads = 0);
```
adamnovak: Rather than enforcing that `omp_set_num_threads()` and `parse_thread_count()` are used together with a checker script, why not just make one function that sets the thread count from a string and always call that?

src/utility.hpp
```diff
@@ -821,6 +835,40 @@ string file_base_name(const string& filename);
...
+/// Only works for files writable by the current user.
+/// Meant to be used to validate output file names.
+/// Will actually open the file, so it will overwrite the current contents.
+bool file_can_be_written(const string& filename);
+
+/// Check if a file exists and return its name (if so) or error.
+/// Uses file_exists() and is intended to be called when parsing arguments.
+string error_if_file_does_not_exist(const string& context, const string& filename);
+
+/// Error if a file looks like it's gzipped.
```
adamnovak, line 834: This should maybe specify that we're looking at the name, and we're not sniffing for the GZip magic numbers/headers, so this is safe to use on a `.bam` but will maybe not do what is expected on a `.tgz`.

src/utility.hpp
```diff
@@ -821,6 +835,40 @@ string file_base_name(const string& filename);
...
+void error_if_file_is_gzipped(const string& context, const string& filename);
+
+/// Check if a file can be written to and return its name (if so) or error.
+/// Uses file_can_be_written() and is intended to be called when parsing arguments.
+string error_if_file_cannot_be_written(const string& context, const string& filename);
```
adamnovak: How do these various file-checking functions interact with "-", which we can use in many places as representing "read from standard input"? (For example, I'm pretty sure you can `vg filter --exact-name -n MyRead reads.gam | vg view -dp graph.vg -A - | dot -Tsvg -o graph_with_read.svg` in order to have `vg view` read a graph from a file and a read to visualize on it from standard input.

Probably we want "-" to *not* appear writable, even though you probably could create and write to `./-`. And we would want it to appear as if it *does* exist, because if we use any of the `get_input_file()` functions to open it, it effectively exists.

But then, are the checkers always used with those accessors that support "-"?

src/utility.hpp
adamnovak: The names of these `error_if_...` functions don't really suggest that you can use them as type-narrowing filters in the middle of expressions (i.e. the argument is a file that may not exist, but the return type is a file constrained to always exist), but that seems to be the way they're mostly used, and probably the best way to use them. Also, the names are a bit long for something we might use in more or less every flag that takes a filename.

I would call these something more like`require_exists()`, `require_writable()`, etc. Not `ensure`, since that sounds like it would make a file exist when it didn't, but also not a whole clause for something that's meant to be used so frequently, if it can be avoided, and ideally something that sounds like part of an expression and not a code block.

src/utility.cpp
```diff
@@ -24,6 +24,67 @@
 
 namespace vg {
 
+void error_and_exit(const string& context, const string& message) {
```
adamnovak: This might be better as a `fatal_error()`?

src/utility.cpp
```diff
@@ -24,6 +24,67 @@
...
+    #pragma omp critical (cerr)
+    {
+        cerr << "error" << context << ": ";
+        emit_with_indent(cerr, message, 7 + context.size());
+    }
+    exit(EXIT_FAILURE);
+}
+
+void emit_warning(const string& context, const string& message) {
```
adamnovak: Would `warning()` or `warn()` alone be too dangerous to have kicking around our namespace?

src/subcommand/sift_main.cpp
```diff
@@ -557,33 +556,27 @@ int main_sift(int argc, char** argv){
...
+        in.open(alignment_file);
+        // if (just_calc_insert) {
         //     vg::io::for_each_interleaved_pair_parallel(in, calc_insert);
         //     exit(0);
         // }
 
-        if (is_paired){
-            cerr << "Processing..." << endl;
+        if (is_paired) {
+            cerr << context << ": Processing..." << endl;
```
adamnovak: Here this is touching essentially info-level logging lines, which is making it smell more and more like it's actually a logging framework.

adamnovak (changes requested):
This is a great idea!

I have a bunch of design comments and quibbles with the names of things.

The biggest issue is that basically now you've written 70-80% of a logging framework. Which maybe should go all the way to 100% of our own logging framework (`log.hpp`, documentation for "How logging should work in vg" and "why we need our own logging framework", etc.), or maybe should go to 0% and have us just import a real logging framework someone else maintains that meets our needs.

src/subcommand/convert_main.cpp
faithokamoto: I've swapped to `omp_get_max_threads()` for now but I can change it later

src/utility.hpp
faithokamoto: * `file_exists()` and thus `error_if_file_does_not_exist()` treat `"-"` as always readable.
* `file_can_be_written()` and thus `error_if_file_cannot_be_written()` now treat `"-"` as always non-writable [I just edited to change this behavior]
* `error_if_file_is_gzipped()` will look at `"-"`, note a lack of a `.gz` suffix, and assume it's probably fine. If it's not fine we couldn't detect that anyhow.

src/utility.hpp
faithokamoto: mostly because it's usually called in addition to `error_if_file_does_not_exist()` but I'll change it for consistency

src/utility.hpp
faithokamoto: I'm swapping to `require_exists()` and `ensure_writable()`. The latter is "ensure" because it does actually make a file exist if it didn't exist before.

src/utility.hpp
faithokamoto: I've added a `pretty_print` flag that's default set to true. Going to go through all the uses of the error/warning functions and switch off pretty-printing whenever the error message includes $BAD_THING (e.g. an exception message or a bad argument) so that $BAD_THING will all appear on the same line.

Within the pretty-print I also added a find-and-replace to convert `\t` characters to four spaces.

Now sure how exactly to check for terminal width in C++. The answers here seem to require libraries we don't have? https://stackoverflow.com/questions/1022957/getting-terminal-width-in-c

src/utility.hpp
faithokamoto: implemented this

src/utility.hpp
faithokamoto: I couldn't find a library that did something like this - logging frameworks tended to be way too fancy and aren't meant to support the simple one-line streams that we want for these functions. So I rolled my own. (Possibly I didn't look hard enough because I had an idea for rolling my own and I liked it... sorry.)

Glad to say that
```
warning(context) << "something is bad" << std::endl;
fatal_error(context) << "something is bad and we'll somehow work out when to kill the program because of it" << std::endl;
```
both now "just work". In actuality those functions will do a set-up log to `cerr` and then return a `cerrWrapper` that passes anything `<<`-ed to `cerr` itself. For `fatal_error()`, once the `cerrWrapper` goes out-of-scope, its destructor handles the clean program exit. For when we want to build up a longer message, we can save the initial `cerrWrapper` object and then stream to it as if it were `cerr`. 

I _wanted_ to make indenting work, but it got annoying and I decided to leave that For Later (TM).  Some things that I found include [a library that handles indenting](https://github.com/asap-projects/asap-textwrap) (but seems to stick everything in an in-between string?) and that, if we figure out how to make it work, I could stick a `stringbuf` inside the `cerrWrapper` to [handle the indenting](https://stackoverflow.com/a/9600752/). The difficult part is figuring out when to do the indents once you move to a paradigm where each char is passed to `cerr` one by one.

src/utility.hpp
faithokamoto: closing as moot since I stopped pretty-printing

src/utility.cpp
faithokamoto: I named it `warning()`, seems fine

src/utility.cpp
faithokamoto: renamed

src/utility.hpp
faithokamoto: (passing everything inside of `cerrWrapper` is meant to let us change it later easily)

faithokamoto: though possibly it makes more sense to return a plain `cerr` for the ones that don't need to kill the program during self-destruct

src/utility.hpp
faithokamoto: Hey, I figured out a way to prevent overwriting. First I check if a file already exists. If so, then I use `access()` to check whether we have write permissions to it. If it doesn't already exist, then I use the original process. Same effect, but no need to actually open the file if it currently exists.

faithokamoto:
Something else I recall, as I slowly double-check all my file changes, is that I replaced some [runtime *errors*](https://github.com/vgteam/vg/blob/dfe612c34248a2b6615cd9ee84b298fb2e887b82/src/subcommand/msga_main.cpp#L616-L619) (as in the things you raise/[yeet](https://github.com/vgteam/vg/pull/2601)) with [vanilla program exits](https://github.com/vgteam/vg/blob/203906ecd10a516f206308165501dff41c31e155/src/subcommand/msga_main.cpp#L617-L619). Should those be changed back?

faithokamoto:
Hey, just over a month later, I think I've addressed your issues @adamnovak 
* Stuff got renamed (`require_exists()`, `ensure_writable()`, `parse_pair()` etc.) and the interfaces jiggered to make them easier to use (`parse_pair()` now uses templates, the logging functions have `<<` streaming that Just Works, `set_thread_count()` parses and sets the thread count, etc.)
* `ensure_writable()` doesn't overwrite files any more - if it detects that a file already exists, it just checks permissions
* No pretty printing because I couldn't figure out a good way to do it. Back to everything being on one line by default. (I considered modifying the magic streaming interface do auto-indent whenever the user streamed a newline, but not sure if that would be wanted.)
* Logging things moved to `log.cpp`/`log.hpp`
Basically, I think I addressed all the last round of comments. No idea why the Mac test failed though :(

adamnovak:
It looks like the mac tests immediately failed linting, but then went on to do a bunch of other stuff because linting and building are in parallel.

```
 src/subcommand/chains_main.cpp: Use require_exists() or ensure_writable() for standardized file checks: this->output_file = optarg;
make: *** [lint] Error 1
make: *** Waiting for unfinished jobs....
```

src/subcommand/view_main.cpp
```diff
@@ -504,7 +505,8 @@ int main_view(int argc, char** argv) {
                     // We match the tag, so dump this message.
                     if ((*it).second.get() != nullptr) {
                         if (verbose) {
-                            cerr << "Message of " << (*it).second->size() << " bytes in matches tag to extract" << endl;
+                            basic_log(context) << "Message of " << (*it).second->size()
```
adamnovak: Since `basic_log(context)` gets used a lot and is noticeably longer to type than `cerr`, and also has no side effects, maybe we want to set it up along with the context in each file? Something like:

```
const string context = "[vg view]";
auto log = basic_log(context);
```

Then you can log stuff with just `log << "My message" << endl;`

The downside is that's less consistent; error and warning calls need the context, and if info-level lines don't need it, people will forget it or accidentally pass it.

src/log.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * - basic_log(): do nothing but append a context string to the front of
+ *   the log message. This is useful for logging progress messages.
+ *   This is a generic logging utility.
+ * - warning(): emit a warning message with a given context
+ * - fatal_error(): emit an error message with a given context
+ * 
+ * Errors/warnings are output to std::cerr as:
+ * error<context>: message
+ * warning<context>: message
+ * where "context" is a string like "[vg inject]"
```
adamnovak: Would it make more sense for the brackets to be the responsibility of the logging code and not the caller? Do we ever not use them?

src/log.hpp
```diff
@@ -0,0 +1,79 @@
...
+    ~cerrWrapper() {
+        if (exit_on_destruct) {
+            exit(EXIT_FAILURE);
+        }
+    }
+};
+
+/// Log to cerr with a standard format
+/// "context" is caller context, e.g. "[vg inject]"
+cerrWrapper basic_log(const std::string& context);
```
adamnovak: What is the `basic_` part doing here? We don't really have an "advanced" version, and the other two of these (`warning` and `fatal_error`) correspond roughly to Python-style log levels. Maybe this wants to be `info`? Or in a namespace as `log::info`? Or I guess it could be plain `log`, since it is the "normal" log level that one would usually use.

I'm worried about having to constantly type `basic_` for no good reason.

src/utility.cpp
```diff
@@ -157,7 +157,7 @@ void choose_good_thread_count() {
             // TODO: If you have >1024 bits in your mask, glibc can't deal and you will get EINVAL.
             // We're supposed to then try increasingly large dynamically-allocated CPU flag sets until we find one that works.
             auto problem = errno;
-            std::cerr << "warning[vg]: Cannot determine CPU count from affinity mask: " << strerror(problem) << std::endl;
+            warning("vg") << "Cannot determine CPU count from affinity mask: " << strerror(problem) << endl;
```
adamnovak: Right now the caller is responsible for the `[]`, right? So this would give you `warningvg:`.

src/log.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * way you don't have to remember to exit after logging an error,
+ * and we can change format in a centralized place.
+ */
+
+namespace vg {
+
+// https://stackoverflow.com/a/25615354/
+class cerrWrapper {
+private:
+    std::ostream* str;
```
adamnovak: This reads as "string" to me, not "stream".

src/log.hpp
```diff
@@ -0,0 +1,79 @@
...
+        if (str) {
+            *str << t;
+        }
+        return *this;
+    }
+
+    cerrWrapper& operator<<(std::ostream& (*manip)(std::ostream&)) {
+        *str << manip;
+        return *this;
+    }
```
adamnovak, line 55: How would we ever get into a situation where `str` was null? The existence of the check suggests that a null `str` is allowed. But then the next function doesn't have a check, which then looks like a way to get a null dereference bug.

Why do we even bother with a pointer to `std::cerr` if we can *only* ever point to the global `std::cerr`? Why not just hardcode using it?

src/log.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+    template <typename T>
+    cerrWrapper& operator<<(const T& t) {
+        if (str) {
+            *str << t;
+        }
+        return *this;
+    }
+
+    cerrWrapper& operator<<(std::ostream& (*manip)(std::ostream&)) {
```
adamnovak, line 52: Do we need this because the template can't match a function pointer?

adamnovak (changes requested):
I think I found one place where the log message won't be quite right.

I think the logging framework makes sense. I'm not sure about naming what are essentially the 3 log levels `basic_log`, `warning`, and `fatal_error`, since those aren't clearly a set. And if there were a good way to get out of always passing the context that would be nice, but I'm not sure there is.

(Python as used in Toil has you set up a global `logger` object at the top of every file, with the file's name as context, and then you call `logger.info()` and `logger.warning()` and so on to log at different levels. But in vg we like to also include the function name. C++ has that in `__func__` but you would need some kind of macro to get at it and it wouldn't always be appropriate.)

src/log.hpp
faithokamoto: We need that to handle `std::endl`.

src/log.hpp
faithokamoto: good point that was a blind copy-paste issue. changed.

src/log.hpp
faithokamoto: Yeaaah I don't love that name either. But both `info` and `log` have name conflicts. Do you other suggestions? `log::info` is also a little long. `report()` has a name conflict in `giraffe_main.cpp`. etc.

src/log.hpp
faithokamoto: well I got rid of the variable anyhow.

src/log.hpp
faithokamoto: `print()` works - what do you think about that?

src/log.hpp
faithokamoto: probably, yeah. I'll change that

src/log.hpp
adamnovak: I think `print(const std::string&)` is worse because that would look weird when you use it as a stream.

adamnovak: I think the C++ approach would be to put all this stuff in a `vg::log` namespace and then `using vg::log;` to make it shorter when we want.

If I designed it I think I would have a `Logger` class you construct with the `const std::string& context`, and then in each file or in each function I would have `Logger log(...);` and then say `log.info() << ...`.

Or maybe I would have macros for the front end to capture the function name magically and I would say `INFO << ...`.

If we just want to name these things, though, I would go with `log_info`, `log_warning`, and `log_fatal_error`, so they all match but you can still tell fatal errors are magically fatal.

src/log.hpp
adamnovak: Well yes, but the question the code raised for me is why `std::endl` won't just have some (function pointer) type that the first `const T& t` template can be instantiated on.

I think the answer is that `std::endl` is itself a template and needs to use template type inference to pick up its template parameters from the argument type of the `operator<<` it is passed to. If the operator and `std::endl` both try to infer off each other, nobody knows what the stream character type is.

We could do something like having an `operator<<` that takes a *template* template parameter, alongside the one that takes a type, but if we have to have two copies anyway we might as well just make a concrete overload for the one thing we actually need.

src/log.hpp
faithokamoto: I figured out a way to give this the Python log like interface - now I set up a logger at the start of each e.g. `main_add()` as `Logger logger(context);` and then later you can use it like
```
logger.error() << "could not open " << vcf_filename << endl;
```
where `logger.error()` will automagically create a `cerrWrapper` object with the kill-program-on-destruct functionality. Currently the things are invoked as `.error()`, `.warn()`, and `.log()` because of difficulties with name overlap which I will look into fixing (for the error and warn ones) once I get everything switched over.

I think I've managed to iterate to basically remaking whatever Python logging system you used elsewhere.

src/subcommand/view_main.cpp
faithokamoto: moot because of logging framework changes

faithokamoto:
Agh and right after I'd merged master into my branch, something else gets merged into master and causes conflicts. I'll wait until you say what other changes might be needed before resolving, Adam.

src/utility.hpp
```diff
@@ -821,6 +821,50 @@ string file_base_name(const string& filename);
...
+/// or check the magic number.
+string require_non_gzipped(const Logger& logger, const string& filename);
+
+/// Check if a file can be written to and return its name (if so) or error.
+/// Uses file_can_be_written() and is intended to be called when parsing arguments.
+string ensure_writable(const Logger& logger, const string& filename);
+
+inline string ensure_writable(const string& context, const string& filename) {
+    return ensure_writable(Logger(context), filename);
+}
```
adamnovak: It is indeed a little weird that some of these take a Logger and some take a context string and some have overloads for both.

The user-facing feature is so we can see `[vg inject]` and not `[require_exists]` as the source of the error.

With the one-argument non-`explicit` constructor, is `Logger` not implicitly-constructable from `string`? Maybe we can always take `Logger` but sometimes pass `string`.

src/index_registry.cpp
```diff
@@ -319,7 +314,7 @@ vector<string> vcf_contigs(const string& filename) {
     
     htsFile* vcf = hts_open(filename.c_str(),"rb");
     if (vcf == nullptr) {
-        cerr << "error:[IndexRegistry] Could not open VCF" << filename << endl;
+        error(context) << "Could not open VCF: " << filename << endl;
```
adamnovak, line 317: I guess having `error()` exit helps us avoid cases like we're fixing here, where an error is detected but not acted on.

But now that this is so simple and in some ways so much like Python logging it's a little astonishing that that happens. Maybe we just have to see if it actually trips people up ever?

src/index_registry.cpp
```diff
@@ -516,9 +506,10 @@ construct_minimizers_impl(
     const MinimizerIndexParameters& params
 ) {
     if (IndexingParameters::verbosity != IndexingParameters::None) {
-        cerr << "[IndexRegistry]: Constructing minimizer index and associated zipcodes." << endl;
-        cerr << "    using parameters -k " << params.k << " -w " << params.w_or_s << (params.use_weighted_minimizers ? " -W " : "")
-            << (params.paths_in_payload ? " with paths" : "") << endl;
+        info(context) << "[IndexRegistry]: Constructing minimizer index and associated zipcodes." << endl;
+        info(context) << "    using parameters -k " << params.k << " -w " << params.w_or_s
```
adamnovak: This looks like the context conversion wasn't quite right.
```suggestion
        info(context) << "Constructing minimizer index and associated zipcodes." << endl;
        info(context) << "    using parameters -k " << params.k << " -w " << params.w_or_s 
```

adamnovak (changes requested):
I think I found one actual wrong line, and a couple things where there might not be anything to do.

src/index_registry.cpp
faithokamoto: I've added an all-caps bit to the docstring:
```
/// Error AND EXIT with a standard format
```
other than that, if someone's calling a function called `error()` and didn't read the docstring, and then is surprised that it led to an error, I think that's on them?

src/utility.hpp
faithokamoto: got rid of the overloads

faithokamoto:
And I have no clue how that non-standard `--help` description managed to sneak past `check_options.py` to be merged in.

src/subcommand/describe_main.cpp
```diff
@@ -145,7 +146,7 @@ int main_describe(int argc, char** argv) {
         std::string filename = argv[i];
         std::ifstream in(filename, std::ios_base::binary);
         if (!in) {
-            std::cerr << "error: [vg describe] cannot open file " << filename << std::endl;
+            logger.info() << "cannot open file " << filename << std::endl;
```
adamnovak, line 149: I might count these as warnings rather than info. But I'm not sure it's worth another round of tests.

adamnovak (approved):

PR #4690, opened 2025-09-03
Cyclic ziptree

faithokamoto:
## Changelog Entry

 * Zip code trees for `vg giraffe`'s chaining mode now have non-heuristic* distances in cyclic snarls [*intra-chain reversals are still not handled at all]
 * As a practical matter, we get significant speedups on hifi and R10 reads (especially for the slowest reads) and a tiny increase in read identity scores (though some increase and some decrease)
 * Adds distance index function `minimum_nontrivial_distance()` which is mostly the same as `minimum_distance()` but will use sidesteps to find a true self-loop distance instead of returning 0

## Description

I rewrote how zip code trees handle snarls in service of removing most of the heuristics we use in cyclic snarls. Notable changes are split apart below by phase. Many various quality-of-life things (e.g. better docstrings, or printing `inf` instead of the `size_t` max for edge distances) are not listed.

### Zip code tree construction

| Current code  | This PR | Reason |
|-|-|-|
| Cyclic snarl child chains are [heuristically ordered, duplicated, and reversed](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.cpp#L2888) based on the sequence, number, and orientation of seeds within the read. | One copy of each cyclic snarl child chain in used with arbitrarily order and orientation. | We traverse all child chains in both directions regardless of starting location. | DAG snarls use `DAG_SNARL_START`/`DAG_SNARL_END` and cyclic snarls use `CYCLIC_SNARL_START`/`CYCLIC_SNARL_END`. | Makes it easier to treat the snarl types differently. Root snarls are considered DAGs for this purpose. |
| A `NODE_COUNT` is [right before the snarl end](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.hpp#L90). | A `CHAIN_COUNT` is [right after the snarl start](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L109). | Necessary to read the distance matrix. |
| Snarl/chain [bounds do not use the `value` field](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.hpp#L127-L130) of `tree_item_t`. | Snarl and child chain bounds store [snarl ID in the `value` field](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L156-L160). The `ZipCodeTree` stores [`snarl_start_indexes`](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L279-L280), a `{ID : snarl start index}` lookup table. | IDs are necessary for [on-the-fly distance matrix lookups](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L1706) by child chains. |
| [Snarl distance matrices are broken up](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.hpp#L92-L93) by child. | [Snarl distance matrices](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L112-L125) are consolidated at the start of the snarl. | We can't guarantee direction of traversal inside cyclic snarls. |
| Cyclic snarls use the same kind of distance matrix as DAG snarls (relative to their Frankenstein-ed chains). | [Cyclic snarls](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L117-L125) store distances relative to _both_ ends of each child chain, plus three [self-loop distances](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L901-L905) per child chain. | *All* child chains are traversed, which includes the origin chain. |
| If the read has multiple minimizers with the same seed position, they are strung out in a chain with distances of `0` in between e.g. `pos/1 0 pos/2 0 pos/3`. | If a read has multiple minimizers with the same seed position, they are [collapsed into the same `tree_item_t` struct](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L326-L354) via its optional [`other_values` vector](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L166-L168)<sup>1</sup>. This is represented as `pos<1/2/3>`. | Reduce tree size & avoid irrelevant 0-distance transitions. |
| We track where bounds index-wise are during construction only, in the `forest_growing_state_t`'s `sibling_indices_at_depth`. | The `ZipCodeTree` stores [`bound_pair_indexes`](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L281-L282), a lookup table of `{index : index}` where, given an index of a snarl/chain bound, you can get the index of its matching bound. | This is necessary for "jumping" during parsing. |

<sup>1</sup> The comment here should read "Other values (for seeds with same position/orientation)" but I forgot to fix it and there's no way I'm making a new commit and updating all of my links.

## Parsing distances from a zip code tree

Updated flowchart is a [Google Drawing](https://docs.google.com/drawings/d/1diKMXCuMteR06fF64BhSttsD5RHh3eTnd8FtyYqs17Y/edit?usp=sharing).

| Current code  | This PR | Reason |
|-|-|-|
| Zip code tree iterators are backed by [actual](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.hpp#L239) [iterators](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.hpp#L294). | Iterators store a [reference](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L347) to the `vector<tree_item_t>` and an [index](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L399) into that vector. | Messy, but I need to have access to the index for "jumping". | 
| Iterators yield a single result for the seed at the current position. | Iterators yield a vector of results for the seed and any `other_values` at the current positions. | Easiest way to deal with handling all `other_values`. |
| [Seed-finding iterator](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.hpp#L208) (`iterator`) simply dootles along and yields each seed in turn. | [Seed-finding iterator](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L341) (`seed_iterator`) also has to keep track of child chain numbers and whether it's inside a cyclic snarl. | Necessary for distance matrix lookups when we exit a chain and find ourselves in a new snarl, and for knowing if we should send off the iterator in both directions in a cyclic snarl. |
| [Transition-finding iterator](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.hpp#L254) (`reverse_iterator`) always goes right to left from a starting seed. | [Transition-finding iterator](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.hpp#L420) (`distance_iterator`) [may go in either direction](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L2136), and may even change directions during parse. Notably, cyclic snarls are [explored](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L2084) left to right then right to left. | Must be able to get to all chains in both directions in cyclic snarls. |
| When the transition-finding iterator first enters a snarl, it stacks up all the distances it can see by [iterating until it hits a bound](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.cpp#L1949). | When the transition-finding iterator first enters a snarl, it stacks up the relevant distance via an [on-the-fly lookup](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L1706). | On-the-fly is necessary since we can't guarantee direction of traversal. |
| Chains are [skipped by iterating](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.cpp#L2064) through them without yielding anything. | Chains are [skipped by jumping](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L1812) to the index of the other bound as soon as we give up. | Optimization. |
| Snarls skip over (via [iterating while ignoring](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.cpp#L2051)) distances for later children if distances are already stacked. | We [jump](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L1893) [around](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L2068) [snarls](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L2110-L2116) when starting/finishing their parsing, in order to ignore distance matrices and the like. | Optimization. |
| [All seeds](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.cpp#L1886-L1895) encountered by the distance-finding iterator are yielded. | Only seeds with a [compatible orientation](https://github.com/vgteam/vg/blob/569faf941ddb18a81381867947bb7d98af55c7f8/src/zip_code_tree.cpp#L1969-L1990) are yielded. | Off-by-one errors caused by changing parsing directions were way too annoying. This way all direction changes at least cancel out. |


src/algorithms/chain_items.hpp
```diff
@@ -319,6 +319,35 @@ transition_iterator zip_tree_transition_iterator(const std::vector<SnarlDistance
...
+    const std::unordered_map<size_t, size_t>& seed_to_starting, 
+    const std::unordered_map<size_t, size_t>& seed_to_ending);
+
+/**
+ * Filter out transitions that can't be used, e.g. not reachable in the read.
+ * 
+ * Used as a helper by zip_tree_transition_iterator().
+ * 
+ * Transitions are taken as (source anchor, destination anchor, graph distance)
+ * and output as (source anchor, destination anchor, graph distance, read distance).
```
adamnovak, line 361: This means this isn't *just* a filtering operation; they also get elaborated a bit. It's not *great* to have one function do two things, but maybe it makes sense here.

src/algorithms/chain_items.cpp
```diff
@@ -214,186 +222,284 @@ transition_iterator zip_tree_transition_iterator(const std::vector<SnarlDistance
...
+    }
+    pos_t cur_pos;
+    string rev_string = "";
+    for (const auto& cur_anchor : dest_anchors) {
+        auto dest_seed = cur_anchor.first;
+        std::cerr << " (S" << dest_seed.seed << "/anchor #" << cur_anchor.second << ")";
+        cur_pos = seeds[dest_seed.seed].pos;
+        rev_string = dest_seed.is_reversed ? "rev" : "";
+    }
+    std::cerr  << " at " << cur_pos << rev_string << std::endl;
```
adamnovak: I would indent these debug-printing blocks at the level of the surrounding code, not 1 level below the ifdef.

Usually we've been putting the ifdefs at zero indentation to show that they happen in their own pass. But we could indent them in as if they were runtime control structures instead.

src/algorithms/chain_items.cpp
```diff
@@ -214,186 +222,284 @@ transition_iterator zip_tree_transition_iterator(const std::vector<SnarlDistance
...
+
+    ZipCodeTree::seed_iterator dest = zip_code_tree.begin();
+    bool right_to_left = true;
+    while (dest != zip_code_tree.end()) {
+        // For each destination seed left to right
+        vector<ZipCodeTree::oriented_seed_t> dest_seeds = *dest;
+
+        if (!right_to_left) {
+            for (auto& cur_dest : dest_seeds) {
+                // Going backwards
```
adamnovak: The iteration logic here is fiddly and it's (EDIT: **not**) immediately obvious whether `right_to_left` e.g. applies to this "For each destination seed left to right" loop we're just starting.

I think what's going on is, we go left to right, and then at each item `dest` we do a right to left pass (which is "forwards"), and then a not-right-to-left pass (which is "backwards"), if it's in a cyclic snarl. If it's not, we do a right-to-left pass only. Which passes we do is controlled by some logic at the end of the `while` loop that either advances to the next `dest` or sets up for a not-right-to-left pass on the current `dest`.

This would be clearer if it was less dependent on having done exactly the right thing already, and the control logic was all brought into one place with some idiomatic `for` loops.

It also seems more idiomatic to do your orientation flag as false and *then* true, if you can manage it. But here we already have a convention of the normal passes being backward (right to left) from the old "reverse" iterator. So maybe we do want `right_to_left` as the flag and to run it through `true`, then `false`? But then `true` (right to left) should be "Going backwards" probably? Maybe we should throw the concepts of "backwards" and "forwards" out of this loop if we can, and have only "left to right" and "right to left".

Maybe:
```
for (ZipCodeTree::seed_iterator dest = zip_code_tree.begin(); dest != zip_code_tree.end(); ++dest) {
    // Visit every possible dest left to right
    for (bool right_to_left_pass : (dest.in_cyclic_snarl() ? {true, false} : {true})) {
        // In cyclic snarls we do a right to left pass and a left to right pass. In non-cyclic snarls we only do a right to left pass
        ...
```


I'm not really sure you can ternary on initializer lists like that, but even building the little bool vector by hand here would make more sense.

src/algorithms/chain_items.cpp
```diff
@@ -214,186 +222,284 @@ transition_iterator zip_tree_transition_iterator(const std::vector<SnarlDistance
...
+    std::vector<std::tuple<size_t, size_t, size_t>> all_transitions;
+
+    ZipCodeTree::seed_iterator dest = zip_code_tree.begin();
+    bool right_to_left = true;
+    while (dest != zip_code_tree.end()) {
+        // For each destination seed left to right
+        vector<ZipCodeTree::oriented_seed_t> dest_seeds = *dest;
+
+        if (!right_to_left) {
+            for (auto& cur_dest : dest_seeds) {
```
adamnovak: If we already have a `dest` we probably don't also want a `cur_dest`; this could just be `seed` or something.

src/algorithms/chain_items.cpp
```diff
@@ -214,186 +222,284 @@ transition_iterator zip_tree_transition_iterator(const std::vector<SnarlDistance
...
+                // Going backwards
+                cur_dest.is_reversed = !cur_dest.is_reversed;
```
adamnovak: It looks like you add a doc comment for this field as `/// Is the seed traversed backwards in the tree?`.

That still sounds like it's meant to be something about the seed and not scratch space for toggling depending on which orientation of traversal we're thinking about at the moment.

If you just toggle it here that's going to affect all the prints that use this. Is that really correct?

Can there be something here to explain why it's correct to flip all the oriented seeds around, rather than using "is the seed reversed" and "are we doing a right-to-left traversal" together later?

src/algorithms/chain_items.cpp
```diff
@@ -214,186 +222,284 @@ transition_iterator zip_tree_transition_iterator(const std::vector<SnarlDistance
...
-            // The zipcode tree is about point positions, but we need distances between whole anchors.
-            // The stored zipcode positions will be at distances from the start/end of the associated anchor.
-            
-            // If the offset between the zip code point and the start of the destination is 0, and between the zip code point and the end of the source is 0, we subtract 0 from the measured distance. Otherwise we need to subtract something.
-            size_t distance_to_remove = dest_anchor.start_hint_offset() + source_anchor.end_hint_offset();
-
+                        if (!cur_source.is_reversed && !cur_dest.first.is_reversed) {
+                            // Both of these are in the same orientation relative to
+                            // the read, and we're going through the graph in the
+                            // read's forward orientation as assigned by these seeds.
```
adamnovak: Doesn't `right_to_left` matter for the justification here?

src/snarl_distance_index.hpp
```diff
@@ -19,6 +19,13 @@ using namespace bdsg;
 //Minimum distance taking a pos instead of id/orientation/offset
 size_t minimum_distance(const SnarlDistanceIndex& distance_index, pos_t pos1, pos_t pos2,
                         bool unoriented_distance = false, const HandleGraph* graph=nullptr); 
+//Minimum distance but if the positions are the same, a true self-loop distance
+//is calculated (i.e. it doesn't default to 0, and may be infinite)
+//This is assumed to always be directed
+//The length of one node may be passed for a minor speed improvement
+size_t minimum_distance_self_nonzero(const SnarlDistanceIndex& distance_index, pos_t pos1, pos_t pos2,
+                                     size_t pos2_length = std::numeric_limits<size_t>::max(),
+                                     const HandleGraph* graph=nullptr);
```
adamnovak: This should probably have `///` comments so Doxygen picks them up.

I'm a little worried about the names `minimum_distance` and `minimum_distance_self_nonzero` as a pair. I gusee for the `_self_nonzero` one it *always* guarantees that the distance from a position to itself is nonzero? (Or at least it does as long as no empty nodes have snuck in?) But it still seems like there should be a word that describes the definition of minimum distance that is being computed, and not the properties of the result. Maybe `minimum_nontrivial_distance` or something?

src/zip_code_tree.hpp
```diff
@@ -74,109 +94,143 @@ class ZipCodeTree {
...
         size_t value : 59;
 
-        //For seeds, is the position of the seed traversed backwards in the tree?
+        /// Is the position of the seed traversed backwards in the tree?
+        /// Ignored for non-seeds and should be set to false
         bool is_reversed;
 
+        /// Other values/reversedness (for seeds on the same position)
+        /// Not allocated unless necessary
+        vector<size_t>* other_values = nullptr;
```
adamnovak: This now looks a little silly because we've carefully packed the type, value, and reverse-ness flag into 64 bits to wring out as much space as possible, and then we've tacked on a 64 bit pointer to *every* entry (even the snarl/chain bounds and the edge data entries) which as far as I can tell is going to be empty most of the time. Presumably overall the PR still reduces tree-size-plus-extra-values-size *in bytes* by not needing to duplicate cyclic snarl contents and by getting rid of zeroes between coinciding seed entries, but this will double the size of what's left.

Are you sure we can't go back to inlining the values, but have the logic scoop up a whole run of adjacent symbols (or a bunch of symbols with a "next one too!" flag set) instead of following a pointer to get the values?
Or maybe we could look up the extra values in a map by symbol index, instead of pointing to them?
Or maybe we could take all the extra-values vectors and store them in a vector themselves, so they all get assigned small numbers, and in `value` here we could pack both the seed index and the index of the extra values vector if any.

src/zip_code_tree.hpp
```diff
@@ -74,109 +94,143 @@ class ZipCodeTree {
...
                    : value;
         }
         bool get_is_reversed() const { return is_reversed; }
+        /// Check whether other_values is allocated
+        bool has_other_values() const { return other_values != nullptr; }
+        /// Add an extra value to the other_values vector
+        /// If other_values is not allocated, it will be allocated
+        void add_extra_value(size_t extra_value) {
+            if (!other_values) {
+                other_values = new vector<size_t>();
```
adamnovak: This looks like `tree_item_t` itself is meant to own this memory: it allocates it internally. But it doesn't use anything like `std::shared_ptr` or `std::unique_ptr`, or implement its own non-default copy/move/assignment/destructor logic. And I don't see anything coming by to do the delete on this. So I *think* this data just never gets deallocated and leaks.

(If this doesn't leak, we should mention in the doc comment for the field where it is meant to be being freed.)

Right now we pass around `tree_item_t` like a nice simple small value type; I think we even have an `inline tree_item_t something_or_other()` that returns one by value. If it's going to own and manage memory itself (instead of having someone responsible for freeing all the allocated extra value vectors when we're done with a zip code tree), that passing becomes more expensive because it needs to do copying of the vectors (if each `tree_item_t` owns its own vector) or reference counting (if copies of an "original" `tree_item_t` all point to the same one, but collectively own it).

faithokamoto:
> Also, if you somehow didn't need to put the snarl ID in the bounding symbols, you could put the offset to the other bound in the bounding symbols, and maybe get out of needing the other-bound lookup map. Whichever of those gets used more should probably be the one that's inline.

That was the first thing I tried - adding a field for pair index to the bound itself. The problem is that it's not uncommon to be moving things _around_ the ziptree. The two cases are when I add a distance matrix and when I splice off a section. In both cases, I need to then iterate through all of the ziptree that I'm moving in order to update the memorized indexes. (I _might_ be able to get away with storing `bound_pair_indexes` separately during construction, and then tacking all of them on to the complete zip trees, but I've only just thought of that.)

src/zip_code_tree.hpp
faithokamoto: Yeah, I could have a run of seeds without distances in between mean that the distance is 0. That probably makes more sense.

src/snarl_distance_index.hpp
faithokamoto: Should I convert all of the other comments in [that file](https://github.com/vgteam/vg/blob/master/src/snarl_distance_index.hpp) to `///`? My use of `//comment` was to maintain the same style.

src/zip_code_tree.hpp
faithokamoto: closing as irrelevant since I'm moving to have the seeds be in a line in the tree itself.

src/algorithms/chain_items.cpp
faithokamoto: That comment is a shortening of the [current version](https://github.com/vgteam/vg/blob/3a1c22a7c502d5a12d40e1d3ef05c7bfef71b0f6/src/zip_code_tree.hpp#L132C9-L132C82) which says `//For seeds, is the position of the seed traversed backwards in the tree?`

I imagine it like a row of tiki statues with faces on one side and backs on the other. `is_reversed` is meant to tell you which direction the face is. Specifically, whether it would be reversed if you walked down the row of statues from right to left (the original default). But if you're walking down the row of statues from left to right, then the relative orientations will be flipped, so I have to flip `is_reversed`.

It could probably do with a better comment and name. Maybe something like `facing_right`.

faithokamoto:
I got rid of `other_values` and now multiple seeds on the same position get stored in a line with no edges between them. `seed_itr` is smart enough to crawl along this line and then return a vector of all the seeds in the same position. `dist_itr` starts from the end of the line and then dootles along, being careful to not yield any seeds at a distance of 0 (which could never be used anyhow so it's not a loss).

In addition, I dealt with a lot of the "iterating over this is so fiddly" issues not necessarily by un-fiddling the iteration but by wrapping it into a clean, simple function which handles everything for the user.

https://github.com/vgteam/vg/blob/dd36a554f95c01da3ad03de29c2c2ff265969574/src/zip_code_tree.cpp#L2106-L2135

This function is used for all three places that need to iterate over the zip code tree (the [unit tests](https://github.com/vgteam/vg/blob/dd36a554f95c01da3ad03de29c2c2ff265969574/src/unittest/zip_code_tree.cpp#L45), the [self-validation](https://github.com/vgteam/vg/blob/dd36a554f95c01da3ad03de29c2c2ff265969574/src/zip_code_tree.cpp#L1321), and the [chaining](https://github.com/vgteam/vg/blob/dd36a554f95c01da3ad03de29c2c2ff265969574/src/algorithms/chain_items.cpp#L268)). Plus, since this new function doesn't have to _do_ anything with the transitions it finds, it's much simpler. There's a minor amount of non-DRYness due to me essentially copy-pasting the distance iteration code to send it in the opposite direction within cyclic snarls. However, I think this is excusable because it does away with a whole lot of conditional logic.

src/snarl_distance_index.hpp
adamnovak: Probably we want to use `/// ` or `/** ... */` comments for doc comments everywhere. It looks like the comments in there are mostly intending to be doc comments, even if they could be elaborated into better ones.

src/algorithms/chain_items.cpp
```diff
@@ -214,186 +222,233 @@ transition_iterator zip_tree_transition_iterator(const std::vector<SnarlDistance
...
+    const std::vector<SnarlDistanceIndexClusterer::Seed>& seeds,
+    const ZipCodeTree& zip_code_tree,
+    size_t max_graph_lookback_bases,
+    const std::unordered_map<size_t, size_t>& seed_to_starting, 
+    const std::unordered_map<size_t, size_t>& seed_to_ending) {
+
+    std::vector<std::tuple<size_t, size_t, size_t>> all_transitions;
+
+    ZipCodeTree::seed_iterator dest = zip_code_tree.begin();
+    bool right_to_left = true;
```
adamnovak: Now this isn't used at all.

faithokamoto: whoops.

faithokamoto:
Okay, so I've tried a few ways of centralizing the iteration logic and concluded that the overhead of giving it its own function/iterator/etc. is too much; that's where a decent amount of the slow-down is coming from. I'll have to go back to iterating separately in the three places and just try to make the iteration logic easier to follow.

faithokamoto:
All the code review things were re-fixed. However, since I decided that it was too much overhead to outsource the iteration logic, instead now I just try to simplify it. Specifically, the poor `seed_iterator` now has to handle both which direction to send the distance iterator off in, as well as dealing with orientations of the dest seeds. 

https://github.com/vgteam/vg/blob/9e01e2c942905bc69554e957cce4af3513033831/src/zip_code_tree.cpp#L1451-L1473

Hiding this in the `seed_iterator` means that outside iterator-using code can just increment both iterators as normal and trust that anything yielded by either is pointing in the right direction.

faithokamoto:
ugh no clue why that failed.

faithokamoto:
Oh, I should write some notable changes since when I first PR'd this thing:
* We now have memorization! The `distance_iterator` keeps a hashmap of {index where we enter a chain : best running distance so far}, and if it encounters the same chain/orientation but with an equal or worse running distance, it skips the chain. This is relevant in nested cyclic snarls.
* `bound_pair_indexes` is no more. Instead, there's a shiny new field [`section_length`](https://github.com/vgteam/vg/blob/8a0e9e6269b81f2b750e4817adb0a29c07d98834/src/zip_code_tree.hpp#L164) which stores the length of the section between two paired bounds. In practice this is set in [`add_close_bound()`](https://github.com/vgteam/vg/blob/8a0e9e6269b81f2b750e4817adb0a29c07d98834/src/zip_code_tree.cpp#L69-L71) since we can guarantee that once a thing is closed it's not changing size. It's then used via [`get_other_bound_offset()`](https://github.com/vgteam/vg/blob/8a0e9e6269b81f2b750e4817adb0a29c07d98834/src/zip_code_tree.hpp#L222-L235), which returns a number you can add to what is presumably a given bound's index, in order to get the index of its pair bound. Reducing hashmap lookups gives a slight speedup to `tick()`.
* `snarl_start_indexes` is also no more. Instead the `value` field (which used to store a snarl ID for snarl & child chain bounds) is used to store a [distance to the snarl start](https://github.com/vgteam/vg/blob/8a0e9e6269b81f2b750e4817adb0a29c07d98834/src/zip_code_tree.hpp#L157-L159). This is set [when a snarl is closed](https://github.com/vgteam/vg/blob/8a0e9e6269b81f2b750e4817adb0a29c07d98834/src/zip_code_tree.cpp#L646-L647) by the same principle of things no longer moving around. Reducing hashmap lookups gives a slight speedup to `tick()`.
* I overloaded the [bool field of `tree_item_t`](https://github.com/vgteam/vg/blob/8a0e9e6269b81f2b750e4817adb0a29c07d98834/src/zip_code_tree.hpp#L166-L170) so now it stores `is_reversed_or_cyclic`. Only seeds can be reversed, and only bounds will care about cyclic-ness, and it was easier to have a single bool field for the constructors and such. The outside code doesn't need to care that this field is overloaded and can simply call the relevant [overlaoded getter](https://github.com/vgteam/vg/blob/8a0e9e6269b81f2b750e4817adb0a29c07d98834/src/zip_code_tree.hpp#L214-L215) when needed. Telling all bounds what kind of cyclic-ness they are means that we don't have to look up the snarl start each time we exit a chain and switch to a snarl-scanning state (`S_SCAN_DAG_SNARL`/`S_SCAN_CYCLIC_SNARL`), improving memory co-locality.
* Since now the snarl bounds store their cyclic-ness in a field, I don't need separate types for DAG/cyclic snarl starts/ends. So now everything is back to just being a `SNARL_START` or `SNARL_END`. I maintained the [differential printing of snarl bounds](https://github.com/vgteam/vg/blob/8a0e9e6269b81f2b750e4817adb0a29c07d98834/src/zip_code_tree.cpp#L23-L34) based on cyclic-ness because it makes debugging easier.

src/zip_code_tree.hpp
```diff
@@ -74,109 +94,166 @@ class ZipCodeTree {
...
+        /// Constructor to set a "false" for is_reversed_or_cyclic
+        tree_item_t (tree_item_type_t type, size_t raw_value) 
+            : tree_item_t(type, raw_value, false) {}
+        /// Constructor to set max for the value
+        tree_item_t (tree_item_type_t type, bool is_reversed_or_cyclic) 
+            : tree_item_t(type, std::numeric_limits<size_t>::max(), is_reversed_or_cyclic) {}
+        /// Constructor for just a type
+        tree_item_t (tree_item_type_t type) 
+            : tree_item_t(type, std::numeric_limits<size_t>::max(), false) {}
+        /// Setters
```
adamnovak: If this is meant to be about more than one thing, it should be a plain comment. Doc comments (`///`) get picked up by Doxygen as documentation for the immediate next thing, so this is documenting `set_value()` as "Setters" and saying nothing about `set_section_length()`.

src/zip_code_tree.hpp
```diff
@@ -74,109 +94,166 @@ class ZipCodeTree {
...
         size_t get_value() const { 
-            return value == ((size_t)1 << 59) - 1
-                   ? std::numeric_limits<size_t>::max()
-                   : value;
+            return value == fake_max() ? std::numeric_limits<size_t>::max()
+                                       : value;
+        }
+        /// Different getters based on context for readability
+        bool get_is_reversed() const { return is_reversed_or_cyclic; }
+        bool get_is_cyclic() const { return is_reversed_or_cyclic; }
```
adamnovak: These should be documented to say what kind of thing they're allowed to be used on. Especially if they aren't going to carry any internal assertions that they are being used right.

src/zip_code_tree.hpp
```diff
@@ -74,109 +94,166 @@ class ZipCodeTree {
...
-        bool is_reversed;
+        /// For a bound, how long the internal section is
+        /// e.g. a chain with one seed would be "1"
+        size_t section_length : 59;
+
+        /// For a bound, if we're walking through the tree from right to left
+        /// (the default), will we traverse this position backwards?
+        /// Or, for a bound is the snarl or parent snarl cyclic?
+        /// Ignored for EDGE/CHAIN_COUNT and should be set to false
+        bool is_reversed_or_cyclic;
```
adamnovak: This says "For a bound" twice; isn't the first one "For a seed" instead?

src/zip_code_tree.hpp
```diff
@@ -74,109 +94,166 @@ class ZipCodeTree {
...
+
+        /// A fake "max" value for internal use
```
adamnovak: What's fake about it and how does having this help internally?

It looks like this is the max value of the `value` field, so it's not "fake" so much as it is "narrow".

src/zip_code_tree.hpp
```diff
@@ -74,109 +94,166 @@ class ZipCodeTree {
...
+        size_t get_other_bound_offset() const {
+            if (section_length == fake_max()) {
+                throw std::runtime_error("Can't get other bound offset of a tree item with no section length");
+            }
+            if (this->type == ZipCodeTree::SNARL_START
+                || this->type == ZipCodeTree::CHAIN_START) {
+                return section_length + 1;
+            } else if (this->type == ZipCodeTree::SNARL_END
+                       || this->type == ZipCodeTree::CHAIN_END) {
+                return -(section_length + 1);
```
adamnovak, line 247: Is this meant to be negative? Because `size_t` is unsigned.

If this function is meant to return negative offsets to go left, then it shouldn't talk about "subtracting" in the docstring (since the caller will always add), and it should return a signed type (maybe `int64_t`).

src/zip_code_tree.hpp
```diff
@@ -74,109 +94,166 @@ class ZipCodeTree {
...
+            } else {
+                throw std::runtime_error("Can't get other bound offset of a tree item that isn't a bound");
+            }
+        }
+        /// We should never need to set reversedness after the fact,
+        /// so the setter is only provided with the is_cyclic name
```
adamnovak: This is a note about a function that *isn't* here, more than it is documentation intended for someone who wants to call the function that *is* here.

src/zip_code_tree.hpp
```diff
@@ -329,99 +486,180 @@ class ZipCodeTree {
...
+        void stack_matrix_value(size_t matrix_start_i, bool has_main_diagonal, size_t row, size_t col);
+
+        // Helper functions for the automaton's state machine
+
         /// Current state of the automaton
         State current_state;
 
         /// Adopt a new state.
-        void state(State new_state);
+        void state(State new_state) { current_state = new_state; }
```
adamnovak: I think if you put function bodies in the header, you need to make sure they are `inline` functions or that you're inside a template class. Otherwise if you include the header in two compilation units you get duplicate implementations in two `.o` files and they aren't allowed to link together. Only `inline` function definitions are allowed to be duplicated like that.

src/zip_code_tree.hpp
```diff
@@ -596,286 +835,310 @@ class ZipCodeForest {
...
+        size_t rank;
+        /// If this seed is in a nested snarl, and we're calculating
+        /// using minimum_distance, then we might need to subtract the
+        /// offset from inner seed to inner snarl edge
+
+        size_t nested_snarl_offset;
+
+        // Pass the seed's index (which is looked up from forest_state.seeds),
+        // whether its position should be reversed, and then a few raw values
+        // In addition, snarl depth & forest_state are used to look up seed info
```
adamnovak: This looks like it wanted to be a doc comment.

src/zip_code_tree.cpp
```diff
@@ -1509,612 +1364,682 @@ void ZipCodeForest::validate_zip_forest(const SnarlDistanceIndex& distance_index
...
+    if (cyclic_snarl_nestedness > 0) {
 #ifdef debug_parse
-    std::cerr << "From " << &*it << " there are " << to_return << " slots after" << std::endl;
+        std::cerr << "Reversing direction in cyclic snarl at seed "
+                  << current_item().get_value() << std::endl;
+#endif
+        if (right_to_left) {
+            // If we're going right to left and in a cyclic snarl,
+            // then we just flip direction and stay at the same seed
+            right_to_left = false;
```
adamnovak: I don't understand how this works; it looks like we'd toggle `right_to_left` before visiting *every* seed in e.g. a chain with a few seeds along it, inside at least one level of cyclic snarl. Are we meant to do that? I feel like there should be an explanation for why the first step of advancing to the next seed inside a cyclic snarl is toggling the direction, instead of doing it at like the end of something.

src/zip_code_tree.cpp
```diff
@@ -1509,612 +1364,682 @@ void ZipCodeForest::validate_zip_forest(const SnarlDistanceIndex& distance_index
...
+        std::cerr << "Tried to initialize a root-level snarl; halting now." << std::endl;
+#endif
+        halt();
+        return false;
+    }
+
+    // Grab distances for this snarl
+    size_t snarl_start_i = index - current_item().get_value();
+    bool is_cyclic = zip_code_tree.at(snarl_start_i).get_is_cyclic();
+    bool original_right_to_left = right_to_left;
```
adamnovak: This is shadowing a member with the same name, which is often confusing. Is this the same sense of "original" as the member variable? Was it actually meant to be a write to the member variable and someone stuck `bool` in here to make it compile because the member is const?

It maybe should have a different name.

src/zip_code_tree.cpp
```diff
@@ -1509,612 +1364,682 @@ void ZipCodeForest::validate_zip_forest(const SnarlDistanceIndex& distance_index
...
+        std::cerr << "Start with leftmost chain" << std::endl;
+#endif
+        // We want to start with the leftmost chain
+        // Jump to CHAIN_COUNT
+        index = snarl_start_i + 1;
+        // Distance matrix has chains & bounds
+        size_t node_count = current_item().get_value() + 1;
+        // Skip past distance matrix
+        index += is_cyclic ? (node_count*2 * (node_count*2 + 1)) / 2
+                           : (node_count * (node_count + 1)) / 2;
```
adamnovak, line 1832: It looks like here and for `within_matrix_i` we're using some dense, similar-but-not-identical math that I kind of have to just hope is the right math for how big the matrices will end up being.

src/zip_code_tree.hpp
```diff
@@ -329,99 +486,180 @@ class ZipCodeTree {
         /// Reverse the top two elements of the stack
         void swap();
 
+        // Methods to look up distances to stack\
```
adamnovak:
```suggestion
        // Methods to look up distances to stack
```

src/zip_code_tree.hpp
```diff
@@ -203,110 +325,145 @@ class ZipCodeTree {
...
-     * Iterator that looks left in the tree from a seed, possibly up to a maximum base distance.
+     * Iterator that looks sideways in the tree from a seed,
+     * possibly up to a maximum base distance.
      *
-     * See https://github.com/benedictpaten/long_read_giraffe_chainer_prototype/blob/b590c34055474b0c901a681a1aa99f1651abb6a4/zip_tree_iterator.py.
+     * Original: 
+     * https://github.com/benedictpaten/long_read_giraffe_chainer_prototype/blob/b590c34055474b0c901a681a1aa99f1651abb6a4/zip_tree_iterator.py.
+     * 
+     * New cyclic snarl handling: 
+     * https://docs.google.com/drawings/d/1diKMXCuMteR06fF64BhSttsD5RHh3eTnd8FtyYqs17Y/edit?usp=sharing
```
adamnovak: Is there a writeup of the approach to iterating over cyclic snarls, besides what's in the PR comments? Someone might come here for a general overview of the approach, and while e.g. the diagram is helpful, it doesn't answer questions like "should we really be toggling orientation first thing in `operator++`?".

adamnovak (changes requested):
I think the design here is right now.

I found some places where the comments don't seem quite right, and it looks like the core `operator++()` is flipping the traversal direction at every step for some reason, which I couldn't find a justification for. Also I think large unsigned numbers are getting used for leftward offsets by abusing overflow.

src/zip_code_tree.cpp
faithokamoto: Whoops, I just needed that because the logic for `stack_snarl_distances` needs to know which direction the iterator was pointing to start the function. renamed to `start_right_to_left`

src/zip_code_tree.hpp
faithokamoto: yeah that should be an int

src/zip_code_tree.hpp
faithokamoto: I meant "fake" in relation to how it's not `std::numeric_limits<size_t>::max()` but we treat one as the other for getting/setting; the outside code doesn't know about the value limit of the internal field. Renamed to `internal_max()`

src/zip_code_tree.hpp
faithokamoto: added more documentation and also sanity checks during calling

src/zip_code_tree.cpp
faithokamoto: I'm sticking more comments in these places. Trust me, this is the easier version of the logic :o it's all based around triangular numbers. Here's the comment for that spot:
```
        // Skip past distance matrix with N chains and thus N+1 nodes
        // A cyclic snarl's matrix has 2 rows per chain plus 2 for bounds,
        // and it also stores self-loop distances (i.e. main diagonal)
        // That means (N+1)*2 rows, so the matrix is the (N+1)*2 triangle number
        // A DAG snarl's matrix has 1 row per chain plus 2 for bounds
        // and it does NOT store self-loop distances (i.e. no main diagonal)
        // That means (N+1) rows, so the matrix is the (N+1) triangle number
```

src/zip_code_tree.cpp
faithokamoto: it's the second time we see this particular seed; I've tried to put an explanation.

src/zip_code_tree.hpp
faithokamoto: made an explanation

PR #4694, opened 2025-09-05
Helptext fix

faithokamoto:
## Changelog Entry

 * Minor helptext copyedits

## Description

Just noticed a few things in the helptext that weren't automatically caught with my script.


faithokamoto:
Looking into the error, it really doesn't seem to be my code's fault? Far as I can tell, the local-build-test-job (and only that) crashed for completely unrelated reasons. The unit test that the code claims to have failed isn't something that the helptext would have touched at all. Plus it passes when I run on the cluster.

PR #4699, opened 2025-09-12
Add the capability of producing supplementary alignments to vg surject

eizengaj-roche:
## Changelog Entry

 * VG mapping tools can now produce supplementary alignments for SAM/BAM output

## Description

There's a good amount of code changes here. All of them affect the surject-based alignment output only, in which they add the capability to produce supplementary alignments. The supplementary alignments are all derived from the graph-space alignment. That is, they are contiguous alignments in graph space that become largely separated when projected onto the reference. This PR does not add the capability to perform graph-space chimeric alignments to any of the VG mapping tools.

Probably the biggest payoff of this PR is that it more-or-less eliminates the issue where some reads are lost during the surject step because they would require an implausibly large realignment between some reference-overlapping anchors. With supplementary alignments activated, these tend to be split into a primary and supplementary alignment.

In all subcommands, the supplementary alignments are off by default and need to be activated by a flag. I tested these features mostly with short reads, but I tried to come up with defaults for key parameters that would make sense even with long reads. It might require some testing and adjustment to get supplementaries to work well with long reads.

One source of friction in the implementation is our I/O interface for paired-end reads. This interface assumes that a read pair produces equal numbers of alignments for each end, so that they can all be emitted as pairs. Supplementary alignments break this assumption, since each end can generate additional supplementary alignments independently. Currently, I am handling this by adding logic in the main functions and expanding the single-end I/O functions so that they can emit single alignments that have a mate. There's an awkward fact that the SAM/BAM output still needs to know information from the mate for some SAM fields. I worked around this rather inelegantly using the annotation system, but a better long-term solution would probably be to generalize the AlignmentEmitter interface to properly handle these cases.


benedictpaten:
Nice! Do you have any idea how much this improves variant calling? Were you
focused on SV discovery? I can imagine this would also be useful for
RNA-seq mapping to a pangenome containing splice junctions.

jeizenga:
I think SV discovery is the most obvious application. I could see it also improving small variant calling around the breakpoints of SVs that were previously causing us to lose reads during surject, but I don't have any data on that. I think RNA-seq probably stands to benefit a bit less, since there's already machinery to chunk up the surject problem on exons so that you don't have to realign across (potentially large) introns.

benedictpaten:
got it, thanks!

PR #4702, opened 2025-09-17
Add subpaths when extracting graph using GAM

adamnovak:
## Changelog Entry

 * `vg find -G` now includes regions of paths touched by the extracted graph

## Description
I noticed a while ago that these path regions were missing and made this change on a branch I eventually threw away. Here it is by itself.

It could perhaps use a test to prove it works.


faithokamoto:
re: "It could perhaps use a test to prove it works." - one of the things on my side project to-do list is to add a check for whether all command-line options are under test in the `test/t/` scripts. Would _requiring_ all options to be used during testing be useful? Or annoying?

PR #4703, opened 2025-09-19
Lr giraffe recombinations

dcmonti:
## Changelog Entry

 * VG Giraffe now implements a recombination aware chaining algorithm

## Description
Most of the code changes are on the Minimizer Index with the definition of a new Payload type for storing path information. Path information has also been added to seed, anchor and chains in order to compute a recombination penalty during the chaining step.  There are two different penalties, one for the fragmenting step and the other one for the chain. 

By default the recombination option is not set and can be activated from CLI, together with the recombination penalties that are by default set to 0.

The main problem is that this PR requires to reconstruct the minimizers and zipcodes indexes. They use a different extension so it shouldn't create any conflict with previous indexes. The option to build this type of minimizer index is not yet implemented in the autoindex command.

### DIFF
--- a/src/subcommand/giraffe_main.cpp
+++ b/src/subcommand/giraffe_main.cpp
@@ -1670,7 +1670,11 @@ int main_giraffe(int argc, char** argv) {
         cerr << "error: [vg giraffe] Paired-end alignment is not yet implemented for --align-from-chains or chaining-based presets" << endl;
         exit(1);
     }
-
+    if (use_path_minimizer && !map_long_reads) {
+        // We don't have file paths to load defined for recombination-aware short-read minimizers.
+        cerr << "error: [vg giraffe] Path minimizers cannot be used with short reads." << endl;
+        return 1;
+    }
     bool haplotype_sampling = !haplotype_name.empty() & !kff_name.empty();
     if (!index_basename_override.empty()) {
         index_basename = index_basename_override;
@@ -1812,15 +1816,9 @@ int main_giraffe(int argc, char** argv) {
         cerr << "Loading Minimizer Index" << endl;
     }
     unique_ptr<gbwtgraph::DefaultMinimizerIndex> minimizer_index;
-    unique_ptr<gbwtgraph::MinimizerIndexXL> path_minimizer_index(new gbwtgraph::MinimizerIndexXL());
+    unique_ptr<gbwtgraph::MinimizerIndexXL> path_minimizer_index;
     if (map_long_reads) {
         if (use_path_minimizer) {
-            // FIXME: This is a temporary hack to load the path minimizers, I can't find a way to use VPKG to load them.
-            // auto path = registry.require("Long Read PathMinimizers").at(0);
-            // ifstream path_in(path, std::ios::binary);
-            // if(!path_in) throw std::runtime_error("cannot open " + path);
-            // path_minimizer_index->deserialize(path_in);
-            // path_in.close();
             path_minimizer_index = vg::io::VPKG::load_one<gbwtgraph::MinimizerIndexXL>(registry.require("Long Read PathMinimizers").at(0));
         } else {
             // Use the long read minimizers
@@ -1909,11 +1907,13 @@ int main_giraffe(int argc, char** argv) {
     unique_ptr<MinimizerMapper> minimizer_mapper_ptr;
 
     if (use_path_minimizer) {
+        crash_unless(path_minimizer_index);
         minimizer_mapper_ptr = std::make_unique<MinimizerMapper>(
             gbz->graph, *path_minimizer_index, distance_index.get(),
             &oversized_zipcodes, path_position_graph
         );
     } else {
+        crash_unless(minimizer_index);
         minimizer_mapper_ptr = std::make_unique<MinimizerMapper>(
             gbz->graph, *minimizer_index, distance_index.get(),
             &oversized_zipcodes, path_position_graph


.gitmodules
```diff
@@ -97,7 +97,7 @@
 	url = https://github.com/vgteam/xg.git
 [submodule "deps/gbwtgraph"]
 	path = deps/gbwtgraph
-	url = https://github.com/jltsiren/gbwtgraph.git
```
adamnovak, line 100: I think we want to keep Jouni's repo as the source here, or maybe make a `vgteam/gbwtgraph` if we want to have lots of people touching it.

Github is pretty good about letting you fetch the right commits from the "wrong" source repo; if you have a PR against `jltsiren/gbwtgraph` open, nobody will have a problem pulling your commits even before the PR merges.

src/algorithms/chain_items.hpp
```diff
@@ -144,34 +144,92 @@ class Anchor {
...
+            exit(1);
+        }
+        start_paths |= (1UL << path);
+        end_paths |= (1UL << path);
+    }
+
+    inline void set_paths(size_t anchor_start_paths, size_t anchor_end_paths) {
+        start_paths = anchor_start_paths;
+        end_paths = anchor_end_paths;
+    }
```
adamnovak: This depends pretty heavily on `std::numeric_limits<size_t>::digits == 64`, uses 64 as a magic number constant, and isn't particularly clear on whether any given `size_t` is meant to represent the *number* of a path, or the *bit flags* for a set of paths.

If we decide that actually we only need to think about 32 paths, or that we really need to expand to 128 paths, all the client code that calls this API will need to change.

I would replace `size_t` as the bit flag field type with a type you invent specifically to represent a bit flag set of paths, and use a constant (rather than a literal) for the number of bits/the max number of paths. And if you need an integer that's specifically a certain number of bits, rather than one that can represent a size or number of things, I would use one of the `uintXX_t` types and not a `size_t`.

So somewhere (maybe in `Anchor`?) you would have:

```
public:
    using path_flags_t = uint64_t;
    static constexpr int MAX_PATHS = 64
```

And then later you could more easily change to:
```
public:
    static constexpr int MAX_PATHS = 256;
    using path_flags_t = std::bitset<MAX_PATHS>;
```

If you decided you needed more.

And then it would be clear in the code whether you were expecting a `path_flags_t` that would be used as flags, or a `size_t` that represents the number/bit index of an individual path.

src/algorithms/chain_items.hpp
```diff
@@ -144,34 +144,92 @@ class Anchor {
...
+        for (size_t path : anchor_paths) {
+            set_path(path);
+        }
+    }
+
+    inline void set_path(size_t path) {
+        if (path >= 64) {
+            cerr << "" << endl;
+            exit(1);
+        }
```
adamnovak: Instead of mysteriously printing a blank line and exiting the program, this should [throw `std::out_of_range`](https://en.cppreference.com/w/cpp/error/out_of_range.html) because that's what `std::bitset` does in a very similar situation.

You could also use `assert`, or consider vg's `crash_unless` from `crash.hpp` if you think that actually this will happen a lot on dodgy reads and want more controlled crash behavior.

src/algorithms/chain_items.hpp
```diff
@@ -144,34 +144,92 @@ class Anchor {
...
+    /// Find and set the haplotypes supported by this anchor
+    /// i.e. haplotypes both in start_pos and end_pos nodes
+    // TODO: remove this and use the updateted minimizer index
+    void find_set_paths(const gbwt::GBWT* gbwt_index) {
+        size_t start_id = id(graph_start()) * 2 + is_rev(graph_start());
+        size_t end_id = id(graph_end()) * 2 + is_rev(graph_end());
+        auto start_paths = gbwt_index->locate(gbwt_index->find(start_id));
+        set_paths(start_paths);
+        update_paths(gbwt_index->locate(gbwt_index->find(end_id)));
+    }
```
adamnovak: If this isn't a template class, then either this needs to be `inline` or the function body needs to live in a `.cpp` file. Otherwise you get in trouble with multiple definitions of the function, when linking multiple `.o` files that include the header.

src/algorithms/chain_items.hpp
```diff
@@ -144,34 +144,92 @@ class Anchor {
...
+    inline void set_paths(const size_t& anchor_paths) {
+        start_paths = anchor_paths;
+        end_paths = anchor_paths;
+    }
+
+    inline void set_paths(const std::vector<size_t>& anchor_paths) {
+        for (size_t path : anchor_paths) {
+            set_path(path);
+        }
+    }
```
adamnovak, line 176: `set_paths()` flags off all paths, and `set_paths(0)` flags off all paths, but `set_paths(std::vector<size_t>())` leaves the paths unmodified, because this function doesn't clear any existing set bits before turning on the ones from the vector.

That's pretty [astonishing](https://softwareengineering.stackexchange.com/a/187462) and so either needs to be prominently documented or (even better) changed to be more consistent.

src/algorithms/chain_items.hpp
```diff
@@ -144,34 +144,92 @@ class Anchor {
...
+        end_paths &= new_paths;
+    }
+
+    inline void update_paths(const std::vector<size_t>& haplotypes) {
+        size_t new_paths = 0;
+        for (size_t haplotype : haplotypes) {
+            new_paths |= (1UL << haplotype);
+        }
+        update_paths(new_paths);
+    }
```
adamnovak: This should explain in more detail what it means to "update" the paths. I might think that "setting" and "updating" the paths are the same operation, but two different methods are provided, so what's the difference? How do I know which one to call?

Ideally one of them should be renamed to a name that's more distinct from the other.

src/algorithms/chain_items.hpp
```diff
@@ -227,6 +287,9 @@ class TracedScore {
     /// Add (or remove) points along a route to somewhere. Return a modified copy.
     TracedScore add_points(int adjustment) const;
     
+    /// Add points and merge paths along a route to somewhere. Return a modified copy.
+    TracedScore add_points_and_paths(int adjustment, std::pair<size_t,size_t> paths_to_add);
```
adamnovak: It looks like this nether "adds" nor "merges" paths, but actually "intersects" paths, and then replaces the original paths if the intersection was empty.

Is there a better word for this operation?

Also, is there a good reason that it needs to be part of the same operation with adding points? It might be a lot clearer to say something like `new_score = old_score.add_points(5).join_paths(successor_paths)` and do it as two operations, and I'm not sure it would be appreciably slower after the optimizing compiler got done with it.

src/algorithms/chain_items.hpp
```diff
@@ -256,6 +319,8 @@ class TracedScore {
     int score;
     // Index of source score among possibilities/traceback pointer
     size_t source;
+    // Supported paths
```
adamnovak: This should probably be a `///` doc comment.

src/algorithms/chain_items.hpp
```diff
@@ -391,6 +456,7 @@ TracedScore chain_items_dp(vector<TracedScore>& chain_scores,
                            const HandleGraph& graph,
                            int gap_open,
                            int gap_extension,
+                           int recomb_penalty = 0,
```
adamnovak: When adding new optional arguments, it often makes the most sense to add them at the *end*, so existing calls that didn't use any optional arguments don't need to change.

Here I'm not sure there are a lot of such calls, though.

src/algorithms/chain_items.hpp
```diff
@@ -479,6 +547,9 @@ int score_best_chain(const VectorView<Anchor>& to_chain, const SnarlDistanceInde
 /// This produces a penalty (positive number).
 int score_chain_gap(size_t distance_difference, size_t average_anchor_length);
 
+/// Apply a recombination penalty if two consecutive anchors are not sharing a common haplotype
+int score_chain_rec(const Anchor& from, const Anchor& to);
```
adamnovak: This says it "applies a penalty", but that makes it sound like it modifies something somewhere else, when I think the caller is instead meant to take the return value and combine it into a total score.

For `score_chain_gap` above I took pains to say that the output is a positive penalty to be *subtracted* from a score. (Arguably I should either flip that or rename the method to something like `chain_gap_penalty` or `chain_gap_badness`.)

This next method *doesn't* specifically say the output is a positive, more-is-worse penalty. So is it actually a proper score to be added in, where worse recombinations are more negative? Or is it still a positive penalty to be subtracted out? The documentation should explain, and if the explanation gets too dubious, the code should change to be more explicable.

src/index_registry.cpp
```diff
@@ -507,6 +507,106 @@ int execute_in_fork(const function<void(void)>& exec) {
 #endif
 }
 
+template<typename PayloadType>
+std::vector<std::vector<std::string>>
+construct_minimizers_dispatch(const vector<const IndexFile*>& inputs,
+                              const IndexingPlan* plan,
+                              const IndexGroup& constructing,
+                              int minimizer_k, int minimizer_w, bool minimizer_W) {
```
adamnovak, line 517: If this is meant to be a helper function just for this file (which it must be, since template functions generally need to be in headers to be useful), you could mark it `static` to make it clearer that it only ever can be used in this CPP file.

It could also use a doc comment here explaining what it is for. From the name I would say it "constructs a minimizers dispatch", but I've never heard of a "minimizers dispatch".

So then I get that it is something that does "dispatching" in the service of constructing minimizers, but that the order of the words in the name has been modified from the most natural one, maybe to make it clear that it's a subfunction of `construct_minimizers`. Which *is* a thing that exists, but only as a local inline lambda function. And it's odd to see a function so simple it should be a local lambda function that also is so complex it needs a big helper.

src/index_registry.cpp
```diff
@@ -507,6 +507,106 @@ int execute_in_fork(const function<void(void)>& exec) {
...
+    assert(gbz_filenames.size() == 1);
+    auto dist_filename = dist_filenames.front();
+    auto gbz_filename = gbz_filenames.front();
+
+    assert(constructing.size() == 2 || constructing.size() == 3);
+    vector<vector<string>> all_outputs(constructing.size());
+    auto minimizer_output = *constructing.begin();
+    auto zipcode_output = *constructing.rbegin();
+    auto& output_name_minimizer = all_outputs[0];
+    auto& output_name_zipcodes = all_outputs[1];
```
adamnovak: This ends up having an optional *middle* value in `constructing`, which is weird. The optional value should be the last one, and the fixed ones should be `constructing[0]` and `constructing[1]` after you've checked the size.

src/index_registry.cpp
```diff
@@ -507,6 +507,106 @@ int execute_in_fork(const function<void(void)>& exec) {
...
+                    } else {
+                        oversized_zipcodes.emplace_back(zip);
+                        zip_index = oversized_zipcodes.size() - 1;
+                        node_id_to_zipcode_index.emplace(id(pos), zip_index);
+                    }
+                }
+                return {0, zip_index}; // extended encoding
+            }
+        } else {
+            // Assume PayloadType == PayloadXL
```
adamnovak: You could back this up with a `static_assert` I think.

src/index_registry.cpp
```diff
@@ -507,6 +507,106 @@ int execute_in_fork(const function<void(void)>& exec) {
...
+    ZipCodeCollection oversized_zipcodes;
+    hash_map<vg::id_t, size_t> node_id_to_zipcode_index;
+
+    std::function<PayloadType(const pos_t&)> payload_lambda =
+    [&](const pos_t& pos) -> PayloadType {
+        ZipCode zip;
+        zip.fill_in_zipcode(*distance_index, pos);
+        auto payload = zip.get_payload_from_zip();
+
+        if constexpr (std::is_same<PayloadType, gbwtgraph::Payload>::value) {
```
adamnovak, line 568: If we're introducing `if constexpr` code, I think we want to change the `Makefile` to use `-std=c++17` on Linux as well as on Mac, since it's definitely a C++17 feature.

G++ might accept it anyway in c++14 mode, but that's sort of cheating.

src/index_registry.cpp
```diff
@@ -4106,131 +4208,43 @@ IndexRegistry VGIndexes::get_vg_index_registry() {
...
+    
     registry.register_recipe({"Short Read Minimizers", "Short Read Zipcodes"}, {"Giraffe Distance Index", "Giraffe GBZ"},
                              [&](const vector<const IndexFile*>& inputs,
                                 const IndexingPlan* plan,
                                 AliasGraph& alias_graph,
                                 const IndexGroup& constructing) {
         return construct_minimizers(inputs, plan, constructing, IndexingParameters::short_read_minimizer_k, 
-                                    IndexingParameters::short_read_minimizer_w, IndexingParameters::short_read_minimizer_W);
+                                    IndexingParameters::short_read_minimizer_w, IndexingParameters::short_read_minimizer_W, false);
     });
```
adamnovak, line 4254: With the guts of the recipe moved out into the template, we don't really need/want the lambda anymore. This right now is in two places in the code calling the same function with two different values for a bool flag, so it can internally check the flag and call one of two different template instantiations. We should just have the call sites be `construct_mimizers<gbwtgraph::Payload>` and `construct_minimizers<gbwtgraph::PayloadXL>` to start with, and name the template `construct_minimizers` and not bother with the lambda anymore.

src/minimizer_mapper.cpp
```diff
@@ -50,22 +50,74 @@ namespace vg {
...
+//     clusterer(distance_index, &graph),
+//     gbwt_graph(graph),
+//     extender(new GaplessExtender(gbwt_graph, *(get_regular_aligner()))),
+//     choose_band_padding(algorithms::pad_band_random_walk()),
+//     fragment_length_distr(1000,1000,0.95) {
+    
+//     // The GBWTGraph needs a GBWT
+//     crash_unless(graph.index != nullptr);
+// }
+
```
adamnovak: This commented-out code should be removed before we merge.
```suggestion
```

src/minimizer_mapper.cpp
```diff
@@ -50,22 +50,74 @@ namespace vg {
...
-    // The GBWTGraph needs a GBWT
+    fragment_length_distr(1000,1000,0.95)
+{   
     crash_unless(graph.index != nullptr);
+    with_index([&](const auto& idx) {
+        // Initialize the k, w, and syncmer use
+        k = static_cast<int32_t>(idx.k());
+        w = static_cast<int32_t>(idx.w());
+        uses_syncmers = idx.uses_syncmers();
+    });
```
adamnovak: This is a neat trick, but instead of doing it, we should make changes upstream to avoid needing or wanting to do it.

Index templates with different payload types *should* share a base class that gives you access to all the fields and methods that *don't* depend on the payload type. Then we'd have one constructor, one member pointer to the base class, and normal code for all the normal member accesses that don't need to know about the payloads.

We'd also have no incentive for breaking the "no forgery principle" (the idea that a good design keeps each piece of data in a single place) here. Right now we copy fields from the index into our own fields to avoid having to ask what the payload type is at every point of access.

src/minimizer_mapper.hpp
```diff
@@ -576,7 +595,9 @@ class MinimizerMapper : public AlignerClient {
         size_t agglomeration_start; // What is the start base of the first window this minimizer instance is minimal in?
         size_t agglomeration_length; // What is the length in bp of the region of consecutive windows this minimizer instance is minimal in?
         size_t hits; // How many hits does the minimizer have?
-        const typename gbwtgraph::DefaultMinimizerIndex::value_type* occs;
+        //const typename gbwtgraph::DefaultMinimizerIndex::value_type* occs;
+        const void* occs = nullptr;
+
```
adamnovak, line 600: If this is a `void*` we should have some explanation here about what it is, not the commented-out old definition. If `Minimizer` is no longer supposed to talk about types in `DefaultMinimizerIndex`, we also need to get rid of the reference to it in `value`.

Maybe instead of a `void*` here, `Minimizer` could be templated on the index type? That might just make things more complicated, though.

We could maybe add a template accessor that casts internally to get the right typed value out of `occs`, instead of doing a bunch of casting at the call site.

src/minimizer_mapper.cpp
```diff
@@ -3482,11 +3538,20 @@ std::vector<MinimizerMapper::Minimizer> MinimizerMapper::find_minimizers(const s
...
+                return idx.minimizer_regions(sequence);
+            });
     for (auto& m : minimizers) {
         double score = 0.0;
-        auto hits = this->minimizer_index.find(get<0>(m));
+        auto hits = with_index([&](const auto& idx) {
+            auto h = idx.find(get<0>(m));
+            return std::pair<const void*, size_t>(
+                static_cast<const void*>(h.first), h.second);
+            //return idx.find(get<0>(m));
```
adamnovak:
```suggestion
```

src/minimizer_mapper.cpp
```diff
@@ -4011,40 +4075,104 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
...
+            //             //If we don't have the oversized payloads, then fill in the zipcode using the pos
+            //             seeds.back().zipcode.fill_in_zipcode(*(this->distance_index), hit);
+            //             seeds.back().zipcode.fill_in_full_decoder();
+            //         }
+            //     } else {
+            //         //If the zipcode was saved in the payload
+            //         seeds.back().zipcode.fill_in_zipcode_from_payload(minimizer.occs[j].payload);
+            //     }
+
+            // }
```
adamnovak:
```suggestion
```

src/minimizer_mapper.cpp
```diff
@@ -4011,40 +4075,104 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
...
+                    if (minimizer.value.is_reverse) {
+                        size_t node_length = this->gbwt_graph.get_length(this->gbwt_graph.get_handle(id(hit)));
+                        hit = reverse_base_pos(hit, node_length);
+                    }
+
+                    // Extract component id and offset in the root chain, if we have them for this seed.
+                    seeds.emplace_back();
+                    seeds.back().pos = hit;
+                    seeds.back().source = i;
+                    seeds.back().paths = gbwtgraph::get_paths_or_zero(occ.payload);
```
adamnovak: This whole block of code is the same for both indexes but needs to be duplicated because `occ` has different types in the two branches. Maybe there's a way to use a template (or lambda that's secretly a template) here?

Or, if `gbwtgraph::PositionPayload<gbwtgraph::PayloadXL>` shared a base class with `gbwtgraph::PositionPayload<gbwtgraph::Payload>`, anything accessing non-payload fields could safely address it by the base type, and we'd only need to cast-and-branch for the parts that actually access the payload.

src/minimizer_mapper.hpp
```diff
@@ -663,7 +684,31 @@ class MinimizerMapper : public AlignerClient {
...
+    auto with_index(Fn&& fn) const
+        -> decltype(fn(*static_cast<const DefaultIndex*>(nullptr)))
+    {
+        if (index_kind == IndexKind::S) {
+            return fn(*minimizer_index_s);
+        } else {
+            return fn(*minimizer_index_xl);
+        }
+    }
+    //const gbwtgraph::DefaultMinimizerIndex& minimizer_index;
```
adamnovak:
```suggestion
```

src/subcommand/giraffe_main.cpp
```diff
@@ -531,6 +531,20 @@ static std::unique_ptr<GroupedOptionGroup> get_options() {
...
+        "penalty for a recombination when chaining (TEMP)",
+        int_is_nonnegative
+    );
+    chaining_opts.add_range(
+        "rec-frag-penalty",
+        &MinimizerMapper::rec_penalty_fragm,
+        MinimizerMapper::default_rec_penalty_fragm,
+        "penalty for a recombination when fragmenting (TEMP)",
+        int_is_nonnegative
+    );
```
adamnovak, line 547: I don't think `_fragm` is the right abbreviation. And we should avoid reordering the words between the fields and the options if we can get away with it.

The other options use a pattern that would have these be `--rec-penalty` and `--fragment-rec-penalty`, with the fragment one defined first with the other fragmenting options, and the chaining one defined later with the chaining options.

Also these shouldn't be `(TEMP)` if we're merging them in.

src/subcommand/giraffe_main.cpp
```diff
@@ -1739,10 +1769,12 @@ int main_giraffe(int argc, char** argv) {
...
-    
-    auto index_targets = map_long_reads
-                       ? VGIndexes::get_default_long_giraffe_indexes()
-                       : VGIndexes::get_default_short_giraffe_indexes();
+    auto index_targets = VGIndexes::get_default_short_giraffe_indexes();
+    if (map_long_reads) {
+        index_targets = use_path_minimizer
+                      ? VGIndexes::get_default_long_path_giraffe_indexes()
+                      : VGIndexes::get_default_long_giraffe_indexes();
+    }
```
adamnovak: It would be neater if we could avoid a set-and-overwrite here. Probably we should type `index_targets` and default-construct it, then fill it in with a bunch of if/else.

src/subcommand/giraffe_main.cpp
```diff
@@ -1776,8 +1808,19 @@ int main_giraffe(int argc, char** argv) {
         cerr << "Loading Minimizer Index" << endl;
     }
     unique_ptr<gbwtgraph::DefaultMinimizerIndex> minimizer_index;
+    unique_ptr<gbwtgraph::MinimizerIndexXL> path_minimizer_index(new gbwtgraph::MinimizerIndexXL());
```
adamnovak: I don't think we want to have this non-null but pointing at an empty index here. We should only construct the index if we're really loading it.

src/subcommand/giraffe_main.cpp
```diff
@@ -1776,8 +1808,19 @@ int main_giraffe(int argc, char** argv) {
...
     if (map_long_reads) {
-        minimizer_index = vg::io::VPKG::load_one<gbwtgraph::DefaultMinimizerIndex>(registry.require("Long Read Minimizers").at(0));
+        if (use_path_minimizer) {
+            // FIXME: This is a temporary hack to load the path minimizers, I can't find a way to use VPKG to load them.
```
adamnovak: You would need to register the type with the VPKG system in a `.cpp` file in `src/io/` in vg, and I think it would also need to have a different magic number than the other minimizer index version.

src/subcommand/giraffe_main.cpp
```diff
@@ -1851,7 +1901,21 @@ int main_giraffe(int argc, char** argv) {
...
+            &oversized_zipcodes, path_position_graph
+        );
+    } else {
+        minimizer_mapper_ptr = std::make_unique<MinimizerMapper>(
+            gbz->graph, *minimizer_index, distance_index.get(),
+            &oversized_zipcodes, path_position_graph
+        );
+    }
+    MinimizerMapper& minimizer_mapper = *minimizer_mapper_ptr;
+    //MinimizerMapper minimizer_mapper(gbz->graph, *minimizer_index, &*distance_index, &oversized_zipcodes, path_position_graph);
```
adamnovak:
```suggestion
```

src/minimizer_mapper.cpp
```diff
@@ -4011,40 +4075,104 @@ std::vector<MinimizerMapper::Seed> MinimizerMapper::find_seeds(const std::vector
...
+
+                    // XL mode: zipcode does not use PayloadXL, workaround: we create a dummy payload.
                     seeds.back().zipcode.fill_in_zipcode(*(this->distance_index), hit);
                     seeds.back().zipcode.fill_in_full_decoder();
```
adamnovak: I don't think the comment is describing what happens here. It looks like the zipcode code depends on the small payload type specifically, and can't be handed just its slice of/info from the payload in a useful way, or the new larger payload type. So we intended to make a fake smaller payload out of the larger payload and hand it to the zipcode code to parse out the zipcode that's really stored in there alongside the path info.

But then instead of doing that we are throwing away the zipcode stored in the payload and building a fresh zipocde from the distance index for every hit, which isn't what we want to be doing.

If the larger payload inherited from the smaller payload, or if the zipcode logic took some type that was inside the payload instead of the whole payload, this would be a lot easier.

src/subcommand/chain_main.cpp
```diff
@@ -30,27 +30,28 @@ void help_chain(char** argv) {
...
+         << "  -r, --rec-pen INT    set recombination penalty (default: 0)" << endl
          << "  -h, --help           print this help message to stderr and exit" << endl;
 }
 
 int main_chain(int argc, char** argv) {
 
     bool show_progress = false;
-    
-    
+    int recomb_penalty = 0;
```
adamnovak: I feel like we have two different levels of abbreviation here, which will lead to confusion when you reach for the option at one level of abbreviation, but it was actually at the other one. These should both be the same thing, maybe `--recombination-penalty` or maybe something shorter but consistent.

src/subcommand/giraffe_main.cpp
```diff
@@ -925,6 +941,9 @@ int main_giraffe(int argc, char** argv) {
 
     // Are we mapping long reads or short reads? According to the parameter preset
     bool map_long_reads = false;
+    
+    // If recombination penalties are set, we are using PathMinimizer, else standard Giraffe.
```
adamnovak: That's not what really happens; this is controlled by a dedicated flag.

(Do the other penalty flags need to have help text about how they don't work without `-E`?)

src/subcommand/minimizer_main.cpp
```diff
@@ -377,61 +377,65 @@ int main_minimizer(int argc, char** argv) {
...
 
-            payload = zipcode.get_payload_from_zip();
-            if (payload != MIPayload::NO_CODE) {
-                //If the zipcode is small enough to store in the payload
                 #pragma omp critical 
                 {
-                node_id_to_payload.emplace(id(pos), payload);
+                //If we've already seen this node before, then return the saved payload
+                if (node_id_to_payload.count(id(pos))) {
+                    payload =  node_id_to_payload[id(pos)];
```
adamnovak: Looks like the original code was missing a level of indentation here, which probably can be fixed.

Also this should probably be just a `find()` within the critical section, and a comparison to `end()` outside it, so you don't need to `[]` at all and you can do just one lookup.

Though I think this code was here before.

src/subcommand/minimizer_main.cpp
```diff
@@ -377,61 +377,65 @@ int main_minimizer(int argc, char** argv) {
...
+                    //If the zipcode is too big and we don't have a file to save the big zipcodes
+                    #pragma omp critical 
+                    {
+                    payload = MIPayload::NO_CODE;
+                    node_id_to_payload.emplace(id(pos), payload);
+                    }
+                    return payload;
+                }
+            };
+        gbwtgraph::index_haplotypes(gbz->graph, index, payload_lambda);
```
adamnovak: Is `vg minimizer` not learning to make the new wider payloads?

src/zip_code.hpp
```diff
@@ -71,6 +71,22 @@ class ZipCode {
...
+        // XL payload: down-convert to classic (keep only first/second).
+        inline void fill_zipcode_from_any_payload(ZipCode& zc,
+                                                const gbwtgraph::PayloadXL& pxl) {
+            gbwtgraph::Payload p;
+            p.first  = pxl.first;
+            p.second = pxl.second;
+            zc.fill_in_zipcode_from_payload(p);
+        }
+
+
```
adamnovak: I think this logic is for the missing actually-using-the-zipcodes-from-the-longer-payloads part and is never called.

adamnovak (changes requested):
There should also probably be a PR for https://github.com/dcmonti/gbwtgraph/compare/f1cb2ba6f1d1f1f9e09edcdf8bc8d67a6b3c5754...e2b4fdbe2b0b3f7ac542e3c47002da470509adba so @jltsiren and I can review those changes too. The diff there doesn't seem too complicated though.

In terms of functionality, I think this mostly works, but the logic to pull zip codes out of the new longer payloads is not plugged in where it's meant to be used, so with the longer payloads the zipcodes are getting recalculated every time.

In terms of design, the `with_index()` workaround is neat, but I think really what we want to do is set up our types so we don't need it. I think we could do it with a base class for the two index types, and then maybe casting to the appropriate index type to call the methods that deal with payloads, or using real polymorphism if we need it.

I think some of the fiddliness with the zipcodes only knowing how to load themselves from the short payload type could be resolved by either making the longer payload extend the shorter payload, or making the zipcodes know how to load themselves from something that sits inside the payload (maybe just from two 64-bit values?).

After reading through everything, I'm wondering if we shouldn't really be dealing with the payloads as `char*`s into the minimizer index's own memory. It would certainly make the type gymnastics easier, and I'm not sure it would be a lot worse on speed. In fact, [the minimizer index lookups already just hand you back a pointer to a block and a count of contiguous results](https://github.com/dcmonti/gbwtgraph/blob/e2b4fdbe2b0b3f7ac542e3c47002da470509adba/include/gbwtgraph/minimizer.h#L802-L837), and I think that basically supplies `Minimizer::occs` already.

.gitmodules
dcmonti: Ok! Do I open a PR on the master or is there a specific branch?

src/algorithms/chain_items.hpp
dcmonti: Changes done

src/algorithms/chain_items.hpp
dcmonti: My bad, I forgot some debugging code. I used `crash_unless` now

src/algorithms/chain_items.hpp
dcmonti: That was some old code no longer used, I deleted it

src/algorithms/chain_items.hpp
dcmonti: I completely agree, I changed the name and they should be less astonishing

.gitmodules
adamnovak: I think @jltsiren is just using the one `master` branch.

src/subcommand/chain_main.cpp
dcmonti: changed with `--recombination-penalty`

src/subcommand/giraffe_main.cpp
dcmonti: Fixed all three definitions

src/subcommand/minimizer_main.cpp
dcmonti: Yes I don't remember changing that part

src/subcommand/minimizer_main.cpp
dcmonti: I have added the option to the minimizer subcommand

src/algorithms/chain_items.hpp
dcmonti: Changed and split the two operations

src/algorithms/chain_items.cpp
```diff
@@ -612,6 +644,10 @@ TracedScore chain_items_dp(vector<TracedScore>& chain_scores,
             // here.
             jump_points = -score_chain_gap(indel_length, base_seed_length) * gap_scale;
 
+            // add recombination penalty if necessary 
+            jump_points -= score_chain_rec(chain_scores[from_anchor], here) * recomb_penalty;
+            //jump_points -= score_chain_rec_jaccard(chain_scores[from_anchor], here, recomb_penalty);
```
adamnovak:
```suggestion
```

src/snarl_seed_clusterer.hpp
```diff
@@ -53,7 +53,8 @@ class SnarlDistanceIndexClusterer {
 
 
     public:
-
+        using paths_flags_t = uint64_t;
+        const static size_t MAX_PATHS = 64;
```
adamnovak, line 57: It looks like these are defined here but also in `chain_items.hpp`. Maybe one of these should pull from the other, or maybe they should both reference a third place where these are defined?

src/algorithms/chain_items.hpp
dcmonti: I've refactor the names

src/algorithms/chain_items.hpp
dcmonti: Ok!

src/algorithms/chain_items.hpp
dcmonti: Put it together with the optional args

src/algorithms/chain_items.hpp
dcmonti: Changed to `check_recombination`, updated the doc

src/index_registry.cpp
dcmonti: Changed name of method plus doc

src/index_registry.cpp
dcmonti: It was a old check for the previous way I was using for building the index, removed

src/index_registry.cpp
dcmonti: I added it, now it is not just assuming the type

src/minimizer_mapper.hpp
dcmonti: I added some informations about it, we can discuss how to change it later

src/minimizer_mapper.cpp
dcmonti: I changed it with a lambda

src/subcommand/giraffe_main.cpp
dcmonti: Fixed it, not the magic number because I think I need to change it gbwtgraph side, right?

src/zip_code.hpp
dcmonti: Removed, but I added the logic for not recomputing the zipcode everytime

src/subcommand/giraffe_main.cpp
dcmonti: changed to rec_penalty_fragments, switch the order in the CLI, changed TEMP to ALPHA

src/subcommand/giraffe_main.cpp
dcmonti: Done

src/algorithms/chain_items.hpp
```diff
@@ -479,6 +577,11 @@ int score_best_chain(const VectorView<Anchor>& to_chain, const SnarlDistanceInde
 /// This produces a penalty (positive number).
 int score_chain_gap(size_t distance_difference, size_t average_anchor_length);
 
+/// Determine if adding the new anchor would cause a recombination event
+/// with respect to the old anchor, given their supported paths.
+/// Returns 0 if no recombination, or 1 if there is a recombination.
```
adamnovak, line 582: It's a bit weird that this is an `int`; I guess it's so you can do math with it without doing weird stuff like multiplying by `bool`?

src/algorithms/chain_items.hpp
```diff
@@ -227,6 +316,9 @@ class TracedScore {
     /// Add (or remove) points along a route to somewhere. Return a modified copy.
     TracedScore add_points(int adjustment) const;
     
+    /// Update the paths supported by this score and return a modified copy
+    TracedScore set_shared_paths(const std::pair<size_t,size_t>& new_paths) const;
```
adamnovak, line 320: This isn't quite a "set" operation because it's doing some logic with the existing paths and both halves of this pair. Going by the way we name things for `Anchor` methods, this would be an `update` since were AND-ing path flags together.

src/io/register_loader_saver_minimizer.cpp
```diff
@@ -36,6 +36,27 @@ void register_loader_saver_minimizer() {
...
 
+void register_loader_saver_minimizer_paths() {
+    std::uint32_t magic_number = gbwtgraph::MinimizerHeader::TAG;
+    std::string magic_string(reinterpret_cast<char*>(&magic_number), sizeof(magic_number));
+
+    Registry::register_bare_loader_saver_with_magic<gbwtgraph::MinimizerIndexXL>("MinimizerIndexXL", magic_string, [](istream& input) -> void* {
+        gbwtgraph::MinimizerIndexXL* index = new gbwtgraph::MinimizerIndexXL();
+
+        // Load it. In case of a failure, this will:
+        // * Throw an exception if sanity checks fail.
```
adamnovak: I think we probably want to call sanity checks something else, since it's considered somewhat rude to people who aren't, strictly speaking, sane.

src/subcommand/giraffe_main.cpp
```diff
@@ -531,6 +531,20 @@ static std::unique_ptr<GroupedOptionGroup> get_options() {
...
     );
+    chaining_opts.add_range(
+        "rec-penalty-chain",
+        &MinimizerMapper::rec_penalty_chain,
+        MinimizerMapper::default_rec_penalty_chain,
+        "penalty for a recombination when chaining, requires -E (ALPHA)",
+        int_is_nonnegative
+    );
+    chaining_opts.add_range(
+        "rec-penalty-fragm",
```
adamnovak: This could match the variable name.
```suggestion
        "rec-penalty-fragment",
```

It could also be `frag`, but I don't think `fragm` is a good abbreviation for "fragment".

src/subcommand/giraffe_main.cpp
adamnovak: It doesn't look to be changed at all.

adamnovak (approved):
There's a couple things I still want to change, but I think I can just change them.

Also it looks like some commits may have gone missing, since I noticed a Github comment saying something changed when it didn't.

PR #4720, opened 2025-10-21
Option for vg inject to treat missing contigs as unmapped

### LINKED ISSUE #4613: vg inject crashed when a read maps to a contig that isn't present in the pangenome graph
Hello. We are building pangenome graphs for the purpose of lifting over reads between assemblies. After graph building with minigraph-cactus, we use vg inject with reads mapped to one assembly as input and vg surject to output a bam file of those reads lifted to a different assembly.

Overall it works great, but I noticed that vg inject crashes if a read in the input file maps to a contig that is not in present in the graph. This happens when (often short) contigs present in the original assembly do not align during graph construction and are left out. We've also come across an example where a contig that isn't particularly short but maps to the acrocentric regions (such that a true mapping is hard to resolve), gets omitted during graph construction.

For now, we can filter our input bam files to exclude reads which map to contigs which do not appear in the graph metadata and that should solve the problem. In the future, it would be great if vg inject had an option to funnel reads which could not be properly injected for this reason into a separate output file, perhaps, and not kill the whole run. I see no obvious reason why the subset of reads which does map to contigs which exist in the graph can't be injected/surjected, but please let me know if there is a reason I'm not considering for this behavior. Thanks again for your help!

vg version v1.65.0 "Carfon"
Compiled with g++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0 on Linux
Linked against libstd++ 20210601
Using HTSlib headers 101990, library 1.19.1
Built by root@buildkitsandbox

faithokamoto:
## Changelog Entry

 * `vg inject` now has the option `--allow-missing-contig`/`-a` which treats reads mapped to missing contigs as unmapped instead of erroring (Resolves #4613)

## Description

Gah is this a wordy thing to write a description for. The longer version is that this changes how `vg inject` treats reads that look like they didn't come from its input graph. Previously, if the input read's contig was not present in the graph, the entire thing would error. But in some cases this behavior might not be desired (e.g. if there is a file consisting of reads mapped to both contigs in the graph and ones left out) as in #4613. This PR adds an option which lets the user explicitly toggle the behavior to treat all such problematic reads as unmapped. Any injectable reads will still be smoothly injected.

### DIFF
--- a/src/alignment.hpp
+++ b/src/alignment.hpp
@@ -103,7 +103,9 @@ void cigar_mapping(const bam1_t *b, Mapping& mapping);
 
 /// Convert a BAM record to an Alignment.
 /// May throw AlignmentEmbeddingError if the BAM record is inconsistent with
-/// the provided graph and set_missing_contig_to_unmapped is false.
+/// the provided graph.
+/// Unless set_missing_contig_to_unmapped is true, will also error if a read
+/// is aligned to a contig which is not in the graph.
 /// If true, then those reads are treated as unmapped instead.
 Alignment bam_to_alignment(const bam1_t *b,
                            const map<string, string>& rg_sample,


src/alignment.hpp
```diff
@@ -103,12 +103,14 @@ void cigar_mapping(const bam1_t *b, Mapping& mapping);
 
 /// Convert a BAM record to an Alignment.
 /// May throw AlignmentEmbeddingError if the BAM record is inconsistent with
-/// the provided graph.
+/// the provided graph and set_missing_contig_to_unmapped is false.
+/// If true, then those reads are treated as unmapped instead.
```
adamnovak, line 109: Do we never throw AlignmentEmbeddingError for reasons other than a missing contig?

adamnovak (approved):
I find the boolean flags on the `alignment.hpp` functions a bit suspicious; often a bool flag really wants to be a separate function, or logic at some other level of the code (like maybe in the lambda instead of in the for-each?).

But we don't have a proliferation of them at these locations yet; maybe we really do need these?

src/alignment.hpp
faithokamoto: Oh, um, double-checking I think we throw embedding errors for everything involving misfits to the graph _except_ missing contigs. (Simple missing contigs we just exit.) I'll fix that.

PR #4726, opened 2025-10-27
Rec giraffe bugfix

dcmonti:
## Changelog Entry

Fixes for `vg minimizer` and `autoindex` after the `rec-mode` changes

## Description
`vg minimizer` now stores the oversized zipcodes if specified and `autoindex` uses weighted minimizers for long reads


src/minimizer_index_helper.hpp
```diff
@@ -0,0 +1,254 @@
+/*
+minimizer_index_helper.hpp
+Shared logic for minimizer index construction.
+*/
+#pragma once
```
adamnovak: I don't think we want to adopt `#pragma once` if we can avoid it, since it's not standard C++. We might be able to jump straight to C++20 modules, which Wikipedia thinks are meant to replace it.

src/minimizer_index_helper.hpp
```diff
@@ -0,0 +1,254 @@
+/*
+minimizer_index_helper.hpp
+Shared logic for minimizer index construction.
```
adamnovak: This should be set up to be recognized by Doxygen as the file doc comment.

```suggestion
/**
\file minimizer_index_helper.hpp
Shared logic for minimizer index construction.
```

src/minimizer_index_helper.hpp
```diff
@@ -0,0 +1,254 @@
...
+#include "snarl_distance_index.hpp"
+#include "vg/io/vpkg.hpp"
+#include "zip_code.hpp"
+
+namespace mi_helper {
+
+using namespace std;
+using namespace vg;
+
+inline size_t trailing_zeros(size_t value) {
```
adamnovak: All the functions in here ought to have doc comments explaining what they do.

src/minimizer_index_helper.hpp
```diff
@@ -0,0 +1,254 @@
...
+    return result;
+  }
+  while ((value & 1) == 0) {
+    value >>= 1;
+    result++;
+  }
+  return result;
+}
+
+inline size_t estimate_hash_table_size(const gbwtgraph::GBZ &gbz,
```
adamnovak: If this isn't a function that's short and called often, it probably shouldn't be `inline`; it should be declared in the header and defined in a cpp file. While it's *possible* to get out of making a cpp file for a unit of code by just making all the functions in it either templates or `inline`, it's not advisable. It means the compiler has to go through the implementation code every time someone wants to have access to the interface, when it wouldn't otherwise need to.

src/minimizer_index_helper.hpp
```diff
@@ -0,0 +1,254 @@
...
+        return MIPayload::NO_CODE;
+      }
+    };
+
+    gbwtgraph::index_haplotypes(gbz->graph, index, payload_lambda);
+  } else {
+    std::function<PayloadType(const pos_t &)> payload_lambda =
+        [&](const pos_t &pos) -> PayloadType {
+      PayloadType payload = PayloadType::default_payload();
+#pragma omp critical
```
adamnovak: We might consider replacing these with a lock scoped to the `build_minimizer_index` function invocation. OMP critical sections like this are just one global lock, so if for some reason we wanted to run two `build_minimizer_index()` calls at once, they would all fight over the one global lock even though they wouldn't be writing to the same data structures.

src/minimizer_index_helper.hpp
```diff
@@ -0,0 +1,254 @@
...
+        {
+          oversized_zipcodes.emplace_back(zipcode);
+          size_t zip_index = oversized_zipcodes.size() - 1;
+          payload = {0, zip_index};
+          node_id_to_payload.emplace(id(pos), payload);
+        }
+        return payload;
+      } else {
+// If the zipcode is too big and we don't have a file to save the big zipcodes
```
adamnovak: I think the indentation is wrong here.

src/minimizer_index_helper.hpp
```diff
@@ -0,0 +1,254 @@
...
+ * be nonempty.
+ *
+ * Logs progress if progress is true.
+ */
+template <typename IndexType, typename PayloadType>
+void build_minimizer_index(const gbwtgraph::GBZ *gbz, IndexType &index,
+                           const bdsg::SnarlDistanceIndex *distance_index,
+                           const std::string &zipcode_name,
+                           const std::string &output_name,
+                           bool progress) {
```
adamnovak: If there was a real `MinimizerIndexHelper` class, it could hold some of these shared arguments (`gbz`, `index`, `progress`) as fields so they don't need to be passed twice. (It could also extend `vg::Progressive` from `progressive.hpp` which is a semi-standard way to manage showing progress in vg.)

It could also centralize the default threshold, `space_efficient_counting`, iteration and hash table size values for `set_frequent_kmers()`, and provide a place for a doc comment explaining the intended sequence of calls to actually build an index.

src/minimizer_index_helper.hpp
dcmonti: It's a little bit messy because of the `#pragma omp critical` block, but according to `clang-format` is indented right

dcmonti:
> This looks pretty good, but I think the code should be revised a little to be more like other code in vg, and probably there ought to be a CPP file for the minimizer indexing helper.
> 
> We could make a real `MinimizerIndexHelper` class, but we might be able to get away with not doing it, since there are only two functions in the interface and a manageable number of parameters for them.

I implented the minimizer index helper as its own class, with the idea you suggested

src/minimizer_index_helper.hpp
adamnovak: I think `clang-format` thinks the comment belongs to the `#pragma`; really it's documenting the whole else case. Maybe it needs to be on the opening line of the `else` in case `clang-format` wants to move it? Or under the `#pragma`? Or maybe it needs a blank line?

adamnovak (approved):
This looks OK to me now.

src/minimizer_index_helper.hpp
dcmonti: I'm not sure actually, but it is not there anymore, but I agree it didn't look nice

PR #4740, opened 2025-11-05
Actually detect and use paths in payloads

adamnovak:
## Changelog Entry

 * `vg giraffe` will again use path payloads from the minimizer index

## Description
@dcmonti noted that path payloads from the minimizer index weren't actually being used, I think after #4727. This adds the missing logic to check the payload type in the `MinimizerMapper` and turn on the flag that says they are there if they are there.


adamnovak:
This seems to fix a bug, but still doesn't give the same results as commit afc234a20fc469660816dc22d684e5acb31d5062 right where we merged #4726.

PR #4745, opened 2025-11-11
Snarl rank reorder

faithokamoto:
## Changelog Entry

 * Remove [duct tape](https://github.com/vgteam/vg/pull/4715) by reordering snarl ranks, which **breaks previous distance indexes**

## Description

Here's the un-duct-taping! The problem was that the zip code tree and the distance index would sometimes disagree on the orientation of a given snarl. The zip code tree relies on snarl ranks to know how to order subchains. Thus, sometimes the chains would be ordered incorrectly, which caused Very Bad things to happen to alignments. This PR convinces the zip code tree and the distance index to make friends.

First, the **distance index stops trying to guess whether a snarl is forward or reversed** relative to the top level chain. (This probably makes the distance index creation process faster, due to less attempted traversals, though I dunno by how much.) Technical issues with how the distance index did those traversals meant that there was no simple way to fix the orientation logic in cyclic snarls. Now, all snarls have ranks ordered from snarl start to snarl end, no matter their orientation.

Second, the **zip code tree uses the distance index's orientations** via the zipcode `get_is_reversed_in_parent()` function. The zip code tree has its own `seed_is_reversed_at_depth()` function which it uses plenty of other places. Trying to replace that one entirely with `get_is_reversed_in_parent()` had very bad consequences that I don't understand. However, it is sufficient to use the zipcode function for the purpose of figure out which direction to sort snarl ranks.

Now everyone agrees about which order the chains should go in. I think. At least I was able to get rid of the duct tape without a massive correctness hit like before.

## Notes

I'm not sure if the distance index _has_ a version field. I did find a [magic number field](https://github.com/vgteam/libbdsg/blob/4ab24e826cb2ef4070b5853562313dbcd69b2742/bdsg/include/bdsg/snarl_distance_index.hpp#L961) in libbdsg; changing that would probably make old indexes break. It is really important to regenerate distance indexes or the snarl ranks will be all confused again.


faithokamoto:
Took out chain rescoring

src/zip_code_tree.cpp
```diff
@@ -2301,20 +2243,34 @@ void ZipCodeForest::sort_one_interval(forest_growing_state_t& forest_state, cons
...
+            const Seed& seed = seeds->at(zipcode_sort_order[sub_interval.interval_start]);
+            for (size_t depth = 0; depth <= sub_interval.depth ; depth++) {
+                if (seed.zipcode.get_is_reversed_in_parent(depth)) {
+#ifdef DEBUG_ZIP_CODE_SORTING
+                    cerr << "Reverse at depth " << depth << ": " << seed.zipcode.get_code_type(depth) << endl;
+#endif
+                    is_reversed = !is_reversed;
+                }
+            }
+        }
```
adamnovak, line 2272: This looks a little suspicious; it wouldn't be needed if different modules didn't have different notions of reversed-ness. I think the note about how `get_reversed_at_depth` can't yet be replaced in order to unify the concepts ought to go here.

src/zip_code_tree.hpp
```diff
@@ -452,6 +452,16 @@ class ZipCodeTree {
      * 
      * New cyclic snarl handling flowchart: 
      * https://docs.google.com/drawings/d/1diKMXCuMteR06fF64BhSttsD5RHh3eTnd8FtyYqs17Y/edit?usp=sharing
+     * 
+     * In order to de-duplicate work, this iterator will not actually look at
+     * _everything_ before it; if it knows that seeds from a given chain would
+     * have already output transitions to its own position. This occurs in
```
adamnovak: There's a missing "then" for this "if" I think. I guess it's "then it will skip the chain"?

src/algorithms/chain_items.cpp
```diff
@@ -268,8 +268,8 @@ transition_iterator zip_tree_transition_iterator(const std::vector<SnarlDistance
             // Emit a transition between a source and destination anchor, or skip if actually unreachable.
             size_t source_anchor_index = std::get<0>(transition);
             size_t dest_anchor_index = std::get<1>(transition);
-            size_t graph_distance = std::get<3>(transition);
-            size_t read_distance = std::get<2>(transition);
+            size_t graph_distance = std::get<2>(transition);
+            size_t read_distance = std::get<3>(transition);
```
adamnovak: Have we been swapping these the whole time?

If so, should we maybe move away from numbered tuples to a little struct or something to make this mistake harder to make? Or add a test to catch it?

faithokamoto: I swapped them back in the first cyclic snarl update PR, and I just swapped them back. I'm not sure it has any actual effect due to the indel penalties being symmetric. A struct is a good idea.

src/zip_code_tree.cpp
faithokamoto: more explanation added

src/algorithms/chain_items.cpp
```diff
@@ -570,30 +562,35 @@ TracedScore chain_items_dp(vector<TracedScore>& chain_scores,
     }
 
     // We will run this over every transition in a good DP order.
-    auto iteratee = [&](size_t from_anchor, size_t to_anchor, size_t read_distance, size_t graph_distance) {
+    auto iteratee = [&](transition_info transition) {
```
adamnovak: If this is now a struct, this might be better as a const reference argument.

src/algorithms/chain_items.hpp
```diff
@@ -381,14 +381,33 @@ ostream& operator<<(ostream& out, const TracedScore& value);
...
+    // Distance between anchors in the read (defaults to max if not set)
+    size_t read_distance = std::numeric_limits<size_t>::max();
+
+    // Constructor without read distance
+    transition_info(size_t from, size_t to, size_t graph_dist)
+        : from_anchor(from), to_anchor(to), graph_distance(graph_dist) {}
+    
+    // Constructor with read distance
+    transition_info(size_t from, size_t to, size_t graph_dist, size_t read_dist)
+        : from_anchor(from), to_anchor(to), graph_distance(graph_dist), read_distance(read_dist) {}
```
adamnovak: These could be one signature with `size_t read_dist = std::numeric_limits<size_t>::max()`.

Also, I think they might need to be `inline` if the bodies are in the HPP here, even if the bodies are empty, or they will violate the one-definition rule.

src/algorithms/chain_items.hpp
```diff
@@ -381,14 +381,33 @@ ostream& operator<<(ostream& out, const TracedScore& value);
...
+
 /**
  * Iteratee function type which can be called with each transition between
  * anchors.
  * 
  * Takes two anchor numbers (source and destination), and their read and graph
  * distances, in that order.
  */
-using transition_iteratee = std::function<void(size_t from_anchor, size_t to_anchor, size_t read_distance, size_t graph_distance)>;
+using transition_iteratee = std::function<void(transition_info transition)>;
```
adamnovak: If you make the transition be passed by reference, this would need to change type.

src/unittest/minimizer_mapper.cpp
```diff
@@ -1020,12 +1020,14 @@ TEST_CASE("MinimizerMapper can make correct anchors from minimizers and their zi
                 // Set up to get all the transitions between anchors in the zip code tree
                 auto transition_iterator = algorithms::zip_tree_transition_iterator(seeds, zip_forest.trees.at(0), std::numeric_limits<size_t>::max(), std::numeric_limits<size_t>::max());
                 // And get them
-                transition_iterator(anchors, distance_index, graph, std::numeric_limits<size_t>::max(), [&](size_t from_anchor, size_t to_anchor, size_t read_distance, size_t graph_distance) {
+                transition_iterator(anchors, distance_index, graph, std::numeric_limits<size_t>::max(), [&](vg::algorithms::transition_info transition) {
```
adamnovak: If you make the transition be passed by reference, this would need to change type.

adamnovak (changes requested):
I think this is pretty good, but I think the non-inline constructor bodies in the header aren't standard, and now that we have a struct instead of a bunch of variables we might want to pass it by reference more. Unless we think that passing 4 size_ts beats a pointer lookup, which it might.

PR #4754, opened 2025-11-18
Chain Visualization Script

adamnovak:
## Changelog Entry

* Explainer explanations for reads now get organized into `explanation_<READ_NAME>` directories.
* Explainer explanations for reads now explain all chains, not just the best one.
* Explainer explanations for reads now include coordinates on all haplotypes, not just references.
* New `scripts/plot_chains.sh` script to plot all the chains for explained reads against all the contigs

## Description
Now if you `vg giraffe --show-work --set-refpos`, and you run `scripts/plot_chains.sh`, you will end up with a bunch of plots in `explanation_<READ_NAME>` directories. You can look at different slices of these plots to get a better understanding of what a read is up to. For example, assuming you have `imgcat`, you can look at all the chains best to worst for read `S40_22229` against contigs in CHM13 with:
```
imgcat explanation_S40_22229/plot-*CHM13*.png
```

Or you can look at the best chain against all the haplotypes' contigs with:
```
imgcat explanation_S40_22229/plot-chain0-*.png
```

This can help you see what's really going on with a chain that goes off-reference. It might also help you see if a chain makes sense in the context of any single haplotypes, versus needing a lot of recombinations between them.

This contains some code generated by Anthropic Claude, to lift the Explainer logic from having just one flat list of explanations into having explanations within a per-read context. I had to rework it extensively because it was overly specific and clumsy with the C API, so I don't think it saved any time over just doing it right the first time.

### DIFF
--- a/scripts/plot_chains.sh
+++ b/scripts/plot_chains.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+# plot_chains.sh: plot read vs. path dotplots for Giraffe chaining dumps
+
+# To use this script, run vg giraffe with --track-correctness or --set-refpos,
+# and --show-work.
+#
+# It will dump a bunch of expalantion_<READNAME> firectories with chaining
+# information in the current directory.
+#
+# Then run this script, which will make plots in those directories of every
+# chain (numbered in decreasing order of score) against every path (including
+# haplotype paths).
+#
+# Then you can view them with commands like:
+#
+#   imgcat explanation_S40_22229/plot-*CHM13*.png
+# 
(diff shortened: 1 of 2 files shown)


src/explainer.cpp
```diff
@@ -5,32 +5,97 @@
...
+        // Use the global numbering
+        return Explainer::next_explanation_number++;
+    } else {
+        // Use the per-thread numbering
+        return Explainer::next_context_explanation_number++;
+    }
+}
+
+
+std::string Explainer::make_filename(const std::string& base_name, const std::string& extension) {
```
adamnovak, line 63: This really should take care of including the explanation number itself.

PR #4758, opened 2025-11-24
User-friendly tweaks to haplotype sampling

faithokamoto:
## Changelog Entry

 * Add `vg haplotypes --ban-contig` option to disallow specific contig by name
 * Allow passing `vg haplotypes --coverage median` to estimate coverage by median instead of mode

## Description

The "user" in the title is me. <strike>These are things that I wanted as part of my usage of haplotype sampling. Logging the specific contig selected (at verbosity level 3/max) is useful for debugging successful or failed sampling runs.</strike> Allowing me to manually ban a specific contig is useful when I want to use the same graph for a bunch of samples, ignoring a different contig for each. The other option is to create a bunch of highly similar graphs that have a single contig removed, which is space-wasteful. The option to estimate coverage by median is empirically better in some cases

### DIFF
--- a/test/t/54_vg_haplotypes.t
+++ b/test/t/54_vg_haplotypes.t
@@ -97,4 +97,5 @@ rm -f indirect.gbz direct.gbz no_ref.gbz
 rm -f diploid.gbz diploid2.gbz diploid3.gbz
 rm -f full.HG003.* default.gam
 rm -f sampled.003HG.* specified.gam
+rm -f GRCh38.HG003.* HG003_GRCh38.gam
 rm -f log.txt


jltsiren:
I don't think adding either is a good idea. The coverage estimation logic should now be robust enough to not need an explicit fallback to median, and banning individual contigs feels too arbitrary. Banning samples would make more sense, and we could use different scoring/sampling models for different top-level chains / graph components (which are usually identified by reference contig names). But a contig name is not a meaningful semantic unit in haplotype sampling, and every new feature in the core logic makes it harder to change in the future.

faithokamoto:
Okay, I can close this, though I'm keeping the tweaks for my local branch of vg. (It really does make more sense to ban a _contig_ instead of a _sample_ for my analysis, so I will need that option.)

PR #4763, opened 2025-12-02
Add back in missing transitions

faithokamoto:
## Changelog Entry

 * Bugfix for `vg giraffe` chaining; improvements to accuracy and minor effect on runtime

## Description

Some problems are simple to describe but require arcane, convoluted sacrifices to fix. This is one of those.

### Problem

The zip code tree's distance iterator was failing to find somewhat less than 1/4 of transitions which cross a cyclic snarl boundary, such as `[A { [Brev] } ]`:

- Seed A occurs to _left_ of the cyclic snarl
- Seed B is oriented _backwards_ relative to seed A (though orientation in a cyclic snarl is supposed to be arbitrary)
- The cyclic snarl is not nested in a larger cyclic snarl

### Example

If there is an edge from the subchain's right side to the start of the snarl, the L->R traversal starting from B would need to turn around and exit the snarl from the opposite side as its original direction. The distance iterator used to only exit in the original direction: the end for a L->R traversal, the start for a R->L traversal.

If A was to the right of the cyclic snarl, then its R->L traversal would (under the old algorithm) make sure to traverse all subchains in the L->R direction. If the cyclic snarl was nested, then A's L->R traversal would similarly traverse subchains in the R->L direction. If B was non-rev, then its R->L traversal would simply pop out of the cyclic snarl and find seed A. But I managed to miss this particular case.

### Solution

It needs to be B's "job" to find this traversal; A isn't in a cyclic snarl and thus can't be expected to care about L->R traversals. So B needs to have its L->R traversal exit the start of the cyclic snarl and go R->L. But it still needs to have its L->R traversal exit the end and go L->R.

The "obvious" solution is to exit in both directions. It only took me a few hundreds of lines of code to add that behavior.

I kept the iterator's external interface the same. Beyond not wanting to force the caller to care about cyclic vs. DAG snarls, there is a significant amount of state information that the "other exit" iterator would need to be created with (e.g. the running stack); it makes most sense to have the iterator duplicate itself at the point where it is about to leave a cyclic snarl.

### Implementation

I added **"saved traversals"**. I keep a stack of memorized iterator states. Once the current traversal hits a point where it would normally have to stop, I check this stack to see if there are any pending traversals. If so, I pop the top one and "restore" to that point:

https://github.com/vgteam/vg/blob/b70ca307b9f99a313f2475dfa6f52deded42f1c2/src/zip_code_tree.cpp#L1955-L1971

### Runtime

Obviously, adding these extra traversals will increase iteration time. I implement two key methods to reduce runtime:

- Traversals will now **_never_ enter a cyclic snarl.** Seeds outside of a cyclic snarl will simply trust that, if they could have possibly transitioned to a seed inside the snarl, that inner seed will handle finding the transition. Thus the iterator will skip past all cyclic snarls.
- The iterator **memorizes optimal running distance for all cyclic snarl exits**. If it calculates that, upon exiting via a cyclic snarl's end, the top of the stack would be 24, it will check whether it has ever exited that particular snarl's end. If so, then the iterator will pretend that the top of the stack would actually be the max `size_t`. This will prevent the traversal from progressing beyond that point.

Without that latter improvement, nested cyclic snarls cause terrible exponential runtime explosions.

### DIFF
--- a/src/zip_code_tree.hpp
+++ b/src/zip_code_tree.hpp
@@ -485,6 +485,12 @@ public:
                           std::stack<size_t> chain_numbers = std::stack<size_t>(), bool right_to_left = true,
                           size_t distance_limit = std::numeric_limits<size_t>::max());
 
+        // Iterators are copyable and movable
+        distance_iterator(const distance_iterator& other) = default;
+        distance_iterator(distance_iterator&& other) = default;
+        distance_iterator& operator=(const distance_iterator& other) = default;
+        distance_iterator& operator=(distance_iterator&& other) = default;
+
         /// Move in right_to_left direction until we hit another seed or the end
         distance_iterator& operator++();
 
@@ -545,8 +551,8 @@ public:
         /// Stored as {snarl start index : (start dist, end dist)}
         std::unordered_map<size_t, std::pair<size_t, size_t>> best_cyclic_snarl_exits;
 
-        /// State of a saved traversal, with various context
-        /// to set the member variables back to
+        /// A specific snapshot of the iterator's position
+        /// which includes necessary state information (e.g. stack)
         struct iteration_position {
             /// Where we are in the stored tree.
             size_t index;
@@ -571,16 +577,11 @@ public:
                   chain_numbers(chain_numbers),
                   state(state) {}
 
-            iteration_position(iteration_position&&) noexcept = default;
-            /// Move assignment operator
-            iteration_position &operator =(iteration_position&& other) noexcept {
-                index = std::move(other.index);
-                right_to_left = std::move(other.right_to_left);
-                stack_data = std::move(other.stack_data);
-                chain_numbers = std::move(other.chain_numbers);
-                state = std::move(other.state);
-                return *this;
-            }
+            // Copyable and movable
+            iteration_position(const iteration_position& other) = default;
+            iteration_position(iteration_position&& other) = default;
+            iteration_position& operator=(const iteration_position& other) = default;
+            iteration_position& operator=(iteration_position&& other) = default;
         };
         /// The iterator's current position (state, direction, stack, etc.)
         iteration_position pos;


src/zip_code_tree.hpp
```diff
@@ -494,7 +511,19 @@ class ZipCodeTree {
...
 
         /// Is the iteration done?
-        inline bool done() const { return index == end_index; }
+        inline bool done() {
+            // Attempt to use saved traversals if we hit the end
+            if (index == end_index) {
+                if (pending_traversals.empty()) {
+                    return true;
+                } else {
+                    use_saved_traversal();
```
adamnovak: It looks like this makes `done()` not only non-`const`, but also not idempotent: it matters how many times you call it. That's probably a bad idea, and also it's not documented here. We should move the restore call to the iterator advance logic instead of the done-ness check.

src/zip_code_tree.hpp
```diff
@@ -529,6 +558,40 @@ class ZipCodeTree {
...
+                           bool right_to_left,
+                           std::stack<size_t> stack_data,
+                           std::stack<size_t> chain_numbers,
+                           State current_state)
+                : index(index),
+                  right_to_left(right_to_left),
+                  stack_data(stack_data),
+                  chain_numbers(chain_numbers),
+                  current_state(current_state) {}
+        };
```
adamnovak: Instead of inventing a new type for this, and having duplicate definitions for all the fields, can we make the state a base class of the iterator?

It's also not a *complete* "state", since now the "state" of an iterator is going to include this stack of previous saved states. It might be worth coining a word here.

src/zip_code_tree.hpp
```diff
@@ -529,6 +558,40 @@ class ZipCodeTree {
...
+        std::unordered_map<size_t, std::pair<size_t, size_t>> best_cyclic_snarl_exits;
+
+        /// State of a saved traversal, with various context
+        /// to set the member variables back to
+        struct iterator_state {
+            size_t index;
+            bool right_to_left;
+            std::stack<size_t> stack_data;
+            std::stack<size_t> chain_numbers;
+            State current_state;
```
adamnovak: A state that has a state inside it named `current_state` is confusing.

src/zip_code_tree.cpp
```diff
@@ -1852,6 +1952,24 @@ void ZipCodeTree::distance_iterator::continue_snarl() {
...
 
+void ZipCodeTree::distance_iterator::use_saved_traversal() {
+    auto next_traversal = pending_traversals.top();
+    pending_traversals.pop();
+    // Restore state
+    stack_data = next_traversal.stack_data;
+    chain_numbers = next_traversal.chain_numbers;
+    index = next_traversal.index;
+    right_to_left = next_traversal.right_to_left;
+    state(next_traversal.current_state);
```
adamnovak: This looks like the domain-specific language stuff from the old PDA code has well and truly escaped. Do we still want this to be `state()` for adopting a state or should it be `set_state()`, or just `state = new_state`?

src/zip_code_tree.cpp
```diff
@@ -1852,6 +1952,24 @@ void ZipCodeTree::distance_iterator::continue_snarl() {
     }
 }
 
+void ZipCodeTree::distance_iterator::use_saved_traversal() {
+    auto next_traversal = pending_traversals.top();
+    pending_traversals.pop();
+    // Restore state
+    stack_data = next_traversal.stack_data;
+    chain_numbers = next_traversal.chain_numbers;
```
adamnovak: This is going to copy both the stacks from the pending traversal stack into `next_traversal`, then copy them again into the object. And we had to copy them again when we put the traversal on the traversal stack.

The restore logic should be `std::move()`, and we should make sure the state type has a real moving move assignment operator. 

We could also compare this approach (which doesn't need to manage memory on push/pop but does need to copy whole stacks) with something based on [the immutable list type we use elsewhere](https://github.com/vgteam/structures/blob/d9e184da87492535c5e69d60706ee02a7eb2e82e/src/include/structures/immutable_list.hpp#L78), which supports this sort of snapshot manipulation in constant time. (Or maybe some wrapper of that, where the stack tail is immutable but whatever's on top is mutable, because we do do things like add to the value on top of the stack.)

src/zip_code_tree.hpp
```diff
@@ -529,6 +558,40 @@ class ZipCodeTree {
...
+                  right_to_left(right_to_left),
+                  stack_data(stack_data),
+                  chain_numbers(chain_numbers),
+                  current_state(current_state) {}
+        };
+        /// Other traversals that we have to do later
+        std::stack<iterator_state> pending_traversals;
+
+        /// Restore to the point of a saved traversal.
+        void use_saved_traversal();
```
adamnovak: The comments here should elaborate more on the mechanism; "have to do later" is vague.

src/zip_code_tree.cpp
faithokamoto: Yeah that syntax never really made sense to me. I'll change it.

src/zip_code_tree.hpp
faithokamoto: Renamed the struct to iteration_position

src/zip_code_tree.hpp
faithokamoto: It didn't feel right to make it a _class_, since it doesn't _do_ anything. It just sits there storing state information. What I did to de-duplicate was instead keep the struct, now renamed `iteration_position`, and have the distance iterator store one of its own (`pos`). So now I access member variables with `pos.index`, `pos.state`, etc.

src/zip_code_tree.cpp
faithokamoto: I expect the stack to work better because we do a lot more push/pop vs. copy/restore: the former is constantly over all the iteration, and the latter is only once per cyclic snarl.

src/zip_code_tree.hpp
faithokamoto: Stuck `use_saved_traversal()` in `move_index()`

PR #4765, opened 2025-12-03
Remove fragmenting from Giraffe

adamnovak:
## Changelog Entry

 * Giraffe now just uses a single chaining pass, instead of a fragmenting pass and then a chaining pass

## Description

To avoid metaphysical angst about why recombination penalties at fragmenting make things worse instead of better, this PR removes fragmenting entirely (on top of some commits merely bypassing it).

Bypassing fragmenting seems to decrease speed substantially on simulated hifi reads, increase accuracy somewhat on simulated hifi reads, and decrease speed somewhat on real hifi reads. (I haven't gotten R10 results yet because my whole-node timing jobs are still in queue.)

This code has been almost all synthesized by Anthropic Claude, using *almost* all of its patience (aka token limit for the day). I reviewed it and it appears to have done what I wanted to do and glommed the two step functions together (even though it did this by writing a new one and *then* deleting the old ones), but this still needs to be tested for mapping and calling accuracy effects (vs. d1625a9f81f1b743003e0933d6f76b362b3301b9).

### DIFF
--- a/src/zip_code_tree.cpp
+++ b/src/zip_code_tree.cpp
@@ -583,8 +583,9 @@ void ZipCodeForest::close_snarl(forest_growing_state_t& forest_state,
             // If it was the snarl itself, then the next child added to the chain
             // will be the next open_chain, but I
             // haven't implemented this yet- it won't change the correctness
-            bool chain_after_snarl = forest_state.open_chains.back().first >= active_zip_tree.size();
-            if (depth > 0 && forest_state.open_chains.size() > 0 && chain_after_snarl) {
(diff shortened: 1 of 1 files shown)


adamnovak:
@faithokamoto We can't get rid of the abstraction of having an `Anchor` represent several seeds unless we also get rid of the gapless extension feature in the chaining codepath, since we use it there too:

https://github.com/vgteam/vg/blob/0c86c8fec3729f647cb880e5974fe7338c33ccd4/src/minimizer_mapper_from_chains.cpp#L1386-L1391

So I think even with the removal of fragmenting, we still have to deal with having seeds in play that are not `Anchor` boundaries.

PR #4767, opened 2025-12-06
Add functionality to compute discontiguous supplementary alignments in short-read Giraffe

eizengaj-roche:
## Changelog Entry

 * `vg giraffe` can compute supplementary alignments with the `--supplementary` option.

## Description

This PR provides the capability to identify supplementary alignments from non-adjacent regions of the graph in `vg giraffe`. It compliments https://github.com/vgteam/vg/pull/4699, which added the capability to preserve contiguous alignments that became separated during surjection as supplementary alignments.

The strategy for identifying supplementary alignments is intentionally fairly conservative. Supplementaries are only reported if it's possible to find a set of alignments that are disjoint on the read and completely cover the read. There are also some conditions on score and size that get at some notion of the supplementary alignment being "significant". These conditions are largely borrowed from BWA-MEM. 

In order to get better candidates for supplementary alignments, the `--supplementary` flag additionally triggers behavior that allows clusters to escape being filtered by the funnel if they cover a part of the read that is disjoint with all alignments computed previously. My testing suggest that this has a negligible impact on mapping speed. 

The current implementation is entirely inside the `map_from_extensions` code path. If this code path is eventually eliminated, there will be a bit of work to port this functionality over to the chaining code path. There's probably some benefit in developing a supplementary alignment identification step in the chaining code path regardless, since split alignments play a bigger role in long read and contig mapping than in short read mapping.


benedictpaten:
Hi Jordan,

This is really cool. We are, of course, very interested in getting such
functionality going for the chaining code path, with Faith on deck to take
it on, specifically targeting longer reads (but we want the chaining code
path to work for both short and long). Mobin, et al. - we should test with
DV.

src/interval_union.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <utility>
+
+namespace vg {
+
+using namespace std;
+
+/**
+ * The union of a collection of intervals
+ */
+class IntervalUnion
```
adamnovak, line 19: This is a candidate for living in `structures`.

src/minimizer_mapper.hpp
```diff
@@ -1404,6 +1465,36 @@ class MinimizerMapper : public AlignerClient {
...
+    template<typename Score = double>
+    void process_until_threshold_e(size_t items, 
+        const function<Score(size_t)>& get_score,
+        const function<bool(size_t, size_t)>& comparator,
+        const function<bool(size_t)>& threshold_escape,
+        double threshold, size_t min_count, size_t max_count,
+        LazyRNG& get_seed,
+        const function<bool(size_t, size_t, bool)>& process_item,
+        const function<void(size_t)>& discard_item_by_count,
+        const function<void(size_t)>& discard_item_by_score) const;
```
adamnovak, line 1493: These are getting pretty complex, and we're now filling in a matrix of possible combinations of features, so adding one feature gives us two new functions.

Maybe these (eventually?) need to become some kind of builder pattern, where you have an object you construct, and then you hand it the various thresholds and sorting functions and discard event listener callbacks with specific methods for each, so you don't need a function version for every combination one might use. Then once you've set it up you run it.

src/surjector.cpp
```diff
@@ -137,47 +137,249 @@ using namespace std;
...
+                                final_edit->set_to_length(final_edit->sequence().size());
+                            }
+                            else {
+                                auto softclip_edit = final_mapping->add_edit();
+                                softclip_edit->set_from_length(0);
+                                softclip_edit->set_sequence(clipped_surj.sequence().substr(subseq_end, clipped_surj.sequence().size() - subseq_end));
+                                softclip_edit->set_to_length(softclip_edit->sequence().size());
+                            }
+                        }
+                    }
```
adamnovak: I think this piece gets a bit deep into the weeds of juggling individual edits to belong in the top-level `::surject()` function. It probably should be split off into a few functions that do the different manipulations and then orchestrated here.

eizengaj-roche: Factored this out into separate functions

adamnovak:
I think I'm positive on this overall, but I'm rushing around before curtailment and kind of not paying attention here, sorry.

I didn't know about the [SAM SA tag](https://samtools.github.io/hts-specs/SAMtags.pdf), and while that in turn raises questions about how it interacts with the SAM supplementary flag, it makes cramming the supplementary records into e.g. GAF tags more reasonable.

`IntervalUnion` is fine to stay where it is for now.

If we are going to move supplementary alignments to real GAM fields in GAM, and we merge/release something with them in annotations instead, we're going to need conversion code kind of forever unless we're willing to silently throw them out, right? Which maybe we are willing to do.

If the same data will be an `sa` tag in GAF, we *could* store it in GAM with the GAF-tag-round-tripping machinery. But I'm not sure those are indexed for access by tag name...

So if you want to merge it with the supplementary alignments in an annotation, you *can*, but I think you should move them to real GAM `Alignment` fields and put the `sa` tag generation in the GAF IO layer.

adamnovak:
@benedictpaten You had some opinions from December about how we ought to do supplementary alignments in GAF, right? I remember you wanted them to not be in tags in GAF but to be their own distinct records to play better with the sorting.

I think @jltsiren was also talking about having an idea for a good way to represent them.

Should any of that come in here? Does anyone want to block this until the right method for representing supplementary alignments in GAF is invented, or can we merge what's there and change it later? Or should we not have GAF storage of supplementaries for now?

eizengaj-roche:
The way I see it, the tension is between the simplicity of positional queries and the simplicity of pairing. We've implicitly depended on read order to encode pairing in most places in VG (including all of the `AlignmentEmitter` classes). BAM gets away with having separate records for each supplementary alignment because they store more read pairing information in the bit-wise flags and the mate fields. If we want to make supplementary alignments as separate records, we should probably also add something similar in the tags to allow us to recover pairing information.

I also think that, even with separate records, we should probably have something like the `sa` tag I made in conjunction. That's how things work in BAM with the `SA` tag, which is important because it allows you to navigate from one alignment to the other supplementary alignment using an additional positional query. Given that, I think we could probably merge this without having separate records fully implemented, since the future changes would be additive rather than modifying (unless we want to consider a different encoding than the one I came up with for `sa`).

PR #4774, opened 2025-12-15
add --ban-sample option to haplotype sampling

faithokamoto:
## Changelog Entry

 * Add option `vg haplotypes --ban-sample`

## Description

Per [Jouni's suggestion on a previous PR](https://github.com/vgteam/vg/pull/4758#issuecomment-3604192669), this partially redos #4758 to add `--ban-sample` as an option which will prevent the sampler from selecting any haplotypes from a particular sample. Useful when you want to do a poor man's leave-one-out experiment without actually rebuilding the entire graph.


faithokamoto:
Yeah, I don't think the small speed differential would be worth the more confusing code. If this ever becomes a bottleneck I'd reconsider. Think I can just merge this, then?

jltsiren:
You can merge this. I'll check before the next release if the performance impact is noticeable.

PR #4780, opened 2025-12-17
have vg paths warn if no paths are matched

faithokamoto:
## Changelog Entry

 * `vg paths` prints a warning if path criteria select 0 paths

## Description

I had to chase down one too many errors whose root cause was an empty `vg paths` output. (For the curious, this means one.) Yeah, it was due to me messing up my selection criteria, but I'd prefer to get an obvious warning that This Probably Isn't What You Expected instead of seeing a non-error execution and assuming everything was fine and dandy.

Since the GBWT & graph code here use different selection mechanisms, I had to add the check in two places. Also stuck in a few unit tests to catch the new behavior.

### DIFF
--- a/src/subcommand/paths_main.cpp
+++ b/src/subcommand/paths_main.cpp
@@ -531,6 +531,9 @@ int main_paths(int argc, char** argv) {
     } else if (graph) {
         
         // Handle queries from the graph
+        if (graph->get_path_count() == 0) {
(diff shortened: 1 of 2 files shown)


faithokamoto:
(It also prints a warning if you try to run with an input graph which has no paths, but in that case you have deeper problems.)

PR #4793, opened 2026-01-07
Count all paths when counting paths in a GBZ

adamnovak:
## Changelog Entry

 * `vg paths` should again work on GBZ files containing only haplotypes

## Description
This adjusts `gbwtgraph` to a version that reports the haplotypes as well as the reference and generic paths in `get_path_count()`.

### DIFF
--- a/test/t/11_vg_paths.t
+++ b/test/t/11_vg_paths.t
@@ -7,7 +7,7 @@ PATH=../bin:$PATH # for vg
 
 export LC_ALL="C" # force a consistent sort order 
 
-plan tests 33
+plan tests 35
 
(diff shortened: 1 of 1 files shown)


adamnovak:
@faithokamoto I've tested and confirmed that vg 1.71.0 as released can do `vg paths` on a GBZ that contains something other than haplotype paths; it only bails out incorrectly if the GBZ is *entirely* haplotype paths. So right now I'm against bothering with a patch release after merging this, because I think the people working with GBZs like that are likely to be building from source anyway.

faithokamoto:
Works for me.

Issue #4724, opened 2025-10-24
vg autoindex crashed trying to index a graph with oversized snarls


faithokamoto:
**1. What were you trying to do?**

Use `vg autoindex --workflow lr-giraffe` on a GBZ that I'd previously made a distance index for. My creation started from a GFA:

```
GRAPH=/private/groups/patenlab/fokamoto/centrolign/graph/unsampled/chr12

# Convert GFA to GBZ
vg convert --gfa-in $GRAPH.gfa | vg mod --chop 1024 - > $GRAPH.pg
vg gbwt --index-paths -x $GRAPH.pg -o $GRAPH.gbwt
vg gbwt --gbz-format -x $GRAPH.pg $GRAPH.gbwt -g $GRAPH.giraffe.gbz

# Index GBZ for haplotype sampling
vg gbwt -r $GRAPH.ri -Z $GRAPH.giraffe.gbz
vg index -t 64 -w 1 -w 2 -j $GRAPH.dist $GRAPH.giraffe.gbz
vg haplotypes -v 3 -t 16 -H $GRAPH.hapl -d $GRAPH.dist -r $GRAPH.ri $GRAPH.giraffe.gbz

# Index GBZ for read alignment
vg autoindex --gbz $GRAPH.giraffe.gbz -w lr-giraffe --prefix $GRAPH
```

The input GFA was the result of Python surgery; I took two GFAs and combined them by adding a dummy source/sink (nodes 1 and 2) that connected to the start/end of every path. Node IDs were properly handled etc. I'm 90% sure that the surgery worked correctly, since it did manage to become a GBZ and all, but if you want to check the surgery code is in `/private/groups/patenlab/fokamoto/centrolign/code/add_dummy_caps.py`.

Notably, in the distance index creation step, I got a warning about the index having oversized snarls. I suspect that's because the graph basically consists of that one giant centromere and it has a bunch of chains due to the aforementioned surgery method. Is this supposed to work with oversized snarls? Am I going to have to increase `--snarl-limit` like it suggested?

**2. What did you want to happen?**

Zipcode/minimizer files to be created.

**3. What actually happened?**

A bunch of fun errors, with this interesting line about `is_regular_snarl()` tossed in:

```
[vg autoindex] Guessing that /private/groups/patenlab/fokamoto/centrolign/graph/unsampled/chr12.dist is Giraffe Distance In
dex
[IndexRegistry]: Constructing minimizer index and associated zipcodes.
[...]

PR #4801, opened 2026-01-14
Autoindex oversized

faithokamoto:
## Changelog Entry

 * Help `vg autoindex` not error when indexing a graph with oversized snarls

## Description

Okay, so this doesn't actually _work_ yet, but it's a step towards addressing #4724. Last August, https://github.com/vgteam/libbdsg/pull/221 changed bdsg so that `is_regular_snarl()` could handle distanceless distance indexes. However, that managed to break stuff for indexes that had distances but also had oversized snarls. In particular, causing [this error](https://github.com/vgteam/libbdsg/blob/8768d38496e5169500a010908d3b92cbb42c0339/bdsg/src/snarl_distance_index.cpp#L244-L246):

```c++
    if ((record_type == SNARL || record_type == OVERSIZED_SNARL) && graph == nullptr) {
        throw runtime_error("error: is_regular_snarl requires a graph if the distance index doesn't contain distances");
    }
```

Great, I thought, why not just pass a graph? If this was trickling down from `vg autoindex`, we already had such a graph in a calling function. I had that graph handed down the line. And now we no longer get the same error. Instead, we get a different error! The [very next one](https://github.com/vgteam/libbdsg/blob/8768d38496e5169500a010908d3b92cbb42c0339/bdsg/src/snarl_distance_index.cpp#L247-L249):

```c++
    if ((record_type == SNARL || record_type == OVERSIZED_SNARL) && !allow_internal_loops) {
        throw runtime_error("error: is_regular_snarl requires distances in the distance index to verify that there are no internal loops");
    }
```

So, uh, yeah. This is still broken. I wanted to document the test case and my attempt at fixing it, though. Possible ways to fix the next error include:
- passing `allow_internal_loop = true`; before we were had the default of `false`, so when I had to be explicit about it I just used that, but we could change it
- dropping `OVERSIZED_SNARL` from the types that trigger the new error; they [_do_ store distances to the boundary nodes](https://github.com/vgteam/libbdsg/blob/8768d38496e5169500a010908d3b92cbb42c0339/bdsg/include/bdsg/snarl_distance_index.hpp#L618-L622) so that should let us calculate whether there's a self-loop, right?

@xchang1 thoughts?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,8 +228,9 @@ else
     $(info OS is Linux)
     $(info Compiler $(CXX) is assumed to be GCC)
 
-    # Linux can have some old compilers so we want to work back to C++14
-    CXX_STANDARD?=14
+	# gbwtgraph uses inline variables and our oldest supported compiler has
+	# C++17, so we should use C++17
+    CXX_STANDARD?=17
 
     # Set an rpath for vg and dependency utils to find installed libraries
     LD_UTIL_RPATH_FLAGS="-Wl,-rpath,$(CWD)/$(LIB_DIR)"
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ On other distros, or if you do not have root access, you will need to perform th
                          liblzma-dev liblz4-dev libffi-dev libcairo-dev libboost-all-dev \
                          libzstd-dev pybind11-dev python3-pybind11 libssl-dev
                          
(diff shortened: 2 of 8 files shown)


xchang1:
I haven't really had a chance to look at this closely yet but to your last two points:

- I think I made `allow_internal_loops=false` the default because that's what it was before I added it. I was worried that making it true by default would break something. But maybe the right thing to do is not give it a default value and make everything that doesn't pass a value false.
- Yeah, that looks right to me.

PR #4821, opened 2026-02-05
pull in new libbdsg with ref overlay fix

glennhickey:
## Changelog Entry

 * Fix (unreleased) memory usage regression in `vg call` and `vg deconstruct`.  

## Description

I've been breaking my head for days now trying to get #4804 tested to the point where I can merge it.  But every time I'd run sanity checks on the hprc v2 gbz it would run out of memory, and I was sure it was due to some new indexes not scaling to whole-genome.

I was comparing to the latest release binary where everything was working fine, but it turns out the issue is in master by way of, I think #4795, and so it took forever for me to clue in...  Since `vg deconstruct` and `vg call`, the two tools I'm testing, happen to use the `ReferencePathOverlay`, and the overlay was apparently changed to index every single path, including haplotypes, in the GBZ, these two tools became unusable at scale regardless of if you have nested paths or not...

This PR updates libdsg to explicitly select for reference paths when creating the overlay to revert back to the original logic (https://github.com/vgteam/libbdsg/pull/236)

### DIFF
--- a/src/subcommand/call_main.cpp
+++ b/src/subcommand/call_main.cpp
@@ -435,6 +435,8 @@ int main_call(int argc, char** argv) {
     // Apply overlays as necessary
     bool need_path_positions = vcf_filename.empty();
     bool need_vectorizable = !pack_filename.empty();
+    // When not using GBWT/GBZ, embedded HAPLOTYPE paths are the sample alleles
+    bool all_paths = gbwt_filename.empty() && !gbz_graph;
     bdsg::ReferencePathOverlayHelper pp_overlay_helper;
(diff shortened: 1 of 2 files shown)


adamnovak:
I think I must have made a mistake implementing the non-hiding of paths, if it resulted in the ReferencePathOverlay indexing *everything* for fast position query. I don't think I'm trying to do something opposed to what you're trying to have happening.

We want the methods that iterate over paths to show you all paths now, not just reference paths. But we don't always want to guarantee that all paths have O(1) position lookups; we only want that for reference paths. The ReferencePathOverlay is meant to do that indexing just for reference paths.

PR #4822, opened 2026-02-06
fix overlay gbz indexing for vg call

glennhickey:
Oops, #4821 only fixed the overlay indexing for `vg deconstruct`.  The patch didn't work for `vg call` because it uses a different overaly -- this PR should resolve that.


adamnovak:
@glennhickey I just merged this now, so it isn't in 1.72.0; do you think we need a 1.72.1 with this?

PR #4824, opened 2026-02-09
Recombination anchors visualization

dcmonti:
## Changelog Entry

Added the total number of recombination in a chain, recombinant anchor are now marked in the chain dump file

### DIFF
--- a/scripts/scatter.py
+++ b/scripts/scatter.py
@@ -913,7 +913,7 @@ def main(args):
         sys.stderr.write("Install the matplotalt package to generate figure alt text\n")
         alt_text = None
     except Exception as e:
-        print("Could not generate alt text due to internal error")
+        sys.stderr.write("Could not generate alt text due to internal error\n")
         traceback.print_exc()
         alt_text = None
    
--- a/src/algorithms/chain_items.cpp
+++ b/src/algorithms/chain_items.cpp
@@ -794,10 +794,14 @@ vector<pair<vector<size_t>, int>> chain_items_traceback(const vector<TracedScore
         // Track the penalty we are off optimal for this traceback
         int penalty = best_past_ending_score_ever - chain_scores[trace_from];
         size_t here = trace_from;
-        //std::cerr << "[REC INFO] Starting traceback at item #" << here << " with recs: " << chain_scores[here].rec_num << " score: " << chain_scores[here].score << std::endl;
+#ifdef debug_chaining
+        std::cerr << "[REC INFO] Starting traceback at item #" << here << " with recs: " << chain_scores[here].rec_num << " score: " << chain_scores[here].score << std::endl;
+#endif
         while (here != TracedScore::nowhere()) {
-            //std::cerr << "\trecs: " << chain_scores[here].rec_num << " paths:\t" << std::bitset<64>(chain_scores[here].paths) << std::endl;
-            //std::cerr << "\t\tanchor #" << here << ": " << to_chain[here] << std::endl;
+#ifdef debug_chaining
+            std::cerr << "\trecs: " << chain_scores[here].rec_num << " paths:\t" << std::bitset<64>(chain_scores[here].paths) << std::endl;
+            std::cerr << "\t\tanchor #" << here << ": " << to_chain[here] << std::endl;
+#endif
             // Mark here as used. Happens once per item, and so limits runtime.
             item_is_used[here] = true;
             size_t next = chain_scores[here].source;
@@ -881,7 +885,9 @@ ChainsResult find_best_chains(const VectorView<Anchor>& to_chain,
                                                              max_indel_bases,
                                                              recomb_penalty,
                                                              show_work);
-    //std::cerr << "[REC INFO] Recombination number for chain: " << best_past_ending_score_ever.rec_num << "\tscore: " << best_past_ending_score_ever.score << "\tpaths: " << best_past_ending_score_ever.paths << std::endl;
+#ifdef debug_chaining
+    std::cerr << "[REC INFO] Recombination number for chain: " << best_past_ending_score_ever.rec_num << "\tscore: " << best_past_ending_score_ever.score << "\tpaths: " << best_past_ending_score_ever.paths << std::endl;
+#endif
     // Then do the tracebacks
     vector<pair<vector<size_t>, int>> tracebacks = chain_items_traceback(
         chain_scores, to_chain, best_past_ending_score_ever, item_bonus, item_scale, max_chains);
@@ -961,24 +967,22 @@ pair<int, vector<size_t>> find_best_chain(const VectorView<Anchor>& to_chain,
                                           double points_per_possible_match,
                                           size_t max_indel_bases) {
                                                                  
-    {
-        ChainsResult cr = find_best_chains(
-            to_chain,
-            distance_index,
-            graph,
-            gap_open,
-            gap_extension,
-            recomb_penalty,
-            1,
-            for_each_transition,
-            item_bonus,
-            item_scale,
-            gap_scale,
-            points_per_possible_match,
-            max_indel_bases
-        );
-        return cr.chains.front().scored_chain;
-    }
+    ChainsResult cr = find_best_chains(
+        to_chain,
+        distance_index,
+        graph,
+        gap_open,
+        gap_extension,
+        recomb_penalty,
+        1,
+        for_each_transition,
+        item_bonus,
+        item_scale,
+        gap_scale,
+        points_per_possible_match,
+        max_indel_bases
+    );
+    return cr.chains.front().scored_chain;
 }
 
 int score_best_chain(const VectorView<Anchor>& to_chain, const SnarlDistanceIndex& distance_index, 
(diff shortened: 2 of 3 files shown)


scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+
+    # Filter seeds to the most common reference path
+    ref_name_counts = Counter(s['ref_name'] for s in seeds)
+    if ref_name_counts:
+        best_ref_name = ref_name_counts.most_common(1)[0][0]
+        seeds = [s for s in seeds if s['ref_name'] == best_ref_name]
+
+    # Build seed lookup by ID
+    seed_by_id = {s['seed_id']: s for s in seeds}
+    seed_ids = set(seed_by_id.keys())
```
adamnovak: This seems like wasted effort; `seed_by_id` is used *only* to get the set of IDs, and the set of IDs is used only to filter the transitions.

scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+        max_score = max_score_by_dest.get(s['seed_id'], 0)
+        seeds_data.append({
+            'index': i,
+            'seed_num': s['seed_num'],
+            'read_pos': s['read_pos'],
+            'ref_pos': s['ref_pos'],
+            'strand': s['strand'],
+            'seed_id': s['seed_id'],
+            'max_score': max_score if max_score > float('-inf') else 0
+        })
```
adamnovak: This might be better, or more succinct, as a `dict()` copy?

scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+
+    # Compress JSON data
+    seeds_compressed = base64.b64encode(gzip.compress(
+        json.dumps(seeds_data).encode('utf-8'))).decode('ascii')
+    transitions_compressed = base64.b64encode(gzip.compress(
+        json.dumps(transitions_data).encode('utf-8'))).decode('ascii')
+
+    # CSS and JS are plain strings (no f-string brace escaping needed).
+    # Only the data script uses f-string interpolation.
```
adamnovak: This speaks to what the code generator was told to do, not what the reader needs to know.

scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+      font-size: 14px;
+      font-weight: bold;
+    }
+  </style>'''
+
+    data_script = (
+        '  <script type="text/javascript">\n'
+        f'    var SEEDS_COMPRESSED = "{seeds_compressed}";\n'
+        f'    var TRANSITIONS_COMPRESSED = "{transitions_compressed}";\n'
+        '  </script>'
```
adamnovak, line 213: I guess this doesn't need the CDATA escaping because we know it contains neither `<` nor `&` inside the script?

scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+        return new ChainData(seeds, transitions);
+      }
+
+      /**
+       * Stores arrays and builds four index Maps.
+       */
+      constructor(seeds, transitions) {
+        /// The raw seed array, for iteration and extent computation.
+        this.seeds = seeds;
+        /// The raw transition array, for iteration and extent computation.
```
adamnovak: Do we really need the transitions to compute the view extent if they can only go between the seeds?

scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+          if (!this._transitionsBySourceIndex.has(t.source_index))
+            this._transitionsBySourceIndex.set(t.source_index, []);
+          this._transitionsBySourceIndex.get(t.source_index).push(t);
+        });
+
+        /// Map from dest_index to the is_max transition arriving there.
+        this._bestTransitionToSeed = new Map();
+        transitions.forEach(t => {
+          if (t.is_max) this._bestTransitionToSeed.set(t.dest_index, t);
+        });
```
adamnovak: This is described as if there will only be one max transition, when in fact there can be multiple maximal transitions and this chooses an arbitrary one (which is the right behavior).

scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+
+      /**
+       * Look up a seed by its seed_id.
+       */
+      seedById(id) {
+        return this._seedById.get(id);
+      }
+
+      /**
+       * The is_max transition arriving at seedIndex, or undefined.
```
adamnovak: This again is written as if we require there to be only one, when we don't.

scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+       */
+      bestTransitionTo(seedIndex) {
+        return this._bestTransitionToSeed.get(seedIndex);
+      }
+
+      /**
+       * A specific transition from sourceIndex to destIndex, or undefined.
+       */
+      transitionFromTo(sourceIndex, destIndex) {
+        const fromSource = this._transitionsBySourceIndex.get(sourceIndex) || [];
```
adamnovak: This probably should call transitionsFrom.

scripts/build-chain-viz.py
```diff
@@ -0,0 +1,837 @@
...
+      }
+
+      #hideSecondaryTransitions(seedIndex) {
+        this.secondaryTransitionLayer.selectAll('.secondary-dest-' + seedIndex).remove();
+      }
+
+      /**
+       * Draw a thick black line from hoveredIndex to the selected seed.
+       */
+      #highlightTransitionToSelected(hoveredIndex) {
```
adamnovak: I'm not sure on the name here; all the transitions to the selected seed are shown, and this is highlighting the one from the hovered seed to the selected one. So shouldn't it be `#highlightTransitionFromHovered`?

scripts/scatter.py
```diff
@@ -885,6 +912,10 @@ def main(args):
     except ImportError:
         sys.stderr.write("Install the matplotalt package to generate figure alt text\n")
         alt_text = None
+    except Exception as e:
+        print("Could not generate alt text due to internal error")
```
adamnovak: This should also go to stderr.

src/algorithms/chain_items.cpp
```diff
@@ -783,7 +794,10 @@ vector<pair<vector<size_t>, int>> chain_items_traceback(const vector<TracedScore
         // Track the penalty we are off optimal for this traceback
         int penalty = best_past_ending_score_ever - chain_scores[trace_from];
         size_t here = trace_from;
+        //std::cerr << "[REC INFO] Starting traceback at item #" << here << " with recs: " << chain_scores[here].rec_num << " score: " << chain_scores[here].score << std::endl;
         while (here != TracedScore::nowhere()) {
+            //std::cerr << "\trecs: " << chain_scores[here].rec_num << " paths:\t" << std::bitset<64>(chain_scores[here].paths) << std::endl;
+            //std::cerr << "\t\tanchor #" << here << ": " << to_chain[here] << std::endl;
             // Mark here as used. Happens once per item, and so limits runtime.
```
adamnovak, line 805: These should probably be `#ifdef debug`-flagged off instead of commented out.

src/algorithms/chain_items.cpp
```diff
@@ -861,24 +881,71 @@ vector<pair<int, vector<size_t>>> find_best_chains(const VectorView<Anchor>& to_
                                                              max_indel_bases,
                                                              recomb_penalty,
                                                              show_work);
+    //std::cerr << "[REC INFO] Recombination number for chain: " << best_past_ending_score_ever.rec_num << "\tscore: " << best_past_ending_score_ever.score << "\tpaths: " << best_past_ending_score_ever.paths << std::endl;
```
adamnovak: This also could be an `#ifdef` and not a comment.

src/algorithms/chain_items.cpp
```diff
@@ -894,21 +961,24 @@ pair<int, vector<size_t>> find_best_chain(const VectorView<Anchor>& to_chain,
...
-        recomb_penalty,
-        1,
-        for_each_transition,
-        item_bonus,
-        item_scale,
-        gap_scale,
-        points_per_possible_match,
-        max_indel_bases
-    ).front();
+    {
```
adamnovak: This looks like an extra layer of braces that isn't doing anything.

src/minimizer_mapper_from_chains.cpp
```diff
@@ -245,7 +251,22 @@ void MinimizerMapper::dump_debug_chains(const ZipCodeForest& zip_code_forest,
...
+                                seedpos.field(seed_anchors.at(seed_num).read_start());
+                                seedpos.field(path_name);
+                                seedpos.field(position.first);
+                                seedpos.field(position.second ? "-" : "+");
+                                seedpos.field(seed_num);
+                                std::stringstream ss;
+                                ss << seed_anchors.at(seed_num);
+                                seedpos.field(ss.str());
+                            }
+                        }
```
adamnovak, line 270: We should avoid this whole loop if `seedpos` is falsey.

scripts/build-chain-viz.py
adamnovak: I think the current name makes more sense when you read it with the argument.

src/minimizer_mapper_from_chains.cpp
adamnovak: Actually the whole function is for debugging so we don't need to.

PR #4846, opened 2026-03-06
remove vg paths -V

faithokamoto:
## Changelog Entry

 * Remove broken `vg paths --extract-vg` option

## Description

This doesn't work; the `.vg` it outputs is malformed and can't be used elsewhere, or at least that happened for my script. So we should just not even offer the option.


src/subcommand/paths_main.cpp
```diff
@@ -228,8 +226,7 @@ int main_paths(int argc, char** argv) {
             break;
 
         case 'V':
-            extract_as_vg = true;
-            output_formats++;
+            logger.warn() << "-V/--extract-vg is deprecated" << endl;
```
adamnovak: I don't think this is deprecated; I think you're removing it now.

We could keep the option and error to say it has been removed, or we could remove the option entirely and just let users of the old option get the not-a-real-option error.

adamnovak (commented):
I think we do want to remove this. I don't think this PR "deprecates" it though; it just removes it.

(It works as designed, it just does a weird thing (generate a fragment of a Protobuf .vg file that indeed can't be independently loaded) which nobody actually wants to do and which is easy to mistake for something more useful.)

faithokamoto:
Had dropped this for a bit because of travel & the paper. Yeah, I have no clue how to fix the failing toil-vg test cases, nor have I figured out how to even _run_ said test cases. (I tried to follow the instructions in the README over there, only succeeded in messing up my Python installations so badly I had to un-install and re-install stuff.) If I could, I would, but at this point our terribly complicated CI is blocking me from making this change, and I'm not sure how useful it is for me to continue to fight it.

In summary: I can't make a simple bugfix (if an option doesn't do what it's supposed to do, that's a bug) because it makes tests fail. I can't fix those tests because I lack the knowledge of how to run them. I have tried to run them and not only did they not run, they messed up my environment. I think Russ might be right and vg is too complicated :p

adamnovak:
I have toil-vg set up, so I think I can fix it to deal with not having `-V`. I'll also fix the problem where the documentation is so terrible that when you try and run the toil-vg tests it eats your Python isntallation instead; we might need something in the vg wiki explaining how to run a single toil-vg test locally and how to patch toil-vg.

faithokamoto:
I mean I could also band-aid here by just removing the option from the helptext but keeping all the code backing it, making it into an undocumented option. We have more than our share of those hanging around.

adamnovak:
No, we should drop as many options as we can get away with. I think you were pretty close on the toil-vg changes; it just needs to be run locally until it actually works.

adamnovak:
OK, I have updated this and https://github.com/vgteam/toil-vg/pull/865 and I *think* it should work.

I've also created https://github.com/vgteam/vg/wiki/Troubleshooting#toil-vg-tests-are-failing on the wiki which should provide a how-to for troubleshooting `toil-vg` tests in the future.

faithokamoto:
Thanks, I'll try that next time those tests cause me issues.

PR #4849, opened 2026-03-10
Single command haplotype sampling

adamnovak:
## Changelog Entry

 * Added `vg giraffe --haplotype-sampling` to automatically count kmers and haplotype-index and haplotype-sample the graph. Make sure to have [kmc](https://github.com/refresh-bio/KMC?tab=readme-ov-file#kmc) installed. Providing either a `--kff-name` or `--haplotype-name` will now also trigger generation of the other. To do one-reference sampling, continue to use `--set-reference`. To do non-diploid sampling with a certain number of haplotypes, use `--no-diploid-sampling` and `--num-haplotypes`.

## Description

This allows Giraffe to be used as a single-command haplotype sampling workflow, without needing separate commands to count kmers, haplotype-index the graph, haplotype-sample the graph, and map.

I introduced a notion of "scopes" for indexes in the registry, so I could attach a scope of the sample name to the FASTQ or KFF and have it propagate and qualify/subscript all indexes that depend on them. This works OK, but to get the "right" filenames we'd been using for the sampled GBZ (`<prefix>.<sample>.gbz`), when it's doubling as a "Giraffe GBZ", I've had to introduce a notion of multiple possible extensions for an index, and then a bunch of logic to work out which of the extensions we actually should use so we don't conflict with older things in the plan.

I think I might be able to replace all that with a notion of "weak" aliases, so the "Giraffe GBZ" can come from the "Haplotype-Sampled GBZ" but use it at the "Haplotype-Sampled GBZ"'s name instead of its own.


adamnovak:
I don't think the "weak" alias approach would work unless the haplotype-sampled GBZ knew it was always sample-scoped and always had `{sample}.gbz` as its suffix, with some notion of a wildcard. Otherwise we would think it would try and fight the plain GBZ for the `gbz` suffix.

adamnovak:
This is now failing due to what I think is a preexisting issue in mainline vg: https://github.com/vgteam/vg/pull/4853#issuecomment-4078440283

PR #4861, opened 2026-03-25
[New Feature] vg convert: add GFAz input support

babyplutokurt:
## Changelog Entry

Added [GFAz](https://github.com/babyplutokurt/GFAz) input support to vg by integrating it into vg's shared GFA parsing and import pipeline.

## Description

This PR adds support for reading [GFAz](https://github.com/babyplutokurt/GFAz) files in vg. Instead of adding a separate copy of the GFA import logic, this change refactors vg's existing GFA parser/import path so that both text GFA and GFAz drive the same semantic conversion code.

The key design change is that GFA parsing is now split into a small hierarchy:

- `GFAParser`: shared base class for parser types, listener registration, shared string/tag types, and event APIs
- `GFATextParser`: existing text-GFA parser logic adapted to the new shared interface
- `GFAzParser`: drives the same listeners from decompressed GFAz elements

This means vg now has a single copy of the logic that turns GFA semantics into vg graph/path semantics, regardless of whether the input is text GFA or GFAz.

In addition, GFAz is now registered as its own input format in `vg::io::Registry`, so generic graph-loading paths can recognize and load `.gfaz` files through the normal `VPKG` mechanism.

## Benchmark

A local benchmark on HPRC `chr1` comparing `vg convert -g` from text GFA vs GFAz into PackedGraph:

| Threads | GFA Wall (s) | GFAz Wall (s) | Speedup | GFA RSS (KB) | GFAz RSS (KB) |
|---:|---:|---:|---:|---:|---:|
| 1 | 434.54 | 182.83 | 2.38x | 11,179,328 | 10,889,244 |

In this workload, GFAz input is about 2.4x faster than text GFA, with slightly lower peak RSS.

## What Changed

New files:
- `src/algorithms/gfaz_to_handle.hpp` / `.cpp`: GFAz-backed parser implementation and import support
- `src/io/register_loader_saver_gfaz.cpp`: registry loader registration for GFAz

Modified files:
- `src/algorithms/gfa_to_handle.hpp` / `.cpp`: refactored shared GFA parsing/import code around a parser hierarchy and shared listener-driven conversion
- `src/io/register_libvg_io.cpp`: registers GFAz loader support
- `src/subcommand/convert_main.cpp`, `src/subcommand/view_main.cpp`, `src/subcommand/sort_main.cpp`, `src/index_registry.cpp`: use the new temporary helper path where these commands still explicitly import GFA-like inputs
- `Makefile`: builds and links GFAz support
- `.gitmodules`: adds `deps/GFAz`
- `test/t/48_vg_convert.t`: adds `.gfaz` import coverage

## Design Notes

This PR follows the review direction of "melting and resynthesizing" the integration instead of duplicating `gfa_to_handle.cpp`.

The main architectural pieces are:

- `GFAz` is registered as its own format in `src/io`, rather than being hidden inside the old text-GFA loader path.
- The old text-GFA parser logic remains, but now lives behind `GFATextParser`.
- `GFAzParser` emits the same parser events as text GFA.
- Shared conversion code attaches listeners to a polymorphic parser and builds vg graphs/paths from those events.
- Path/walk/overlap event APIs were refactored so listeners receive iterator-style callbacks instead of raw text slices that must be reparsed manually.

This keeps GFA semantics handling in one place while allowing different physical encodings of GFA records.

## GFAz-Specific Semantics

GFAz restores segment names as 1-based dense contiguous numeric IDs. vg can use these IDs directly, instead of building the text-GFA name-to-ID translation map normally required for arbitrary string segment names.

As a result:
- GFAz node IDs are imported directly
- unnecessary text-name remapping is avoided
- the shared import code still handles the usual GFA semantics for links, paths, walks, and rGFA tags

## Registry Integration

This PR adds a real `GFAZ` registry entry, so generic graph-loading code can auto-detect `.gfaz` input through `vg::io::VPKG`.

That said, this PR does not fully rewrite all legacy explicit GFA command paths to go through the registry. Some existing `convert` / `view` / `sort` / index-registry code paths still use temporary helper wrappers for "load GFA or GFAz". The parser/import refactor is the main goal of this change; broader unification of all GFA input onto registry/VPKG can be done separately.

## Tests

This PR extends `test/t/48_vg_convert.t` with coverage for:

- explicit `-g` import from `.gfaz`
- automatic `.gfaz` detection through the registry-backed generic graph load path
- equivalence of imported graph/path content between `.gfa` and `.gfaz`
- `-x` conversion from `.gfaz`
- translation-file behavior for `.gfaz`
- stability across different thread counts

## Current Scope / Limitations

- This PR adds `GFAz` input support, not `GFAz` output support.
- Output for `GFAHandleGraph` remains normal text GFA.
- Some legacy explicit GFA import call sites still use temporary filename-based dispatch helpers instead of fully routing through `VPKG`.
- This PR intentionally avoids a broader redesign of vg's full graph-loading API beyond what is needed for shared GFA/GFAz import.

## Rationale

The goal of this change is to add GFAz support without creating a second copy of vg's GFA import logic. By making GFAz another parser implementation that feeds the same semantic conversion layer, the integration stays maintainable and keeps future GFA behavior changes shared across both formats.


adamnovak:
It looks like the GFAz library pulls in zstd as a submodule dependency, but vg already has zstd as a system-level dependency through libgbwtgraph that needs it.

Is there a good reason the GFAz build system can't build against a system-installed zstd? Or should this change the vg build to link against the submodule zstd provided by GFAz?

(And shouldn't `-lzstd` come after `-lgfa_compression_core` in the linker command?)

Finally, it looks like you've used a highly productive AI-enabled development process to come up with GFAz very quickly, so even though it can do all this cool stuff it looks like it might have only existed at all for about a week. What is the project's plan for adoption and maintenance, beyond just vg integration? Is the vg team going to end up having to come in and bump its versions of zstd, pybind11, etc. to newer versions as breaking changes happen, or is there a plan for long-term maintenance of the library from your side? Is there a plan for some kind of preprint/paper or writeup about the format that might lead to it being adopted by more tools than *just* vg? Or that might help us understand or explain how it works?

babyplutokurt:
> It looks like the GFAz library pulls in zstd as a submodule dependency, but vg already has zstd as a system-level dependency through libgbwtgraph that needs it.
> 
> Is there a good reason the GFAz build system can't build against a system-installed zstd? Or should this change the vg build to link against the submodule zstd provided by GFAz?
> 
> (And shouldn't `-lzstd` come after `-lgfa_compression_core` in the linker command?)
> 
> Finally, it looks like you've used a highly productive AI-enabled development process to come up with GFAz very quickly, so even though it can do all this cool stuff it looks like it might have only existed at all for about a week. What is the project's plan for adoption and maintenance, beyond just vg integration? Is the vg team going to end up having to come in and bump its versions of zstd, pybind11, etc. to newer versions as breaking changes happen, or is there a plan for long-term maintenance of the library from your side? Is there a plan for some kind of preprint/paper or writeup about the format that might lead to it being adopted by more tools than _just_ vg? Or that might help us understand or explain how it works?

Thanks for the thoughtful questions, Let me address them one by one:
## 1. `zstd` Dependency (System vs. Submodule) and Linker Order
Yeah I agree, building against the system-installed `zstd` is definitely the right approach to avoid symbol conflicts.

## 2. Long-term Maintenance and Stability
The `vg` team will not be left to maintain `GFAz` or deal with breaking dependency bumps; I am committed to long-term upstream maintenance of the library. 

The key thing to note is that since `vg` primarily works with GFA v1.x input, the `GFAz` integration will be extremely stable. I already have full compatibility and testing across GFA v1.0, v1.1, and v1.2. Because those specifications are essentially finalized, the core implementation powering the `vg` integration will rarely need to change. Moving forward, active development and feature updates will mainly be focused on supporting the GFA v2.0+ format, which won't disrupt the core v1.x pipeline you rely on.

## 3. Adoption, Preprints, and Explanation of the Format
I definitely plan for `GFAz` to extend far beyond just `vg`. As a proof point of our adoption strategy, I have already successfully integrated `GFAz` across **all subcommands in ODGI**, and I've verified speedups across subcommands.

I am also planning to release a formal techinical paper, It will thoroughly detail the grammar-based compression algorithm, structural benchmarks. In the meantime, if you're curious about how it works under the hood, we've included a high-level overview of our architecture and serialization pipelines in the `workflow.md` file in our repository.

src/algorithms/gfaz_to_handle.cpp
```diff
@@ -0,0 +1,494 @@
...
+namespace algorithms {
+
+// Deserialize + decompress a GFAZ file to its in-memory GfaGraph
+// representation.
+static GfaGraph load_gfaz_graph(const string &filename, int num_threads) {
+  if (filename == "-") {
+    throw invalid_argument("GFAZ input from stdin (-) is not supported");
+  }
+
+  CompressedData compressed = deserialize_compressed_data(filename);
```
adamnovak: This is coming in from the library? Shouldn't the library have a namespace?

Makefile
```diff
@@ -918,6 +921,16 @@ $(INC_DIR)/mio/mmap.hpp: $(MIO_DIR)/include/mio/*
...
 	+cp -r $(ATOMIC_QUEUE_DIR)/include/atomic_queue/* $(CWD)/$(INC_DIR)/
 
+$(INC_DIR)/GFAz/gfa_parser.hpp: $(LIB_DIR)/libgfa_compression_core.a
+
+$(LIB_DIR)/libgfa_compression_core.a: $(GFAz_DIR)/CMakeLists.txt $(wildcard $(GFAz_DIR)/src/*) $(wildcard $(GFAz_DIR)/src/gpu/*) $(wildcard $(GFAz_DIR)/include/*) $(wildcard $(GFAz_DIR)/include/gpu/*)
+	+rm -f $(CWD)/$(LIB_DIR)/libgfa_compression_core.a
+	+rm -Rf $(CWD)/$(INC_DIR)/GFAz
+	+cd $(GFAz_DIR) && rm -Rf build && mkdir build && cd build && cmake -DCMAKE_C_COMPILER="$(CC)" -DCMAKE_CXX_COMPILER="$(CXX)" -DCMAKE_CXX_FLAGS="-fPIC $(CXXFLAGS) $(CPPFLAGS)" -DBUILD_PYTHON_BINDINGS=OFF -DBUILD_CLI=OFF -DBUILD_BENCHMARKS=OFF -DBUILD_TESTS=OFF -DGFAZ_USE_SYSTEM_ZSTD=ON .. && $(MAKE) $(FILTER) gfa_compression_core && cp libgfa_compression_core.a $(CWD)/$(LIB_DIR)/
+	+mkdir -p $(CWD)/$(INC_DIR)/GFAz
+	+cp -r $(GFAz_DIR)/include/* $(CWD)/$(INC_DIR)/GFAz/
```
adamnovak: Do we have to think about the one HPP header as a Make target at all? I think with the library already in LIB_DEPS we do the build and install the headers.

src/algorithms/gfaz_to_handle.cpp
adamnovak, line 1: This whole file looks like a clone of `gfa_to_handle.cpp`. Some of the functions (like `write_gfaz_translation()`) look like complete duplicates of the corresponding GFA-side operations.

I don't think that's the right approach; we want to share common code for converting from the two similar formats.

src/algorithms/gfa_to_handle.cpp
adamnovak, line 1: I don't think hooking into the GFA parsing functions and dispatching on extension is the right design. Instead, the GFA parsing functions should only parse GFA, and we should have an `src/io/register_loader_saver_gfaz.cpp` that describes how to identify a GFAz stream based on magic numbers and registers the format with the `vg::io::Registry`, so that all vg commands will understand the format and things like `vg stats tiny/tiny.gfaz` will work.

scripts/benchmark_convert_gfa_vs_gfaz.py
adamnovak, line 1: Do we need to keep this script? It indeed looks to do what it says, but it powers through every problem with more code, which means it takes a lot of reading versus how much it helps to have around. Maybe it wants to live in GFAz or wherever its paper scripts will live?

src/subcommand/convert_main.cpp
```diff
@@ -315,7 +315,8 @@ int main_convert(int argc, char** argv) {
             logger.warn() << "currently cannot convert GFA directly to XG; "
                           << "converting through another format" << endl;
             vg::algorithms::gfa_to_path_handle_graph(input_stream_name, &intermediate,
-                                                 input_rgfa_rank, gfa_trans_path);
+                                                 input_rgfa_rank, gfa_trans_path,
+                                                 nullptr, num_threads);
```
adamnovak: Rather than passing around thread counts, the way vg likes to do thread count management is by using OMP's global state. We use `set_thread_count()` from `src/utility.hpp` to tell OMP how many threads we want for the program, and then OMP gives everything the right number of threads automatically when parallel sections start, even when they're nested or themselves starting from other threads.

Since GFAz already uses OMP, we *shouldn't* need to pass it a thread count; we should just tell it to use the current configured OMP number of threads.

It looks like GFAz doesn't know how to do that, though; the thread count here goes into a [`ScopedOMPThreads`](https://github.com/babyplutokurt/GFAz/blob/0ae68b0fe407c347e068146e213ea881982b3bbd/include/utils/threading_utils.hpp#L48), which really wants to save off all that global state and replace it with either a particular provided value or [whatever `resolve_omp_thread_count()` autodetects from the environment](https://github.com/babyplutokurt/GFAz/blob/0ae68b0fe407c347e068146e213ea881982b3bbd/include/utils/threading_utils.hpp#L12) if 0. It looks like GFAz doesn't support just using the currently configured thread count.

Ideally it would be taught (maybe a new `-1` sentinel value for "just use the current OMP configuration and don't change it"?), but if it can't be we would want to read the thread count with `get_thread_count()` from `src/utility.hpp` and pass that every time we call into GFAz.

src/subcommand/convert_main.cpp
adamnovak, line 348: It turns out we no longer need any of this logic to handle GFA input specially.

At https://github.com/babyplutokurt/vg/blame/842fdc5cef32eac62fda9feaade0c34e10f5fc7b/src/subcommand/convert_main.cpp#L304-L306 we decided we needed to handle GFA here separate from the normal load-a-HandleGraph codepath so that we could keep the filename around and use the disk-backed GFA loader, and also so we could stream straight to XG without a HandleGraph in memory. But then we decided to set up the HandleGraph load path so it *could* track filenames (see https://github.com/vgteam/vg/blame/bf85ab04b251e7a2bc308750d6c8c44afda213f5/src/io/register_loader_saver_gfa.cpp#L96 and https://github.com/vgteam/libvgio/blame/65e0ba22246c7443b879f40d0bb69ac96c467107/include/vg/io/registry.hpp#L315-L322). And I had to remove the straight-to-XG code and use an intermediate HandleGraph anyway.

So now nothing important or special actually happens when `input == input_gfa`, and we should just treat it the same as `input == input_handlegraph` and use the `vg::io::VPKG::load_one<HandleGraph>()` approach. (We might want a warning if we load the HandleGraph and it is *not* an instance of `GFAHandleGraph` when we thought it would be.)

If we do that cut as part of this, then we don't need to worry about having some kind of interface *other* than the `vg::io` stuff that knows how to load either-GFA-or-GFAz to a HandleGraph; everything will go through the registry.

We'd also want to combine the corresponding cases in `vg view`: https://github.com/vgteam/vg/blob/bf85ab04b251e7a2bc308750d6c8c44afda213f5/src/subcommand/view_main.cpp#L557-L578

adamnovak (commented):
I've looked at this again, and I think the integration code needs a bit of melting and resynthesizing.

Rather than hooking into the GFA read functions, GFAz ought to be registered as its own format with the `vg::io::Registry`, with a sniffer or magic number, via files in `src/io` and a call in `src/io/register_libvg_io.cpp`.

Rather than duplicating everything in `src/aglorithms/gfa_to_handle.cpp` and slightly modifying it, we should take `GFAParser` and turn it into a 3-class hierarchy, with a base class (`GFAParser`?) that defines the various string-view types and `tag_list_t` and handles attaching and managing the listeners, and then two implementations (`GFATextParser` for GFA and `GFAzParser` for GFAz?) that drive the listeners. For GFA we would keep the existing in-class parsing code, and for GFAz we add code to drive the listeners by iterating over the decompressed GFAz elements.

To make this work, we need to refactor the event listener signatures slightly: instead of representing things like visits and overlaps as `GFAParser::chars_t` that the listener processes with `GFAParser::scan_p_visits`, `GFAParser::scan_w_visits`, etc., we need to present them to the listener by providing a function that can iterate over them. Then the listener would use the iteratee it's currently passing to `GFAParser::scan_w_visits()` or whatever with the new iterator function, while the existing scan functions would end up being used to construct the iterator functions for `GFATextParser`.

Then we would have one copy of all the code that deals with converting GFA semantics to vg semantics (which attaches listeners to a polymorphic parser and gets and acts on all the events, and also deals with managing node ID translation data), and it would be able to be driven by encapsulated decoders for either data source.

Also, due to advancements in the `vg::io::Registry` system, a lot of code we've had that special-cases reading a GFA so we can track the source filename is no longer necessary, so we can run all the GFA input (except some that hands GFAs straight to `libgbwt`) through the registry system and not need any other logic to identify if something is GFA or GFAz.

src/algorithms/gfaz_to_handle.cpp
babyplutokurt: Addressed by updating GFAz upstream, adding gfaz:: namespace.

Makefile
babyplutokurt: Removed ```$(INC_DIR)/GFAz/gfa_parser.hpp: $(LIB_DIR)/libgfa_compression_core.a``` and ```DEPS += $(INC_DIR)/GFAz/io/gfa_parser.hpp```

scripts/benchmark_convert_gfa_vs_gfaz.py
babyplutokurt: Moved to GFAz/script.

src/subcommand/convert_main.cpp
babyplutokurt: Addressed by adding -1 sentinel value on GFAz side to directly inherit OMP configuration from host side.

Now GFAz threads resolver has three options:
> n: using n threads
== 0: using default(8) threads
== -1: using Host OMP configurations


Actually GFAz integration is using single-thread, never uses OMP. Even for GFAz decompression, it only use OMP for multi-stream Zstd decompression, most pipeline is single-threaded. It still have GB level throughput due to O(N) time complexcity

src/subcommand/convert_main.cpp
babyplutokurt: I agree with the direction here.

This PR already adds `GFAz` as its own registered format in `vg::io::Registry`, and the generic `HandleGraph` load path can now auto-detect and load `.gfaz` through `VPKG::load_one<HandleGraph>()`. That path is covered in the tests with direct `vg convert <file.gfaz> ...` input.

I kept the existing explicit `input_gfa` command paths for now because this PR already does a fairly large refactor in the shared import layer: it introduces a `GFAParser` base class, splits the physical parsers into `GFATextParser` and `GFAzParser`, and moves both formats onto the same shared semantic conversion code. I wanted to keep the command-level changes narrower than the parser refactor itself.

For the explicit `-g` path, the current behavior is that we still honor `-g` as “treat this as GFA-style input”, but the actual loader now auto-detects text `GFA` vs `GFAz` internally through the temporary helpers `load_gfa_or_gfaz_to_handle_graph()` / `load_gfa_or_gfaz_to_path_handle_graph()`. That behavior is also covered in the tests with `vg convert -g <file.gfaz> ...`.

So the current state is:

- `vg convert <file.gfaz> ...` exercises the registry/VPKG path
- `vg convert -g <file.gfaz> ...` exercises the explicit GFA import path, which now auto-detects `GFA` vs `GFAz` internally

I agree that the long-term cleanup should be to collapse the remaining `input_gfa` special cases into the normal `input_handlegraph` / `VPKG::load_one<HandleGraph>()` path in `vg convert`, and similarly combine the corresponding cases in `vg view`, so that the registry is the only place making the GFA vs GFAz format decision.

For this PR, I was aiming to keep the focus on:
- registering `GFAz` properly in `src/io`
- eliminating duplicated GFA/GFAz semantic import code
- making both formats drive the same shared import pipeline


Also I improved the GFAz streaming decompression, the GFAz integration to vg not only speed up the subcommands, also with less peak RSS

babyplutokurt: @adamnovak

adamnovak: OK, we can keep the GFA special-casing out of this and make that its own change. I've opened #4880 for that.

babyplutokurt: > OK, we can keep the GFA special-casing out of this and make that its own change. I've opened #4880 for that.

Cool, I can work on #4880 and push another PR, do you have any suggestions on current GFAz integration?

babyplutokurt: @adamnovak ^^

adamnovak: I'll go through and re-review it.

adamnovak:
I never did re-review this, but now https://dl.acm.org/doi/full/10.1145/3797905.3807870 is out, which means that GFAz has gotten some review from third parties, and which addresses some of my initial concerns around documentation for the format and general staying power.

So now I should, indeed, actually re-review this.

babyplutokurt:
> I never did re-review this, but now https://dl.acm.org/doi/full/10.1145/3797905.3807870 is out, which means that GFAz has gotten some review from third parties, and which addresses some of my initial concerns around documentation for the format and general staying power.
> 
> So now I should, indeed, actually re-review this.

Thanks, @adamnovak really appreciate you circling back, and glad the
published version helps on the format-documentation and longevity fronts.
Happy to pick the re-review back up whenever suits you.

A few updates from our side:

- **Rebase needed.** Both GFAz and vg has new commits and cahnegs, I'll rebase latest GFAZ on top of the latest vg later this week and ping you in this thread

- **New capabilities since the paper.** Beyond the format/decode path this PR
  exposes, GFAz now has a compute engine that runs common pangenome analyses
  *directly on the compressed container*, with no decompression back to GFA.
  Measured against the standard tools at 16 threads:
  - `deconstruct` (GFA→VCF): up to **71×** faster than `vg deconstruct`
  - growth curves: up to **~400×** faster than Panacus
  - PAV: up to **613×** faster than `odgi pav`

  with **3–25× lower peak memory** across the three, and outputs that match the
  baselines (deconstruct records within ~0.2% / ~99.99% position concordance,
  growth curves identical at every point, PAV matrices structurally identical).
  None of this is part of *this* PR, it stays scoped to GFAz input for
  `vg convert`, but I mention it since it speaks to where the format is headed
  and its staying power.

- I've submitted an abstract to the T2T face-to-face in UC Santa Cruz
  (https://sites.google.com/ucsc.edu/t2t-f2f) for a short talk on this latest
  work, happy to chat in person if you'll be around.

I'll follow up here once the rebase is up

adamnovak:
Are you sure you want to do a rebase? I just did a back-merge of the vg changes locally and there aren't a lot of conflicts, and I wanted to see if I could just do any extra refactoring I might want myself.

src/io/register_loader_saver_gfaz.cpp
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+    if (!stream) {
+        stream.clear();
+    }
+    return algorithms::GFAzParser::has_magic(buffer, buffer_used);
+}
+
+void register_loader_saver_gfaz() {
+    Registry::register_bare_loader_saver_with_header_check<GFAHandleGraph, MutablePathDeletableHandleGraph, MutablePathMutableHandleGraph, MutableHandleGraph, PathHandleGraph, HandleGraph>("GFAZ", sniff_gfaz, [](istream& input, const string& filename) -> void* {
```
adamnovak: Since GFAz uses a more traditional fixed-length magic number, we don't need the same "header check" we need for GFA. We can [use `register_loader_saver_with_magic()` like for other similar formats](https://github.com/vgteam/vg/blob/6193310e3bc538bcb56baa281528f940db3d109e/src/io/register_loader_saver_lcp.cpp#L19-L22).

src/io/register_loader_gfaz.cpp
```diff
@@ -0,0 +1,86 @@
...
+            exit(1);
+        }
+        if (!temp_name.empty()) {
+            unlink(temp_name.c_str());
+        }
+        return (void*) gfa_graph;
+    }, [](const void* gfa_graph_void, ostream& output) {
+        assert(gfa_graph_void != nullptr);
+        graph_to_gfa((const GFAHandleGraph*) gfa_graph_void, output);
+    });
```
adamnovak, line 69: This looks identical to the [GFA output code](https://github.com/vgteam/vg/blob/6193310e3bc538bcb56baa281528f940db3d109e/src/io/register_loader_saver_gfa.cpp#L127-L131), and we didn't do anything to `GFAHandleGraph` to remember if it came from a GFAz, so I think this isn't really registering GFAz output support.

I might need to add a `register_bare_loader_with_magic` to really make this work right, if writing GFAz isn't meant to be possible as part of this.

babyplutokurt:
@adamnovak I update the GFAz integration to GFAz head while leave the same commit in VG, I think it's better for you to do the merge and refactor on VG side

src/algorithms/gfa_to_handle.hpp
```diff
@@ -281,8 +319,51 @@ class GFAParser {
...
+
+/// Drive the shared node/edge import logic from an already-configured parser.
+void parser_to_handle_graph(GFAParser& parser,
+                            MutableHandleGraph* graph);
+
+/// Drive the shared node/edge/path import logic from an already-configured parser.
+void parser_to_path_handle_graph(GFAParser& parser,
+                                 MutablePathMutableHandleGraph* graph,
+                                 int64_t max_rgfa_rank = numeric_limits<int64_t>::max(),
+                                 unordered_set<PathSense>* ignore_sense = nullptr);
```
adamnovak: I don't think these do what they say they do. `parser_to_handle_graph()` in particular only calls `add_graph_listeners()` with the same arguments. Neither of these actually call `parse()` which I would think "drives" the parsing. It looks more like these are configuring the parser.

src/algorithms/gfaz_to_handle.cpp
```diff
@@ -0,0 +1,521 @@
...
+    for (const auto& visit : path_info.visits) {
+      callback(visit.node_id, visit.offset, visit.length, path_name, path_info.rank);
+    }
+  }
+}
+
+void GFAzParser::parse(istream& in) {
+  (void)in;
+  throw invalid_argument("GFAZ input from streams is not supported directly");
+}
```
adamnovak, line 424: But the first thing [the deserialization does](https://github.com/babyplutokurt/GFAz/blob/5dc0a3303a76e87d752731f52c6b2d9faf0018cf/src/core/codec/serialization.cpp#L289-L290) is open an ifstream; it wouldn't be hard to add an overload in the library to work on any seekable stream.

And the library seems to use the end position *only* for checking that you aren't reading past EOF, so it ought to be eventually possible to adjust it to detect unseekable streams and treat them as having an all-1s size or something.

src/algorithms/gfaz_to_handle.cpp
```diff
@@ -0,0 +1,521 @@
...
+  gfaz_paths.original_walk_offsets = build_offsets(compressed.original_walk_lengths);
+
+  size_t encoded_path_count =
+      gfaz_paths.path_offsets.empty() ? 0 : gfaz_paths.path_offsets.size() - 1;
+  size_t path_count = min(gfaz_paths.path_names.size(), encoded_path_count);
+  for (size_t i = 0; i < path_count; ++i) {
+    vector<NodeId> visits = decode_sequence_at_index(
+        gfaz_paths.paths_flat, gfaz_paths.path_offsets,
+        gfaz_paths.original_path_offsets, i, gfaz_paths.rules_first,
+        gfaz_paths.rules_second, gfaz_paths.min_rule_id, compressed.delta_round);
```
adamnovak: Aren't there helper functions somewhere in GFAz to help decompress the right compressed blocks with the right target types? This function in vg needs to know a lot about the internals of GFAz to work; it's not going through a narrow interface.

I would want to try and split the decompression concern from the concern of explaining the GFAz contents to the rest of vg, if possible. If the narrow interface isn't in GFAz I would want to build it here.

src/algorithms/gfa_to_handle.hpp
```diff
@@ -133,6 +164,13 @@ class GFAParser {
     // And a type for a collection of GFA tags.
     // This could become a range or list of ranges if we wanted to copy less.
     using tag_list_t = vector<string>;
+    // And a callback shape for iterating path/walk visits without exposing
+    // the source encoding. The node_name is a view into a buffer owned by the
+    // caller and is valid only for the duration of the call. If node_id is
+    // nonzero, it is the pre-resolved graph node ID and the listener should
+    // use it directly instead of looking up node_name; text parsers pass 0.
+    using visit_step_t = function<bool(int64_t rank, nid_t node_id, std::string_view node_name, bool is_reverse)>;
+    using visit_iteratee_t = function<void(const visit_step_t&)>;
```
adamnovak: I think what's `visit_step_t` here is really the iteratee; the "iteratee" in vg terminology is the function that gets called with each item in some collection (here the visits).

So then `visit_iteratee_t` is not actually the iteratee but the function that *takes* the iteratee; it would be better described as something like a "visit source".

And then the `make_gfaz_visit_iteratee()` function later would really be more like a `make_gfaz_visit_source()`.

src/algorithms/gfaz_to_handle.cpp
```diff
@@ -0,0 +1,521 @@
...
+    }
+  }
+
+  vector<vector<NodeId>> seqs(1);
+  seqs[0] = std::move(decoded);
+  for (int round = 0; round < delta_round; ++round) {
+    Codec::inverse_delta_transform(seqs);
+  }
+  return std::move(seqs[0]);
+}
```
adamnovak: If these functions are really the guts of understanding GFAz, they might benefit from comments so the reader can understand the guts of GFAz (or at least what the functions promise to do and what the different inputs need to be).

src/algorithms/gfa_to_handle.cpp
```diff
@@ -168,29 +208,24 @@ static void add_path_listeners(GFAParser& parser, MutablePathMutableHandleGraph*
...
-                                             bool step_is_reverse) {
+        visits([&](int64_t step_rank, nid_t step_node_id, std::string_view step_name, bool step_is_reverse) {
             if (step_rank >= 0) {
                 // Not an empty path sentinel.
-                // Find the node ID to visit.
-                nid_t n = GFAParser::find_existing_sequence_id(GFAParser::extract(step_name), parser.id_map());
-                // And add the step.
+                // Use the pre-resolved ID if the parser supplied one; otherwise look it up by name.
+                nid_t n = step_node_id != 0 ? step_node_id
+                                            : GFAParser::find_existing_sequence_id(step_name, parser.id_map());
```
adamnovak: The lookup here is all in terms of the base `GFAParser` and things it has. And the code that's ultimately calling this to loop over things already knows if we need to do a lookup through the map (when we're a `GFATextParser`) or not (when we're a `GFAzParser`).

So why do we provide for handling both a string view *and* a node ID in the path visit interface? Why not make the listener always use a numeric ID, and make the code for calling the listener on the text parser side do the `find_existing_sequence_id()` call? It doesn't really need to be generic here at runtime because we always do one thing for one implementation and the other thing for the other implementation.

src/algorithms/gfa_to_handle.hpp
```diff
@@ -95,6 +111,21 @@ void gfa_to_path_handle_graph(istream& in,
...
+                                           int64_t max_rgfa_rank = numeric_limits<int64_t>::max(),
+                                           unordered_set<PathSense>* ignore_sense = nullptr);
+
+/// Overload which serializes its translation to a file internally.
+void load_gfa_or_gfaz_to_path_handle_graph(const string& filename,
+                                           MutablePathMutableHandleGraph* graph,
+                                           int64_t max_rgfa_rank,
+                                           const string& translation_filename,
+                                           unordered_set<PathSense>* ignore_sense = nullptr);
+
```
adamnovak: Can't the new one-or-the-other functions here completely replace the GFA-only ones?

src/algorithms/gfaz_to_handle.cpp
```diff
@@ -0,0 +1,521 @@
...
+      sn_col = &col;
+    } else if (col.tag == "SO" && col.type == 'i') {
+      so_col = &col;
+    } else if (col.tag == "SR" && col.type == 'i') {
+      sr_col = &col;
+    }
+  }
+  if (!sn_col || !so_col || !sr_col) {
+    return;
+  }
```
adamnovak: There's no comments explaining this, but it seems odd to me that we need to pull out all the tags in `collect_optional_tags_for_row`, and then also have different code to go through just the rGFA tags all at once at the end.

Is this because we're required to produce the visits in sorted order and it's easier to parse them out twice than to remember them when we're doing the general tag code?

src/algorithms/gfaz_to_handle.cpp
```diff
@@ -0,0 +1,521 @@
...
+    vector<NodeId> visits = decode_sequence_at_index(
+        gfaz_paths.paths_flat, gfaz_paths.path_offsets,
+        gfaz_paths.original_path_offsets, i, gfaz_paths.rules_first,
+        gfaz_paths.rules_second, gfaz_paths.min_rule_id, compressed.delta_round);
+    auto visit_iteratee = make_gfaz_visit_iteratee(visits);
+    handle_duplicate_paths(*this, 'P', [&]() {
+      for (auto& listener : this->path_listeners) {
+        listener(gfaz_paths.path_names[i], visit_iteratee, GFAParser::tag_list_t());
+      }
+    });
```
adamnovak: This looks a little odd in use, but I think it's a good idea. But we could probably make it cleaner if it was a member, so we didn't need to pass `*this`.

adamnovak (commented):
OK, I went through this again. I think I could probably make the changes I want, but one of the things I'm wondering about is whether all the logic in here that's dealing with zstd decompression and interpreting rules is *really* going through the front door API of GFAz, which I think only you can answer @babyplutokurt.

The other thing that would need GFAz changes is that I'd really like to be able to read a graph from an unseekable stream (if not right now, then eventually), and it seems like the reason we can't is just that GFAz lacks the stream overload and really wants to double-check its reads against the file length.

src/algorithms/gfa_to_handle.hpp
adamnovak: This might want to happen when we do #4880.

src/algorithms/gfaz_to_handle.cpp
adamnovak: I think we can't fix this on the vg side until something happens on the GFAz side.

src/algorithms/gfaz_to_handle.cpp
adamnovak: I organized this part a bit more so it looks a bit more like utilities to help you interpret GFAz.

It's still more tightly coupled to the guts of the library/format than I really want it to be, but we can just stay stuck on the version we're on until changes get made to support new versions.

src/algorithms/gfaz_to_handle.cpp
adamnovak: I gave up and put up with the second pass through the tags to do the rGFA visits.

adamnovak (approved):
This now pulls in https://github.com/vgteam/libvgio/pull/91 and I think I have knocked it into a shape I'm mostly happy with.

We still need GFAz support for loading from unseekable streams/standard input, but I don't think that will bee too hard to do later.

babyplutokurt:
> OK, I went through this again. I think I could probably make the changes I want, but one of the things I'm wondering about is whether all the logic in here that's dealing with zstd decompression and interpreting rules is _really_ going through the front door API of GFAz, which I think only you can answer @babyplutokurt.
> 
> The other thing that would need GFAz changes is that I'd really like to be able to read a graph from an unseekable stream (if not right now, then eventually), and it seems like the reason we can't is just that GFAz lacks the stream overload and really wants to double-check its reads against the file length.


Thanks for going through this and for the refactoring.

The current vg integration does not materialize a complete intermediate GfaGraph; it decodes the required columns and emits records through vg’s listener interface. However, you’re right that it currently uses exported low-level GFAz helpers and still understands more of the compressed-column and grammar-rule representation than an external consumer ideally should.

GFAz’s existing streaming interface writes GFA text directly. I think the missing front-door API is therefore a callback/visitor-style decoder that keeps the current streaming behavior while letting GFAz own the format interpretation.

I agree about adding an std::istream overload as well. For unseekable streams, we should validate the length-delimited reads directly rather than requiring seekg()/tellg() to determine the complete file size first.

Do you want me make changes to GFAz right now so we can land ths changes in this PR or we could make another PR about this?

babyplutokurt:
Cool, I will ship another PR since you have already approved the merge, let's wait for tests ckean.

adamnovak:
Something's wrong in the CI tests:
https://ucsc-ci.com/vgteam/vg/-/jobs/112852#L4096

Our `vgci/vgci.py::VGCITest::test_sim_chr21_snp1kg_trained` [test](https://github.com/vgteam/vg/blob/6193310e3bc538bcb56baa281528f940db3d109e/vgci/vgci.py#L1395) is failing because the mapping AUC has dropped to `0.8992689634481723`.

I don't know how that would happen, because those tests all use `.vg` format Protobuf graphs; I don't think they use GFA, let alone GFAz.

I can try rerunning that test on mainline vg to see if something is broken overall for it now.

adamnovak:
The corresponding mainline vg run is https://ucsc-ci.com/vgteam/vg/-/jobs/112856

adamnovak:
According to the stdout log for the failing test from https://github.com/vgteam/vg/commit/6a80a5f4533e8b51736128b6cb44c47c7d6c400e#commitcomment-192020487, the problem isn't even vg; the AUC is way down for bwa-mem.

```
<VGCI name = "map eval results" tsv = "True">
Method	Acc.	Baseline Acc.	AUC	Baseline AUC	Max F1	Baseline F1
bwa-mem-pe	↑ 0.99995	0.9999	↓ 0.89927	0.99986	↑ 0.99997	0.99995
primary-mp-pe	↓ 0.99989	0.9999	↑ 0.99996	0.99989	↓ 0.99994	0.99995
primary-pe	↑ 0.9999	0.99988	↑ 0.99976	0.99972	↑ 0.99995	0.99994
snp1kg-mp-pe	↑ 0.99985	0.99982	↑ 0.99994	0.99983	↑ 0.99992	0.99991
snp1kg-pe	↑ 0.99986	0.99977	↑ 0.99952	0.99944	↑ 0.99993	0.99988
</VGCI>
```

[Nothing has happened for about a year](https://quay.io/repository/biocontainers/bwa?tab=tags&tag=0.7.19--h577a1d6_1) to the [Docker image we use for BWA](https://github.com/vgteam/toil-vg/blob/df65e86373e1f7502de4a08c80fc62311d82a9a4/src/toil_vg/vg_config.py#L176).

I'm going to rerun the test to see if this is some new and exciting nondeterministic failure mode. I think we do simulate new reads every time, so the results are a *little* different every time, but it's not supposed to be like 0.1 below its baseline.

adamnovak:
Here's the report from the current failed test run, since I think if I rerun it will overwrite it.
[report.zip](https://github.com/user-attachments/files/29910822/report.zip)

babyplutokurt:
looks like a flakky test run results?

The failed BWA run had only five incorrect mappings, but one had MAPQ 60, which was enough to reduce AUC to about 0.899.

adamnovak:
OK, it looks like everything worked when I reran it.

I'm not sure how to change the test do it can't fail like that in the future; MAPQ 60 wrong reads from BWA-MEM aren't impossible, and we probably want the reads to be randomized for the test, but if we do see unusually low performance we want to fail the test. But if we make it big enough to have a stable average number of wrong MAPQ 60 reads, then it will be slow. Maybe the least bad option is to stop randomizing the reads and fix a seed.

None of that matters for this PR, so I'm going to go ahead and merge it.

faithokamoto:
Oh cool I guess this got merged in time for our journal club on the GFAz paper.

babyplutokurt:
> Oh cool I guess this got merged in time for our journal club on the GFAz paper.

@faithokamoto The current GFAz codebase includes many optimizations that were not covered in the paper.

Compression throughput has improved by 1.8× compared with the version reported in the paper. We have also introduced several new compute-over-compression subcommands. For example, GFAz deconstruct achieves a 70× speedup over vg deconstruct while using 6× less memory. (https://github.com/babyplutokurt/GFAz#compute-engine-performance)

Feel free to give it a try! :)

PR #4862, opened 2026-03-25
Score seeds correctly

adamnovak:
## Changelog Entry

 * Giraffe no longer ignores the parts of seeds that extend outside their graph nodes to the left when scoring them. Note that this can **reduce R10 read variant calling accuracy** versus the previous release of vg.
 * Giraffe `hifi` mapping preset has been re-tuned for new seed score distribution.
 * Chain visualizations no longer need to be panned or zoomed to show changes to the traceback.
 * Chain visualizations no longer accumulate more and more transition lines when mousing in and out of a selected node.
 * Giraffe no longer tries to position-index all haplotypes when showing work. If you need all haplotypes position-indexed for debugging chaining against them, use `--haplotype-positions`.

## Description

This fixes a read @faithokamoto reported that came unmapped around when we got rid of two-phase chaining. It turns out we weren't scoring seeds right and, without fragmenting, we preferred to lit off after a well-scored seed acorss a giant gap, instead of collecting a poorly-scored seed.

We still might need to limit transitions' required indels to the length of the longest detectable gap where they occur, to avoid that sort of giant-cheap-chaining-gap-makes-alignment-score-negative behavior in other circumstances. But if we no longer arbitrarily don't credit minimizers for their score points, it should happen less.

We might need to re-tune essentially everything in chaining for the increased average seed score, and we should make sure we weren't leaning on this as a pseudo-heuristic with some acceptance testing before merging.

### DIFF
--- a/scripts/build-chain-viz.py
+++ b/scripts/build-chain-viz.py
@@ -17,7 +17,10 @@ from collections import defaultdict, Counter
 def parse_seeds_file(filepath):
     """
     Parse a chain seeds file.
-    Returns list of dicts with keys: read_pos, ref_name, ref_pos, strand, seed_num, seed_name
+    Returns list of dicts with keys: read_pos, ref_name, ref_pos, strand,
+    seed_num, seed_name. Seeds that don't lie on any linear reference path
+    have empty strings for ref_name, ref_pos, and strand; ref_pos is stored
+    as None in that case.
     """
     seeds = []
     with open(filepath, 'r') as f:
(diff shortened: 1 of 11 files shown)


adamnovak:
I'm now pretty sure this isn't adding new and exciting Giraffe bugs, due to sufficient testing.

adamnovak:
I got the HiFi HG002 results down to "just" 30 more total calling errors than mainline vg with this "bugfix". I wasn't able to find an R10 parameter set that was more optimal then the current defaults. Even though fixing the seed scoring makes the R10 calling results for HG002 worse overall, I'm going to merge this, because Benedict thinks we should merge recombination-aware mapping on top and *then* try and optimize R10 calling accuracy by tuning the parameters.

PR #4870, opened 2026-04-07
Hublabel

electricEpilith:
## Changelog Entry

 * Add hub labeling to distance index, which allows efficient exact shortest distance queries even in "oversized" snarls
 * Bug fix for minimizer, significant speed improvement

## Description
Adds hub labeling functionality to the snarl distance index.


src/subcommand/minimizer_main.cpp
```diff
@@ -91,6 +91,12 @@ int main_minimizer(int argc, char** argv) {
             logger.info() << "Loading SnarlDistanceIndex from " << config.distance_name << std::endl;
         }
         distance_index = vg::io::VPKG::load_one<SnarlDistanceIndex>(config.distance_name);
+        // Preload the index eagerly to establish it as recently-used in the OS
+        // page cache. Even though kmer counting may evict some pages, we
+        // re-preload right before cache_payloads. The double-preload is
+        // necessary: a single preload just before cache_payloads isn't enough
+        // to keep the index resident under the memory pressure of 32 parallel
+        // threads and the remaining in-memory data structures.
```
adamnovak: I don't think that's how page caching works; if it's paged in it's paged in, right? It can't possibly get *more* paged in if you add a second, earlier copy of the step where it got paged in.

We shouldn't be doing magic; if somehow this genuinely does improve performance we need to be able to explain why, in terms of something like a page eviction algorithm we can link to that's trying to be clever and really does care how long something has been cached.

src/unittest/support/snarl_decomposition_fuzzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+        [&](handle_t h) { events.push_back({ET::END_SNARL, h}); }
+    );
+
+    if (events.empty()) {
+        return;
+    }
+
+    // Step 2: Build pairing vector mapping each begin to its matching end
+    // and vice versa, using separate stacks for chains and snarls.
+    std::vector<size_t> pair_of(events.size());
```
adamnovak: `pair_of` is a bad name for this, it ought to be something like `other_end`.

(I'm pretty sure I put it in though.)

src/unittest/support/snarl_decomposition_fuzzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+#include <cassert>
+#include <stack>
+
+namespace vg {
+namespace unittest {
+
+using ET = DecompositionEventType;
+
+// SnarlDecompositionFuzzer deterministic constructor
```
adamnovak:
```suggestion
```

This is more confusing than enlightening. We have documentation for the constructors in the HPP that explains how we have one that gets pre-loaded and one that doesn't, using enough words.

src/unittest/support/snarl_decomposition_fuzzer.hpp
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * A HandleGraphSnarlFinder that wraps another HandleGraphSnarlFinder and
+ * randomly flips chains in the snarl decomposition. Flipping a chain reverses
+ * the entire chain including all children; if a child chain is also selected
+ * for flipping, it gets flipped again (canceling the parent's flip for that
+ * child).
+ *
+ * For deterministic testing, chains_to_flip can be provided, which is a set
+ * of (begin_handle, end_handle) pairs identifying chains to flip. When
+ * provided, p_flip and the generator are ignored.
```
adamnovak:
```suggestion
 * For non-randomized testing, the specific chains to flip can be 
 * pre-identified and provided on construction.
```

Even the randomized testing is meant to be deterministic. And we shouldn't start talking about parameter names as if they're real things here in the class doc comment.

src/unittest/support/snarl_decomposition_fuzzer.hpp
```diff
@@ -0,0 +1,157 @@
...
+        const std::function<void(handle_t)>& begin_snarl,
+        const std::function<void(handle_t)>& end_snarl
+    ) const override;
+
+private:
+    /// The wrapped snarl finder
+    const HandleGraphSnarlFinder* wrapped;
+
+    /// Function that decides whether to flip a chain, given either of its
+    /// bounding node IDs. May be nondeterministic.
```
adamnovak:
```suggestion
    /// bounding node IDs. May produce different results when called
    /// multiple times with the same input.
```

It still really ought to be a deterministic sequence of results for a sequence of calls and a random seed somewhere.

src/unittest/support/snarl_decomposition_fuzzer.hpp
```diff
@@ -0,0 +1,157 @@
...
+ */
+class ReplaySnarlFinder : public HandleGraphSnarlFinder {
+public:
+    /// Alias for shared event types
+    using EventType = DecompositionEventType;
+    using Event = DecompositionEvent;
+
+    /**
+     * Construct a replay finder that will emit the given events.
+     * The graph pointer can be null since we never actually use it.
```
adamnovak:
```suggestion
```

What graph pointer? No graph pointer here!

src/unittest/support/snarl_decomposition_fuzzer.hpp
```diff
@@ -0,0 +1,157 @@
...
+        const std::function<void(handle_t)>& end_chain,
+        const std::function<void(handle_t)>& begin_snarl,
+        const std::function<void(handle_t)>& end_snarl
+    ) const override;
+
+private:
+    std::vector<Event> events;
+};
+
+// Template implementation
```
adamnovak:
```suggestion
```

There's only one, it doesn't need a delimited section.

src/unittest/randomly_flipped_nodes.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+namespace vg {
+namespace unittest {
+
+/// Get the canonicalized set of edge sequence pairs from a graph.
+/// Each edge is represented as a pair of sequences (left_seq, right_seq) read
+/// in the orientation of the edge. To canonicalize, we compare each pair
+/// against its reverse complement (RC(right_seq), RC(left_seq)) and keep the
+/// lexicographically smaller one.
+static set<pair<string, string>> canonical_edge_pairs(const HandleGraph& graph) {
```
adamnovak, line 24: This is kind of a hack, but I guess it works for potentially catching mistakes.

I'm pretty sure I added this.

src/unittest/randomly_flipped_nodes.cpp
```diff
@@ -0,0 +1,190 @@
...
+    return result;
+}
+
+TEST_CASE("randomly_flipped_nodes preserves graph structure on a simple linear graph", "[randomly_flipped_nodes]") {
+    bdsg::HashGraph graph;
+    handle_t h1 = graph.create_handle("AAA", 1);
+    handle_t h2 = graph.create_handle("CGC", 2);
+    handle_t h3 = graph.create_handle("TTT", 3);
+    graph.create_edge(h1, h2);
+    graph.create_edge(h2, h3);
```
adamnovak: If we're doing tests involving 50% of the nodes, we could stand to have a bigger graph here with more nodes.

src/unittest/randomly_flipped_nodes.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    auto original_edges = canonical_edge_pairs(graph);
+
+    SECTION("flipping no nodes preserves edges exactly") {
+        default_random_engine gen(test_seed_source());
+        auto flipped = randomly_flipped_nodes(graph, 0.0, gen);
+
+        REQUIRE(flipped.get_node_count() == 3);
+        REQUIRE(flipped.get_edge_count() == 2);
```
adamnovak: These in all the test cases could compare against the relevant statistics from `graph` instead of constants.

src/unittest/snarl_decomposition_fuzzer.cpp
```diff
@@ -0,0 +1,448 @@
...
+}
+
+TEST_CASE("ReplaySnarlFinder replays events faithfully", "[snarl_decomposition_fuzzer]") {
+    // Build a small graph to get real handles
+    bdsg::HashGraph graph;
+    graph.create_handle("A", 10);
+    graph.create_handle("C", 12);
+    graph.create_handle("G", 15);
+    graph.create_handle("T", 20);
+    graph.create_handle("AA", 22);
```
adamnovak: There's no good reason to creatively assign graph node IDs in these tests.

src/unittest/snarl_decomposition_fuzzer.cpp
```diff
@@ -0,0 +1,448 @@
...
+    std::vector<Event> events = {
+        {ET::BEGIN_CHAIN, h10f},
+        {ET::BEGIN_SNARL, h10f},
+        {ET::BEGIN_CHAIN, h12r},
+        {ET::END_CHAIN, h15r},
+        {ET::END_SNARL, h20f},
+        {ET::BEGIN_SNARL, h20f},
+        {ET::END_SNARL, h22f},
+        {ET::END_CHAIN, h22f},
+    };
```
adamnovak: All these handle-getting calls could be inlined in the expected events vector and it would be probably *more* readable.

src/unittest/snarl_decomposition_fuzzer.cpp
```diff
@@ -0,0 +1,448 @@
...
+        std::vector<pair<ET, handle_t>> expected = {
+            {ET::BEGIN_CHAIN, h22r},
+              {ET::BEGIN_SNARL, h22r},
+              {ET::END_SNARL, h20r},
+              {ET::BEGIN_SNARL, h20r},
+                {ET::BEGIN_CHAIN, h15f},
+                {ET::END_CHAIN, h12f},
+              {ET::END_SNARL, h10r},
+            {ET::END_CHAIN, h10r},
+        };
```
adamnovak: We could also just change `capture_events` to spit out event, node ID, orientation tuples, and then we could write cleaner expected values and not need to load handles into weirdly named variables to re-use them.

src/unittest/snarl_decomposition_fuzzer.cpp
```diff
@@ -0,0 +1,448 @@
...
+    ReplaySnarlFinder replay(events);
+
+    SECTION("flipping an empty chain") {
+        std::unordered_set<nid_t> flips {5};
+        SnarlDecompositionFuzzer fuzzer(&graph, &replay, flips);
+
+        auto captured = capture_events(fuzzer);
+
+        // Flipped: begin_chain(flip(5fwd)) = begin_chain(5rev)
+        //          end_chain(flip(5fwd)) = end_chain(5rev)
```
adamnovak: I'm not sure this tells us anything.
```suggestion
```

src/unittest/snarl_decomposition_fuzzer.cpp
```diff
@@ -0,0 +1,448 @@
...
+        std::unordered_set<nid_t> flips {1, 6};
+        SnarlDecompositionFuzzer fuzzer(&graph, &replay, flips);
+
+        auto captured = capture_events(fuzzer);
+
+        // Outer flipped: snarls reversed, and inner chains also reversed
+        // as part of the parent flip.
+        // Chain 5f->5f reversed to 5r->5r.
+        // Chain 2f->3f (containing snarl 2f->3f) reversed to 3r->2r
+        // (containing snarl 3r->2r).
```
adamnovak: This is just re-explaining how the flip logic works again.
```suggestion
```

src/unittest/snarl_decomposition_fuzzer.cpp
```diff
@@ -0,0 +1,448 @@
...
+    SECTION("flip outer and inner chain 2->3") {
+        std::unordered_set<nid_t> flips {1, 6, 2, 3};
+        SnarlDecompositionFuzzer fuzzer(&graph, &replay, flips);
+
+        auto captured = capture_events(fuzzer);
+
+        // Outer flipped (reversing everything, including chain 2f->3f to
+        // 3r->2r and chain 5f->5f to 5r->5r), AND THEN inner chain 2f->3f
+        // is flipped again back to its original orientation 2f->3f.
+        // Chain 5f->5f is NOT in flip set, so it stays reversed as 5r->5r.
```
adamnovak: This has too many moving parts to be so boring. If nothing interesting is happening we shouldn't carefully describe everything as if something is.
```suggestion
```

src/unittest/snarl_distance_index.cpp
```diff
@@ -192,7 +197,80 @@ namespace vg {
...
+            graph.create_edge(h2, h5);
+
+            IntegratedSnarlFinder snarl_finder(graph);
+
+            SECTION("Snarl classifications are correct") {
+                SECTION("Distance index") {
+                    SnarlDistanceIndex distance_index;
+                    fill_in_distance_index(&distance_index, &graph, &snarl_finder);
+                    REQUIRE(distance_index.is_regular_snarl(distance_index.get_parent(distance_index.get_parent(distance_index.get_node_net_handle(graph.get_id(h3))))));
+                } SECTION("Distanceless index") {
```
adamnovak:
```suggestion
                }
                SECTION("Distanceless index") {
```

src/unittest/snarl_distance_index.cpp
```diff
@@ -192,7 +197,80 @@ namespace vg {
...
+            graph.create_edge(h3, graph.flip(h3));
+
+            IntegratedSnarlFinder snarl_finder(graph);
+
+            SECTION("Snarl classifications are correct") {
+                SECTION("Distance index") {
+                    SnarlDistanceIndex distance_index;
+                    fill_in_distance_index(&distance_index, &graph, &snarl_finder);
+                    REQUIRE(!distance_index.is_regular_snarl(distance_index.get_parent(distance_index.get_parent(distance_index.get_node_net_handle(graph.get_id(h3))))));
+                } SECTION("Distanceless index") {
```
adamnovak:
```suggestion
                }
                SECTION("Distanceless index") {
```

src/gbwtgraph_helper.cpp
```diff
@@ -519,8 +534,18 @@ gbwtgraph::DefaultMinimizerIndex build_minimizer_index(
...
         // A zipcode only depends on the node id.
-        hash_map<nid_t, payload_t> node_id_to_payload;
+        vg::hash_map<nid_t, payload_t> node_id_to_payload;
         node_id_to_payload.reserve(gbz.graph.max_node_id() - gbz.graph.min_node_id());
+        // Re-preload the distance index right before use. find_frequent_kmers
+        // runs for a long time and may evict the mmap'd index pages from the OS
+        // page cache. We also preload eagerly right after loading the index (in
+        // minimizer_main.cpp) so the kernel treats those pages as recently-used;
+        // together the two preloads prevent cache_payloads from page-faulting on
+        // every node under the memory pressure of 32 parallel threads.
```
adamnovak: This doesn't really explain how two passes of preloading could possibly help, either.

src/snarl_distance_index.cpp
```diff
@@ -782,31 +788,62 @@ SnarlDistanceIndex::TemporaryDistanceIndex make_temporary_distance_index(
...
+ * Also responsible for filling in min_length, distance_start_start, and distance_start_end on the TemporarySnarlRecord when a distance matrix is used.
+ */
+static void populate_distance_matrix_row(SnarlDistanceIndex::TemporaryDistanceIndex& temp_index, const SnarlDistanceIndex::temp_record_ref_t& snarl_index, SnarlDistanceIndex::TemporaryDistanceIndex::TemporarySnarlRecord& temp_snarl_record, const SnarlDistanceIndex::temp_record_ref_t& start_index, const HandleGraph* graph, size_t start_rank, bool is_internal_node, size_t size_limit); 
+
+/** 
+ * Fills in required distance matrix rows for each child
+ * - Normal snarl: all rows
+ * - Oversized snarl: boundaries and tips
+ * - size_limit == 0: no distances in index, so no rows
+ * - Top-level chain distances only: ???
```
adamnovak: It would be good to figure this out and fill this in.

src/snarl_distance_index.cpp
```diff
@@ -1063,52 +1098,201 @@ void populate_snarl_index(
...
+      // port.
+      temp_snarl_record.min_length = promote_distance<size_t>(hhl_query(temp_snarl_record.hub_labels.begin(), bgid(0, false, true), bgid(1, false, false)));
+      temp_snarl_record.distance_start_start = promote_distance<size_t>(hhl_query(temp_snarl_record.hub_labels.begin(), bgid(0, false, true), bgid(0, true, false)));
+      temp_snarl_record.distance_end_end = promote_distance<size_t>(hhl_query(temp_snarl_record.hub_labels.begin(), bgid(1, true, true), bgid(1, false, false)));
+      // TODO: Should this be here or should it be part of populate_hub_labeling()? Or its own function?
     } else {
+      if (size_limit == 0 || only_top_level_chain_distances) { 
         temp_snarl_record.include_distances = false;
+      }
+      //Also fills in min_lenght, distance_start_start, and distance_start_end, and sets is_simple to false if snarl isn't simple
```
adamnovak:
```suggestion
      // Also fills in min_length, distance_start_start, and distance_start_end, and sets is_simple to false if snarl isn't simple
```

src/snarl_distance_index.cpp
```diff
@@ -1063,52 +1098,201 @@ void populate_snarl_index(
...
+  for (auto e : boost::make_iterator_range(edges(ov))) {
+    const EdgeProp& ep = ov[e];
+    std::cerr << "Edge " << source(e, ov) << " -> " << target(e, ov)
+              << ": contracted=" << (ep.contracted ? "true" : "false")
+              << " weight=" << ep.weight
+              << " arc_cover=" << ep.arc_cover
+              << " ori=" << (ep.ori ? "true" : "false") << std::endl;
+  }
+  std::cerr << "=== End CHOverlay Dump ===" << std::endl;
+#endif
```
adamnovak, line 1189: I think I put it it here, but this could stand to become a CHOverlay method or debug function.

src/snarl_distance_index.cpp
```diff
@@ -1118,357 +1302,588 @@ void populate_snarl_index(
...
-             start_rank != 0 && start_rank != 1))) {
+        //traversal start is not a tip or a boundary node
+        bool start_normal_child = (!start_is_tip && start_rank != 0 && start_rank != 1);
+ 
+        if ( (temp_snarl_record.node_count > size_limit || size_limit == 0 || only_top_level_chain_distances) && (temp_snarl_record.is_root_snarl || start_normal_child)) {
             //If we don't care about internal distances, and we also are not at a boundary or tip
             //TODO: Why do we care about tips specifically?
             continue;
         }
+        //getting here means snarl is not oversized
```
adamnovak: We only even get into this function now if we're not oversized, or if `size_limit` is 0 and we're not including distances at all. This should be reworked to understand that.

adamnovak (commented):
Here's my review, including a bunch of stuff I now want to change about code I committed.

The changes to `libbdsg` also need to be reviewed.

src/subcommand/minimizer_main.cpp
electricEpilith: This actually improves speed (at the cost of more memory usage). idk why yet

src/unittest/support/snarl_decomposition_fuzzer.hpp
adamnovak: Actually the graph here is an inherited field, and it can't be null or `HandleGraphSnarlFinder::find_snarls_unindexed()`, which we don't override, won't work. We can't just not use a method we inherit and get out of needing to follow base class invariants.

adamnovak:
This goes with https://github.com/vgteam/libbdsg/pull/239, for reference.

src/snarl_distance_index_child_graph.hpp
```diff
@@ -0,0 +1,69 @@
...
+//
+// Encapsulates the "get outgoing handle → follow_edges → map to snarl child"
+// kernel shared by topo_sort_children and populate_distance_matrix_row.
+// Both callers need the same three-step operation:
+//   1. compute the handle pointing out of `child` in direction `go_left`
+//   2. follow graph edges from that handle
+//   3. for each landing node, resolve its snarl-level ancestor + direction
+//
+// Dijkstra priority-queue management and topo-sort BFS logic stay with the
+// respective callers.
```
adamnovak: This speaks to how this class got made but does not explain what it *is*, and might amount to noise rather than signal to someone trying to figure out what's going on here. The stuff about this 3-step operation really properly belongs to the documentation fro a method, and the stuff about material that *wansn't* moved during a refactor probably should be deleted.

It can make sense to mentions the expected callers of a method as a see-also, but it doesn't often make sense to define something in terms of where else it is used.

src/snarl_distance_index_child_graph.hpp
adamnovak, line 1: Should this just be `snarl_child_graph.hpp`? It's not actually a real inner class of the SnarlDistanceIndex.

src/snarl_distance_index_child_graph.hpp
```diff
@@ -0,0 +1,69 @@
...
+class SnarlChildGraph {
+public:
+    using temp_record_ref_t  = SnarlDistanceIndex::temp_record_ref_t;
+    using TempIndex          = SnarlDistanceIndex::TemporaryDistanceIndex;
```
adamnovak: Are these here because we really can expect users of `SnarlChildGraph` to not need to know that a `SnarlChildGraph::TempIndex` happens to be the same as a `SnarlDistanceIndex::TemporaryDistanceIndex`? Or are they just here to save the fingers of the code generator from having to type more letters? If it's the second one we probably should ditch them and just use the full type names, because adding more distinct names for the same things will confuse the humans.

If it turns out we need all this temporary record stuff in a bunch of places, we might want to move it out of the enclosing class in `libbdsg`.

src/snarl_distance_index_child_graph.hpp
adamnovak: This should be a real `///` or `/** */` doc comment.

src/snarl_distance_index_child_graph.hpp
```diff
@@ -0,0 +1,69 @@
...
+    //                     (chain traversed right-to-left, or node reversed)
+    //   edge_distance   — traversal length of `neighbor`:
+    //                     TEMP_NODE → sequence length of the landing node;
+    //                     TEMP_CHAIN → chain.min_length (∞ if disconnected)
+    //   arriving_node_id — graph node id of the immediate landing handle
+    //                      (before ancestor resolution; needed by callers that
+    //                       must preserve the original follow_edges semantics
+    //                       for is_simple detection)
+    //
+    // Boundary nodes ARE included in the callback; callers filter as needed.
```
adamnovak: This should be a real doc comment.

src/snarl_distance_index_child_graph.hpp
```diff
@@ -0,0 +1,69 @@
...
+    using TempIndex          = SnarlDistanceIndex::TemporaryDistanceIndex;
+
+    // `children` may be empty if the caller doesn't need the children() accessor.
+    // All other methods only require temp_index, snarl_index, and graph.
+    SnarlChildGraph(TempIndex& temp_index,
+                    temp_record_ref_t snarl_index,
+                    std::span<const temp_record_ref_t> children,
+                    const handlegraph::HandleGraph* graph);
+
+    std::span<const temp_record_ref_t> children() const noexcept;
```
adamnovak: So this class stores the children *just* to hand them back to you?

It looks like the `children()` accessor is never called, and nothing else in this class uses the children.

src/snarl_distance_index_child_graph.hpp
```diff
@@ -0,0 +1,69 @@
...
+    //   edge_distance   — traversal length of `neighbor`:
+    //                     TEMP_NODE → sequence length of the landing node;
+    //                     TEMP_CHAIN → chain.min_length (∞ if disconnected)
+    //   arriving_node_id — graph node id of the immediate landing handle
+    //                      (before ancestor resolution; needed by callers that
+    //                       must preserve the original follow_edges semantics
+    //                       for is_simple detection)
+    //
+    // Boundary nodes ARE included in the callback; callers filter as needed.
+    void for_each_outgoing(
```
adamnovak: There might want to be a noun in this function: for each outgoing *what*?

I think this whole class does a bad job of representing the idea of the net graph among a snarl's children. It seems to exist *just* to shave 3 arguments (the temporary index, the snarl itself, and the graph) off of this method, and I'm not sure it's worth it.

If we want to keep it, then this class should do a better job of really representing a net graph for a snarl. It shouldn't call it the "child net-graph" because that makes it sounds like it belongs to one of the children and not the snarl itself. And it shouldn't say, here are the children in the net graph but you don't really have to provide them and they never get used. It should provide more of the basic operations that are actually needed to use a net graph, and expect you to use those, rather than calling a callback with the results of very specific queries. `ancestor_of_node_in_snarl` and `child_side_reversed` would be methods of the class that you call if you need them, and not extremely specific static helpers. This iteration method should have the iteratee structured so that we just tell where you end up (maybe in terms of the destination node ID and orientation?), and you query back to this class for the other things you need to know about.

You want to be building construction kits for the software you need to build, not a whole Galapagos of highly specialized creatures each with its own niche.

src/snarl_distance_index_query.cpp
```diff
@@ -0,0 +1,705 @@
...
+          if ((SnarlDistanceIndex::sum(next_max_length, current_distance) !=
+                   std::numeric_limits<size_t>::max() &&
+               SnarlDistanceIndex::sum(next_max_length, current_distance) >=
+                   min_distance)) {
+            if (distance_index.is_node(next)) {
+              size_t curr_distance_end = SnarlDistanceIndex::minus(
+                  SnarlDistanceIndex::sum(next_max_length, current_distance),
+                  1);
+          // If its a node that puts us over, add the node to the subgraph, then
+          // start the search from that node
```
adamnovak: This seems to be at the wrong indentation level.

src/snarl_distance_index_query.cpp
```diff
@@ -0,0 +1,705 @@
...
+    handle_t bound = distance_index.get_handle(current_node, super_graph);
+
+    super_graph->follow_edges(bound, false, [&](const handle_t &next_handle) {
+      search_start_nodes.emplace_back(next_handle, current_distance);
+      seen_nodes.erase(make_pair(super_graph->get_id(next_handle),
+                                 super_graph->get_is_reverse(next_handle)));
+    });
+    // seen_nodes.erase(make_pair(super_graph->get_id(bound),
+    // super_graph->get_is_reverse(bound))); search_start_nodes.emplace_back(
+    // bound, current_distance);
```
adamnovak: Rather than rewrapping this commented-out code,w e might drop it.

src/snarl_distance_index_query.cpp
adamnovak, line 1: Is the idea here to pull out all the definitions (but not the declarations) for the functions for this query operation into their own source file, since they're related to each other?

Could this be made into something in `vg::algorithms` instead, with its own header? Then it would be a much more clearly separate module. We could also turn the helper functions into static ones, and hide them from the header, since they're only used for the one main function and it now has its own translation unit.

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+
+struct DistanceRowContext {
+    SnarlDistanceIndex::TemporaryDistanceIndex&                       temp_index;
+    temp_record_ref_t                                                 snarl_index;
+    SnarlDistanceIndex::TemporaryDistanceIndex::TemporarySnarlRecord& snarl_record;
+    temp_record_ref_t                                                 start_index;
+    size_t                                                            start_rank;
+    bool                                                              is_internal_node;
+    size_t                                                            size_limit;
+    SnarlChildGraph&                                                  child_graph;
```
adamnovak: These are probably too lined up given the length of one of the types.

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    // Pass an empty children span — for_each_outgoing doesn't use it.
+    SnarlChildGraph child_graph(temp_index, snarl_index,
+                                std::span<const SnarlDistanceIndex::temp_record_ref_t>{}, graph);
+    DistanceRowContext ctx{temp_index, snarl_index, temp_snarl_record, start_index,
+                           start_rank, is_internal_node, size_limit, child_graph};
+    for (bool start_rev : starting_directions(ctx)) {
+        run_dijkstra_from_side(ctx, start_rev);
+    }
+    finalize_internal_node_contribution(ctx);
+}
```
adamnovak: OK, so we've successfully moved all the complexity out of `populate_distance_matrix_row()`, which now just has to say that it does some Dijkstras and then fills in some stuff based on them.

But it doesn't look like we've learned much about the algorithm by breaking it up on the context object. None of the new pieces have the sorts of doc comments that you would expect to be obligated to add in order to set out an understanding of the algorithm sufficient to accomplish this transformation. What does it really mean to "finalize" the "contribution" of internal node(s)?

If we really need this many moving parts, and a struct to keep state between function calls, we probably don't want everything kicking around an anonymous namespace at the tail end of a file. I would make the context into a `DistanceMatrixRowBuilder`, make everything else here into private inner types and methods, and give it a constructor to set it all up and a `build()` method to perform the side effects. And I would document all the fields and methods because when the whole component is this big the whole point of splitting it up like this is to create a place where you can write about the interconnections. The whole thing could get its own translation unit if we want.

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    } else if (!ctx.snarl_record.is_root_snarl && ctx.start_rank == 1 &&
+               current_index != ctx.start_index &&
+               !(next_index.first == SnarlDistanceIndex::TEMP_NODE &&
+                 next_index.second == ctx.snarl_record.start_node_id)) {
+        ctx.snarl_record.is_simple = false;
+    }
+}
+
+[[nodiscard]] NeighborSide resolve_neighbor_side(const DistanceRowContext& ctx,
+        temp_record_ref_t next_index, bool next_rev) {
```
adamnovak: I don't think it makes a lot of sense to start marking things `[[nodiscard]]` without some notion of why they would need to be, or some documentation of the semantics of the operation. Usually you use it when e.g. you have an error code that people shouldn't be allowed to just not react to. We want the compiler to trigger a warning when someone throws away the result, to encourage them to come back here and properly understand the docs about how they're supposed to be holding the operation. But there's no docs here.

Are we worried someone would ask us to figure out which neighbor side they need to use, and then for some reason throw the answer away? How do we think they think they are getting any useful work done in that case?

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+        rank = next_index.first == SnarlDistanceIndex::TEMP_NODE
+            ? ctx.temp_index.get_node(next_index).rank_in_parent
+            : ctx.temp_index.get_chain(next_index).rank_in_parent;
+    }
+    bool is_boundary = !ctx.snarl_record.is_root_snarl && (rank == 0 || rank == 1);
+    bool reversed    = is_boundary ? false : next_rev;
+    return NeighborSide{rank, reversed, is_boundary};
+}
+
+[[nodiscard]] bool record_distance(DistanceRowContext& ctx,
```
adamnovak: This is `[[nodiscard]]` and only seems to be used at one point, where the result is promptly discarded.

What does the result even mean? Is it an indication of an error? How does it need to be handled?

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    SnarlChildGraph&                                                  child_graph;
+};
+
+struct NeighborSide {
+    size_t rank;
+    bool   reversed;
+    bool   is_boundary;
+};
+
+[[nodiscard]] vector<bool> starting_directions(const DistanceRowContext& ctx) {
```
adamnovak: Is this `[[nodiscard]]` to try and catch people who think maybe it does something to the context that would help them, instead of returning an answer?

In that case it might also help to give it a name that makes it clear it is meant to go compute something and give it to you, like `get_startting_directions()`.

Documentation for how it's meant to be use would probably help more than hints to the compiler on how it shouldn't be used.

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+        //fill in all distances for a row
+        populate_distance_matrix_row(temp_index, snarl_index, temp_snarl_record, start_index, graph, start_rank, is_internal_node, size_limit);   
+    }                                                                                                                    
+}      
+      
+    
+                        
+namespace {
+
+using temp_record_ref_t = SnarlDistanceIndex::temp_record_ref_t;
```
adamnovak: Again, if we're constantly aliasing this type we might just want to bump it up into `bdsg::`, where we can `using` it per file.

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+ * A simple snarl is always regular.
+ */
+bool check_regularity(const SnarlDistanceIndex::TemporaryDistanceIndex& temp_index, const SnarlDistanceIndex::temp_record_ref_t& snarl_index, const SnarlDistanceIndex::TemporaryDistanceIndex::TemporarySnarlRecord& temp_snarl_record, const vector<SnarlDistanceIndex::temp_record_ref_t>& all_children, const HandleGraph* graph);
+
+// ---------------------------------------------------------------------------
+// Phase helpers for populate_snarl_index (all file-static)
+// ---------------------------------------------------------------------------
+
+// Step 1: Walk up the snarl tree from curr_index until we find the direct
+// child of ancestor_snarl_index that contains curr_index.
```
adamnovak: If these are phases or steps of a larger process, they shouldn't be floating around as static helpers sibling to static helpers for the phases and steps of a *different* process. If these are real helper functions, they should be documented in terms of what they require and guarantee: if this finds the child of `ancestor_snarl_index` that contains `curr_index`, just say that. Then it makes sense to potentially use this anywhere one would need that operation, instead of just as one step in one particular process.

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+            }
+        }
+
+        callback(neighbor, neighbor_rev, edge_distance, arriving_nid);
+        return true;
+    });
+}
+
+// Phase 1: Mark tip nodes and set is_simple=false if any tip is found.
+static void identify_tips(
```
adamnovak: This should be explained more in terms of what fields it promises to fill in on what.

src/snarl_distance_index_build.cpp
```diff
@@ -0,0 +1,1708 @@
...
+            temp_snarl_record.include_distances = false;
+        }
+        // Also fills in min_length, distance_start_start, distance_start_end, sets is_simple=false if not simple.
+        populate_distance_matrix_if_needed(temp_index, snarl_index, temp_snarl_record, all_children, graph, size_limit, only_top_level_chain_distances);
+    }
+}
+
+// Phase 4: For simple snarls, record child node orientations.
+// IMPORTANT: iterates temp_snarl_record.children[0..node_count), NOT all_children —
+// boundary nodes appended in Phase 3 must not be included here.
```
adamnovak: Does this function *only* ever make sense to call immediately after `compute_snarl_distances()`? Or can what it requires and guarantees be articulated more clearly?

adamnovak (changes requested):
I like the idea of breaking up all the stuff in `snarl_distance_index.cpp` into more distinct units. But I think the design and especially documentation of the units needs work, and we also need some overarching organization of those units that isn't just several cpp files that all belong to one header.

I think some of these pieces might make sense as a few different algorithms in `vg::algorithms`. There each algorithm could have its own header defining the interface, and then all the helper stuff to power each could be nicely tucked away in the cpp file for just that algorithm.

If instead we want to keep them organized under the theme of building a snarl distance index, we might give them a folder and a namespace that reflects that, and then again give each piece its own header. Then that whole namespace would have something that sort of constitutes its external interface (`populate_snarl_index()`?) and around there we would have some documentation on how the different responsibilities are spread across these modules.

adamnovak:
I think there's also still outstanding stuff to do in https://github.com/vgteam/libbdsg/pull/239 around dropping code from the oracle algorithms that aren't being used, and making sure we have a fresh file format version number, which I think is covered by the current review there.

It looks like CI is failing here because the hash-constraining tests weren't actually dropped yet.

electricEpilith:
Dropped the hash tests. Moved refactor commits to `hublabel-refactor`.

src/subcommand/minimizer_main.cpp
```diff
@@ -91,6 +91,12 @@ int main_minimizer(int argc, char** argv) {
...
         distance_index->preload(true);
```
adamnovak: @electricEpilith If I remember right, you did some research and found out that only the second preload is actually important for speed, so we can drop this first one, right?

```suggestion
```

src/gbwtgraph_helper.cpp
```diff
@@ -537,8 +552,18 @@ gbwtgraph::DefaultMinimizerIndex build_minimizer_index(
...
```
adamnovak:
```suggestion
        // Preload the distance index right before use.
```

src/snarl_distance_index.cpp
```diff
@@ -1118,357 +1281,588 @@ void populate_snarl_index(
...
```
adamnovak:
```suggestion
```

adamnovak (changes requested):
This looks pretty good. I found a couple places where it's not quite done and made suggestions.

I think the libbdsg changes also need to be finished, and then this needs to be made to point at the final libbdsg commit. Right now it's behind https://github.com/vgteam/libbdsg/pull/239 by a few commits.

src/snarl_distance_index.cpp
electricEpilith: think it's fine here

src/snarl_distance_index.cpp
electricEpilith: done

src/snarl_distance_index.cpp
electricEpilith: done

src/subcommand/bench_dist_query_main.cpp
```diff
@@ -0,0 +1,233 @@
...
+        if (show_progress) {
+            cerr << "Loading distance index from " << dist_path << "..." << endl;
+        }
+        SnarlDistanceIndex distance_index;
+        distance_index.deserialize(dist_path);
+        cerr << "Loaded distance index from " << dist_path << endl;
+
+        // Pull the whole index into the OS page cache so timings reflect query
+        // cost, not first-touch I/O for indexes that exceed free RAM.
+        distance_index.preload(true);
```
adamnovak: What does the index exceeding free RAM have to do with anything? I'm not sure this sentence understands itself.
```suggestion
        // Pull the whole index into the OS page cache so timings reflect
        // unavoidable query cost, not avoidable first-touch I/O
        distance_index.preload(true);
```

deps/libbdsg
adamnovak, line 1: This doesn't seem to be pulling in https://github.com/vgteam/libbdsg/pull/239/changes/e0adffb0050d066a24c4af31d00db4c799d2ba42 which makes a breaking change to the distance index format. So we definitely don't want to merge this without that change.

adamnovak (changes requested):
This looks pretty good now. The only thing I think we *really* need to change is making sure we point at the actual libbdsg commit we want.

deps/libbdsg
electricEpilith: updated libbdsg commit

.gitlab-ci.yml
```diff
@@ -319,7 +319,11 @@ report-job:
 docs-job:
     interruptible: true
     stage: build
+    before_script:
+      - sudo apt-get -q -y update
+      - sudo apt-get -q -y install --no-upgrade doxygen graphviz rsync git
```
adamnovak, line 325: Are we sure we need these (including `git` and `rsync` which we never installed before?)

If we install doxygen here, do we need it in the general before_script anymore?

src/unittest/snarl_distance_index.cpp
```diff
@@ -7869,6 +7869,50 @@ namespace vg {
...
+            REQUIRE(graph->has_node(154));
+
+            IntegratedSnarlFinder snarl_finder(*graph);
+            SnarlDistanceIndex distance_index;
+            fill_in_distance_index(&distance_index, graph.get(), &snarl_finder, 1);
+
+            size_t fwd = distance_index.minimum_distance(103, false, 0, 154, true, 3, false, graph.get());
+            size_t rev = distance_index.minimum_distance(154, true, 3, 103, false, 0, false, graph.get());
+            cerr << "[overreport] fwd(103fd0->154rev3) = " << fwd
+                 << "   rev(154rev3->103fd0) = " << rev << endl;
```
adamnovak: We don't want to keep debugging prints in the tests if we can avoid it.
```suggestion
```

src/unittest/snarl_distance_index.cpp
```diff
@@ -7869,6 +7869,50 @@ namespace vg {
...
         }
 
+        TEST_CASE( "Distance index hub labeling does not over-report unreachable distances",
+                  "[snarl_distance][snarl_distance_hub_label_overreport]" ) {
+            // Regression repro captured from the random hub-label fuzzer. In the dumped
+            // graph, 103fd0 -> 154rev3 is unreachable (confirmed by bidirected BFS /
+            // Dijkstra) but the reverse 154rev3 -> 103fd0 IS reachable. The hub-label
+            // query was returning the reverse direction's finite distance (264) for the
+            // unreachable forward query.
+            unique_ptr<HandleGraph> graph = vg::io::VPKG::load_one<HandleGraph>("hublabel_overreport.vg");
```
adamnovak: We don't want the unit tests to depend on data files; they should pass if you just download the static binary and run `vg test`.

This might want to become a CLI test, or we might want to include the offending graph (or, ideally, a minimally-sized version of that graph with as much stuff removed or simplified as possible) as JSON here to load for the test.

adamnovak (changes requested):
I think this is good except that we don't want to add unit tests that depend on on-disk data files. We want the unit tests to pass when all you have is the binary, since we ship them in the binary.

adamnovak (approved):

PR #4879, opened 2026-04-20
Giraffe server engine

parsaeskandar:
## Changelog Entry

## Description

This PR adds a server-oriented in-process Giraffe mapping API and extends giraffe-server to emit structured graph-path payloads that middleware can forward to coordinate-translation code.


src/subcommand/giraffe_server_main.cpp
```diff
@@ -0,0 +1,234 @@
...
+        static struct option long_options[] = {
+            {"gbz", required_argument, 0, 'Z'},
+            {"minimizer", required_argument, 0, 'm'},
+            {"distance", required_argument, 0, 'd'},
+            {"zipcodes", required_argument, 0, 'z'},
+            {"threads", required_argument, 0, 't'},
+            {"max-multimaps", required_argument, 0, 'M'},
+            {"batch-size", required_argument, 0, 'b'},
+            {"emit-header", no_argument, 0, 1000},
+            {"framed-output", no_argument, 0, 1001},
```
adamnovak: I think using raw numbers here without any `constexpr int EMIT_HEADER_OPT = 1000;` sort of variable baffles @faithokamoto's option linting script and makes the Mac CI tests fail with an error inside the script.

Either this needs to change to use some variables to name the numbers and make sure they're the same numbers here and in the `switch`, or the linting script needs to change to understand this (or to prohibit it). Probably both should change.

src/subcommand/giraffe_server_main.cpp
```diff
@@ -0,0 +1,234 @@
...
+         << "                               (may be repeated; required for per-read targets)" << endl
+         << "  -h, --help                   show help" << endl
+         << endl
+         << "stdin protocol (one read per line):" << endl
+         << "  SEQUENCE" << endl
+         << "  NAME<TAB>SEQUENCE" << endl
+         << "  NAME<TAB>SEQUENCE<TAB>QUALITY" << endl
+         << "  NAME<TAB>SEQUENCE<TAB>QUALITY<TAB>SURJ_TARGET" << endl
+         << "                               (QUALITY may be empty when SURJ_TARGET is set)" << endl
+         << "  FLUSH_NOW   (force immediate processing of buffered reads)" << endl
```
adamnovak: OK so this is less of a "flush" for data that might be stuck in output buffers, and more of a go-align-the-current-batch command?

src/subcommand/giraffe_server_main.cpp
```diff
@@ -0,0 +1,234 @@
...
+         << "      --emit-header            emit GAF header lines before output" << endl
+         << "      --framed-output          emit read-grouped framed output for middleware" << endl
+         << "  -S, --surject-target NAME    pre-index this haplotype path for surjection" << endl
+         << "                               (may be repeated; required for per-read targets)" << endl
+         << "  -h, --help                   show help" << endl
+         << endl
+         << "stdin protocol (one read per line):" << endl
+         << "  SEQUENCE" << endl
+         << "  NAME<TAB>SEQUENCE" << endl
+         << "  NAME<TAB>SEQUENCE<TAB>QUALITY" << endl
```
adamnovak, line 56: Is the quality encoded as phred+33 (as in a FASTQ) or as raw phred bytes or as something else?

src/subcommand/giraffe_server_main.cpp
```diff
@@ -0,0 +1,234 @@
...
+                read.name = fields[0];
+                read.sequence = fields[1];
+                read.quality = fields[2];
+            } else {
+                read.name = fields[0];
+                read.sequence = fields[1];
+                read.quality = fields[2];
+                read.surjection_target = fields[3];
+            }
+
```
adamnovak, line 346: There's no validation here to ensure that e.g. sequence and quality are the same length, or that we haven't gotten something that can't be a read but also isn't a known command.

src/subcommand/giraffe_server_main.cpp
```diff
@@ -0,0 +1,234 @@
...
+            if (read.sequence.empty()) {
+                continue;
+            }
```
adamnovak, line 359: So if a request to map a read with a specified name but no sequence is provided, it will never be answered.

src/subcommand/giraffe_server_main.cpp
```diff
@@ -0,0 +1,234 @@
...
+         << "  -m, --minimizer FILE         minimizer index path" << endl
+         << "  -d, --distance FILE          distance index path" << endl
+         << "  -z, --zipcodes FILE          zipcode index path" << endl
+         << endl
+         << "optional:" << endl
+         << "  -t, --threads N              mapping threads [1]" << endl
+         << "  -M, --max-multimaps N        max mappings per read [1]" << endl
+         << "  -b, --batch-size N           input reads per mapping batch [256]" << endl
+         << "      --emit-header            emit GAF header lines before output" << endl
+         << "      --framed-output          emit read-grouped framed output for middleware" << endl
```
adamnovak, line 48: Since we explain the input format in the help, we should probably also explain the output format in the same place. What should a person expect to see come out in framed mode and in non-framed mode?

src/haplotype_assignment.hpp
```diff
@@ -0,0 +1,92 @@
...
+/** \file
+ * Post-processing utility that, for a final Giraffe Alignment, computes the
+ * set of haplotypes that traverse the alignment's graph path, optionally
+ * extended outward to the nearest enclosing snarl boundaries.
+ *
+ * Uses the already-loaded GBZ (GBWT + GBWTGraph) and SnarlDistanceIndex that
+ * the engine owns. No new index files are required.
+ *
+ * Thread-safe after construction: assign() is const and does not mutate any
+ * member state. The precomputed node_to_enclosing_snarl table is read-only.
```
adamnovak, line 13: I don't know what any of these things are yet, and I don't think it makes sense to try and race a constructor against other methods, so it might make sense to just say the class is thread-safe here.

src/haplotype_assignment.hpp
```diff
@@ -0,0 +1,92 @@
...
+    /// capping by max_haplotypes_to_report.
+    size_t candidate_count = 0;
+    /// Left snarl-boundary node (in forward orientation) if the alignment was
+    /// extended to a boundary. 0 if no extension was requested, or if
+    /// extension walked off the graph, hit a branch, or was not performed.
+    nid_t left_boundary_node = 0;
+    /// Right snarl-boundary node, analogous to left_boundary_node.
+    nid_t right_boundary_node = 0;
+    /// True if seq_ids was truncated to max_haplotypes_to_report.
+    bool truncated = false;
```
adamnovak, line 76: This just tells you if the length of `seq_ids` is less than `candidate_count`, and is never read.

src/giraffe_engine.cpp
```diff
@@ -0,0 +1,410 @@
...
+                auto& m = mapped[j];
+                auto gaf = io::alignment_to_gaf(gbz_->graph, m);
+                stringstream ss;
+                ss << gaf;
+                if (!hap_results.empty()) {
+                    const auto& hap = hap_results[j];
+                    ss << "\thp:Z:" << haplotype_seq_ids_csv(hap)
+                       << "\thc:i:" << hap.candidate_count
+                       << "\thl:i:" << hap.left_boundary_node
+                       << "\thr:i:" << hap.right_boundary_node
```
adamnovak, line 627: Isn't this going to let placeholder values of the struct fields out as essentially-meaningless tags, when we could omit those tags instead? Or do we assume `assign_haplotypes_extend_to_snarls` is always staying true here?

src/haplotype_assignment.hpp
```diff
@@ -0,0 +1,92 @@
...
+    const SnarlDistanceIndex* dist_index_;
+
+    /// node_to_enclosing_snarl_[id] is the dense integer id of the snarl that
+    /// immediately encloses the node with the given node id. 0 is a sentinel
+    /// meaning "no enclosing snarl" (nodes directly under the root).
+    std::vector<uint32_t> node_to_enclosing_snarl_;
+
+    /// Return the enclosing-snarl id for the given node id, or 0 if the node
+    /// is out of range.
+    uint32_t enclosing_snarl(nid_t node_id) const;
```
adamnovak, line 147: Can't the SnarlDistanceIndex answer this question fast enough to avoid needing this extra table?

src/giraffe_engine.hpp
```diff
@@ -0,0 +1,165 @@
...
+};
+
+/// Convert a status enum value into a short stable token suitable for
+/// emitting in GAF tags or framed protocol fields.
+const char* haplotype_surjection_status_token(HaplotypeSurjectionResult::Status status);
+
+/**
+ * Owns a Surjector configured against a PathPositionHandleGraph that already
+ * has the desired target paths indexed for positions. Used by GiraffeEngine
+ * to project graph alignments onto a specific GBZ-known haplotype path.
```
adamnovak, line 81: OK but what *is* it? Is it just a widget that does some of the things a `Surjector` does but in a way that `GiraffeEngine` finds more convenient? Should it be not just in this file but maybe an inner class of `GiraffeEngine`?

Do we really get mileage out of this class, or would its functionality be better expressed as methods on `GiraffeEngine` that can do surjection in terms of numbered target sequences instead of whatever `Surjector` takes to specify that?

src/giraffe_engine.hpp
```diff
@@ -0,0 +1,165 @@
...
+struct GiraffeFastqRead {
+    std::string name;
+    std::string sequence;
+    std::string quality;
+    /// Optional: name of a pre-indexed haplotype path to surject this read
+    /// onto. Empty means "no surjection". If set, the engine will surject
+    /// every reported alignment for this read onto the given path and
+    /// append the result as GAF tags.
+    std::string surjection_target;
+};
```
adamnovak, line 136: I don't think this really represents the concept of a FASTQ read, or that there needs to be a special abstraction for FASTQ reads for Giraffe or GiraffeServer purposes.

Usually we use `Alignment` as the internal format for even our yet-to-be-aligned reads. If we need to bring along the surjection target information or we want something that maps really closely to the input format for the server we probably want something like this but representing a `GiraffeServerRequest`.

src/giraffe_engine.hpp
```diff
@@ -0,0 +1,165 @@
...
+// Forward declaration so we can mention Alignment in surjection types
+// without forcing every includer to drag in vg.pb.h.
+class Alignment;
+
+struct GiraffeEnginePaths {
+    std::string gbz_path;
+    std::string minimizer_path;
+    std::string distance_path;
+    std::string zipcode_path;
+};
```
adamnovak, line 33: Does this *just* exist to limit the number of arguments that `GiraffeEngine::load()` takes? Maybe it should be closer to the function it belongs to? Why isn't this attached to the `GiraffeEngineConfig` somehow?

src/giraffe_engine.hpp
```diff
@@ -0,0 +1,165 @@
...
+    bool is_loaded() const;
+
+    const GiraffeEngineConfig& config() const;
+
+    /// If true, map_reads() post-processes each final Alignment to annotate
+    /// it with the set of haplotypes that traverse its path.
+    bool assign_haplotypes = true;
+    /// If true and assign_haplotypes is true, the haplotype search is
+    /// extended outward to the nearest enclosing snarl boundaries.
+    bool assign_haplotypes_extend_to_snarls = true;
```
adamnovak, line 171: Why are these public members, while other similar things belong to the config?

adamnovak (commented):
@parsaeskandar You wanted me to check this for overly tight coupling to what Giraffe does. I don't *think* it should have to change too much when Giraffe changes. We won't be able to avoid touching it if we e.g. add a new thing `MininizerMapper` needs to have loaded in order to work, but I'm not sure there's a great way to avoid that.

But I think you might not connect to the Giraffe machinery quite enough, because there's no way here to e.g. set the mapping parameters that need to change between HiFi and R10 reads (and there's no code to set them to good values for pretending to be BLAT). I'm not sure we're even turning on the chain-based mapping. The Giraffe preset system is all tied in with its option parsing system, which makes using it from here difficult, but maybe it needs to be broken out somehow?

There also seems to be a profusion of little classes defined mostly in relation to each other, like `HaplotypeSurjectionResult` and `GiraffeEnginePaths`, which could do with more explanation of what they really represent or mean for someone who does not, to begin with, care about a `HaplotypeSurjector` or a `GiraffeEngine`.

PR #4882, opened 2026-04-21
add new "sampling" workflow to vg autoindex

faithokamoto:
## Changelog Entry

 * `vg autoindex` has a `-w sampling` workflow to make indexes for haplotype sampling 

## Description

Resolves #4872

That was easier than expected. I figured out the various places that would need to be poked in order to register a new workflow (thankfully all of the individual pieces were already known as recipes) and slapped this together. Should make it easier to point folks who need to make these indexes because they can't index the full graph.

### DIFF
--- a/src/index_registry.cpp
+++ b/src/index_registry.cpp
@@ -4670,10 +4670,8 @@ vector<IndexName> VGIndexes::get_default_long_path_giraffe_indexes() {
 
 vector<IndexName> VGIndexes::get_default_haplotype_sampling_indexes() {
     vector<IndexName> indexes{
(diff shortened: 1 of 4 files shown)


jeizenga:
The r-index and distance index are only needed temporarily, right? If so, you can leave them off of the default requested indexes. The indexing plan will figure out that it needs them, and then it will also delete the temporary files at the end on the run.

faithokamoto:
Oh cool, thanks.

faithokamoto:
The test that failed was just a timeout, so I'll just bring this branch up-to-date and let that all rerun.

PR #4886, opened 2026-04-23
Make `vg giraffe` error when the minimizers or zipcodes are older than the distance index

adamnovak:
## Changelog Entry

 * `vg giraffe` will now stop with an error when the minimizers or zipcodes are older than the distance index they were supposedly generated from.

## Description

We wasted some time trying to map reads with the wrong distance index, which didn't go with the minimizers and zipcodes files, because it had been replaced later.

This adds an error to Giraffe in that case. We might want to demote it to a warning if we run into a lot of trouble with downloading these files from places in the wrong order and not preserving their modification times, inside of workflows.

### DIFF
--- a/src/index_registry.cpp
+++ b/src/index_registry.cpp
@@ -551,6 +551,10 @@ construct_minimizers_impl(
         *gbz, distance_index.get(), &oversized_zipcodes, params
     );
 
+    // Close the distance index so it can't appear to be modified after the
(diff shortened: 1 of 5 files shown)


adamnovak:
We probably should extend/replace this with hashing the distance index and hiding the hash in the minimizer/zipcode file tags, like we do for the "graph name".

PR #4889, opened 2026-04-23
remove unnecessary return info

faithokamoto:
## Changelog Entry

 * Simplify an internal return value for `align_sequence_between()`

## Description

I saw a TODO and thought I could tackle it for kicks. Indeed, while `align_sequence_between()`/`align_sequence_between_consistently()` used to return a pair of values, the only thing we actually ever used those return values for was checking whether the first was nonzero or not. So now I simply have it return whether that first value would have been zero. Makes it easier to interpret. Also possibly saves a negligible amount of overhead that would have been spent constructing the pair.

### DIFF
--- a/src/minimizer_mapper.hpp
+++ b/src/minimizer_mapper.hpp
@@ -1081,11 +1081,11 @@ protected:
      *
      * For connecting alignment, restricts the alignment to use <= max_dp_cells
      * cells. If too many DP cells would be used, produces an Alignment with
-     * and empty path.
+     * an empty path.
      *
-     * Returns the number of nodes and bases in the graph aligned against.
+     * Returns whether a graph was aligned against or not.
      */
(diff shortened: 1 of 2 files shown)


faithokamoto:
Mac failed with `Could not resolve host: raw.githubusercontent.com` when I think it was trying to download something. @adamnovak do I just re-run or is there something I should fix?

adamnovak (approved):
This looks like a good simplification.

adamnovak:
I don't think there's anything to fix; our test runners sometimes inexplicably cannot talk to the Internet. I'm hoping this will go away when OpenStack is destroyed in June and I'm forced to migrate them to live somewhere else; I don't *think* the underlying problem is https://github.com/systemd/systemd/issues/32045 because the networking comes back and I don't remember setting up any watchdog stuff on the Gitlab runners.

PR #4891, opened 2026-05-01
new test file

faithokamoto:
## Changelog Entry

 * Added a little test file of reads for new Quickstart page

## Description

I wanted some reads to tell the Quickstart user to align, but I wasn't able to find any that would be easily understandable and directly related to a useful graph. So I simulated some. If there's a better option I'm happy to hear it.


faithokamoto:
@adamnovak unless you have some objection to adding another test file?

adamnovak:
My only concern would be the file size; we might not want to add 1000 reads to the Git history if 100 would do. But I think it's probably still small enough, and writing it *out* of the history to shrink it would be more work than it's worth.

PR #4895, opened 2026-05-05
update gbwtgraph to path index bugfix

dcmonti:
## Changelog Entry
## Description
Update gbwtgraph submodule to solve bug in path indexing in the minimizers.
Previously a minimizers with same position and key would have overwritten the paths of all previous occurences. The bugfix avoids it by merging their paths


jltsiren:
This was included in #4897 that was just merged into master.

PR #4898, opened 2026-05-08
Scoring Refactor

adamnovak:
## Changelog Entry

 * Alignment scoring and mapping quality computation have been broken out of `GSSWAligner` and moved to `AlignmentScorer` and `MappingQualityCalculator`.

## Description

This breaks the alignment-scoring and mapping quality computation responsibilities off of the aligner class hierarchy and gives them their own class hierarchies.

Now there's alignment scorers for normal, quality-adjusted, and Minimap2-style logged-gap approaches. Each of them knows how to produce the "log base" that we use for mapping quality computation and hand it off to the MAPQ calculator. The aligners can use the appropriate ones of these scoring classes for the actual DP.

A lot of this code was synthesized via Anthropic Claude, and I've gone over *most* of it to try and make it acceptable.

This *should* produce exactly the same answers as before.


adamnovak:
I think this is finally really ready; I've removed all the things that Anthropic Claude wanted me to do that I don't actually want to do, after several passes and several days.

PR #4901, opened 2026-05-12
Avoid upgrading packages and try to update emulator

adamnovak:
## Changelog Entry

 * vg CI builds of ARM containers should no longer segfault when upgrading libc

## Description

We've been getting CI ARM Docker build crashes like this: https://ucsc-ci.com/vgteam/vg/-/jobs/109050#L2786

This should avoid https://github.com/docker/buildx/issues/314 in two ways:
* We no longer upgrade packages from the base image when building the Docker image in CI. They *should* be up to date enough in the base image (but we might need to make sure that https://stackoverflow.com/q/77924205 is still true).
* We install a new binfmt-misc emulator that maybe won't have the crash bug. We're *supposed* to uninstall the old one and install the new one, but when we move the workload to shared machines we don't want the emulator getting uninstalled while in use.


adamnovak:
I'm going to merge this but it might need to be backed out because CI doesn't actually run the Gitlab job this touches for PRs.

PR #4903, opened 2026-05-13
Add --cluster-min-len option to vg deconstruct

glennhickey:
## Changelog Entry

 * Add `--cluster-min-len` option to `vg deconstruct` that sets threshold for when allele-merging (`-L`) is applied
 *

### DIFF
--- a/src/deconstructor.cpp
+++ b/src/deconstructor.cpp
@@ -864,26 +864,33 @@ bool Deconstructor::deconstruct_site(const handle_t& snarl_start, const handle_t
         // If a minimum-allele-length gate is set, skip clustering at sites
         // where every used traversal's interior (non-boundary) sequence is
         // shorter than the threshold.  This lets the clustering be SV-only
-        // (default 50bp matches the standard SV size cutoff) while leaving
-        // small variants represented exactly as they are.
-        bool gate_clustering = false;
+        // (setting it to 50bp matches the standard SV size cutoff) while
+        // leaving small variants represented exactly as they are.
+        bool do_clustering = true;
         if (cluster_min_allele_len > 0 && cluster_threshold < 1.0) {
-            int64_t max_alt_len = 0;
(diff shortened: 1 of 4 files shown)


src/deconstructor.cpp
```diff
@@ -859,11 +859,45 @@ bool Deconstructor::deconstruct_site(const handle_t& snarl_start, const handle_t
...
-                                                               trav_cluster_info,
-                                                               unused_child_snarl_mapping);
+        vector<vector<int>> trav_clusters;
+
+        // If a minimum-allele-length gate is set, skip clustering at sites
+        // where every used traversal's interior (non-boundary) sequence is
+        // shorter than the threshold.  This lets the clustering be SV-only
+        // (default 50bp matches the standard SV size cutoff) while leaving
+        // small variants represented exactly as they are.
+        bool gate_clustering = false;
```
adamnovak: `gate_clustering` controls whether we *don't* run clustering, but the name is ambiguous. We probably want either `do_clustering` or `skip_clustering`.

src/deconstructor.cpp
```diff
@@ -859,11 +859,45 @@ bool Deconstructor::deconstruct_site(const handle_t& snarl_start, const handle_t
...
+                const Traversal& t = travs[i];
+                int64_t len = 0;
+                if (t.size() > 2) {
+                    for (size_t k = 1; k + 1 < t.size(); ++k) {
+                        len += graph->get_length(t[k]);
+                    }
+                }
+                if (len > max_alt_len) max_alt_len = len;
+            }
+            gate_clustering = max_alt_len < cluster_min_allele_len;
```
adamnovak: Instead of taking the max length of any traversal and checking whether that max is under the threshold, shouldn't we look at the length of each traversal and stop when we find one that hits the threshold? We don't actually need the max length; we need to know whether any traversal exists that is long enough.

adamnovak (changes requested):

PR #4909, opened 2026-05-14
Minimizer index construction speed up

dcmonti:
## Changelog Entry

 * Speed up minimizer index construction.

## Description
Change the locate in minimizer index in order to use r-index. Parallelized payload caching

### DIFF
--- a/src/gbwtgraph_helper.cpp
+++ b/src/gbwtgraph_helper.cpp
@@ -449,7 +449,8 @@ std::vector<key_type> find_frequent_kmers(const gbwtgraph::GBZ& gbz, const Minim
 void cache_payloads(
     const gbwtgraph::GBZ& gbz,
     const SnarlDistanceIndex& distance_index,
-    hash_map<nid_t, payload_t>& node_id_to_payload,
+    std::vector<payload_t>& node_id_to_payload,
+    nid_t min_node_id,
     ZipCodeCollection* oversized_zipcodes,
     bool progress
 ) {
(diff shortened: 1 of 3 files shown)


src/gbwtgraph_helper.cpp
```diff
@@ -536,18 +541,20 @@ gbwtgraph::DefaultMinimizerIndex build_minimizer_index(
         gbwtgraph::index_haplotypes(gbz, index, [](const pos_t&) { return nullptr; });
     } else {
         // Cache payloads before building the index.
-        // A zipcode only depends on the node id.
-        hash_map<nid_t, payload_t> node_id_to_payload;
-        node_id_to_payload.reserve(gbz.graph.max_node_id() - gbz.graph.min_node_id());
-        cache_payloads(gbz, *distance_index, node_id_to_payload, oversized_zipcodes, params.progress);
+        // A zipcode only depends on the node id. The map is indexed by
```
adamnovak: There's no map anymore.

src/gbwtgraph_helper.cpp
```diff
@@ -449,7 +449,8 @@ std::vector<key_type> find_frequent_kmers(const gbwtgraph::GBZ& gbz, const Minim
 void cache_payloads(
     const gbwtgraph::GBZ& gbz,
     const SnarlDistanceIndex& distance_index,
-    hash_map<nid_t, payload_t>& node_id_to_payload,
+    std::vector<payload_t>& node_id_to_payload,
```
adamnovak: We probably want a comment on the function, and a new name for this, to convey how it doesn't actually get indexed by actual node ID anymore.

src/gbwtgraph_helper.cpp
dcmonti: Removed it in last commit

src/gbwtgraph_helper.cpp
dcmonti: Comment for `cache_payload` updated, name changed for `node_id_to_payload`

adamnovak (approved):

PR #4912, opened 2026-05-21
Add tests for getting tags from FASTQ through chaining Giraffe and into BAM

adamnovak:
## Changelog Entry

 * `--comments-as-tags` is now under test with `vg giraffe`'s chaining codepath
 * Surject tests now test SAM tags in GAM with an actual `vg surject` command line
 * `vg surject` now preserves unrecognized GAF tags as tags on output alignments (and GAF input in general retains tags)
 * `vg giraffe` chaining mode now properly retains input tags on unmapped reads

## Description

~There was concern that tags might be falling out on the way from FASTQ to GAM, when Giraffe is using its chaining codepath.~

I found some ways that tags could indeed fall off reads (specifically, when a read was unmapped on the chaining codepath, or if we went through GAF format and then `vg surject`).

This adds tests for those cases and fixes them.

### DIFF
--- a/src/minimizer_mapper.hpp
+++ b/src/minimizer_mapper.hpp
@@ -884,7 +884,7 @@ protected:
      * If we do gapless extension, turn good full-length gapless extensions into alignments and return them in alignments
      * Gapless extensions are considered good enough if they have fewer than default_max_extension_mismatches mismatches
      */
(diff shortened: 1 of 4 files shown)


adamnovak:
This uses https://github.com/vgteam/libvgio/pull/90

PR #4915, opened 2026-05-26
Reduce excessive path indexing

adamnovak:
## Changelog Entry

 * `vg find -Q/--paths-named` is now deprecated due to its partial-Protobuf output
 * `vg find` will now index its target paths but not other haplotype paths.
 * vg should no longer position-index haplotype paths unnecessarily in commands using the `PathPositionOverlayHelper`.

## Description

This should fix cases where using the `PathPositionOverlayHelper` was making us index all haplotype paths.

I'm not sure if more commands need to be patched like `vg paths` to send along the names of haplotype paths they're actually interested in, to let queries on haplotype paths work. But we've only been indexing all the paths for a few months, and before that I think the behavior before that was for those queries to mysteriously give wrong answers.

### DIFF
--- a/src/subcommand/clip_main.cpp
+++ b/src/subcommand/clip_main.cpp
@@ -306,7 +306,7 @@ int main_clip(int argc, char** argv) {
             // If we want all paths matching some prefixes
             graph->for_each_path_handle([&](path_handle_t path_handle) {
                 // Look at all ther paths
-                std::string path_name = pp_graph->get_path_name(path_handle);
+                std::string path_name = graph->get_path_name(path_handle);
                 subrange_t subrange;
                 // And get their base path names
                 path_name = Paths::strip_subrange(path_name, &subrange);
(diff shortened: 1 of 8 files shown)


adamnovak:
I think we might need to fix `vg stats --snarl-sample`; I'm not sure it's under test with non-reference samples.

adamnovak:
This uses https://github.com/vgteam/libbdsg/pull/241 in libbdsg.

adamnovak:
OK I went through all the subcommands using `PathPositionOverlayHelper` and I think I've fixed them all to ask to index the right paths.

The one I couldn't do yet was `vg inject`; now it can only inject into reference paths, since we don't have a great way to load the BAM header without the graph being ready.

We already have a lockout for that:
https://github.com/vgteam/vg/blob/1d90c0b6d54c219bfa427a06fddf3213cc9ff694/src/alignment.cpp#L590-L600

This should be ready to merge when CI passes.

PR #4916, opened 2026-05-27
Leave low-scoring MAPQ 0 reads unmapped

faithokamoto:
## Changelog Entry

 * Add new `vg giraffe` filter for low-scoring MAPQ 0 R10 reads

## Description

Per Slack discussion about how ridiculously low scoring MAPQ 0 R10 reads are invariably mapped incorrectly, here is my attempt at an implementation. It managed to successfully not align a test read. Basically this gives us a way to refuse to use an alignment if it looks terrible and we have no confidence in it. It's only on for R10 mode.

### DIFF
--- a/src/minimizer_mapper_from_chains.cpp
+++ b/src/minimizer_mapper_from_chains.cpp
@@ -1059,7 +1059,10 @@ vector<Alignment> MinimizerMapper::map_from_chains(Alignment& aln) {
     }
 
     if (track_provenance) {
-        funnel.substage("demapping");
+        funnel.stage("demapping");
+        for (size_t i = 0; i < mappings.size(); i++) {
+            funnel.project(i);
+        }
     }
 
     if (mapq == 0 && !scores.empty() && scores.front() < min_mapq0_score) {
@@ -1086,7 +1089,7 @@ vector<Alignment> MinimizerMapper::map_from_chains(Alignment& aln) {
     } else if (track_provenance) {
         // Pass all remaining mappings
         for (size_t i = 0; i < mappings.size(); i++) {
-            funnel.pass("mapq0-score", i, scores.front());
+            funnel.pass("mapq0-score", i);
         }
     }
 
--- a/src/minimizer_mapper_from_chains.cpp
+++ b/src/minimizer_mapper_from_chains.cpp
@@ -1058,18 +1058,22 @@ vector<Alignment> MinimizerMapper::map_from_chains(Alignment& aln) {
         }
     }
 
+    if (track_provenance) {
+        funnel.substage("demapping");
+    }
+
     if (mapq == 0 && !scores.empty() && scores.front() < min_mapq0_score) {
         if (show_work) {
             #pragma omp critical (cerr)
             {
-                cerr << log_name() << "Failing MAPQ 0 alignment for having score "
-                    << scores.front() << " which is below " << min_mapq0_score << endl;
+                cerr << log_name() << "Failing MAPQ 0 alignment for having top score "
+                     << scores.front() << " which is below " << min_mapq0_score << endl;
             }
         }
         if (track_provenance) {
             // Fail all remaining mappings
-            for (size_t i = 0; i < alignments.size(); i++) {
-                funnel.fail_if_needed("mapq0-score", i, scores.front());
+            for (size_t i = 0; i < mappings.size(); i++) {
+                funnel.fail("mapq0-score", i, scores.front());
             }
         }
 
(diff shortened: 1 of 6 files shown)


faithokamoto:
Yeah this is currently borked for track provenance stuff, sigh. I am trying to divine what magic the funnel system wants from me.

faithokamoto:
@adamnovak one of the tests is failing with 
```
Exception: Error running rtg tools. Return code was 1, output:  / Cannot determine system memory!Could not automatically choose percentage based memory allocation, check configuration. Using Java default.nullError: RTG has encountered a difficulty, please contact support@realtimegenomics.com
```
and I don't think that's my fault?

src/subcommand/giraffe_main.cpp
```diff
@@ -339,6 +339,12 @@ static std::unique_ptr<GroupedOptionGroup> get_options() {
         MinimizerMapper::default_mapq_score_scale,
         "scale scores for mapping quality"
     );
+    comp_opts.add_range(
+        "min-mapq0-score",
+        &MinimizerMapper::min_mapq0_score,
+        MinimizerMapper::default_min_mapq0_score,
+        "refuse to align lower scoring MQ 60 reads"
```
adamnovak: Nope that's not right.

It might make more sense to explain this as discarding the alignments or leaving the reads unaligned, rather than refusing to align them, because by the time we know the score we've computed the alignment.

src/minimizer_mapper_from_chains.cpp
```diff
@@ -1058,6 +1058,34 @@ vector<Alignment> MinimizerMapper::map_from_chains(Alignment& aln) {
...
+            {
+                cerr << log_name() << "Failing MAPQ 0 alignment for having score "
+                    << scores.front() << " which is below " << min_mapq0_score << endl;
+            }
+        }
+        if (track_provenance) {
+            // Fail all remaining mappings
+            for (size_t i = 0; i < alignments.size(); i++) {
+                funnel.fail_if_needed("mapq0-score", i, scores.front());
+            }
```
adamnovak, line 1077: This is a little weird because we're going back and failing the `alignments` on filters after we've already [projected](https://github.com/vgteam/vg/blob/d189bf889ce143a58f2e5d449863353d111459d0/src/minimizer_mapper_from_chains.cpp#L2397) some of them to entries in `mappings`. And since some of them might have failed other filters and *not* become mappings, we need the `_if_needed` versions of the pass and fail functions.

I think the right answer is probably to introduce another stage; we pick `mappings` from `alignments` in [the "winner" stage](https://github.com/vgteam/vg/blob/d189bf889ce143a58f2e5d449863353d111459d0/src/minimizer_mapper_from_chains.cpp#L916), but then right before we check the filter we should do `funnel.stage("demapping");` or something, which will put us in a new stage where the objects to be passed or failed are the entries in `mappings`. Then we can fail or pass them all without worrying about the `alignments` that might have already failed other filters.

adamnovak (changes requested):
I think the option documentation is wrong, and I think the funnel stuff would work a lot more neatly if we declared that this post-filtering is happening in a new funnel stage.

I'm not sure what's wrong with the CI tests; something about the new CI runners seems to have broken test job 7 everywhere. I will see if I can fix that somehow.

src/minimizer_mapper_from_chains.cpp
```diff
@@ -1058,6 +1058,41 @@ vector<Alignment> MinimizerMapper::map_from_chains(Alignment& aln) {
         }
     }
 
+    if (track_provenance) {
+        funnel.stage("demapping");
+        for (size_t i = 0; i < mappings.size(); i++) {
+            funnel.project(i);
```
adamnovak: We're meant to figure out if the item at the previous stage fails any filter, or if it passes all filters, and then only project to an item at this stage if it passed all the filters. See:
https://github.com/vgteam/vg/blob/21c53097507b4abea00b65a6c3fdefe88a1d22fd/src/funnel.hpp#L139-L140

So the project call should probably be right after the `pass()` call, instead of here.

adamnovak (changes requested):
I think the funnel setup wasn't explained enough, so this still isn't quite right.

I'm not having good luck in https://github.com/vgteam/vg/pull/4918 so we might need to merge this over the failure of that CI test (or maybe just xfail it until it can be fixed).

adamnovak (approved):

faithokamoto:
@adamnovak `ERROR: failed to push quay.io/vgteam/vg:cache-mapq0-score-build: denied: System is currently read-only. Pulls will succeed but all write operations are currently suspended.`

adamnovak:
Quay has returned, apparently.

PR #4918, opened 2026-05-29
Use vg WDL that gives RTG a defined memory limit

adamnovak:
## Changelog Entry

 * vg CI should no longer fail with complaints from `rtg`

## Description
This will hopefully fix the CI problem noted in https://github.com/vgteam/vg/pull/4916#issuecomment-4569185586 where rtg will fail on the new CI runners for mysterious and possibly memory-limit-related reasons.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,7 @@ RUN apt-get -qq -y update && apt-get -y install --no-upgrade \
 ###DEPS_END###
 
 FROM packages AS build
+ARG THREADS=8
 
 RUN echo build > /stage.txt
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,8 @@ RUN apt-get -qq -y update && apt-get -y install --no-upgrade \
 ###DEPS_END###
 
 FROM packages AS build
+# TODO: We're *supposed* to inherit ARGs from our base stage according to the
+# Docker documentation, but we don't.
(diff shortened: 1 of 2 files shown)


adamnovak:
Setting up the memory limit in advance didn't help, but this should now avoid the failing step (building an RTG SDF template inside of hap.py) by just passing in the one the workflow already made.

Then I made a bunch of changes to fix problems I encountered trying to build the vg Docker and run the test outside of CI, which I never actually managed to do.

adamnovak:
OK, the underlying problem is what was encountered at https://github.com/Illumina/hap.py/issues/189 where `rtg` just always fails with `null` for nearly anything. (I guess *except* making an SDF?)

I can test for it with:
```
docker run -ti --rm quay.io/adamnovak/hap.py:v0.3.15 /opt/hap.py/libexec/rtg-tools-install/rtg index
```
(Or more or less any `rtg` command that isn't just help.)

When I run this on Courtyard it works. When I run this on our new Ubuntu 26.04 CI runners with kernel 7.0 and a similarly fresh Docker daemon, it fails.

It was suggested in https://github.com/Illumina/hap.py/issues/189#issuecomment-3817745798 that replacing the `rtg` in the container with version 3.13 might help. I *think* 3.13 needs a newer Java than we have in the container, but the example approach doesn't seem to bother updating it.

PR #4929, opened 2026-06-11
Fix compatibility with current GNU Make (and Ubuntu 26.04)

adamnovak:
## Changelog Entry

 * vg now builds with GNU Make 4.4.1, as on Ubuntu 26.04
 * vg now uses system libelf/libdw
 * vg now ships its own Cairo and pixman libraries as submodues, to enable static builds on modern Ubuntu. These build with Meson; if needed, a new enough Meson will be installed on older Ubuntu via pipx by `make get-deps`.
 * Reverted #4926 due to crashes in the compiler in the emulated builds
 * vg Docker containers are now built on top of Ubuntu 26.04

## Description

GNU Make [removed the hack](https://stackoverflow.com/a/76870264) we were using to run a rule at Makefile startup (claiming that it would generate an include-able Makefile fragment, marking it `.PHONY` so it would always run, and telling Make not to complain if the file to be included didn't exist). It's not mentioned in the project's `NEWS` file, but it happened between 4.3 and 4.4.1.

This converts half of that target to `$(shell )` calls (the half that needs a bunch of Make-level variables), and the other half to a shell script (the half where we need to enforce success).


adamnovak:
This still doesn't fix 26.04, because the compiler there doesn't like something in libelf:

```
riscv_disasm.c: In function 'riscv_disasm':
riscv_disasm.c:1259:46: error: initialization discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
 1259 |                   struct known_csrs *found = bsearch (&key, known,
      |                                              ^~~~~~~
```

adamnovak:
This will need https://github.com/vgteam/structures/pull/15 and https://github.com/vgteam/libbdsg/pull/242

adamnovak:
I don't think we'll be able to get the Docker build logs from CI without fixing or working around https://github.com/greg7mdp/sparsepp/issues/98.

When I run the Docker build locally on my Mac I get a complaint at the final link step that there's something wring with `libvcfh.a`:
```
3.971 /usr/bin/aarch64-linux-gnu-ld.bfd: /vg/lib/libvcfh.a: error adding symbols: archive has no index; run ranlib to add one
3.990 collect2: error: ld returned 1 exit status
3.990 make: *** [Makefile:551: lib/vg_is_static] Error 1
```

This happens even if I stick a `ranlib` call into the command used to build it.

adamnovak:
If we want the Docker to ship a static binary, which we pull out and use for our static ARM release and for putting in a new, smaller Docker container, then we can't upgrade past 22.04 until https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1104888 is fixed and shipped in an Ubuntu release, since in 24.04 and on Ubuntu no longer packages a `libcairo.a` anywhere.

Or else we need to pull in Cairo as a submodule or otherwise provide a source build of Cairo for the Docker container.

adamnovak:
OK I have no idea why the docs build job can't find https://github.com/vgteam/libbdsg/commit/54299954575a980c8506624e52e1593333a2a6d1 which is merged in over there. And it looks like the updated libelf is not happy back on 22.04?

I will have to come back to this.

adamnovak:
I have finally fixed the docs job, but my approach makes the docs job almost twice as long as the fix in #4870. So we probably want to use that one when they conflict.

adamnovak:
There's still a problem where the Meson in the CI base image is too old to build Cairo/pixman. I think I need to also switch the CI prebake image over to a newer Ubuntu.

Issue #2199, opened 2019-04-07
vg sift -1: Segmentation fault if multithreaded


evanbiederstedt:
I was fiddling with `vg sift`, using the option `-1` whereby

```
-1 / --calc-insert        Calculate and print the insert size mean / sd every 1000 reads.
```

It looks like this works by default or if `-t 1`, but if users try more than one thread, there's a `Segmentation Fault`. That's somewhat reasonable, as the options state:

```
 -t / --threads  <MTHRDS>    number of OMP threads (not all algorithms are parallelized).
```

Perhaps this is something to polish at some point (I could consider a PR if there's time). 

Here's an example, whereby `file1.gam` is approx. 45GB:

```
$ vg sift -t 2 -1 /path/file1.gam

Filtering /path/file1.gam

ERROR: Signal 11 occurred. VG has crashed. Run 'vg bugs --new' to report a bug.
Stack trace path: /tmp/vg_crash_SN6StQ/stacktrace.txt
Segmentation fault
```
and then

```
$ cat /tmp/vg_crash_SN6StQ/stacktrace.txt
Crash report for vg v1.14.0-206-gc9ba0364 "Quadrelle"
Stack trace (most recent call last):
#9    Object "", at 0xffffffffffffffff, in 
#8    Object "", at 0x7efe808bcd0e, in 
#7    Object "", at 0x7efe8585e4a3, in 
#6    Object "", at 0x7efe80da06b5, in 
#5    Object "", at 0x56303e9b6260, in 
#4    Object "", at 0x7efe80da2f1b, in 
#3    Object "", at 0x7efe80d9a4ae, in 
#2    Object "", at 0x56303e9b5cae, in 
#1    Object "", at 0x56303e9b5a07, in 
#0    Object "", at 0x56303e911c4d, in 
```

Similar behavior if `-t N` s.t. `N>1`, but never if `N==1`.

edawson:
Thanks for catching this! I'll see what I can do about cleaning it up.

faithokamoto:
vg sift was removed

PR #4930, opened 2026-06-12
Remove `vg sift` (been deprecated since 2020)

faithokamoto:
## Changelog Entry

 * `vg sift` is removed

## Description

`vg sift` was deprecated in 2020 by #3136 and has since only really been "updated" by me, when I tweaked all the subcommand files for better option parsing etc. It is at its fundamental level just a GAM filterer. We already have `vg filter` for that, and indeed you could use `vg filter --tsv-out` plus some extra parsing to get e.g. reads with a bunch of softclips.

So this ends up being a bunch of code that we don't maintain nor use, for over five years. I propose deleting it. `vg sift` depended largely on a `Filter` object from `src/filter.cpp`/`.hpp`, and a lot of that got tossed too. The remainder is a small collection of functions that are less filtery and more standard Alignment handling; might make more sense to hand them over to the `Alignment` class?


faithokamoto:
Oh also `vg sift` is at least somewhat broken, e.g. see #2199

faithokamoto:
I realized the remaining parts of `src/filter.*` were only used in a `genotype_svs()` function which seems to have never been implemented, just stubbed in via comments. So I took that out too. If we want to have that functionality, we should put it in properly.

adamnovak (approved):
It looks like this was all part of an attempt at a structural variant genotyper by @edawson, which I'm not sure was finished. We can't usefully support it and probably don't need it anymore.

Issue #2688, opened 2020-03-25
Make vg combine handle subpaths and link chunked graphs (for chunked augmentation)


jmonlong:
I would like to be able to chunk a graph, modify the chunks (e.g augmenting with variants in the "middle" of the chunk) and combine them back together.

I think `vg combine` works with disjoined chunks (different chromosomes) but when I try it on chunks that share a path, e.g. a chromosome split in multiple chunks, it complains about "duplicated rank 1 in path ..." and wouldn't link the chunks together anyway (AFAIK).

Assuming that I was careful and the tails of my chunks are unchanged by whatever I'm doing to the chunks, could it be possible to combine chunks and have `vg combine`:
- stitch the "duplicated" subpaths if they spanned multiple chunks, back into the reference path for example.
- have the chunks linked, i.e. adding an edge from the end or the previous chunk to the beginning of the next one, or something of the like.

What do you think? Am I missing something?

I would have to combine each chromosome separately and make sure the new nodes created when working on a chunk don't conflict with other chunks. I think `vg ids -j` would handle that.

glennhickey:
vg concat may already work on simple cases for this.  I'm going through to
make it a bit more robust by adding a --reference-path option to give a
path to join on.

glennhickey:
I think there's some work to be done to make `vg chunk` play more nicely with this.  As you mentioned on the call, if it chunks in the middle of a node, you'll get an overlap between chunks which would then be dupliated afted `concat`. Perhaps if `chunk` adopted a convention where split nodes always end up in the left chunk that would be sufficient?  Or just cutting the node itself I guess.

Issue #2515, opened 2019-10-22
How to join multiple vg files


RenzoTale88:
Hello,
I've got a very simple question. I've got multiple vg files generated from single chromosome cactus alignments. I would like to combine them in a single vg archive so to process it separately.

Is it ok if I proceed as follow:

```
for n in {1..N}; do
    vg ids -s GRAPH/chr${n}.tmp.vg > GRAPH/chr${n}.final.vg
done

vg ids -j  $(for i in $(seq 1 n); do echo chr${i}.final.vg; done)

vg join $(for i in $(seq 1 n); do echo chr${i}.final.vg; done) > all.vg
```
Is it correct?

Thank you again for your help,
Andrea

ekg:
We really need to remove vg join because it does not do what users expect!
It links the graphs to a single head node, and was used during testing I
did five years ago. This is getting urgent because we are having one to two
support requests a week about this.

It used to be that we used vg ids to make the id space of each graph not
overlap the others and then just use `cat` to write them into a single
graph. I think this still works.

vg ids -j a.vg b.vg c.vg
cat a.vg b.vg c.vg >x.vg

RenzoTale88:
Oh wow, that's even easier! 
Thank you for the quick reply and support,

Andrea

glennhickey:
Agreed on removing vg join.  And a vote for removing vg concat for the same
reasons.

faithokamoto:
this seems to have been handled

Issue #2502, opened 2019-10-10
vg augment crashed


verne91:
I am trying to augment a graph with the `.gam` file using the following command line.
```
vg augment -p -t 12 merged.vg x.gam -C -A x.aug.gam > x.aug.vg
```
The standard output is:
```
Reading input graph
vg: src/augment.cpp:327: std::unordered_map<long int, std::set<std::tuple<long int, bool, long unsigned int> > > vg::forwardize_breakpoints(const HandleGraph*, const std::unordered_map<long int, std::set<std::tuple<long int, bool, long unsigned int> > >&): Assertion `graph->has_node(node_id)' failed.
ERROR: Signal 6 occurred. VG has crashed. Run 'vg bugs --new' to report a bug.
Stack trace path: /tmp/vg_crash_PnjvsC/stacktrace.txt
```

The `stacktrace.txt` is:
```
Crash report for vg v1.19.0 "Tramutola"
Stack trace (most recent call last):
#13   Object "/home/cnsun/software/bin/vg", at 0x4d6a79, in _start
#12   Object "/home/cnsun/software/bin/vg", at 0x1ac8318, in __libc_start_main
#11   Object "/home/cnsun/software/bin/vg", at 0x40a7d2, in main
#10   Object "/home/cnsun/software/bin/vg", at 0x9a94e7, in vg::subcommand::Subcommand::operator()(int,
char**) const
#9    Object "/home/cnsun/software/bin/vg", at 0x96dc2b, in main_augment(int, char**)
#8    Object "/home/cnsun/software/bin/vg", at 0xa4538c, in vg::get_input_file(s
td::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::function<void (std::istream&)>)
#7    Object "/home/cnsun/software/bin/vg", at 0x96a2f1, in std::_Function_handler<void (std::istream&), main_augment(int, char**)::{lambda(std::istream&)#7}>::_M_invoke(std::_Any_data const&, std::istream&)
#6    Object "/home/cnsun/software/bin/vg", at 0xa7d887, in vg::augment(handlegraph::MutablePathMutableHandleGraph*, std::istream&, std::vector<vg::Translation, std::allocator<vg::Translation> >*, std::ostream*, bool, bool, bool)
[...]

PR #4932, opened 2026-06-12
Remove vg join (been deprecated since 2020)

faithokamoto:
## Changelog Entry

 * `vg join` is removed

## Description

Another of the subcommands we deprecated back in 2020 with #3136. `vg join` is problematic: https://github.com/vgteam/vg/issues/2515#issuecomment-544903318 (comment from 2019):

> We really need to remove vg join because it does not do what users expect!
It links the graphs to a single head node, and was used during testing I
did five years ago. This is getting urgent because we are having one to two
support requests a week about this.

That issue thread also suggests removing `vg concat`, which _also_ does unexpected things https://github.com/vgteam/vg/issues/2502#issuecomment-540712178 -- except `vg concat` does some useful things too, e.g. see it being recommended here #2688. So I am not sure whether it should be murdered with fire. Thus this PR only removes `vg join`.


faithokamoto:
Also, you can do exactly what vg join is for by just manipulating GFAs. I did that for Centrolign and it worked fine.

adamnovak:
Since we're working on substantial improvements to `vg combine` (which will resolve the missing features mentioned in #2688), I think we want to remove `vg join` now and then probably remove `vg concat` when the `vg combine` changes land.

adamnovak (approved):

Issue #281, opened 2016-03-30
vg circularize


edawson:
I think it'd be nice to have a single command to circularize a graph by creating an edge between its head and tail nodes. Right now I do this manually, taking the graph through GFA, but it turns out I'm doing it more often than I initially expected.

I'll work on this tonight; it's probably all of five lines of code.

ekg:
Need to take from the command line the head and tail nodes to join. The
problem is that we can't always do this automatically; we don't necessarily
want to link all the head and tails nodes.

Another related problem is indicating that a specific alignment is
circular. This would be a flag on the path.

edawson:
So I have a thought or two on this. As @ekg handily pointed out, we may have some paths in a graph which are circular and some that are not. One example I can imagine of this is if we place an HPV decoy in a human graph. It is therefore important to circularize only specific paths.

I could solve this by taking in a file that looks like this:
`ChromA    circular
ChromB     circular
ChromC     linear`

but it's a bit ugly and I want some opinions on it. In the mean time, I'll try generating a circular pan-HPV graph using MSGA to see if it's feasible/useful.

ekg:
Maybe it would be simpler to have a vg mod command that takes a list of paths in a file and ensures that they are circular in the graph. So this would mean setting a flag in the path object to say if it is circular or not and also adding some edges to allow the circular path to be fully embedded in the graph.

For msga we could take an argument which lists the circular paths by name. This would be ideal as it makes the assembly circular genome aware. Otherwise we will get divergent alignments at the starts and ends of the sequences.

So I guess vg::VG should have a function which circularizes an embedded path and ensures that we can traverse all parts of it in the graph by linking the head and tail positions with an edge if such does not exist already.

ekg:
And also vg.proto should be patched to have a is_circular boolean flag on the Path object.

ekg:
I've got some more extensions to this.

1. Visualization--- I want to look at circular paths.
2. Path::is_circular (vg.proto)--- we need to be able to serialize graphs and remember that the paths are circular.

I'm on it!

ekg:
![screenshot from 2016-04-18 00-22-04](https://cloud.githubusercontent.com/assets/145425/14590475/a26cf91e-04fb-11e6-8219-5972e9c5b007.png)

ekg:
Next extension: MSGA should handle circular paths.

PR #4934, opened 2026-06-15
vg circularize tweaks

faithokamoto:
## Changelog Entry

 * Smoother interface & errors for `vg circularize`

## Description

I saw something that I could improve, and who's going to tell me no?

A collection of minor tweaks to `vg circularize`. It was implemented over a decade ago by #281 as a minor utility and seems to have been mostly left alone.
- If the `--head` ID was larger than the `--tail` ID, then nothing happened but that wasn't obvious. Now we print an explicit error.
- If `--head`/`--tail` were used, circularizing paths by name wasn't allowed. I don't really see why that restriction should exist so I removed it.
- Now `--path` can be used multiple times, just in case that's useful.
- Added a few unit tests.
- No need to have `--describe` here for listing path names when we already have a command purpose-built to list path names, i.e. `vg paths --list`.

### DIFF
--- a/src/subcommand/circularize_main.cpp
+++ b/src/subcommand/circularize_main.cpp
@@ -31,7 +31,6 @@ void help_circularize(char** argv) {
          << "  -P, --pathfile FILE     circularize all paths in the provided file" << endl
          << "  -a, --head ID           circularize a head and tail node (must provide a tail)" << endl
          << "  -z, --tail ID           circularize a head and tail node (must provide a head)" << endl
-         << "  -d, --describe          list all the paths in the graph"   << endl
          << "  -h, --help              print this help message to stderr and exit" << endl;
     exit(1);
 }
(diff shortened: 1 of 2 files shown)


src/subcommand/circularize_main.cpp
```diff
@@ -98,27 +98,24 @@ int main_circularize(int argc, char** argv) {
         }
     }
 
-    vector<string> paths_to_circularize;
-    if (!((head * tail) > 0)) {
-        help_circularize(argv);
+    if ((head == DEFAULT_ID) != (tail == DEFAULT_ID)) {
         logger.error() << "Both a head and tail node must be provided" << endl;
+    } else if (tail < head) {
+        logger.error() << "Tail " << tail << " is smaller than head " << head << endl;
```
adamnovak, line 104: This restriction doesn't make all that much sense to have at all. But if we have it we do need the error.

adamnovak (approved):

Issue #1984, opened 2018-11-12
error installing in Ubuntu 16.04


nllg:
Please describe:

1. What you were trying to do
2. What you wanted to happen
3. What actually happened
4. What data and command line to use to make the problem recur, if applicable

```
Format code blocks or terminal copy-pastes like this, between triple backticks.
```

nllg:
Hello,

I am trying to install vg in my Ubuntu 16.04
I followed the instructions:
git clone --recursive https://github.com/vgteam/vg.git

I get the following error:
 fatal: unable to connect to sourceware.org
fatal: clone of 'git://sourceware.org/git/elfutils.git' into submodule path 'deps/elfutils' failed

Do you have any suggestions?

Thank you,

Nathaly

adamnovak:
Hello,

I've just tried the command you tried, and it currently works for me. I suspect that sourceware.org, which hosts one of our dependencies, may have been down when you ran the command.

Can you try the same command again?

steinbrl:
Hi,

I got the same error:
clone of deps/elfutils failed

The git page is not down. It is possible to navigate manually to this page. I don't see the problem.

Do you have any suggestions?

Greetings,

Lars

adamnovak:
Hello @steinbrl 

I just did a `git clone git://sourceware.org/git/elfutils.git` and it seemed to work. Maybe Sourceware was down, or the git part of it was dfown, but it came back up? What happens if you `rm -Rf deps/elfutils` and `git submodule update --init --recursive`?

If that still doesn't work, and you can still get to Sourceware's web pages, what happens if you go to some other folder and try to `git clone --verbose git://sourceware.org/git/elfutils.git`? Does it give any more information on why it can't clone the repo beyond just that it failed? (Like a complaint that you are out of disk space or something?)

ekg:
Are our releases including tarballs?

adamnovak:
Yeah, that's another option. The latest one is
https://github.com/vgteam/vg/releases/download/v1.20.0/vg-v1.20.0.tar.gz
and it comes with all the submodules pre-downloaded.

steinbrl:
Hi,
I tried the recommended tar-ball. The building process failed with this Message:

make[1]: Entering directory '/home/lars/Desktop/NGS/vg_1/vg-v1.20.0/deps/snappy'
[...]

PR #4936, opened 2026-06-16
Remove snappy dependency

faithokamoto:
## Changelog Entry

 * Remove submodules `snappy`, `libVCFH`, and `fermi-lite`

## Description

Saw in #1984 that removing the snappy dependency was suggested. RocksDB, also suggested for removal, was axed by #3137. I looked around for snappy and genuinely could not find anywhere we were using it. Deleting that submodule seemed to work fine for me. (I built from scratch etc.) I guess having the CI take a look will be the real test though.


faithokamoto:
Okay, so removing only snappy seems to be fine. I pushed a thing that also removes `libVCFH` and `fermi-lite`, but I'm less confident that those will be OK. If they fail I'll revert back to the commit that passed CI.

adamnovak (approved):
This seems like a good idea to me!

Issue #1904, opened 2018-10-02
Integration test 02_vg_construct fails


emikoifish:
Wanted to pass all integration tests. 

```prove -v t/02_vg_construct.t ``` fails test 17.

```
t/02_vg_construct.t: line 82: ../deps/vcflib/bin/vcf2tsv: No such file or directory
(standard_in) 2: parse error
not ok 17 - unnamed test
#          got: '1074'
#     expected: 'the graph contains all the sequence in the reference and VCF'
```
Running on a Mac.

faithokamoto:
Dependencies no longer needed, see above PR.

PR #4942, opened 2026-06-25
Remove vcf2tsv and fastahack binary need in tests

faithokamoto:
## Changelog Entry

 * Unit tests no longer use `vcf2tsv` and `fastahack`

## Description

Was working on something unrelated and noticed these. They can be done with simple non-external commands so I switched over to that. Why not? Makes the tests less finicky.


faithokamoto:
no idea why the mac failed, nor do I know how to test on mac without a mac. Running the mac tests again with an extra print

faithokamoto:
This should also fully resolve #1904 since no longer do we fail the test if those dependencies aren't there.

PR #4943, opened 2026-06-26
Remove `vg msga`

faithokamoto:
## Changelog Entry

 * `vg msga` is removed

## Description

Messed with a lot of test stuff, but it doesn't rely in `vg msga` any more. You know what _does_ rely on it, though? The wiki.

- https://github.com/vgteam/vg/wiki/Complex-graph-visualization
- https://github.com/vgteam/vg/wiki/RDF%3A-for-VG
- https://github.com/vgteam/vg/wiki/VG-RDF%2C-the-Ensembl-bacteria-E.-coli-genome-hack-attack

I'm not entirely sure what to do about those buggers. I guess for the first two I could just run the command and save the output? But the last one is that sort of broken page that I would now be breaking even further.

@adamnovak since you asked for it :)


faithokamoto:
It failed some doc job `/builds/vgteam/vg/doc/../README.md` that I have no idea where it is. Looks like we'd just need to delete some line that tests msga.

faithokamoto:
Oh, silly. That means the repo base level README.

adamnovak:
For the last one (https://github.com/vgteam/vg/wiki/VG-RDF%2C-the-Ensembl-bacteria-E.-coli-genome-hack-attack) we could try `vg map` or `vg giraffe`, if they work on such long inputs, to get a GAM, and then `vg augment` to add the GAM to the graph.

faithokamoto:
I think the last one is begging for me to actually try to rerun the entire example. I'll put it back on the wiki update list...

faithokamoto:
Reworked the tutorial to not use `vg msga`: https://github.com/vgteam/vg/wiki/Graph-&-annotations-for-two-E.-coli-strains

Will plug it into everywhere relevant tomorrow, I'm about to go watch Project Hail Mary again :)

faithokamoto:
@adamnovak anything else we might need to worry about before this gets merged in?

adamnovak:
I think that's probably it.

PR #4945, opened 2026-06-26
Remove `vg explode`

### LINKED ISSUE #2855: Remove vg explode
When you run `vg explode` it prints help that implies it still works, but when you try to use it it complains about being deprecated and fails. We should just take it out now that it's been deprecated for a while.

faithokamoto:
## Changelog Entry

 * `vg explode` is removed

## Description

Resolves #2855, from over six years ago, which is frankly a ridiculous amount of time to get rid of a subcommand that, when run, does nothing but error and tell you to do something else.

And! About "tell you to do something else". I have added a simple map lookup for removed subcommands. If we think that there should be a warning message printed when you try to use something we've axed, now we can have that message printed even in the absence of the subcommand.


src/subcommand/subcommand.hpp
```diff
@@ -66,6 +66,12 @@ enum CommandCategory {
     DEPRECATED
 };
 
+const static std::map<std::string, std::string> REMOVED_CMD_MESSAGES{
+    {"explode", std::string("Please use \"vg chunk -C source.vg -b part_dir/component\" ")
+                 + std::string("for the same* functionality as \"vg explode source.vg part_dir\"\n")
+                 + std::string("* (unlike explode, the output directory must already exist when running chunk, though)")}
```
adamnovak: Can't we avoid these string concatenations?

```suggestion
    {"explode", std::string("Please use \"vg chunk -C source.vg -b part_dir/component\" "
                            "for the same* functionality as \"vg explode source.vg part_dir\"\n"
                            "* (unlike explode, the output directory must already exist when running chunk)")
    }
```

PR #4946, opened 2026-06-26
Correct autocomp

faithokamoto:
## Changelog Entry

 * Update bash autocomplete files, and force them to remain up-to-date

## Description

I admit, I have no idea what these darlings do! I certainly haven't noticed vg doing autocompletes for me. Both autocomplete files were last touched over a decade ago. The list of subcommands has changed drastically since.

I figure: either these things are useless, in which case maybe this PR reminds Adam that they exist, or they're useful, and in that case we should keep them up-to-date. Proceeding assuming the latter I have brought them up to modern vg and have added another test into my option checking script.

(I swear I have tried to figure out what these things are for, but if there are details somewhere they're buried. Doesn't help that `autocomplete.sh` seems to be some sort of AI tool now.)

### DIFF
--- a/autocomp.bash
+++ b/autocomp.bash
@@ -3,7 +3,7 @@ _vg_complete()
 	 local cur prev
 	 COMPREPLY=()
 	 cur="${COMP_WORDS[COMP_CWORD]}"
-	 opts="construct deconstruct view vectorize index find paths align map stats join ids concat kmers sim mod surject msga pileup call genotype compare scrub circularize validate version"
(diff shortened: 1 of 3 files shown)


adamnovak:
I think you're meant to `source autocomp.bash` and than get help autocompleting vg commands in that shell. And then you could put that in your bashrc.

It *could* get packaged to drop in to some kind of system directory for autocomplete information, but [for Debian/Ubuntu it does not](https://packages.ubuntu.com/resolute/amd64/vg/filelist).

@mr-c You do the Debian packaging, right? How would we ship this so people could use it?

mr-c:
> I think you're meant to `source autocomp.bash` and than get help autocompleting vg commands in that shell. And then you could put that in your bashrc.
> 
> It _could_ get packaged to drop in to some kind of system directory for autocomplete information, but [for Debian/Ubuntu it does not](https://packages.ubuntu.com/resolute/amd64/vg/filelist).
> 
> @mr-c You do the Debian packaging, right? How would we ship this so people could use it?

Looks like bash completion files get installed to `/usr/share/bash-completion/completions` by `dh_bash-completion` by listing the path to the completion file (`autocomp.bash`) in `debian/package.bash-completion` and adding `dh-sequence-bash-completion` to the `Build-Depends` list in `debian/control`.

PR #4950, opened 2026-06-30
Dup names

### LINKED ISSUE #547: vg construct from fasta path name selection could be better.
When building a VG from an Ensembl Bacteria the selected path name is not ideal.

The FASTA header is
```
>Chromosome dna:chromosome chromosome:ASM1942v1:Chromosome:1:4686137:1 REF
```
This gives after VG construct

```
vg construct --reference ecoli/Escherichia_coli_str_k_12_substr_dh10b.ASM1942v1.dna.chromosome.Chromosome.fa -m 1000 > ecoli/Escherichia_coli_str_k_12_substr_dh10b.ASM1942v1.vg
```

The following path name
```
 vg view ecoli/Escherichia_coli_str_k_12_substr_dh10b.ASM1942v1.vg | grep "^P" | head -n 1

P       1       Chromosome      1       +       1000M
```

This is not ideal, as then all paths if building a VG of Ensembl bacteria have the same name.

Desired resolution, entire fasta headers should be preserved by vg construct

faithokamoto:
## Changelog Entry

 * Descriptive error when creating a graph from FASTAs with duplicate names

## Description

Resolves #547 in a way. Now instead of non-obvious problems when a sequence name is duplicated, we give a big beautiful obvious error. Old behavior was that the last of the duplicately named sequences was used. New behavior is that we tell you that there are duplicate names. Test cases also added.


faithokamoto:
great it can't find my commit (to be fair I am also confused about where the commit lives)

faithokamoto:
This is failing now because it needs some fixes I put in #4945

adamnovak:
OK, here's the problem:
```
[anovak1@mustard test]$ rm -f xx.fa xx.fa.fai
sed "s/y/x/" small/xy.fa > xx.fa
vg construct -r xx.fa > /dev/null
index file xx.fa.fai not found, generating...
[anovak1@mustard test]$ echo $?
0
[anovak1@mustard test]$ vg construct -r xx.fa > /dev/null
Sequence "x" appears multiple times
[anovak1@mustard test]$ echo $?
1
```
The check for duplicate entries in the FASTA is in the index load code. If you run the test more than once locally without deleting the index, the FASTA is already indexed and the index load code detects the duplication. If you run the test in a clean repo (like on the Mac CI), the index gets built and we don't actually go through the index load code, so the duplicate check never happens.

adamnovak:
This now uses https://github.com/vgteam/vcflib/pull/29

adamnovak:
If someone indexes a FASTA with duplicate records with samtools, which warns and drops duplicates from the index, then vg won't be able to detect that there are duplicates in the file when we open it, because we'll just query the index.

That might be good enough.

PR #4953, opened 2026-07-01
Allow long-read Giraffe to take self-loops/reversals between seeds

faithokamoto:
## Changelog Entry

 * Allow long-read Giraffe (`-b hifi` / `-b r10`) to take self-loops/reversals between seeds

## Description

PR'ing this since I don't see any more how to get it to work better. It decreases the number of incorrectly mapped reads, at least.

Notable changes:
- Snarl or node levels in zipcodes now store whether it is possible to self-loop back as two bools bit-packed into metadata fields.
    - For a forward reversal, that means leave in the forward direction, then go back and enter the same bit in the reverse direction. Reverse is opposite.
    - Old zipcodes are compatible with the new code. The new algorithm will assume that they have no reversals
- Ziptrees store reversals using a `LOOP` item.
    - Effort is made to de-duplicate `LOOP`s as much as possible.
    - During traversal, whenever we hit into a `LOOP` we save a position using the new `S_TAKE_LOOP` state. When we reload this position we immediately take the loop by turning around and then keep going.
    - Simple best-so-far memorization is used for LOOP positions to avoid infinite bouncing back between each other.
- Graph distances between seeds are saved from chaining and are propagated all the way through to setting up base-level alignment.
- We can extract a connecting graph even if it requires traversing one of the end nodes in both directions (i.e. taking a loop). Before we forced all end nodes to become tips. That prevented taking loops through an end now. Now if necessary a copy of an end node is made and then only one copy is made into a tip.


src/algorithms/chain_items.hpp
```diff
@@ -415,12 +420,28 @@ struct ChainScoringScheme {
...
+    // The index of the item
+    size_t index;
+    // The graph distance from the previous thing
+    // std::numeric_limits<size_t>::max() if the first thing
+    // or otherwise unknown (e.g. gapless extension anchors)
+    size_t graph_distance;
+
+    TracedItem (size_t idx, size_t graph_dist) : index(idx), graph_distance(graph_dist) {}
+    TracedItem (size_t idx) : index(idx), graph_distance(std::numeric_limits<size_t>::max()) {}
+};
```
adamnovak: What does "traced" mean here?

For `TracedScore`, I meant "traced" to mean that, in addition to the score value, we knew where the score came from and could "trace" it back.

That's not really what's going on here.

src/algorithms/chain_items.hpp
```diff
@@ -291,8 +291,11 @@ class Anchor {
...
 ostream& operator<<(ostream& out, const Anchor& anchor);
 
-// For doing scores with backtracing, we use this type, which is a
-// score and a number for the place we came from to get it.
+// For doing scores with backtracing, we use this type, which is
+// - a score
+// - a number for the place we came from to get it
+// - the graph distance from that point (ignored for comparison purposes
+//   as it's just an annotation for later lookup, not part of the tracing)
 class TracedScore {
```
adamnovak, line 299: I don't know that we want to just list the fields it has, and this is still missing the recombination info.

I think this isn't *really* a traced score anymore; it's a chaining DP table cell entry and it holds a bunch of information that we need to bring through the DP in addition to the score and the traceback pointer.

src/algorithms/chain_items.hpp
```diff
@@ -436,14 +457,6 @@ struct ChainWithRec {
     std::vector<std::pair<size_t, size_t>> rec_intervals;
 };
```
adamnovak, line 453: Looking at this as a whole, it's weird that we would store the graph distances in with the anchors. It makes sense that we might tack them onto here to avoid recomputing them later or having to fill in a big map of them just to look them up again. But they should probably be their own field on here.

It's also weird that it includes a bunch of data about the chain, but only `Rec` makes it into the name.

src/algorithms/chain_items.hpp
```diff
@@ -562,11 +575,11 @@ TracedScore chain_items_dp(vector<TracedScore>& chain_scores,
  *  won't be possible to recombine two tracebacks to get a higher score; no
  *  edges followed between items will ever need to be cut.
  */
-vector<pair<vector<size_t>, int>> chain_items_traceback(const vector<TracedScore>& chain_scores,
-                                                        const VectorView<Anchor>& to_chain,
-                                                        const TracedScore& best_past_ending_score_ever,
-                                                        const ChainScoringScheme& scheme = ChainScoringScheme(),
-                                                        size_t max_tracebacks = 1);
+vector<pair<vector<TracedItem>, int>> chain_items_traceback(const vector<TracedScore>& chain_scores,
```
adamnovak: Probably we should change this from `vector, penalty` pairs to just whole `ChainWithRec`s, and move the penalty-to-score conversion into this function.

Or, if we don't want unset fields we could invent a new `Chain` type that doesn't have the recombination fields that `find_best_chains()` populates. Then we make `ChainWithRec` extend it to add those fields, and `find_best_chains()` becomes copying all the chains from the base to the derived class and filling in the recombination-related parameters.

src/algorithms/extract_connecting_graph.hpp
```diff
@@ -47,12 +47,15 @@ namespace algorithms {
     ///
     /// Returns: a map from node ids in the extracted graph to the node ids in
     /// the original graph. The map and the graph will have the same number of
-    /// entries.
+    /// entries, plus two if track_new_ids is used:
+    ///   * -std::numeric_limits<id_t>::max() -> pos_1's new edge node ID
+    ///   * std::numeric_limits<id_t>::max() -> pos_2's new edge node ID
```
adamnovak: These aren't just weird keys but also are new-graph values in a map that's meant to map to original-graph values.

Instead of having an optional boolean for whether to include these, I would have an optional pointer or pointers, defaulting to null, of where and where to write these. Maybe `nid_t* tip1 = nullptr, nid_t* tip2 = nullptr`.

Then instead of smuggling them through and fishing them out of the map, they can have their own channel.

src/algorithms/extract_connecting_graph.hpp
adamnovak: Aren't we changing the semantics here? Or should we explain the subtlety that was overlooked before?

src/algorithms/extract_extending_graph.hpp
```diff
@@ -33,8 +33,12 @@ namespace algorithms {
     ///  pos                     extend from this position
     ///  backward                extend in this direction
     ///  preserve_cycles_on_src  if necessary, duplicate starting node to preserve cycles after cutting it
+    /// Returns: a map from node ids in the extracted graph to the node ids in
+    /// the original graph. The map and the graph will have the same number of
+    /// entries, plus one if track_new_id is used:
+    ///   std::numeric_limits<id_t>::max() -> pos's new edge node ID
     unordered_map<id_t, id_t> extract_extending_graph(const HandleGraph* source, DeletableHandleGraph* into, int64_t max_dist, pos_t pos,
-                                                      bool backward, bool preserve_cycles_on_src_node);
+                                                      bool backward, bool preserve_cycles_on_src_node, bool track_new_id = false);
```
adamnovak: Similarly I think here we should use a pointer to a place to optionally write the ID instead of hiding it in the map.

src/minimizer_mapper_from_chains.cpp
```diff
@@ -251,42 +251,42 @@ void MinimizerMapper::dump_debug_chains(const ZipCodeForest& zip_code_forest,
...
                             seedpos.field("");
                             seedpos.field("");
-                            seedpos.field(seed_num);
+                            seedpos.field(seed_num.index);
                             std::stringstream ss;
-                            ss << seed_anchors.at(seed_num);
+                            ss << seed_anchors.at(seed_num.index);
                             seedpos.field(ss.str());
                         }
                     }
```
adamnovak, line 292: This whole section is currently using vectors of seeds that *don't* represent chains, and so it shouldn't be using the new chain types.

src/minimizer_mapper_from_chains.cpp
```diff
@@ -299,10 +299,10 @@ void MinimizerMapper::dump_debug_chains(const ZipCodeForest& zip_code_forest,
             const std::string& marker = kv.first;
             for (size_t run_number = 0; run_number < kv.second.size(); run_number++) {
                 // For each run of seeds in it
-                const std::vector<size_t>& included_seeds = kv.second[run_number];
+                const std::vector<algorithms::TracedItem>& included_seeds = kv.second[run_number];
                 for (size_t idx = 0; idx < included_seeds.size(); ++idx) {
                     // For each seed in the run (index-based so we can consult chain flags)
-                    size_t seed_num = included_seeds[idx];
+                    size_t seed_num = included_seeds[idx].index;
```
adamnovak: These are also working with collections of seeds that aren't chains.

src/minimizer_mapper_from_chains.cpp
```diff
@@ -1196,7 +1196,7 @@ vector<Alignment> MinimizerMapper::map_from_chains(Alignment& aln) {
 void MinimizerMapper::do_chaining_on_trees(const Alignment& aln, const ZipCodeForest& zip_code_forest,
     const std::vector<Seed>& seeds, const VectorView<MinimizerMapper::Minimizer>& minimizers,
     const vector<algorithms::Anchor>& seed_anchors,
-    std::vector<std::vector<size_t>>& chains, std::vector<std::vector<bool>>& chain_rec_flags,
+    std::vector<std::vector<algorithms::TracedItem>>& chains, std::vector<std::vector<bool>>& chain_rec_flags,
```
adamnovak: Here we're using `TracedItem` pretty far from `algorithms/chain_items.hpp`, and with slightly different semantics. Here the index refers into seed space, not into anchor space for the chaining problem like when used in the chaining module.

This makes it hard to have a good comment on `TracedItem`'s `index` field that says how to interpret it; it would really want to say that sometimes these are anchor indexes in a chaining problem but in other contexts the indexes are seed indexes.

I feel like this should be a different type, especially since I'm hoping we could get rid of `TracedItem` in the chaining module.

src/minimizer_mapper_from_chains.cpp
```diff
@@ -2846,11 +2867,15 @@ Alignment MinimizerMapper::find_chain_alignment(
...
-                            cerr << log_name() << "Not bothering to skip to " << *skip_to_it 
+                            cerr << log_name() << "Not bothering to skip to " << (*skip_to_it).index
                                  << " because total gaps are only " << gap_lengths << endl;
                         }
                     }
 #endif
+                    total_graph_distance = (*next_it).graph_distance;
+                    if (total_graph_distance == std::numeric_limits<size_t>::max()) {
+                        total_graph_distance = algorithms::get_graph_distance(*here, *next, *distance_index, gbwt_graph);
+                    }
```
adamnovak: This little pattern around `get_graph_distance()` is being added 3 times. Maybe `get_graph_distance()` could itself take an optional precomputed hint distance and include the logic to maybe use it?

src/minimizer_mapper_from_chains.cpp
```diff
@@ -3368,118 +3427,44 @@ void MinimizerMapper::with_dagified_local_graph(const pos_t& left_anchor, const
             ss << " with max path length of " << max_path_length;
             throw ChainAlignmentFailedError(ss.str());
         }
+        
+        local_left_anchor_id = local_to_base.at(-std::numeric_limits<id_t>::max());
+        local_right_anchor_id = local_to_base.at(std::numeric_limits<id_t>::max());
+        local_to_base.erase(-std::numeric_limits<id_t>::max());
+        local_to_base.erase(std::numeric_limits<id_t>::max());
```
adamnovak: If we used output parameters here we would be able to simplify this map smuggling and just fill in the variables with the call.

src/zip_code.hpp
```diff
@@ -197,9 +197,12 @@ class ZipCode {
...
 
-        // This will be 0 for irregular snarls, 1 for regular, and 2 for non-dag irregular snarls
-        // cyclic snarls will be identical to irregular snarls except for SNARL_IS_REGULAR
-        const static size_t SNARL_IS_REGULAR_OFFSET = 0; 
+        // Ones bit: 0 for irregular snarls, 1 for regular
+        // Twos bit: 0 for non-cyclic snarls, 1 for cyclic snarls
+        // cyclic snarls are identical to irregular snarls except for this bit
+        // Fours bit: 1 for a forward loop < LOOP_DISTANCE_STORAGE_THRESHOLD, 0 otherwise
+        // Eights bit: 1 for a reverse loop < LOOP_DISTANCE_STORAGE_THRESHOLD, 0 otherwise
+        const static size_t SNARL_METADATA_OFFSET = 0;
```
adamnovak: I'm not sure this is more metadata-flavored than e.g. the child count.

Maybe this should be `SNARL_FLAGS_OFFSET`?

src/zip_code.hpp
```diff
@@ -254,6 +263,12 @@ class ZipCode {
         //Return a vector of size_ts that will represent the snarl in the zip code
         snarl_code_t unpack_snarl_code(size_t zipcode_level) const;
 
+        /// Check for a particular bit being set
+        /// Used for the bitpacked SNARL/NODE_METADATA_OFFSET values
+        inline bool bit_is_set(size_t value, size_t bit_position) const {
+            return (value & (1 << bit_position)) != 0;
+        }
```
adamnovak: Instead of this and a bunch of comments like "we're accessing the twos bit which represents blah", we could have a bunch of constants `SNARL_FLAG_IS_REGULAR`, `SNARL_FLAG_FORWARD_LOOP`, etc. defined with the right bit set, and then use `if (zip_value & SNARL_FLAG_FORWARD_LOOP) { ... }`.

src/zip_code_tree.cpp
```diff
@@ -499,9 +581,135 @@ void ZipCodeForest::add_child_to_chain(forest_growing_state_t& forest_state, con
...
+        }
+
+#ifdef DEBUG_ZIP_CODE_TREE
+        cerr << "\tAdding a forward loop of value " << new_loop_value << endl;
+#endif
+        trees[forest_state.active_tree_index].zip_code_tree.back().set_forward_loop_length(new_loop_value);
+        trees[forest_state.active_tree_index].zip_code_tree.emplace_back(ZipCodeTree::LOOP, new_loop_value, false);
+        // Update memory
+        cur_chain.last_forward_loop_index = trees[forest_state.active_tree_index].get_tree_size() - 1;
+        cur_chain.dist_since_forward_loop = 0;
```
adamnovak: I think this is maybe referring to *a* "memory" (of something), but I might find another word for it. To me, if we're talking about "memory" we mean we're working with generic blocks of memory because we're doing memory management, or we have to clear a region of memory in bulk, or something. `Update memory` doesn't mean much in that context.

src/zip_code_tree.hpp
```diff
@@ -559,9 +583,10 @@ class ZipCodeTree {
         bool original_right_to_left;
         /// References to the zip code tree to let us look up distance matrices
         const vector<tree_item_t>* zip_code_tree;
-        /// Best distance for each snarl exit we've seen
-        /// Stored as {snarl start index : (start dist, end dist)}
-        std::unordered_map<size_t, std::pair<size_t, size_t>> best_cyclic_snarl_exits;
+        /// Best distance for each cyclic snarl exit or loop we've seen
+        /// For cyclic snarls: {snarl start index : (start dist, end dist)}
+        /// For loops: {loop index : (right_to_left dist, left_to_right dist)}
+        std::unordered_map<size_t, std::pair<size_t, size_t>> distance_memory;
```
adamnovak: Maybe this should be a "cache"? Is it a cache?

adamnovak (changes requested):
I looked mostly at the chaining changes and not as much at the actual loop logic, but I think there has to be a neater way to bring the graph distances out of chaining than hiding each one in a struct with the index of the seed.

I would add them to what is right now the `ChainWithRec` structure.

That makes them harder to carry through when we translate to chains of seeds and not chains of anchors. But maybe we could have a new structure for that?

Ultimately they really need to be available to `get_graph_distance()` to accelerate it, right? If they're too hard to bring everywhere with the chains, we could try caching them all in a table during chaining and giving the table to `get_graph_distance()`.

src/minimizer_mapper_from_chains.cpp
faithokamoto: swapped back commit 384b54733

src/minimizer_mapper_from_chains.cpp
faithokamoto: swapped back commit 384b54733

src/zip_code_tree.hpp
faithokamoto: renamed to `best_so_far_distance_cache` commit fb6a4ea85

src/zip_code_tree.cpp
faithokamoto: changed to refer to "hints" in commit 2e96d57c3

src/zip_code.hpp
faithokamoto: renamed to that in commit 858348a27

src/zip_code.hpp
faithokamoto: swapped to that format in commit f684ab98d

src/algorithms/extract_connecting_graph.hpp
faithokamoto: no more smuggling (using pointers) as of commit b24676665

src/minimizer_mapper_from_chains.cpp
faithokamoto: no more smuggling (using pointers) as of commit b24676665

src/algorithms/extract_extending_graph.hpp
faithokamoto: no more smuggling (using pointers) as of commit b24676665

src/algorithms/extract_connecting_graph.hpp
faithokamoto: documented in commit 3c01571a9

src/minimizer_mapper_from_chains.cpp
faithokamoto: done in commit 3ba5d6580

src/algorithms/chain_items.hpp
faithokamoto: handled in commit b0b8fe45a

src/algorithms/chain_items.hpp
faithokamoto: tacked on as mentioned and renamed to `SparseAnchorChain` in commit b0b8fe45a

src/algorithms/chain_items.hpp
faithokamoto: got rid of in commit b0b8fe45a

src/minimizer_mapper_from_chains.cpp
faithokamoto: got rid of in commit b0b8fe45a

src/algorithms/chain_items.hpp
faithokamoto: yeah I did that in commit 43d998421fba952da526f87202987aa7032a5102

PR #4955, opened 2026-07-03
Remove unused code

faithokamoto:
## Changelog Entry

 * Remove oodles of unused code without change to functionality

## Description

Hey, if its only use is that a unit test checks whether it works, then it seems we no longer use it. A small collection of things I found that I could delete without breaking the build.


adamnovak:
I liked `IndexedVG` at the time, but now probably anything we'd actually want to do with it around graph random access would be better handled by learning to read https://github.com/jltsiren/gbz-base.

This also rips out the code that was meant to be an MCMC-based genotyper. We now have Glenn's off-reference calling work and a lot invested in the surject/DeepVariant pipeline instead, and we haven't wanted to pursue that, but if we decide we want to revive it we can come back here and get it again.

Issue #4974, opened 2026-07-21
error[vg call] No reference paths found


majssssa:
1. I built a pangenome graph using PGGB and then generated the Giraffe indexes with:

```
vg autoindex --workflow giraffe -r Jiujiu_Ma_1.0.fa -v PGGB.vcf.gz -p PGGB.vcf.gz
```

I mapped paired-end reads and generated a pack file with:

```
vg giraffe -t 64 -Z ../PGGB.vcf.gz.giraffe.gbz -m ../PGGB.vcf.gz.shortread.withzip.min -d ../PGGB.vcf.gz.dist -f SRR12330022_1.fastq.gz -f SRR12330022_2.fastq.gz > SRR12330022.gam

vg pack -x ../PGGB.vcf.gz.giraffe.gbz -g SRR12330022.gam -t 64 -Q 5 -o SRR12330022.pack
```

Finally, I tried to call and genotype variants with:

```
vg call -t 32 ../PGGB.vcf.gz.giraffe.gbz -k SRR12330022.pack -s SRR1233002 -z > SRR12330022.vcf
```

2. I expected `vg call` to produce a VCF file containing the called variants and genotypes.

3. The output VCF was empty, and I got this error:

```
error[vg call] No reference paths found
```

4. No stack trace was generated.

5. The input files are `Jiujiu_Ma_1.0.fa`, `PGGB.vcf.gz`, `SRR12330022_1.fastq.gz`, and `SRR12330022_2.fastq.gz`. I can provide them if needed. The commands required to reproduce the problem are listed above.

faithokamoto:
Looks like you need to give a reference path. You can check the paths in your graph with `vg paths --metadata -x PGGB.vcf.gz.giraffe.gbz`, and you can mark things as reference following the instructions in the wiki: https://github.com/vgteam/vg/wiki/Changing-References

There are also `--ref-path`, `--ref-sample`, `--path-prefix` options to tell `vg call` which samples to treat as references. Though even then the paths need to be marked as such. (Come to think of it, that should probably be in the error message.)

faithokamoto:
I've PR'd an update to make better error messages. As for how to solve your problem right now, you should do the `vg paths --metadata` thing, and if you don't see things marked as REFERENCE when they should be reference, then use the wiki page I linked to set an appropriate reference sample.

majssssa:
Thank you very much for your help. I followed your suggestion and marked `Jiujiu_Ma_1.0` as the reference sample.

[...]

PR #4975, opened 2026-07-21
Better vg call errors when no reference paths are found

faithokamoto:
## Changelog Entry

 * `vg call` errors about not finding reference paths are more descriptive

## Description

For #4974 the error message is rather unhelpfully vague. Now the errors should be more specific about what exactly is the problem, why it is still a problem even if you've helpfully passed e.g. a `--ref-sample` which lacks REFERENCE sense paths, and also how to fix the problem (i.e. the wiki page).

### DIFF
--- a/src/subcommand/call_main.cpp
+++ b/src/subcommand/call_main.cpp
@@ -593,7 +593,8 @@ int main_call(int argc, char** argv) {
             }
         });
         if (ref_paths.empty()) {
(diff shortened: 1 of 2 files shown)


adamnovak (approved):
This is an improvement.

But we could also add code to use a reference overlay here to make the selected paths become references, so the user doesn't need to do it themselves at all. We should think about doing that.

Issue #510, opened 2016-10-13
Clarify which kinds of graph chunked_call works on


glennhickey:
chunked_call has some hidden assumptions that can trigger cryptic assertion errors
* node size must be < 1000
* node ids must be sorted
* (as we've seen in #469) reference path must be acyclic for vg call

I also don't know if it's ever been tested on graphs with reverse-direction mappings in the reference path.  vg call should work on this case, but should verify if the chunking id logic holds up on this as well.  At the very least, this should be spelled out somewhere in the documentation or help, and have more useful error messages.

faithokamoto:
Script is deprecated.

Issue #2153, opened 2019-03-08
Chunked_call/whole genome variant calling


zchacwi76:
I have been trying to use vg call to call variants from a whole genome graph. I have attempted to use vg call on the whole genome without chunking, but the output vcf file was empty. I then tried using the chunked_call script, however it is unclear how to use it. 

chunked_call [-h] [--chunk CHUNK] [--overlap OVERLAP]
                    [--filter_opts FILTER_OPTS] [--pileup_opts PILEUP_OPTS]
                    [--call_opts CALL_OPTS] [--threads THREADS] [--overwrite]
                    xg_path gam_path path_name path_size sample_name out_dir

vg/scripts/chunked_call wholegenome.xg wholegenome_aln.gam chrY 57227415 SAMPLE001 ~/Scratch/graph2/chunks/

I am not sure that I have specified the path name and size correctly, as I have encountered the following error while using chunked_call, and am not sure how to proceed:
```
ValueError: invalid literal for int() with base 10: '"94899485"\n'
```
Any help with this issue would be really appreciated as I would really like to be able to call variants on the whole genome!

ekg:
The main problem is that you can't just run VG call on a whole (large) genome. There are two obstacles. The first is VG's in memory graph model. It is very heavy, and makes it impossible to directly run call on the graph. We should be able to fix this by replacing this class with [odgi](https://github.com/vgteam/odgi) (og?). @glennhickey is call completely handle graph dependent now? We could make the swap directly in call and it's dependents. The second problem is GAM indexing. Do we have the right kind of GAM index to work directly from that and the compressed graph model? If so then we can avoid chunking and make the calling process appear single step to the user in the sense that we make the chunks dynamically.

glennhickey:
The chunked_call script is out of date.  You can use toil-vg call (
https://github.com/vgteam/toil-vg) instead.   There is a bit more
information here https://github.com/vgteam/vg/issues/2122 and here
https://github.com/vgteam/toil-vg/wiki/Whole-genome-calling, but I'm still
behind on the documentation (please feel free to ask questions in the
meantime).

@ekg We've been working on SV genotyping, and I think the results are
finally good enough as of last week to freeze the code and start writing up
in earnest.  After that's done, I really want to take a stab at de-coupling
the process from vg chunk.  Not only does it suck to have to run these
complicated scripts to call, the chunking step is a huge time sink as
well.  I think the indexes we have now should be enough to do it
[...]

PR #4977, opened 2026-07-22
Have `chunked_call` note that it's deprecated

faithokamoto:
## Changelog Entry

 * `scripts/chunked_call` prints out a deprecation warning

## Description

As an alternative to something like #510 where we were trying to improve documentation for this script, I think now we want people to just not use it. Hasn't been touched in over half a decade.


faithokamoto:
ref for deprecation: https://github.com/vgteam/vg/issues/2153#issuecomment-471563849

adamnovak (approved):
This might be slightly better on standard error, but standard output is fine.

PR #4978, opened 2026-07-23
Teach vg call to do overlays for non-ref HAPLOTYPE paths, if we ask

### LINKED ISSUE #4976: When you tell `vg call` to use non-ref paths as a reference, it should pretend for you
> But we could also add code to use a reference overlay here to make the selected paths become references, so the user doesn't need to do it themselves at all. We should think about doing that.

_Originally posted by @adamnovak in https://github.com/vgteam/vg/pull/4975#pullrequestreview-4756713959_

The `--ref-path`, `--ref-sample`, and `--path-prefix` options tell `vg call` to call against those paths. However, if those paths exist but as non-reference sense, then it still doesn't work. It's possible for the user to address this by marking the path as reference in advance, but we could also temporarily pretend the path is a reference. The user very clearly _wants_ the path to be a reference for at least this invocation.

faithokamoto:
## Changelog Entry

 * `vg call` will call against a non-reference path if you explicitly ask

## Description

In the thrilling sequel to my previous minor vg call PR, we now dump some of the warnings I just added because they're no longer relevant. If you tell `vg call` that you want to call against a HAPLOTYPE-sense path, it will shrug and apply an overlay for you. Resolves #4976


src/subcommand/call_main.cpp
```diff
@@ -709,6 +717,17 @@ int main_call(int argc, char** argv) {
...
+            pretend_ref_paths.emplace(ref_path);
+        }
+    }
+
+    // Use an overlay if needed
+    if (!pretend_ref_paths.empty()) {
+        bdsg::ReferencePathOverlayHelper overlay_helper;
+        if (show_progress) logger.info() << "Applying overlay to treat HAPLOTYPE paths as REFERENCE" << endl;
+        graph = overlay_helper.apply(graph, pretend_ref_paths);
     }
```
adamnovak: This isn't going to work because the overlay helper will go out of scope at the end of the block, leaving `graph` as a pointer to the deallocated memory. You need it to be in the same scope as `graph` is, outside the `if`.

```suggestion
    bdsg::ReferencePathOverlayHelper overlay_helper;
    if (!pretend_ref_paths.empty()) {
        if (show_progress) logger.info() << "Applying overlay to treat HAPLOTYPE paths as REFERENCE" << endl;
        graph = overlay_helper.apply(graph, pretend_ref_paths);
}
```

src/subcommand/call_main.cpp
```diff
@@ -579,7 +579,7 @@ int main_call(int argc, char** argv) {
 
     // Process path prefixes to find ref paths
     if (!ref_path_prefixes.empty()) {
-        graph->for_each_path_of_sense({PathSense::REFERENCE, PathSense::GENERIC}, [&](const path_handle_t& path_handle) {
+        graph->for_each_path_of_sense({PathSense::REFERENCE, PathSense::GENERIC, PathSense::HAPLOTYPE}, [&](const path_handle_t& path_handle) {
```
adamnovak, line 582: I wish we had a way to query paths by prefix, since adding hundreds of HPRC haplotype paths to the scan doesn't feel great, but I don't think we do. And maybe it's fine.

src/subcommand/call_main.cpp
```diff
@@ -593,15 +593,19 @@ int main_call(int argc, char** argv) {
             }
         });
         if (ref_paths.empty()) {
-            logger.error() << "No REFERENCE or GENERIC paths (see vg paths -M) found matching prefix(es)\n"
-                           << "Also see: https://github.com/vgteam/vg/wiki/Changing-References" << endl;
+            logger.error() << "No non-alt paths found matching prefix(es) paths (see vg paths -M)" << endl;
```
adamnovak: Does `vg paths -M` provide any insight here anymore? Maybe you want `--list` instead?

src/subcommand/call_main.cpp
adamnovak: The comment says we "use them all", but actually we've been using all reference/generic paths, or the reference paths belonging to the reference sample.

Now we want to use all reference/generic paths, or the paths of any kind belonging to the reference sample.

This would probably be simpler if we [used `for_each_path_of_sample` or `for_each_path_matching`](https://github.com/vgteam/libhandlegraph/blob/07fd981f0f43e9d4d8aa61a84169a9eeda69152a/src/include/handlegraph/path_metadata.hpp#L177-L189) to do the query when we have a sample name, and then if that doesn't work we could scan all the reference paths. Right now we're scanning all haplotype paths ourselves looking for that sample, and it would be better if we used the method that lets the graph drive that query in the fastest way it knows how.

We might want one lambda that we can pass to the appropriate query function.

adamnovak, line 670: I think you could change this to go through `ref_path_set` and [call `for_each_subpath_of()`](https://github.com/vgteam/vg/blob/b21f13969166b66eb68022a6f87581b52df207d9/src/path.hpp#L435-L446) to get the subpaths of each reference path by breaking it down into sample, locus, etc. and querying those.

I think it's possible for that to end up requiring *more* full scans of the graph path set, if the graph doesn't have its paths indexed by sample *and* locus. That could be handled by adding a bulk version that takes a collection of base paths and queries for all the samples, and loci at once and filters down.

This is probably fine as is too; we don't want to spend too much time chroming the queries we do once at startup.

src/subcommand/call_main.cpp
```diff
@@ -709,6 +717,17 @@ int main_call(int argc, char** argv) {
             }
         }
         ref_path_ploidies.push_back(path_ploidy);
+
+        if (graph->get_sense(graph->get_path_handle(ref_path)) == PathSense::HAPLOTYPE) {
+            pretend_ref_paths.emplace(ref_path);
+        }
```
adamnovak, line 726: You don't have to do bookkeeping to put *only* the haplotype paths in here; you could just turn the whole ref path list into a set and pass it along to the overlay helper as like `apply(graph, {ref_paths.begin(), ref_paths.end()})` I think.

adamnovak (changes requested):
This looks pretty good, except I think the overlay is getting deallocated while in use and the declaration needs to move.

I think there are ways to potentially improve the path querying using some of the lesser-known `PathMetadata` query methods, or possibly by using some of the subpath-related helpers in vg's `path.hpp` (though the right bulk versions might be missing there).

src/subcommand/call_main.cpp
faithokamoto: handled commit ef1ee4c2f67109e5392286cf05f8a6815313be4b

src/subcommand/call_main.cpp
faithokamoto: handled commit 6b466a6db022c0fa298e3e23cd6be428e1bfccb6

src/subcommand/call_main.cpp
faithokamoto: That would require swapping `graph` from a `PathHandleGraph` to a `PathPositionHandleGraph`, which seems like probably way bigger of a change than I wanted this PR to be.

faithokamoto: Split the `ref_sample` logic elsewhere in commit b26e7f0faf65a67685db6aea6a1caf83b995ced4

faithokamoto:
Yeah, seems I have provided more overhead in this load step, in multiple places. We have to loop over all haplotype paths both to look up paths by a prefix (if a prefix is given), and then later to check in on our all ref paths (if we pass a prefix, a sample, or just path names). It would be more efficient to skip that final everything-loop (the `else` block that starts with `if paths are given, we convert them to subpaths so that ref paths list corresponds`) but then I'd need to duplicate the subrange handling code a bunch. And this is one-time at startup anyhow.

faithokamoto:
Had to revert to doing the overlay only for non-true-ref paths, because it was causing test failures.

src/subcommand/call_main.cpp
adamnovak: I think the function might be asking for too strong of an interface; it doesn't actually appear to use position queries in there.

adamnovak (approved):

PR #4979, opened 2026-07-25
Force manpage to have all non-deprecated subcommands

faithokamoto:
## Changelog Entry

 * manpage generator is required to include non-deprecated subcommands (and is updated to do so)
 * We have a `CONTRIBUTING.md`!

## Description

I have no idea why our [manpage](https://github.com/vgteam/vg/wiki/vg-manpage) is missing subcommands that we actually use. My best guess is that when those commands were added we forgot to update the manpage generator. This does make the list longer. Though, that's sort of the point of a manpage: to have all the information right here. I did make a concession to leave out deprecated subcommands since we don't want to point people to them.

This PR updates the `scripts/check_options.py` unit test script (which is rapidly becoming the "penalize practices Faith doesn't like" script) to force the manpage to include those subcommands, and also updates the manpage to past that test.


faithokamoto:
I'm not really expecting this to get in before the next version release, but once it does I'll update the wiki page.

adamnovak (approved):
This is a good idea.

At some point we're going to want to rename `check_options.py` to maybe something more like `lint.py`.

And rather than pestering people with failing tests after the fact, we ought to make some effort (in a `CONTRIBUTING` document?) to make sure people know they need to do this in advance.

Or, ideally, we should make this part of the man page *also* generate automatically, and keep the One True Copy of this documentation with the subcommand itself.

Maybe we carve out a place in the subcommand system for a Markdown tagline, and set up a `vg help` that lets us pull it out?

faithokamoto:
I'll work on documenting the errors in a CONTRIBUTING file before I merge this.

faithokamoto:
@adamnovak I've tossed together a `CONTRIBUTING.md`. I have ideas for how to get subcommands to know themselves how they should be described in the manpage (convenient that we tend to make all the wiki links in the same format). Question: should I set that up as its own command (e.g. `vg man`) or as an option within `vg help`, e.g. `vg help --man`?

adamnovak:
Rolling it into `help` might make things simpler because then we have fewer weird subcommands just for us lying around.

adamnovak (approved):

faithokamoto:
Merging this in for now, will add `vg help --man` later

PR #4983, opened 2026-07-27
Fix static binary

glennhickey:
vg [1.76.0](https://github.com/vgteam/vg/releases/download/v1.76.0/vg) doesn't work on my computer at home.  This reproduces when I `make static` on mustard.  Seems likely that it'll affect other users as well.  

It seems that this was a known issue before, and the work-around was undone in #4929 in an effort to add support for Ubuntu 26.  This caused a regression that apparently broke binary compatiblity. 

This PR reverts this change, bringing back the submodule, and fixing the build for it.  It will necessitate at patch release.

Some more details from the chatbot who did the work: 

#4929 swapped the deps/elfutils submodule for the system libdw, dropping the --disable-libdebuginfod --disable-debuginfod it was configured with, and the comment saying why:

    We need to disable libdebuginfod or the static binary will try and load it
    at runtime and pull in incompatible libs it depends on on whatever system
    it's running on.

Distro libdw has debuginfod enabled, and elfutils dlopens libdebuginfod.so.1 from a constructor before main. In a -static binary that maps the host's libc.so.6 in beside the one we linked and dies in __ctype_init, so every invocation segfaults. v1.76.0 does this on a glibc 2.39 host; v1.75.1, the last release before the change, does not.

Restoring the rule also needs the two libraries 0.195 split out and the 0.186 layout it was written for did not have: libeu (eu_tsearch and friends) and libdwfl_stacktrace (dwflst_tracker_cache_elf). Both were already built by the rule, neither was installed or linked. That is probably why #4929 gave up on the submodule, since b87ebaa9b bumped 0.186 -> 0.195 a few commits earlier and would have broken the link exactly this way.

Verified by a full static build from clean on Ubuntu 26.04 (GNU Make 4.4.1, gcc 15.2, glibc 2.43) and on 22.04: no libdebuginfod string in the binary, and it runs on a glibc 2.39 host where both v1.76.0 and an unpatched local build segfault.

## Changelog Entry

 * Revert back to elftools submodule (instead of system install) in order to fix static build.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -130,3 +130,6 @@
 [submodule "deps/pixman"]
 	path = deps/pixman
 	url = git://anongit.freedesktop.org/git/pixman
+[submodule "deps/elfutils"]
+	path = deps/elfutils
+	url = https://github.com/vgteam/elfutils.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ RUN apt-get -qq -y update && apt-get -y install --no-upgrade \
     raptor2-utils rasqal-utils bison flex gawk libgoogle-perftools-dev liblz4-dev liblzma-dev \
     libffi-dev libfontconfig-dev libfreetype-dev libglib2.0-dev libpcre2-dev libpng-dev \
     libprotobuf-dev libboost-all-dev tabix bcftools libzstd-dev pybind11-dev \
-    python3-pybind11 pandoc libssl-dev libjitterentropy3-dev kmc libdw-dev meson
+    python3-pybind11 pandoc libssl-dev libjitterentropy3-dev kmc meson
 ###DEPS_END###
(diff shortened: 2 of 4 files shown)


adamnovak:
Sorry!

If we don't want to only rely on the more-strongly-worded comment (which probably will work), we would want to add something to the CI tests that tries out the static binary on an older (or at least different) distro/glibc than it is built on.

This is not the first time elfutils has reorganized everything on us; maybe depending on a bunch of details of the layout of an 0.xxx project is a bad idea. We could look at https://github.com/bombela/backward-cpp#libraries-to-read-the-debug-info and do some testing and see if we can't move to... maybe libdwarf, since while the one in Ubuntu also is in 0.x at least there are supposedly several API-compatible implementations. IIRC I tested a bunch of them a long time ago and could really only get libdw working, but maybe that's changed.

I think we might strip most of the good stuff that this reads when we `strip -d` our static release binaries anyway.

PR #4984, opened 2026-07-28
add warning to `vg augment -i` for nameless paths

### LINKED ISSUE #500: cyclic alignments with MEM threading / should path inclusion in vg mod warn when an unnamed path is used?
@jeizenga and I were just investigating what happens when we give overlapping MEMs in the threading aligner. Turns out I fixed the problem, but we got confused because when we used `vg mod -i` to edit the graph we didn't get a graph back with the path we were interested in.

```
vg construct -r tiny/tiny.fa >t.vg
vg index -x t.xg -g t.gcsa -k 16 t.vg
vg map -Q cycle -s CAAATAAGGCTTGGAAATTTTCTGGAGTTCTACTGGCGTTCTATTATATTCCAACTCTCTG -x t.xg -g t.gcsa -a -D >t.gam
vg mod -i t.gam t.vg | vg view -dp - | dot -Tpng -o x.png
```

This produces something that seems semi-reasonable:

image

Removing the `-Q cycle` means the path doesn't have a name, which results in a graph with the path edits but not the embedded path.

I configured things this way but it's very confusing. Would a warning that an anonymous path is being used for editing be helpful? I guess we might often want to do this.

Also, when the repeated sequence in the alignment is exact, we get a nice loop directly out of the MEM threading mode:

image

And another problem we ran into (which is visible above) is that the alignment picks a weird 3bp MEM rather than a larger one. We can fix this by adding `-L 10` to the `vg map` call:

image

faithokamoto:
## Changelog Entry

 * `vg augment -i` warns when it is told to embed a path without a name

## Description

Resolves #500 (the `-i` option moved from `vg mod` to `vg augment` in the interim). I could easily be convinced that this should be an error instead.

### DIFF
--- a/src/augment.cpp
+++ b/src/augment.cpp
@@ -302,6 +302,10 @@ void augment_impl(MutablePathMutableHandleGraph* graph,
             Path added = add_nodes_and_edges(graph, simplified_path, node_translation, added_seqs,
                                              added_nodes, orig_node_sizes);
 
+            if (aln.name() == "") {
+                cerr << "warning[augment] embedding a path with no name; downstream operations may fail." << endl;
+            }
+
             // Copy over the name
             *added.mutable_name() = aln.name();
 


adamnovak (approved):
I don't think there's anything about generic paths that says the empty string *can't* be a path name. But it is indeed kind of an Air Bud situation.

jltsiren:
If we want to maintain compatibility with GFA, generic path names must be non-empty, as they have to match regex `[!-)+-<>-~][!-~]*`.

adamnovak:
Hmm. Unless we want to fail at GFA export, maybe we should indeed make this warning an error.

PR #5003, opened 2026-08-08
test: strip BSD wc padding so the test suite is usable on macOS

benedictpaten:
## The problem

BSD `wc -l` right-aligns its count in a fixed-width field; GNU `wc -l` does not. So an assertion like

```bash
is $(grep -v "^#" out.vcf | wc -l) 3 "..."
```

compares `"       3"` against `"3"`. It passes on Linux and fails on macOS.

The suite already knows about this and works around it ad hoc in two different ways — `| tr -d ' '` in a handful of places, `| sed 's/^[[:space:]]*//'` in another handful. **503 sites have neither.**

## Why it matters

Whole test files fail unconditionally on macOS, which makes them useless as a regression signal. Measured before and after:

| file | before | after |
|---|---|---|
| `18_vg_call.t` | 43 failures | 0 |
| `32_vg_snarls.t` | 7 | 0 |
| `36_vg_annotate.t` | 7 | 0 |
| `40_vg_gamcompare.t` | 5 | 0 |
| `53_clip.t` | 5 | 0 |
| `03_vg_view.t` | 3 | 0 |

`09_vg_concat.t`, `12_vg_kmers.t` and `14_vg_mod.t` passed before and after — the change is not simply making failures disappear.

I hit this while working on something unrelated in `18_vg_call.t`: four assertions I had genuinely broken were invisible behind 43 pre-existing failures nobody could read past.

## The change

Mechanical. `wc -l` → `wc -l | tr -d ' '` everywhere it was unguarded, leaving the existing `sed`-based sites alone so the diff does not churn them.

## Verification and its limits

- The six files above were measured FAIL → PASS, three more confirmed unaffected.
- Every test file still parses under `bash -n`.
- The remaining files are the same substitution and are **not** individually verified — several need fixtures or runtimes I do not have locally.
- On Linux this cannot change behaviour: the padding it strips is never present, so `tr -d ' '` is a no-op on every value CI sees.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/t/37_vg_gbwt.t
+++ b/test/t/37_vg_gbwt.t
@@ -33,9 +33,9 @@ is $(vg gbwt -c x.gbwt) 2 "chromosome X: 2 paths"
 is $(vg gbwt -C x.gbwt) 1 "chromosome X: 1 contig"
 is $(vg gbwt -H x.gbwt) 2 "chromosome X: 2 haplotypes"
 is $(vg gbwt -S x.gbwt) 1 "chromosome X: 1 sample"
-is $(vg gbwt -T x.gbwt | wc -l) 2 "chromosome X: 2 path names"
-is $(vg gbwt -C -L x.gbwt | wc -l) 1 "chromosome X: 1 contig name"
-is $(vg gbwt -S -L x.gbwt | wc -l) 1 "chromosome X: 1 sample name"
+is $(vg gbwt -T x.gbwt | wc -l | tr -d ' ') 2 "chromosome X: 2 path names"
+is $(vg gbwt -C -L x.gbwt | wc -l | tr -d ' ') 1 "chromosome X: 1 contig name"
+is $(vg gbwt -S -L x.gbwt | wc -l | tr -d ' ') 1 "chromosome X: 1 sample name"
 
 rm -f x.gbwt parse parse_x.gbwt
 rm -f parse_x parse_x_0_1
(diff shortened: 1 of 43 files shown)


adamnovak (approved):
I think in places we left this unquoted, the shell has been saving us by parsing the output; I'm not sure the example int eh PR comment really shakes out the way it's explained.

And I think on CI we're not hitting this because we probably have GNU wc installed; we might want to *remove* that so that CI actually enforces that we work with BSD wc.

But this looks right to me.

adamnovak:
I've pulled this in to run the tests as `test-wc-portability`.

test/t/37_vg_gbwt.t
```diff
@@ -371,7 +371,7 @@ vg gbwt -g gfa2.gbz -G graphs/components_walks.gfa
 is $? 0 "GBZ construction from GFA"
 is $(md5sum gfa2.gbz | cut -f 1 -d\ ) 74d119e982352d10a3bd29f23db0ccc3 "GBZ was serialized correctly"
 vg gbwt -Z gfa2.gbz -o /dev/null --translation gfa2.trans
-is $(wc -l < gfa2.trans) 0 "no chopping: 0 translations"
+is $(wc -l | tr -d ' ' < gfa2.trans) 0 "no chopping: 0 translations"
```
adamnovak, line 374: Looks like Claude was slightly *too* mechanical here and wit this in the wrong place with respect to the input redirection.

PR #5018, opened 2026-08-21
Automatic manapge generation with `vg help --man`

faithokamoto:
## Changelog Entry

 * `vg help --man` will output a fully formed Markdown manpage
 * `vg help subcommand` will output the same thing as running `vg subcommand`, i.e. the default helptext

## Description

BEHOLD MY WORKS. Ahem. Extension of #4979 but now the Better (TM) manpage is generated by `vg help --man` instead of a separate script. I've diffed it against the current manpage and what changed were:
- A few subcommand blurbs were edited to be better (e.g. `vg giraffe`'s dropped the "short" from before "read alignment")
- Things are correctly alphabetized
- Deprecated subcommands still have their helptext printed, they're just not in the lists at the top.

Because `vg help` now needs to be able to print helptexts for other commands, I have `Subcommand`s save a helptext function. Some subcommands' helptext functions require extra arguments. I worked around this by defining extra standard-argument functions, e.g. see `help_augment_default()`.

The main thing I'm worried about here is `vg giraffe`. I did some terrible no good very bad hacks to make its default helptext function work because the previous one really wanted to be passed a list of presets. If you have a better way to do that I'm all ears.


faithokamoto:
Un-hacked `vg giraffe` so now the helptext prints a fixed list of presets instead of dynamic generation. To make the list short enough I got hid `sr-chaining`, but we don't want people to use it anyhow. I also removed one of the default help function uses. `vg augment` used a completely separate system for a lot of its options and I don't see why it needed it. My best guess is that it was supposed to auto-generate stuff to force options to be registered correctly, but the system never really got used elsewhere and now the `scripts/lint.py` will enforce the rules so I just dumped `src/option.hpp`/`.cpp`

src/subcommand/giraffe_main.cpp
```diff
@@ -772,6 +761,12 @@ void help_giraffe(char** argv, const BaseOptionGroup& parser, const std::map<std
     }
 }
 
+// A version taking only argv to make the subcommand registry happy
+void help_giraffe_default(char** argv) {
+    GroupedOptionGroup empty_parser;
+    help_giraffe(argv, empty_parser, false);
+}
+
```
adamnovak, line 769: I don't think this will really work; that parser object is where about half the options actually live, including most of the ones that change what the mapping algorithm does.

src/option.hpp
adamnovak, line 1: I liked this option system, but apparently not enough to use it.

src/subcommand/subcommand.hpp
```diff
@@ -91,30 +146,38 @@ class Subcommand {
...
      * Make and register a subcommand with the given name and description, in
-     * the given category, with worst priority, which calls the given main
-     * function when invoked.
+     * the given category, with the given manpage entries,
+     * with the given help function, with worst priority,
+     * which calls the given main function when invoked.
      */
     Subcommand(std::string name, std::string description,
         CommandCategory category,
+        std::vector<manpage_item> manpage_entries,
```
adamnovak, line 168: Do we expect one subcommand to have multiple entries? I couldn't find any, just some that have an entry and some that don't. If we only want 0 or 1 of something, there's now `std::optional` for that.

src/subcommand/subcommand.hpp
```diff
@@ -66,6 +72,55 @@ enum CommandCategory {
...
+    MANIPULATE_GRAPH,
+    // Conversion between formats
+    CONVERT_FORMAT,
+    // Subgraph extraction
+    EXTRACT_GRAPH,
+    // Extremely specific analyses
+    RARE_NEEDS,
+    // Developer tools
+    DEV_TOOLS
+};
```
adamnovak, line 99: We're adding this second set of categories so we can more or less exactly reproduce the current manpage. But then we end up with both a command category and maybe a categorized manpage section on each command, except that deprecated-category commands don't need manpage sections. Some of these (`DEVELOPMENT` and `DEV_TOOLS`) sure sound like they mean the same thing.

I would try and combine these two systems into one. We could make it hierarchical and put each command in exactly one leaf category (like `FORMAT_CONVERSION`), which then can fall into a broader category (probably what is now `WIDGET`). Or we can give each command 0 to n tags. Or we can throw away the existing command categories and use only the categories we're currently using for the manpage, plus `DEPRRECATED`.

Also, we probably want these to be all verbs or all nouns; `GET_STATS` and `RARE_NEEDS` are different types of phrase.

src/subcommand/subcommand.hpp
```diff
@@ -31,12 +29,20 @@
...
+ *              << endl;
+ *     }
+ * 
  *     int main_frobnicate(int argc, char** argv) {
  *         return 0;
  *     }
  *
  *     static Subcommand vg_frobnicate("frobnicate", "frobnicate nodes and edges",
- *         main_frobnicate);
+ *                                     help_frobnicate, main_frobnicate);
```
adamnovak, line 45: This could be an example that includes manpage information.

src/subcommand/subcommand.hpp
```diff
@@ -66,6 +72,55 @@ enum CommandCategory {
...
+};
+
+/// Represents a list item in the manpage
+struct manpage_item {
+    /// Sub-list to put this item in
+    ManpageSection section;
+    /// Description of subcommand usage
+    std::string blurb;
+    /// A relevant wiki page (optional)
+    std::string wiki_link;
```
adamnovak, line 108: I think if we put some constructors on the manpage item, we could avoid a bunch of empty strings in the subcommand code when there's no wiki page link.

src/subcommand/surject_main.cpp
```diff
@@ -945,4 +945,6 @@ int main_surject(int argc, char** argv) {
 }
 
 // Register subcommand
-static Subcommand vg_surject("surject", "map alignments onto specific paths", main_surject);
+static Subcommand vg_surject("surject", "map alignments onto specific paths", WIDGET,
+                             vector<manpage_item>{{CONVERT_FORMAT, "project graph alignments onto a linear reference (gam/gaf->bam/sam/cram)", ""}},
+                             help_surject, main_surject);
```
adamnovak, line 950: We have the existing subcommand short descriptions and the existing manpage short descriptions, but are they really different enough in the job that they do that we shouldn't just use one string for both?

I think what we really want here is a single short description for when the command appears in reference lists, and then a way to include a paragraph or two of explanation about the command. The wiki page links, when they exist, can sort of do that, but I don't think we aspire to have a separate 2-paragraph wiki page about every subcommand. When someone runs `vg surject --help` and still doesn't get it, they might check the vg surject manpage next, and it would be good if we could offer more detail.

I think ultimately we'd also want to have longer-than-fit-in-the-list explanations for individual options as well, but we don't have to do that yet.

adamnovak (changes requested):
I think this is a good idea.

But I don't think the square peg/round hole adapter for the Giraffe help function is going to work, because we actually need the option group object that gets built to produce help for half the options.

One approach to fix that could be not taking help functions at all and just running the main functions with "--help" as an argument. Another approach could be taking the help functions but shimming the difficult cases with a fake help function that calls into main. And the third approach is refactoring difficult cases like this so both the help function and the main function can get ahold of the objects they want to share (with globals or an enclosing class or a builder function or something).

I also think we could harmonize the manpage category/description system and the help category/description system, and that we could carve out a place to put a longer description instead of having two short ones.

src/subcommand/subcommand.hpp
faithokamoto: handled commit a8272d926d84a35e3db5eb58ec577905955cfa8f

src/subcommand/subcommand.hpp
faithokamoto: handled commit 5ee8bf7bba9b12d7d29cacc549ebfdc6ab7b26cd

src/subcommand/subcommand.hpp
faithokamoto: Some do have multiple, e.g. see `vg paths` with a ridiculous four entries.

src/subcommand/surject_main.cpp
faithokamoto: For the ones that have multiple entries (e.g. `vg paths`) we use tailored descriptions for each use case. I mean the core problem there is that some commands do ridiculously many things, but short of splitting them up, we do want to be able to trumpet all functionalities

src/subcommand/giraffe_main.cpp
faithokamoto: But do we _want_ to print the full help? Currently we don't have the manpage use the `--help` helptext; we only have the result of running `vg giraffe` without arguments, which is what I'm replicating here.

src/subcommand/subcommand.hpp
faithokamoto: This is complicated due to the subcommands which show up in multiple sublists, such as `vg paths`. How do you think those should be handled?

src/subcommand/giraffe_main.cpp
adamnovak: I think we want the manpage documentation to include every available option, so it can be a reference for them. Real manpages in the actual `man` manual seem to do it that way.

src/subcommand/subcommand.hpp
adamnovak: OK, I looked at the wiki page and when we have something like vg call, which we put in multiple categories with multiple blurbs for its multiple use cases, clicking through from the listing in each category to the actual `vg paths` section is not at all helpful for figuring out how to do anything but the most obvious use case, because it just shows the command help.

`vg paths` seems to have wiki page links for how to use it for the different use cases, but e.g. `vg gbwt` has all the wiki links pointing to the main GBWT wiki page.

Maybe we should say a subcommand has a main use case and 0 or more alternative use cases. You can set up the subcommand with the main use case alone, or with the main one and a list of alternate ones. Each alternate one needs to have a link to something explaining how to actually do that alternate use case with the subcommand.

Then we can categorize the subcommands by their main use case, and then use the alternate use cases for these extra cross-listings.

adamnovak: The main use case blurb and the full command blurb should probably be the same string, or at least we should have an overload that lets them be the same string.

src/subcommand/surject_main.cpp
adamnovak: I think if we make having extra use cases more of a special thing, we can make things cleaner for the case we want to be doing, which is one use case per command.

src/subcommand/giraffe_main.cpp
faithokamoto: swapped over as of commit 7789951fd107d2600a2dd70501a8f682082d0727

Issue #2098, opened 2019-02-08
Augmenting graph within genotype uses more memory


glennhickey:
As discovered in #2079, CI test 1 runs out of memory when using a toil-vg that augments the graph inside of genotype (as opposed to re-running augment first).  This could be
* normal: it keeps something around that costs a few extra bytes, just pushing the tests over the limit
* genotype being inefficient:  maybe something big is hanging around in a member variable somewhere beyond when it's needed
* memory leak in edit(): this would be more problematic than the other two, and should be checked soon

Issue #1365, opened 2018-01-11
Remove dependence on defray filter 


glennhickey:
A lot of time is spent on making xg indexes of graph chunks for calling in the current pipeline.  See test_call_chr21_snp1kg results here : http://cgl-pipeline-inputs.s3.amazonaws.com/vg_cgl/vg_ci/jenkins_reports/commit/38ef47f9c9f201de9a2b67ef656bedd6946c2257/index.html

The only reason that's being done is to run vg filter -D.  Would be nice if we can get genotype robust enough to not rely on this or any other nontrivial gam filters.  

An alternative would be to revert toil-vg back to using the full index for calling chunks, but this was clunky for whole genomes.

Issue #571, opened 2016-11-29
genotype should at least consider min and max length path through bubble


ekg:
Right now, the genotyper can't be used on SVs because it only considers paths supported by reads.

While I think this is a smart decision, maybe in some cases we could relax it and consider even paths that are too long to be covered by a single read? Would this be easy to fit into the new genotypekit framework? How could I go about doing so?

PR #5019, opened 2026-08-24
Deprecate `vg genotype`

faithokamoto:
## Changelog Entry

 * Mark `vg genotype` as deprecated

## Description

For @benedictpaten as I said in meeting today.

Resolves (as stale) the following issues because we'd not be bothering to update/evaluate `vg genotype`:

- #442
- #571
- #1242
- #1272
- #1365
- #2098

It seems we have some `toil-vg` tests that involve `vg genotype`? Those would need to be removed or reworked once we actually nuke it.


adamnovak:
We're deprecating this in favor of a project that will roll alignment-based (as opposed to pileup-based) calling into `vg call`. We're hoping to drop `vg genotype` at the same time we merge that, so we're wanting to mark it deprecated now.

adamnovak (approved):

PR #4997, opened 2026-08-04
Handle X-dropped ranges in dozeu more consistently

jeizenga:
## Changelog Entry
To be copied to the [draft changelog](https://github.com/vgteam/vg/wiki/Draft-Changelog) by merger:

 * `dozeu` respects X-drop definition more strictly
 
## Description

We were alerted that the VG fork of dozeu did not strictly respect the X-dropped column ranges during traceback. This PR updates dozeu to a commit that fixes this bug.

PR #5020, opened 2026-08-25
Fix Dozeu X-drop a different way

adamnovak:
## Changelog Entry

 * `dozeu` respects X-drop definition more strictly

## Description

This takes some fixes that @jeizenga and @mgawan hammered out that fix a bug in traceback in Dozeu, using @mgawan's code.

This goes with https://github.com/vgteam/dozeu/pull/17.

I tried to also merge in https://github.com/vgteam/dozeu/pull/16, but I ended up with a Dozeu commit with, as far as I could tell, no non-comment code changes versus https://github.com/vgteam/dozeu/pull/17 but which made vg fail the `vg test "[xdrop]"` tests. So this is taking @mgawan's version as merged into Dozeu. @mgawan said some things I didn't really understand about the changes not actually working on our mainline Dozeu, which differed from what they were built on by what looks to me like a no-op merge commit, so if something goes wrong here we might need to do some serious Git surgery to unravel what's going on.

This closes #4997, a competing PR to do the same basic thing


adamnovak:
I don't want to merge this until I'm sure that @mgawan is happy with that Dozeu commit actually going into the project. If not, we'd want to back it out again.

mgawan:
All is good from our end. Thanks for working on it. 

I believe I had missed Jordan's more recent commits on his PR. But looks like we both ended up with the same end fix.

PR #5026, opened 2026-08-31
vg surject : --diploid-map and other code changes

Sagorikanag:
## Purpose

This draft combines the complete diploid map surjection work into one branch for review.

This has not been rebased over #5025 yet. The paired-surjection implementation overlaps with that PR, and will be changed after reviewing by @adamnovak.

## Main changes

- Add `vg surject --diploid-map`.
- Compare alternative graph placements jointly.
- Select haplotype-local and global primary surjections.
- Add `hp`, `hq`, `GlobalQ`, and `aq` annotations.
- Score paired surjections using alignment and fragment-length evidence.
- Learn the fragment-length distribution before enabling parallel processing.
- Handle paired reads with no compatible path/strand placement.
- Prune mapper-declared tail-region anchors.
- Classify overlapping alternatives as secondary and disjoint alternatives as
  supplementary.
- Rank paths by their best alignment.
- Realign anchors containing substitutions.
- Validate anchor sliding on target paths.
- Preserve tail annotations through GAF.
- Add implementation and evaluation documentation.

----
- ## Dependencies

This PR depends on two open `libvgio` PRs:

- vgteam/libvgio#93 adds grouped parallel input for alternative placements of unpaired reads.
- vgteam/libvgio#95 adds grouped parallel input for alternative placements of interleaved paired-end reads.
- Existing smaller surjection PRs may be superseded by this consolidated PR.


doc/diploid-map-evaluation.md
adamnovak, line 1: This report and `diploid-map-surjection.md` are both very thorough, and could help people understand this PR, but I don't think they make a lot of sense to carry in the vg repository, at least without organizing things so it's clear that these are reports about a particular project that was done at a particular time.

Some of the material in these would be useful on the vg wiki, because it covers how `vg suject`'s diploid mode now works, and what the new tags mean.

But a lot of this is about how the changeset in particular was tested, and what particular changes to surjection are part of this project, and there are a lot of particular accuracy statistics and references to particular experiments you or I did at some point. That is all of limited utility to future readers of the vg repository, who will only care about how everything works at that future point, and would be better placed in a manuscript about your project.

src/subcommand/surject_main.cpp
```diff
@@ -40,25 +40,24 @@ void help_surject(char** argv) {
          << "options:" << endl
          << "  -x, --xg-name FILE        use this graph or XG index (required)" << endl
          << "  -t, --threads N           number of threads to use" << endl
+         << "  -d, --diploid-map NAME    like --into-ref, but also turns on --multimap and" << endl
+         << "                            emits hp/hq/aq tags for haplotype-aware" << endl
+         << "                            diploid surjection" << endl
```
adamnovak, line 45: `--into-ref` is described as just `surject into this reference assembly`, so we don't save a lot of space by referring to it, whereas it makes the user need to find another option later in the list to understand this one.

It's not quite clear from this if "haplotype-aware diploid surjection" is the name of what this turns on, or some downstream process that consumes the tags, or some other thing that can be flagged on separately and, if turned on, will emit those tags when this flag is used.

src/subcommand/surject_main.cpp
```diff
@@ -605,6 +564,114 @@ int main_surject(int argc, char** argv) {
...
+                                                     std::move(pair.second),
+                                                     max_frag_len);
+                    }
+                    if (!result.singles.empty()) {
+                        alignment_emitter->emit_singles(std::move(result.singles));
+                    }
+                    total_reads_surjected += 2;
+                };
+
+                auto process_fragment = [&](PairedSurjector::fragment_group_t& placements) {
```
adamnovak, line 583: This then gets passed to e.g. `vg::io::gam_paired_grouped_for_each_parallel_after_wait`, which doesn't know about/depend on `PairedSurjector::fragment_group_t` and always wants to give you a vector of pairs of `Alignment`s. That makes it seem like there might be some confusion around exactly what piece of code owns the idea of a `fragment_group_t`, in case we wanted to someday change it to a real struct or something.

I would describe the things that need to attach to `libvgio` in the same terms that `libvgio` does, and if that's too complicated and we need some abstractions for the types, the definitions should live in `libvgio` if we can get away with it.

src/subcommand/surject_main.cpp
```diff
@@ -727,16 +839,80 @@ int main_surject(int argc, char** argv) {
...
+                } else {
+                    auto gaf_checking_lambda = [&](Alignment& src) {
+                        check_gaf_aln(src);
+                        return lambda(src);
+                    };
+                    vg::io::gaf_unpaired_for_each_parallel(*xgidx, file_name, gaf_checking_lambda);
+                }
             }
         }
     } else if (input_format == "GAMP") {
```
adamnovak, line 918: What if someone provides `--diploid-map` with GAMP input? It doesn't look like we document that it won't work, or like we produce an error when asked to do that, but it also doesn't look like it actually activates diploid mode. So a user can think that they've run diploid surjection when they haven't.

src/subcommand/surject_main.cpp
```diff
@@ -727,16 +839,80 @@ int main_surject(int argc, char** argv) {
...
+                        // Compute hp/hq within this graph placement.
+                        surjector.annotate_hap_tags(placement, surjected);
+
+                        all_surjected.insert(all_surjected.end(),
+                                             std::make_move_iterator(surjected.begin()),
+                                             std::make_move_iterator(surjected.end()));
+                    }
+
+                    // Compute GlobalQ and select one primary across all graph placements.
+                    surjector.annotate_global_mapq_and_primary(placements.front(), all_surjected);
```
adamnovak, line 878: I think most of this (maybe except for the `ensure_alignment_is_for_graph()` and/or `set_metadata()` calls) might want to be a method on `Surjector` (maybe an overload or variant of `surject()`?) because you need all these pieces to accomplish the operation of taking a collection of graph placements and producing the corresponding collection of linear placements. This is a piece we might want to be able to write a unit test for.

src/subcommand/surject_main.cpp
```diff
@@ -727,16 +839,80 @@ int main_surject(int argc, char** argv) {
...
+
+                    alignment_emitter->emit_singles(std::move(all_surjected));
```
adamnovak, line 880: Does `emit_singles()` promise no other threads' reads will interpose themselves between the reads in this batch? If not, maybe it should be made to promise that, because the user probably wants all the surjections coming out grouped by read if the inputs were.

src/subcommand/surject_main.cpp
```diff
@@ -605,6 +564,114 @@ int main_surject(int argc, char** argv) {
...
+
+                auto emit_result = [&](PairedSurjector::result_t&& result) {
+                    for (auto& pair : result.pairs) {
+                        alignment_emitter->emit_pair(std::move(pair.first),
+                                                     std::move(pair.second),
+                                                     max_frag_len);
+                    }
+                    if (!result.singles.empty()) {
+                        alignment_emitter->emit_singles(std::move(result.singles));
+                    }
```
adamnovak, line 579: I'm pretty sure that other threads are allowed to interpose their output between what we write with an `emit_pair` call and what we later write with an `emit_singles` call. Is that a problem? Or something we need to warn downstream readers of the file about?

src/subcommand/surject_main.cpp
```diff
@@ -727,16 +839,80 @@ int main_surject(int argc, char** argv) {
                     report_exception(ex);
                 }
             };
+
+            // Group all graph placements for a read so haplotype and global
+            // primary selection are performed jointly.
+            std::function<void(std::vector<Alignment>&)> set_lambda = [&](std::vector<Alignment>& placements) {
```
adamnovak, line 845: First I thought that `set_lambda` was setting a value somewhere, either to a lambda or as itself a lambda that sets values. Then I thought that this was for that logic in libvgio I saw one time that groups input reads based on a lambda for how to key the group, especially since the comment says that the verb here is "Group".

But really this is a lambda for operating on a set of all the placements for a particular read at once, and surjecting them all together.

It might be possible to name or document this more clearly.

src/surjector.hpp
```diff
@@ -93,133 +92,53 @@ using namespace std;
...
-                                     bool allow_negative_scores = false,
-                                     bool preserve_deletions = false) const;
-
-        /// Populate refpos on Alignments from a separate collection of
-        /// positions.
-        void set_refpos(vector<Alignment>& alns, const vector<tuple<string, int64_t, bool>>& positions) const;
-        
-        /// a local type that represents a read interval matched to a portion
-        /// of the alignment path
+        /// Compute MAPQ for the selected candidate relative to the competing scores.
```
adamnovak, line 96: I recently learned [we actually need 2 lines](https://www.doxygen.nl/manual/docblocks.html) of `///` comment for Doxygen to pick it up as a doc comment.

src/surjector.hpp
```diff
@@ -93,133 +92,53 @@ using namespace std;
...
+        int32_t compute_mapping_quality_from_scores(const Alignment& source,
+                                                    const std::vector<double>& scores,
+                                                    size_t best_idx,
+                                                    bool fast_approximation = false) const;
+
+        /// Choose the best non-supplementary haplotype surjection and annotate
+        /// candidates with hp and hq.
+        void annotate_hap_tags(const Alignment& source,
+                               std::vector<Alignment>& alns) const;
```
adamnovak, line 105: The name of this function make sit sound like it's annotating tags, but I think it's really annotating the alignment with the tags, or creating or adding the tags.

src/surjector.hpp
```diff
@@ -93,133 +92,53 @@ using namespace std;
...
+
+        /// Choose one overall primary across non-supplementary surjections and
+        /// annotate candidates with GlobalQ and aq.
+        void annotate_global_mapq_and_primary(const Alignment& source,
+                                              std::vector<Alignment>& alns) const;
```
adamnovak, line 110: This function also could probably have a better name.

src/surjector.hpp
```diff
@@ -93,133 +92,53 @@ using namespace std;
...
+
+        /// Score a read pair using alignment scores and, when available,
+        /// fragment-length likelihood.
+        double score_alignment_pair(const Alignment& source,
+                                    const Alignment& first,
+                                    const Alignment& second,
+                                    int64_t template_length,
+                                    double fragment_mean,
+                                    double fragment_stddev,
+                                    bool use_fragment_length) const;
```
adamnovak, line 126: If there's a separate `PairedSurjector`, and this is only ever called from `PairedSurjector`, what is this doing on the plain `Surjector`?

If we feed in a first read alignment and a seocnd read alignment, what is `source`, the third alignment of the pair, for? Is it the source of the first or the source of the second?

src/surjector.hpp
```diff
@@ -397,17 +320,22 @@ using namespace std;
         template<class AlnType>
         string path_score_annotations(const unordered_map<pair<path_handle_t, bool>, vector<pair<AlnType, pair<step_handle_t, step_handle_t>>>>& surjections) const;
         
-        // helpers to choose one among supplementary alignments to be the primary
+        /// Select the highest-scoring surjection as primary, classify overlapping
+        /// alternatives as secondary, and retain disjoint alternatives as supplementary
```
adamnovak, line 324: This means overlapping and disjoint in the read, not in the reference? It could be clearer.

src/surjector.hpp
```diff
@@ -533,410 +461,80 @@ using namespace std;
...
-            // Move in the new vectors.
-            alns1 = std::move(new_alns1);
-            positions1_out = std::move(new_positions1);
-            alns2 = std::move(new_alns2);
-            positions2_out = std::move(new_positions2);
-        }
-    }
+    /// Surject paired graph placements, score compatible pairs, and select
+    /// haplotype and global primaries using alignment and fragment-length evidence.
+    class PairedSurjector {
```
adamnovak, line 470: I hear all the cool kids are favoring composition over inheritance now, but it seems a little odd to me that a `PairedSurjector` is not a `Surjector`.

I think splitting it up like this lets the main method work better though, because we don't need to create the base class or the derived class before we fill in all the configuration fields on the `Surjector`.

Maybe this should be a `PairSurjector` instead, in the sense of something that surjects pairs?

src/surjector.hpp
```diff
@@ -533,410 +461,80 @@ using namespace std;
...
+    public:
+        using alignment_pair_t = pair<Alignment, Alignment>;
+        using fragment_group_t = vector<alignment_pair_t>;
+
+        struct result_t {
+            vector<alignment_pair_t> pairs;
+            vector<Alignment> singles;
+        };
```
adamnovak, line 478: If these are the organizing abstractions they probably could use some documentation. What's important about a `fragment_group_t` that won't be true of just *any* vector of pairs of `Alignment`s? What's the relationship between the pairs and the singles in a `result_t`? What's the required order?

src/surjector.hpp
```diff
@@ -533,410 +461,80 @@ using namespace std;
...
+
+        PairedSurjector(Surjector& surjector,
+                        const PathPositionHandleGraph* graph,
+                        size_t maximum_sample_size = 1000,
+                        size_t reestimation_frequency = 100,
+                        double robust_estimation_fraction = 0.95,
+                        size_t maximum_buffered_fragments = 50000);
```
adamnovak, line 485: This is a different approach than what we did for all the tunable parameters on `Surjector`, where we default them in the class and allow overwriting from outside.

It's probably fine, but if we start to have a lot of CLI options that control these we might want to make things more consistent.

src/surjector.hpp
```diff
@@ -533,410 +461,80 @@ using namespace std;
...
+            vector<Alignment> second;
+            vector<pair_candidate_t> candidates;
+        };
+
+        int64_t compute_template_length(const Alignment& first,
+                                        const Alignment& second) const;
+        size_t choose_best(const Alignment& first_source,
+                           const Alignment& second_source,
+                           const vector<double>& scores) const;
+        void finalize_learning();
```
adamnovak, line 527: These could probably use documentation.

src/surjector.cpp
```diff
@@ -117,6 +120,587 @@ using namespace std;
...
+            begin = end + 1;
+        }
+
+        if (!new_tags.empty()) {
+            new_tags.push_back('\t');
+        }
+        new_tags += tag + ":" + type + ":" + value;
+
+        set_annotation<std::string>(aln, "tags", new_tags);
+    }
```
adamnovak, line 189: This *works*, but it's not (and probably can never be) efficient to call several times in a row. Each time you set a tag, it scans throught he existing tags, and then it does a bunch of string concatenation which in C++ ends up copying the strings every time. The net result is `O(n^2)` to add `n` tags.

And then later it gets called multiple times in a row, to set multiple tags.

Instead of working with the 2-character SAM tags on `Alignment`s, we should define either GAM annotations or real GAM fields for the new information, and then add code to serialize them to/deserialize them from SAM and GAF into the IO layer. Then we can access them efficiently internally.

src/surjector.cpp
```diff
@@ -117,6 +120,587 @@ using namespace std;
...
+            && fragment_length_distribution->std_dev() > 0.0) {
+            fragment_length_distribution->force_parameters(
+                fragment_length_distribution->mean(), fragment_length_distribution->std_dev());
+            fragment_scoring_available = true;
+        }
+        else {
+            learning_abandoned = true;
+            fragment_scoring_available = false;
+        }
+    }
```
adamnovak, line 377: I'm wondering if there needs to be more to `FragmentLengthDistribution` so that this doesn't look so much like https://github.com/vgteam/vg/blob/3a37aa335c671c0dd8dbd2a895f000fe707ee2c3/src/minimizer_mapper.hpp#L538-L549

Why do we need to both have the `FragmentLengthDistribution` around and also wrap and expose all these methods? Does the abandonment feature need to be pushed down into `FragmentLengthDistribution`?

Maybe we should be inheriting from `PairedEndMapper` (which might need to be renamed to encompass other kinds of paired-end read processing)?
https://github.com/vgteam/vg/blob/3a37aa335c671c0dd8dbd2a895f000fe707ee2c3/src/mapper.hpp#L144-L168

src/surjector.cpp
```diff
@@ -117,6 +120,587 @@ using namespace std;
...
+                                               second_offset, path_length, 0);
+        if (first_cigar.empty() || second_cigar.empty()) {
+            return numeric_limits<int64_t>::max();
+        }
+        auto lengths = compute_template_lengths(first_offset, first_cigar,
+                                                second_offset, second_cigar);
+        return llabs(static_cast<int64_t>(lengths.first));
+    }
+
+    // Select the highest-scoring candidate, shuffling ties deterministically.
```
adamnovak, line 412: These comments maybe want to be in the header?

src/surjector.cpp
```diff
@@ -117,6 +120,587 @@ using namespace std;
...
+                                         work.front().source_second, global_scores);
+        size_t best_placement = global_index[best_global].first;
+        size_t best_candidate = global_index[best_global].second;
+        auto& winning_placement = work[best_placement];
+        auto& winning_candidate = winning_placement.candidates[best_candidate];
+
+        // Learn fragment length from confident pairs and buffer ambiguous fragments.
+        if (!ready_for_parallel()) {
+            bool can_train = winning_placement.source_first.mapping_quality() == 60
+                && winning_placement.source_second.mapping_quality() == 60
```
adamnovak, line 551: Should we accept input from other mappers with MAPQs higher than 60?

src/surjector.cpp
```diff
@@ -117,6 +120,587 @@ using namespace std;
...
+                                  bool preserve_deletions,
+                                  uint64_t maximum_fragment_length,
+                                  result_t& result) {
+        result = result_t();
+        if (placements.empty()) {
+            return true;
+        }
+
+        fragment_scoring_available = fragment_length_distribution->is_finalized();
+        vector<placement_t> work;
```
adamnovak, line 437: If this is some kind of scratch that we use through the whole function, it might make sense to explain how we make sure that it has the values we need it to have at any given point, or what it's supposed to look like after every pass through it.

src/surjector.cpp
```diff
@@ -117,6 +120,587 @@ using namespace std;
...
+                            mate = &placement.second[placement.candidates[found->second].second];
+                        }
+                    }
+                    set_annotation(aln, "mate_info",
+                        mate
+                            ? mate_info(mate->refpos(0).name(), mate->refpos(0).offset(),
+                                        mate->refpos(0).is_reverse(), false)
+                            : mate_info("", -1, false, false));
+                }
+            }
```
adamnovak, line 661: I don't really understand this. I think it was like this before, but don't we need to figure out which alignment of this read the supplementary alignment supplements, and then figure out which alignment of the pair partner is actually paired with it?

Otherwise we might have a pair of primary non-supplementary alignments at the start of a chromosome, and a pair of secondary non-supplementary alignments at the end, and then all the supplementary alignments saying they belong to the pair at the start of the chromosome even if they themselves appear at the end, right?

adamnovak (changes requested):
I've looked at the overall design here and the backing abstractions, and made some comments about them. I found some things about how this is written out that I would want to change (I don't think we want to merge the big Markdown reports, for example, and some things might want renaming).

I haven't really dug much into the core algorithms; I think they really need to be constrained by tests. If they give the right answers for weird edge cases I will be more confident in the core algorithms.

PR #5027, opened 2026-09-01
avoid underflow

faithokamoto:
## Changelog Entry

 * Avoid underflowing multiplicities in `vg giraffe` chaining mode (caused erroneous MAPQ 0)

## Description

We had been managing to underflow `chain_count_by_alignment` here such that multiplicities got set as just around the maximum value for a `size_t`. Such high multiplicity guaranteed a MAPQ of 0.

This will not change the alignments output. It _will_ likely "rescue" some MAPQ 0s, including some that represent incorrect alignments, and that will change variant calling. I have not tested the extent of the change.

### DIFF
--- a/src/minimizer_mapper_from_chains.cpp
+++ b/src/minimizer_mapper_from_chains.cpp
@@ -2249,7 +2249,9 @@ void MinimizerMapper::do_alignment_on_chains(const Alignment& aln, const std::ve
                                      seeds[chains[chain_i].back()],
                                      seeds[chains[chain_j].front()],
                                      seeds[chains[chain_j].back()])) {
-                --chain_count_by_alignment[i];
+                // chain i is basically chain j, so don't count it against chain j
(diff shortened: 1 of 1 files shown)


faithokamoto:
In my tests, this changes between 0.2% to 0.8% of mapping qualities, most of which used to be 0. In R10 reads some of those are rescued all the way to 60. HiFi reads tend to get rescued to single-digit MQs. Unfortunately it rescues more wrong MQ0s than correct MQ0s... I'll run variant calling tests next.

faithokamoto:
Fixing the bug lead to a likely-trivial increase in variant calling errors, but then we can't exactly leave an ugly underflow in, can we?

adamnovak (approved):

adamnovak:
I think we have to just hope DeepVariant can train away any calling dis-improvement from this. Unless it's big enough that we can extract a useful and sensible heuristic from the bug we had.

PR #5028, opened 2026-09-02
Update README.md to include Clang version

adhirakunnath-hue:
Noticed when I tried to install vg, it didn't work cause my Clang version was too low. Wanted to add this in the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,16 @@ The easiest way to get vg is to download one of our release builds for Linux. We
 
 **For MacOS**, see [Building on MacOS](#building-on-macos).
 
+## Pre-requisite
+
+Before doing these steps make sure to check your Clang version is greater than 21.0: 
+
+    clang -v
+
+Otherwise update Clang:
+
+    sudo softwareupdate --install "Command Line Tools for Xcode 26.6-26.6"
+
 ### Building on Linux
 
 If you don't want to or can't use a pre-built release of vg, or if you want to become a vg developer, you can build it from source instead.
--- a/README.md
+++ b/README.md
@@ -53,16 +53,6 @@ The easiest way to get vg is to download one of our release builds for Linux. We
 
 **For MacOS**, see [Building on MacOS](#building-on-macos).
 
-## Pre-requisite
-
-Before doing these steps make sure to check your Clang version is greater than 21.0: 
-
-    clang -v
-
-Otherwise update Clang:
-
-    sudo softwareupdate --install "Command Line Tools for Xcode 26.6-26.6"
-
 ### Building on Linux
 
 If you don't want to or can't use a pre-built release of vg, or if you want to become a vg developer, you can build it from source instead.
@@ -156,9 +146,17 @@ The first step is to clone the vg repository:
     git clone --recursive https://github.com/vgteam/vg.git
     cd vg
 
-#### Mac: Install Dependencies
+#### Mac: Install and Update Dependencies
 
 VG depends on a number of packages being installed on the system where it is being built. Dependencies can be installed using either [MacPorts](https://www.macports.org/install.php) or [Homebrew](http://brew.sh/).
+
+Before doing these steps make sure to check your Clang version is greater than 21.0: 
+
+    clang -v
+
+Otherwise update Clang:
+
+    sudo softwareupdate --install "Command Line Tools for Xcode 26.6-26.6"
 
 ##### Using MacPorts
 


faithokamoto:
If that's a dependency, shouldn't it be listed under [the other dependencies](https://github.com/vgteam/vg#linux-install-dependencies)? We don't want to have separate sections for installing every separate dependency.

faithokamoto:
FWIW you'd need to rebase against current master (i.e. merge the most recent PR) or this couldn't pass CI tests. I can't say for sure if this is the correct way to add info to the README, seeing as I don't understand compilers nor Macs. Adam will theoretically come along eventually to approve this.

README.md
```diff
@@ -146,10 +146,18 @@ The first step is to clone the vg repository:
...
     cd vg
 
-#### Mac: Install Dependencies
+#### Mac: Install and Update Dependencies
 
 VG depends on a number of packages being installed on the system where it is being built. Dependencies can be installed using either [MacPorts](https://www.macports.org/install.php) or [Homebrew](http://brew.sh/).
 
+Before doing these steps make sure to check your Clang version is greater than 21.0: 
+
+    clang -v
```
adamnovak, line 155: We could note that the `clang -v` version check command might prompt you to install the Xcode Command Line Tools if you don't have them already, and if so you should do that.

README.md
```diff
@@ -146,10 +146,18 @@ The first step is to clone the vg repository:
...
+
+Otherwise update Clang:
+
+    sudo softwareupdate --install "Command Line Tools for Xcode 26.6-26.6"
```
adamnovak, line 159: That command for updating Clang looks like it only applies to exactly this Clang update; when they release a new patch for Clang that will say `26.11` or something and the `26.6` update might not be available anymore. We need to explain the general process.

```suggestion
Otherwise update Clang, by updating the Xcode Command Line Tools. First, find the "Label" of the available Command Line Tools update:

    softwareupdate --list

This will be something like `Command Line Tools for Xcode 26.6-26.6`. Take that label and ask `softwareupdate` to install it:

    sudo softwareupdate --install "Command Line Tools for Xcode 26.6-26.6"

You can also install the update by opening System Settings, clicking on General in the left sidebar, then on "Software Update" in the main pane, then finding the pending Command Line Tools update and clicking on its "Update Now" button.
```

adamnovak (changes requested):
Contrary to what @faithokamoto suggested, I don't think it quite makes sense to count Clang as a "dependency", since it's the compiler and not one of the things we can install through Macports or Homebrew.

I would move all the new stuff to a new `Mac: Check Compiler` section before `Mac: Install Dependencies`. In there, we could note that on Mac, we only test with Apple's version of Clang, and then give the minimum version, the command for checking the version, and the command(s) for updating.

README.md
```diff
@@ -146,7 +146,27 @@ The first step is to clone the vg repository:
...
+
+Before doing these steps make sure to check your Clang version is greater than 21.0: 
+
+    clang -v
+
+Otherwise update Clang:
+
+    sudo softwareupdate --install "Command Line Tools for Xcode 26.6-26.6"
+
+Otherwise update Clang, by updating the Xcode Command Line Tools. First, find the "Label" of the available Command Line Tools update:
```
adamnovak: This doesn't make sense as written now; it says "Otherwise update Clang" twice. Is the second way supposed to be an alternative to the first way? The text should explain better.

README.md
```diff
@@ -146,7 +146,27 @@ The first step is to clone the vg repository:
     git clone --recursive https://github.com/vgteam/vg.git
     cd vg
 
-#### Mac: Install Dependencies
+#### Compiler
```
adamnovak: I would prefix the section with `Mac: ` like the other ones (which is useful if we want to link to a section that has Mac and Linux versions, because the heading becomes the link target).

Also, the other headings are in the imperative, while this is a noun. So maybe:

```suggestion
#### Mac: Check Compiler
```

adamnovak (changes requested):

README.md
```diff
@@ -146,7 +146,27 @@ The first step is to clone the vg repository:
...
+
+    sudo softwareupdate --install "Command Line Tools for Xcode 26.6-26.6"
+
+Another way to do this is by updating the Xcode Command Line Tools. First, find the "Label" of the available Command Line Tools update:
+
+    softwareupdate --list
+    
+This will be something like `Command Line Tools for Xcode 26.6-26.6`. Take that label and ask `softwareupdate` to install it:
+
+    sudo softwareupdate --install "Command Line Tools for Xcode 26.6-26.6"
```
adamnovak, line 165: Aren't these both ways of updating the Xcode Command Line Tools? They both end up with the same final command example.

PR #4969, opened 2026-07-16
Fix `make clean` and put under CI test

faithokamoto:
## Changelog Entry
To be copied to the [draft changelog](https://github.com/vgteam/vg/wiki/Draft-Changelog) by merger:

 * Fix `make clean` and put under CI test

## Description

Resolves #1536. Instead of having to keep track of dependencies and hoping that each can clean itself up properly, now `make clean` takes the nuclear option of deleting everything and re-cloning. It may take longer, but at least it works.

PR #5032, opened 2026-09-09
Move make clean test

faithokamoto:
## Changelog Entry

 * `make clean` no longer rebuilds from complete scratch, and its test isn't in `test/t`

## Description

because people complained about how I went about #4969 :)


adamnovak:
I'd rather not have us insulting ourselves with [terms borrowed from the eugenics movement](https://en.wikipedia.org/wiki/Henry_H._Goddard#Vineland) in the commit history, if we can avoid it.

adamnovak (approved):

PR #5033, opened 2026-09-09
Allow duplicates to not count against multiplicity

faithokamoto:
## Changelog Entry

 * `vg giraffe` chaining mode removes overlapping chains from multiplicity calculation

## Description

Wheee another MAPQ bugfix. I'll run alignment & variant calling comparisons before merging.

We use multiplicity to count the number of independent as-good things. However we only check for independence if we get all the way to alignment; there's an earlier step where we toss out chains if they overlap a previous alignment. That seems like a pretty non-independent event. Thus here we make sure to decrement multiplicity as makes sense.

### DIFF
--- a/src/minimizer_mapper_from_chains.cpp
+++ b/src/minimizer_mapper_from_chains.cpp
@@ -1985,8 +1985,8 @@ void MinimizerMapper::do_alignment_on_chains(const Alignment& aln, const std::ve
     // Track what positions were used in previously generated alignments, so we
     // can fish out alignments to different placements.
     // Use pairs since we can't hash tuples.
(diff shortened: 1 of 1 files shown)


faithokamoto:
This is a very minor change, on the order of <100 reads in a million for R10, and zero HiFi reads in my tests. Most of those changes are rescuing things to MQ 60, and essentially all of them are correct reads.
- 55 real R10 reads change MQ (44 new MQ60)
- 72 sim R10 reads change MQ (58 new MQ60; 57 correct & 1 wrong)
    - 66 correct, 4 off reference, 2 wrong

